PR #7, opened 2018-08-29
Add empty dist folders to packages 

janiskelemen:
Build:packages fails due to an issue with rollup not creating the folders automatically so I added them to the repo.


philippkuehn:
I do not like to have dist folders in the repository. I've updated the build script so missing dist folders will be created for builds.

PR #12, opened 2018-09-03
add init event

Raerten:
In order to use debug tool [prosemirror-dev-tools](https://github.com/d4rkr00t/prosemirror-dev-tools) we need access to ProseMirror view object. This PR add Editor component event `init` with this variable.


hanspagel:
less than 15 minutes from pull request to merge 💪

PR #14, opened 2018-09-06
UnderlineMark added

Chrissi2812:
Added support to underline text.

Markdown Syntax changes:
`__bold__` will now be `__underlined__`

image


philippkuehn:
Hey @Chrissi2812, thanks for that!

I'm not sure about this one. I do not use markdown that often but I know others do :)
Underline has always been very controversial because it conflicts with bold and italic.

The current implementation of markdown is:

\*\*bold\*\*
\_\_bold\_\_
\*italic\*
\_italic\_
\~strike\~

with your change it is:

\*\*bold\*\*
\_\_underline\_\_
\*italic\*
\_italic\_
\~strike\~

Slack, for example, goes a completely different way:

\*bold\*
\_italic\_
\~strike\~

Maybe we could something like this:

\*\*bold\*\*
\_underline\_
\*italic\*
\~strike\~

Hmm… 🤔

Chrissi2812:
It was inspired by [Discord's Syntax](https://support.discordapp.com/hc/en-us/articles/210298617-Markdown-Text-101-Chat-Formatting-Bold-Italic-Underline-?page=6), but I also don't write Markdown frequently (besides Github Issues 😅). 

I didn't want to reinvent the wheel and use something thats unfamiliar like `=` or `-`.

Chrissi2812:
@philippkuehn I removed the Markdown Syntax for now.

PR #27, opened 2018-09-14
Supporting align left/right/center/justify

jaysaurus:
as the title suggests, it'd be good to implement RTF style left/right/center/jusitfy per node.  This can probably be implemented classically at the node level; in a similar fashion to the placeholder's "is-empty" class.  

If the author is in agreement, I would suggest that nodes with "tiptap-right" class will be text-align: right, "tiptap-center": center, "tiptap-justify": justify [anything else]: left.


philippkuehn:
Hmm, I do not see your commit. And could you please rebase your PR to the current master?

jaysaurus:
I'm proposing it first before I fork since you had a suggestions branch.
I'm keen to make sure you agree with the suggested approach before I start
coding :)

PR #30, opened 2018-09-18
make doc prop reactive fix #29

### LINKED ISSUE #29: change doc dynamically
it would be great to have something similar to `this.$refs.editor.clearContent()` that we can use to replace doc dynamically. eg. `this.$refs.editor.replaceDoc(doc)`

kopipejst:


philippkuehn:
Hey, thanks for your PR. There were some limitations with this PR. I released 0.12.0 with a better solution (see #29). 👍

PR #32, opened 2018-09-20
Alignment sample

jaysaurus:
**This is not a finished merge!**

### problem

I'm at a bit of an impasse with setting up buttons for alignment.  

1) I don't want to create new nodes, rather, edit the classes of the existing block nodes wherever the caret just was/a selection was made. 

2) I can update the contents with a plugin as I type but all efforts to have a button fire the event have failed - and believe me I have tried hours of different approaches and scoured the prose mirror documentation. 

### about the changes
In the sample attached I've created an "events" folder in `tiptap-extensions`.  all events should be wired into the eventMap file (inspired by the vuex mapActions/mapGetters/etc.) You can see it in action 

the key amendments are
```
|
|- Examples > Components > Routes > Alignment > Index.vue
|
|- Packages > tiptap > components > editor.render()
|                                `> utils > buildMenuAction (now returns state)
|
|- Packages > tiptap-extensions > extensions > Alignment
|                              `> events > align
|                                       `> eventMap
|
```

The only course I can think to take is to use the indexes provided by the document's descendants to manually amend the dom elements but I had hoped to be able to do that by exposing the "view" in the editor's slot-scope.

Any suggestions?


philippkuehn:
Hey @jaysaurus, thanks for your PR. I know that diving into Prosemirror is a tough job 😅 Currently I'm on vacation so I do not have much time to work on tiptap.

But here are some thoughts:

- you're using decorations to add classes for alignment. Keep in mind that decorations are not exported with `getJSON()` or `getHTML()`. They are useful to change styling (e.g. for search highlighting) without changing its content. But alignment should be saved in the document for each node, right?

- why not adding align as an attribute to the paragraph node?

philippkuehn:
I'll close this PR due to inactivity.

PR #42, opened 2018-10-01
Add cross-env for Windows 

narian:


philippkuehn:
Hey @narian. Thanks for your PR!

I‘m not sure why cross-env helps because all parameters are still being read with minimist.

But you also changed paths in the package
json. Maybe this is already the fix for Windows? I do not have a Windows PC for testing at the moment. Maybe you could check this? Thanks.

philippkuehn:
Closed because of #50.

PR #49, opened 2018-10-05
fix class name

Chrissi2812:
Wanted to do this earlier,and in the meantime you already have fixed one of the wrong class names 😃


philippkuehn:
Argh, thank you! 😅

PR #62, opened 2018-10-13
refactor(tiptap-extensions): Do not import the full `lowlight` library.

### LINKED ISSUE #16: "Tree-Shaking" support for TipTap
It look like "tiptap-extensions" is load all extensions even i did not use all of them.

```js
import {
    BlockquoteNode,
    BulletListNode,
    HardBreakNode,
    HeadingNode,
    ImageNode,
    ListItemNode,
    OrderedListNode,
    BoldMark,
    ItalicMark,
    HistoryExtension,
  } from 'tiptap-extensions'
```

I use only these extensions but in my vendor file, there is "highlight.js" dependency.

It would be nice if we can import only what we use.

erickwilder:
**BREAKING CHANGE**: `CodeBlockHighlight` was importing the full `lowlight` libraries, including _all_
syntax highlighting definitions from `highlight.js`. The new behavior changes the signature of
`CodeBlockHighlight` to accept an object with all syntax highlighting definitions. This means that
now the user of the library __MUST__ import languages themselves and tiptap will no longer
bundle the full `highlight.js` in itself.

---
### Testing: 

* Given that you have a working copy from the test repo[1]
* Run `npm run build` in the test repo and check if the highlight.js library is bundled with it.
* Checkout this branch and run `yarn build:packages`
* Copy[2] the contents of `tiptap-extensions` to the test repo:
  ```
  cp -R packages/tiptap-extensions /path/to/test/repo/node_modules/tiptap-extensions
  ```
* Run `npm run build` again on the rest repo
* You shouldn't see traces of `highlight.js` anymore.

*Expected result:*
screenshot 2018-10-13 at 16 11 20

---

This closes #16 

[1] https://github.com/erickwilder/tiptap-treeshaking
[2] I had some issues with `npm link` and just to be quick I've copied the locally compiled `tiptap-extensions` over the files installed vie `npm install` in the test repo.

### DIFF
--- a/examples/Components/Routes/CodeHighlighting/examples.js
+++ b/examples/Components/Routes/CodeHighlighting/examples.js
@@ -26,3 +26,25 @@ body, .usertext {
     content: attr(href)
   }
 }`
+
+
+export const explicitImportExample = `import javascript from 'highlight.js/lib/languages/javascript'
+import { Editor } from 'tiptap'
+
+export default {
+  components: {
+    Editor
+  },
+  data() {
+    return {
+      extensions: [
+        new CodeBlockHighlightNode({
+          languages: {
+            javascript,
+            css
+          }
+        })
+      ]
+    }
+  }
+}`;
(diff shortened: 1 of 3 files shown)


erickwilder:
@philippkuehn It would be nice to add the `hacktoberfest` label in the affected issue (#16, or this PR)😄 

Let's make the `1.0.0` roadmap happen! 🍻

PR #65, opened 2018-10-15
feat(tiptap): Add TypeScript type definitions for `tiptap` package.

erickwilder:
This PR adds some basic TypeScript type definitions for the `tiptap`  package. I decided to open separate PRs for each package as I can collect early feedback about it and learn in the process so that the next packages have better definitions.

I intentionally added the most basic form here and I _really_ need some guidance/reviews from someone with more experience with it.

### DIFF
--- a/packages/tiptap/package.json
+++ b/packages/tiptap/package.json
@@ -8,9 +8,11 @@
   "module": "dist/tiptap.esm.js",
   "unpkg": "dist/tiptap.js",
   "jsdelivr": "dist/tiptap.js",
+  "types": "types/index.d.ts",
   "files": [
     "src",
-    "dist"
+    "dist",
+    "types/*.d.ts"
   ],
   "repository": {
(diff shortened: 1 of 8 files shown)


philippkuehn:
Hey @erickwilder, thank you very much for this one. 👍 Currently I'm planning of how to replace scoped slots with components in v1.0. And I think this will need a complete rewrite of this package so maybe it makes more sense to wait for this PR?

erickwilder:
Yeah. I'm fine with that.

whosdustin:
I love the approach of rewriting and adding Typings to the v1.0. Is there a schedule of when that will happen? If not soon, @erickwilder would you have a suggestion on how to add these typings to a current project?

yekver:
Is there any chance this PR will ever be merged? It's really painful to use tiptap with TS without type defs right now :(

Issue #64, opened 2018-10-14
Font Size and Family


saloustrosm:
Hello, great plugin with great extensibility.
I would like to ask, is there a way (hint) how to implement an extension for making buttons for font size and family, font colors would be great too.
I could not find a solution from prosemirror and i suppose this project inherits this issue too?

beyersito:
Is anyone on this? I could give it a try, any ideas on implementation?

saloustrosm:
I probably have a solution for this, i certainly am not the hardcore Vue developer nor a prosemirror expert but i think it works quite well. I will upload tomorrow (in about 12 hours) the plugins (fontColor, fontFamily, fontFillColor and fontSize) and my editor.vue wrapper component hoping it will help someone.

philippkuehn:
Hey @saloustrosm! Do you want to share you solution? 🙌

saloustrosm:
Hey sorry guys, the few hours became few days, it is a crazy period.

So i am attaching the plugins that i made for the editor, also because i am a novice in vue, sounds reasonable to upload my editor.vue file where it binds everything together and i can simply then create an editor by calling it as component (<editor :text="oldText" @update="newText = $event.getHTML()" />)

Please keep in mind that i am using bubble skin, Quasar framework which made easy buttons and dropdowns, a swatch component that i made and keeps all colors (yours if ask), all font-sizes are in 'em' (easily changeable).

I will gladly provide more info. Hope it helps someone...

[tiptapPlugins.zip](https://github.com/heyscrumpy/tiptap/files/2533188/tiptapPlugins.zip)
[editor.zip](https://github.com/heyscrumpy/tiptap/files/2533207/editor.zip)

philippkuehn:
Hey, I decided to not support *styling* nodes in this core packages anymore.

jeremybphillips:
@saloustrosm could you please make available your swatch component?

Thanks!

saloustrosm:
Hey @jeremybphillips, ofcourse no problem, swatch component is attached inside .zip

Ofcourse a dropdown or popup will be needed to give it some good taste
I attach the way i call it bellow, it will be usefull if you use my plugins in the previous comments
`<swatch :value="('fontTextColor' in marks.fontTextColor.attrs) ? marks.fontTextColor.attrs.fontTextColor : '#000000'" @input="marks.fontTextColor.command({ fontTextColor: $event })" />`

Have fun, i wll gladly help if you need something else

[...]

PR #84, opened 2018-11-07
Add Font Family Extension for #64

lenamdn:
Adding Font Family Extension as the request #64


philippkuehn:
After some thinking I decided to **not** supporting all styling nodes like text alignment, font family or font color in this core packages. Maybe you can publish this as a npm package.

PR #106, opened 2018-11-23
Update README.md

RobinMglsk:
I think that markAttrs needs to be getMarkAttrs


philippkuehn:
You are correct. Thanks!

PR #121, opened 2018-12-07
Escape Suggestion trigger char

vanpav:
This fix enables to use $-char or any special char without infinite cycle :-)


philippkuehn:
Thanks!

PR #123, opened 2018-12-07
Register mention based on matcher char

AveryBibeau:
There needs to be a unique identifier for registered mentions when using multiple instances with different matcher characters. Otherwise the `mention` node in the schema is overwritten by the last registered `Mention` instance

### DIFF
--- a/examples/Components/Routes/Suggestions/index.vue
+++ b/examples/Components/Routes/Suggestions/index.vue
@@ -8,6 +8,10 @@
             <icon name="mention" />
             <span>Insert Mention</span>
           </button>
+           <button class="menubar__button" @click="commands.mention_hashtag({ id: 1, label: 'Dogs' })">
+            <span># &nbsp;</span>
+            <span>Insert Hashtag</span>
+          </button>
         </div>
       </editor-menu-bar>
       <editor-content class="editor__content" :editor="editor" />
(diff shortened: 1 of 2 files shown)


AveryBibeau:
While updating the examples with a second Mention instance I realized `get name` is used in naming commands and other fields so `commands.mention` would have to be `commands['mention-@']`
Any ideas for naming that would uniquely identify Mention instances while not disrupting syntax? I'm also worried changing `get name` could break existing code for others

AveryBibeau:
Added a `tag` option to `Mention` for identifying separate instances. Defaults to just using `mention` as the instance name so existing code like `commands.mention( ... )` still works. Adding `tag: 'hashtag'` to the Mention options will name the instance `mention_hashtag`. Also adds `data-mention-tag` to the DOM element for identifying what kind of mention it is.

philippkuehn:
Hey @AveryBibeau. Sorry for the late response. I'm not sure about the naming. You've added a `tag` option. Maybe `type` is a better name for it? So type could basically `user` or `hashtag` or whatever.

AveryBibeau:
@philippkuehn That's a fair point, I've changed the option name from `tag` to `type` now

AlecTroemel:
First of all thanks for this awesome project! Is there any update on this PR? Would love to see this merged!

philippkuehn:
Hey, sorry for my late response but I'm still thinking about this one. What I do not like is that we do some work here that is only usable for this one extension.

```js
export default class Mention extends Node {

  get name() {
    return this.options.type === 'default' ? 'mention' : `mention_${this.options.type}`
  }

}
```

So I was thinking of a method available for all extensions. Something like that:

```js
new Mention({ options }).setName('hashtag')
```

But that means, that we can't use a simple getter function for `name()` because the name attribute can be overwritten after initialization. Maybe we could implement a `init()` method instead? This would be a breaking change.

```diff
export default class Mention extends Node {

-  get name() {
-    return 'hashtag'
-  }

+  init() {
+    this.setName('hashtag')
+  }

}
```

What do you think? Any suggestions for better method names?

philippkuehn:
I don't want to change the current behavior. You can always extend extensions if you want to overwrite names.

```js
import {Mention} from 'tiptap-extensions'

export default class Hashtag extends Mention {
  get name() {
    return 'hashtag'
  }
}
```

iamcodemaker:
To make this work I had to at minimum set the following properties using both the `extend()` and `configure()` methods. Obviously you'll need other properties to have a fully working editor instance, but these were the minimum set need to get the desired behavior. I did this along side another mention instance using the defaults (`@` as the char) with no issues. This also worked with the `generateJSON()` method to generate json from html to hydrate the editor state. I'm working in typescript.

```typescript
import { Editor } from '@tiptap/core'
import Mention from '@tiptap/extension-mention'
import { PluginKey } from 'prosemirror-state'

const editor = new Editor({
  extensions: [
    Mention
      .extend({
        name: 'hashtag',
        suggestion: {
          char: '#',
        },
      })
      .configure({
        suggestion: {
          char: '#',
          pluginKey: new PluginKey('hashtag'),
        },
      }),
  ],
})
```

PR #139, opened 2018-12-17
Change renderless section title in readme

kajuberdut:
Love the project. I'm pretty new to javascript and github so this is the only thing I felt I could contribute to right now.

The construction "What means x?" asks the same as "what other word(s) have the meaning of x?".

"What does x mean?" asks for the definition of x. I think this is more fitting for the title of the renderless section.

Additional reference: https://www.englishforums.com/English/WhatMeansWhatDoesMean/bgzpcx/post.htm


hanspagel:
Thanks! 🤓

PR #141, opened 2018-12-18
added drop cursor

### LINKED ISSUE #140: Drop indicator for dragged elements like images
It would be great to see a cursor for drag operations on where the dragged element would be inserted on release.

Example - not working: In https://tiptap.scrumpy.io/images drag the image to another place in the text.

Example- working: In https://prosemirror.net/examples/dino/ drag the dinos. A simple text cursor becomes visible.

Solution: Add `dropCursor` and `gapCursor` plugins like here: https://github.com/ProseMirror/prosemirror-example-setup/blob/master/src/index.js#L56

Chrissi2812:
This will close #140


Chrissi2812:
added configuration for dropCursor to the Editor:

```js
new Editor({
    dropCursor: {
        width: 2,
        color: '#ccc',
    },
});
```

philippkuehn:
Hey @Chrissi2812. Sorry for the late response. I did not added the dropcursor plugin to the core because I do not like it's styling. For some reason there is added a border-right when dragging something between two nodes.

bad

I think it would be way better to do something like that (it's from notion.so):

good

Unfortunately there is no option for that in the official dropcursor plugin so my plan was to create a custom one.

philippkuehn:
EDIT: okay. my fault. this is already the default behavior in the dropcursor plugin. My example above was for inline and block elements 🤦‍♂️ I'll merge this.

Chrissi2812:
No Problem 😃

The behavior of notion.io looks promising too.

Would be cool for todo lists and tables, as currently it's not really working 😅
tiptap

philippkuehn:
Ah good point. Currently `draggable` is set to `false` for todo items in the schema. When set to `true` it's working fine:
dec-20-2018 10-31-10

PR #178, opened 2019-01-25
Update examples.js

iliyaZelenko:
I stumbled upon the [documentation "Code Highlighting"](https://tiptap.scrumpy.io/code-highlighting), it confused me. 
It looks like it should be as I fixed if you rely on [this code](https://github.com/scrumpy/tiptap/blob/21c8ad852a8cf6602153948ffe40415794a318a4/examples/Components/Routes/CodeHighlighting/index.vue).

Also, the highlighting did not work for me until I connected one of the themes:
```js
import 'highlight.js/styles/github.css'
```
This is if I do not copy styles that are in the `<style>` tag for the second link.


philippkuehn:
Thanks!

PR #185, opened 2019-01-31
added markdown rule for horizontal line

Chrissi2812:
This allows to use `---` as an alias for a `<hr />`

Also `___` and `***` are supprted but must be terminated with a space to take effect, otherwise __bold__ and _italic_ won't work at start of line.


philippkuehn:
nice! thank you!

PR #202, opened 2019-02-18
seperate handleDrop from handlePaste

Chrissi2812:
Because sometimes you maybe just want to handle drop events 😅

This will however break if someone uses `onPaste` to stop drop events.


philippkuehn:
Ah, thank you! :D

PR #210, opened 2019-02-25
Reactive extension options

### LINKED ISSUE #153: Placeholder text is not reactive
@vanpav Mentioned it in issue #127 that is not reactive yet. However, it is not clear if their is a work around for it.

Chrissi2812:
This allows something like this which will call the update method of that extensions.
```js
this.editor.extensions.options.placeholder.emptyNodeText = newValue
```

This will close #153 

# Example
placeholder

# Known Bugs
~~If the content is **not** empty the state get's mismatched, and everything is screwed up 😩~~ fixed with 89e6436 and 0f44e74


philippkuehn:
This one is great! Thanks a lot!

PR #213, opened 2019-02-26
Add support to history custom settings

### LINKED ISSUE #211: Support history custom settings
Currently, Tip Tap doesn't support custom settings on the history extension, like depth and the delay time to create a new history group.

It seems that **prosemirror-history** [has support for this feature](https://github.com/ProseMirror/prosemirror-history/blob/master/src/history.js#L388). I think tip tap could support this too.

I could work on this, if you guys are ok with it!

MatheusRich:
Fixes #211.

Let me know if there's something more to do! :smile:

### DIFF
--- a/packages/tiptap-extensions/src/extensions/History.js
+++ b/packages/tiptap-extensions/src/extensions/History.js
@@ -7,6 +7,13 @@ export default class History extends Extension {
     return 'history'
   }
 
+  get defaultOptions() {
+    return {
+      depth: '',
+      newGroupDelay: '',
+    }
+  }
+
   keys() {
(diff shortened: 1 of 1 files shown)


packages/tiptap-extensions/src/extensions/History.js
```diff
@@ -23,7 +30,10 @@ export default class History extends Extension {
 
   get plugins() {
     return [
-      history(),
+      history({
+        depth: this.config.depth,
```
Chrissi2812: It's **`this.options.depth`** not `this.config.depth`

MatheusRich: @Chrissi2812 Whoops! Fixed! :smiley:

philippkuehn:
Nice! Thank you!

PR #228, opened 2019-03-07
Fix Menu bubble position

### LINKED ISSUE #215: Wrong bubble position
ezgif-1-a99eb749f278

browser: latest chrome
os: macOS

Chrissi2812:
# Bug
`EditorView.cordsAtPos`  returns wrong cordinates if we are at a linebreak.
wrong-position-2

# Fixed
right-position

This will fix #215

### DIFF
--- a/examples/Components/Routes/MenuBubble/index.vue
+++ b/examples/Components/Routes/MenuBubble/index.vue
@@ -1,6 +1,6 @@
 <template>
   <div class="editor">
-    <editor-menu-bubble :editor="editor">
+    <editor-menu-bubble :editor="editor" :keepInBounds="keepInBounds">
       <div
         slot-scope="{ commands, isActive, menu }"
         class="menububble"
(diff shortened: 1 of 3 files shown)


pangrr:
Anyone can review this? It's really helpful.

Chrissi2812:
As mentioned by @bvkimball in #239 the menuBubble was sometimes placed outside the editor bounding box.

When enabled via:

```html
<editor-menu-bubble :editor="editor" :keepInBounds="true">
```

It will keep the menuBubble inside it's offsetParent element.

Currently the `keepInBounds` option is **NOT** reactive.
This will be part of another PR because this involved a rewrite of the Plugin to an Extension so it's handled by the `ExtensionManager`.

philippkuehn:
Thanks for that PR, Chris! I'm not sure if it's a good idea to overwrite core functions of ProseMirror but I will merge that for now.

PR #239, opened 2019-03-18
fix(MenuBubble): MenuBubble doesn't update position when it resizes

bvkimball:
fix(MenuBubble): MenuBubble doesn't stay within bounding box of editor

## BUG
The menu bubble will not detect proper left alignment because of translateX(-50%)
In links demo the menu bubble resizes to show link form, this also causes offset issues.

UxPnkqxUiv

## FIXED

N4X8MnQwb9

## Missing

- [ ] Does not handle right alignment issues
- [ ] Will cause conflicts with #228


philippkuehn:
This should be fixed with #228.

ahstro:
I'm still experiencing this bug. Did #228 solve this for you @bvkimball?

PR #243, opened 2019-03-20
fix(deps): remove high severity vulnerability and add extra travis check for PRs

erickwilder:
The only exception is `webpack-svgstore-plugin` that still points to some older package versions
with `low` severity vulnerabilities and apparently, the repo is not active that [the actual fix](https://github.com/mrsum/webpack-svgstore-plugin/pull/172)
could be merged. Because of that, I'm limiting the scope to just `high` severity issues.

Note: I decided to include `audit-ci` because `yarn` does not provide (yet) a way to filter
violations by severity level.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,5 +14,6 @@ install:
   - yarn build:packages
 
 script:
+  - yarn audit-ci
   - yarn lint
   - yarn test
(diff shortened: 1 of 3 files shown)


erickwilder:
@hanspagel / @philippkuehn 

If you have some time later to review this one it would be awesome.

philippkuehn:
Thanks, Erick! This is great!

PR #246, opened 2019-03-27
Update config.js

iliyaZelenko:
It looks like it can be optimized through the [concat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/concat) method.


NikitaIT:
`concat` very slow, if you're really concerned about performance use `push` instead, but in this case `.flat()` or `.flatMap()`

philippkuehn:
I like destructuring, sorry!

PR #248, opened 2019-03-27
build: different module `output.name`

iliyaZelenko:
Hi, you did a good package, great job! 

At the moment I am developing my package based on yours. But I had a little problem.

I make a package that uses `tiptap` and `tiptap-extensions` packages.  But these two packages are written in one variable - `tiptap`. 

In my rollup config in `output -> globals` I have to specify two different names, not

```
globals: {
  'tiptap': 'tiptap',
  'tiptap-extensions': 'tiptap'
}
```

Would be right:

```
globals: {
  'tiptap': 'tiptap',
  'tiptap-extensions': 'tiptapExtensions'
}
```

My PR removes the problem with the name (variables) conflict when you use the _UMD_ module as _IIFE_.

### DIFF
--- a/build/packages/config.js
+++ b/build/packages/config.js
@@ -41,7 +41,7 @@ function genConfig(opts) {
       file: opts.file,
       format: opts.format,
       banner,
-      name: 'tiptap',
+      name: opts.outputName,
     },
   }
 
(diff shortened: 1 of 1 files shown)


iliyaZelenko:
To be honest, I did not understand the reason for the error CI. Is my code exactly to blame?

philippkuehn:
there was an issue with your PR. this should be fixed in https://github.com/scrumpy/tiptap/commit/d0fc40413943cc70c19a4d27964a93d3fc3ba6f4

Issue #168, opened 2019-01-18
Preserve &nbsp; (Non-breaking whitespace)


ustych:
I have a problem with outputting html after setContent(). When i add non-breaking space with html entity (nbsp;) Editor replaces all non-breaking spaces to default spaces. I need to preserve that special chars. What can i do in this situation?

litvinchuk:
Same problem
I think it's referring to https://prosemirror.net/docs/ref/#model.ParseOptions.preserveWhitespace
But there is no method to pass this option

Bicazian:
Same Problem.
To have the option added to setContent would be great

mayacoda:
Same issue when passing content to the `Editor` constructor. I'll submit a PR to add passing `ParserOptions` to fix this issue.

philippkuehn:
fixed with #254

zalbastakiNetsells:
Note: the options is `parseOptions` and not `parserOptions` 😛had to look into the diff to figure that one out

PR #254, opened 2019-04-08
allow parseOptions to be passed to setContent and Editor constructor

mayacoda:
As requested in #168, I added the possibility to pass `parserOptions` to both the `Editor` constructor and when calling `.setContent()`.  Passing to the constructor was necessary for loading content that contains whitespaces which need to be preserved.


philippkuehn:
Thanks, Maya!

mayacoda:
Thanks for the speedy merge! Do you know when this will be published on npm?

philippkuehn:
It's published now!

Issue #217, opened 2019-03-02
Improve or Override ComponentView@stopEvent


HollyIT:
This was an entire day of figuring out for a few little line changes. 

Problem:
Inputs in Vue components for NodeView is funky. An example can be seen with the Embeds example. If you make it draggable and selectable, you can't change ht URL. If you drop the selectable, you can drag it and change the URL, BUT if it is the top node and you type or paste into the input field, it creates a paragraph above the iFrame node. No matter what, it seems ProseMirror wanted that paragraph.

Solution:
Figuring out a way to better work stopEvent in the ComponentView. For the project I'm working on, I just abstracted that class out on an as need basis, but it would be nice to find a way to handle that better. I've got a fork of TipTap with my modified stopEvent (and comments) you can view[ here](https://github.com/HollyIT/tiptap/blob/master/packages/tiptap/src/Utils/ComponentView.js#L110). There's also an example page for Captions you can fire up and see how it works. My goal was to get close to the way Wordpress' Gutenberg handles images and captions, and this seems to be getting there. There's even buttons to change the alignment on top.

Off the top of my head, I can't figure out any place where these modifications might hurt, but I'm sure there is. One solution might be to check if the Vue Component has a stopEvent method and defer to that. If it returns TRUE or FALSE, then that is what's returned out of ComponentView@stopEvent. If NULL is returned, then let ComponentView@stopEvent login determine the outcome.

jasonvarga:
I'm finding that I need the same functionality.
I'm trying to make a component `draggable`, but only via a drag handle.

The solution outlined at https://discuss.prosemirror.net/t/make-only-part-of-a-nodeview-draggable/1145/7 says to override stopEvent, but we can't from within our node.

philippkuehn:
Fixed with #257

PR #257, opened 2019-04-09
Allow extensions to override stopEvent

jasonvarga:
This PR addresses #217 

While the modifications linked in that issue describes a solution to fix drag events in general, this PR lets you just override it completely. This way, if you really need to change the behavior, go for it.


philippkuehn:
Thanks, Jason!

PR #272, opened 2019-04-24
add option to disable input/pasterules

### LINKED ISSUE #271: How to disable markdown behaviour and input/paste rules in Tiptap?
Hi,

I have noticed that because of the markdown conversion the copy/paste behaviour is not 100% correct. When I paste for example a file name 'file_name_example.pdf' I get a string without underscores but with one part marked as italic (which is kinda obvious). The thing is, I do not want that or any other paste/input rules because my client simply doesn't need it, but needs an easy to understand copy-paste behaviour instead.

The question is: **how to disable the markdown and input/paste rules in Tiptap?** I have found some ways to do this in ProseMirror, but I have no idea where to put it in my Tiptap component. I have tried without success:
1. setting autoInput to false (as mentioned here: https://prosemirror.net/docs/ref/version/0.3.0.html#autoInput) but have no idea where it should be defined in my tiptap instance
2. manipulate the copy/paste action on handlePaste and transformPasted
3. changing format markdown to false
4. setting the editor pasteRules and inputRules to empty arrays/null onInit

I do not want to edit each extension's js file and I am sure there is a way to set it in editor as a props or something similar, without changing the core code.

None of the ways mentioned above works and I am really running out of ideas how to do it. I am having problem understanding how to use proseMirror props/settings etc. in Tiptap. I will appreciate any hints and answers!

Chrissi2812:
This ads two new options to the editor class to allow input/paste rules to be disabled, by their extension name (`bold`, `italic`, `code_block`, etc.) 

```js
new Editor({
  disabledInputRules: ['bold'],
  disabledPasteRules: ['bold'],
})
```

Fixes #271

_Maybe I'm wrong and this is alredy possible via another setting ([autoInput](https://prosemirror.net/docs/ref/version/0.3.0.html#autoInput)???)_

### DIFF
--- a/packages/tiptap/src/Editor.js
+++ b/packages/tiptap/src/Editor.js
@@ -31,8 +31,8 @@ export default class Editor {
         }],
       },
       useBuiltInExtensions: true,
-      disabledInputRules: [],
-      disabledPasteRules: [],
+      disableInputRules: false,
+      disablePasteRules: false,
       dropCursor: {},
       parseOptions: {},
       onInit: () => {},
@@ -123,14 +123,14 @@ export default class Editor {
   createInputRules() {
     return this.extensions.inputRules({
       schema: this.schema,
-      excludedExtensions: this.options.disabledInputRules,
+      excludedExtensions: this.options.disableInputRules,
     })
   }
 
   createPasteRules() {
     return this.extensions.pasteRules({
       schema: this.schema,
-      excludedExtensions: this.options.disabledPasteRules,
+      excludedExtensions: this.options.disablePasteRules,
     })
   }
 
--- a/packages/tiptap/src/Editor.js
+++ b/packages/tiptap/src/Editor.js
@@ -31,6 +31,8 @@ export default class Editor {
         }],
       },
       useBuiltInExtensions: true,
+      disabledInputRules: [],
+      disabledPasteRules: [],
       dropCursor: {},
       parseOptions: {},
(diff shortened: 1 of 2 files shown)


Chrissi2812:
The build errors shuldn't be there. 
Netlify gives and error over a missing AWS key and Travis fails with service temporary unavailable.

Chrissi2812:
I had to force push this to trigger the checks again, nothing was changed.

philippkuehn:
I'm still not sure if it's *too much* to configure. Maybe a simple `disableInputRules: true` and `disablePasteRules: true` is enough.

Chrissi2812:
Yeah I wasn't sure either, but if someone just wants to disable the markdown rules but allow pasting of links, then why not give them the option 🤔

Now a simple `true` will disable all input rules while `false` allows all, and via `['bold']` you will disable only that specific rule.

gabriel-maus:
Hi, on a project I'm doing I need to disable specific extensions when a user is pasting a text. I first found the issue #643 where @philippkuehn says about disabling certain extensions using a disablePastRules prop, but I didn't found it in the code and after that I found this issue right here.

My question is: is it possible to add the first functionality described on this issue? Adding the possibility to disable specific extensions on paste while keeping the others

Haraldson:
I can’t find any occurrences of `disableInputRules` in this repository, was this removed again?

PR #276, opened 2019-04-29
fix #275: link click handler no longer open on selection click

### LINKED ISSUE #275: Clicking selected paragraph opens a random link in said paragraph
Whenever I select a chunk of text and click on a part of said text it opens one of the links inside the text. This can be very irritating whenever someone acidentally triple-clicks on a chunk and opens a link without the intent to do so.

This issue can be reproduced with the example at [https://tiptap.scrumpy.io/links](https://tiptap.scrumpy.io/links). Just select the whole block and click on a selected part.

If my explanation was unclear I'll be able to provide footage if needed.

Chrissi2812:
This will fix #275


philippkuehn:
Thanks again Chris!

PR #277, opened 2019-05-03
Placeholder Example: replace watcher with v-model only

Chrissi2812:
The watcher is actually unnecessary.

So using v-model with `editor.extensions.options.placeholder.emptyNodeText` directly is far more convenient.


philippkuehn:
Thanks Chris!

PR #279, opened 2019-05-03
Add issue templates

Chrissi2812:
As issues increase it' maybe worth adding some sort of issue templates this are the default issue templates, with some small changes, nothing fancy.


philippkuehn:
Good idea Chris!

Issue #156, opened 2019-01-07
Link extension selection


panayotisk:
There is an issue on the link extension, when clicking on the start of the link. You may check it directly in the online example if you try to put the cursor just before the first "w" of the text "world wide web". The selection is other than the link itself. When I try to use links in my code I get the following error if I click at the exact character at the start of the link (I do not know if you also get a similar error):

> Uncaught RangeError: Position ... out of range
    at Function.resolve (app.js?id=8c53831357ab342c5c52:37971)
    at Function.resolveCached (app.js?id=8c53831357ab342c5c52:37994)
    at Node.resolve (app.js?id=8c53831357ab342c5c52:38265)
    at Plugin.handleClick (app.js?id=8c53831357ab342c5c52:59570)
    at app.js?id=8c53831357ab342c5c52:50478
    at EditorView.someProp (app.js?id=8c53831357ab342c5c52:51728)
    at handleSingleClick (app.js?id=8c53831357ab342c5c52:50478)
    at MouseDown.up (app.js?id=8c53831357ab342c5c52:50610)

Could you check if you have the same issue?
Thank you very much,
Panayotis

philippkuehn:
Can you tell my in which browser you get this error? I do not see an error in Chrome and Firefox.

panayotisk:
I get it in Chrome 71.0.3578.98 (64-bit). The same issue is seen on Firefox too (the selection extends after the link). The error is raised if there is no text after the link.

panayotisk:
It seems the error is caused by trying to select more text and the line 
var $end = doc.resolve(range.to);
in handleClick() causes the error, trying to reach a position that does not exist?

panayotisk:
Summing up from the previous posts:
In order to reproduce the bug go to https://tiptap.scrumpy.io/links and
1) Remove the text after the link. (The text: "By default every link will get a rel="noopener noreferrer nofollow" attribute.")
2) Click **on the 1st** "w" of the link "world wide web" with the Chrome Dev Tools open (or Firefox Web console).

If you do not do step 1 and do only step 2 you will notice that the selection in this case contains 2 x times the number of characters of the link.
The bug can be traced to the range that is returned by the call to getMarkRange() in line 60 of the file Link.js in tiptap-extensions. The function seems to return a double range in this specific case.
@philippkuehn I hope you could check it, since it should be easier for you to find the exact source of the error and provide an appropriate correction. In case you do not have enough time tell me to check it further myself.

[...]

PR #280, opened 2019-05-03
Fix getMarkRange

mariux:
reverting fixes for #156 and trying to fix the detection of endPos.

**ATTENTION**
since i cannot reproduce #156 in the first place please verify that this will not reintroduce the issue from #156..

**what use case gets fixed by this?**
having marks like `<a>aaa<b>bbb</b>ccc</a>` will always select `aaa` even if clicking on `ccc`
expected behaviour: when clicking `ccc` `aaabbbcc` will be selected.


mariux:
how to fix the checks?

philippkuehn:
Netlify was broken for the last days. PR looks good. Thank you!

Issue #225, opened 2019-03-06
isActive.image() always return false


RayPS:
<img width="512" alt="screen shot 2019-03-06 at 19 39 52" src="https://user-images.githubusercontent.com/1941540/53878748-a2d16b80-4047-11e9-9752-c6f2c031a182.png">

I want to hide certain items on menu-bubble when an image is selected.

Because we have no reason to Italic an image.

But `isActive.image()` seems always return false.

Am I missing something?

Alecyrus:
Try to get the selected node's name, like 'text', 'cell', or 'image' that you can use to hide or show menu items dynamically.

Alecyrus:
I think that a better way to implement the context menu for `image` node is to use NodeView (Vue Component) that will enable you to handle more complex things, like resizing, adding image title and so on.

RayPS:
Thanks for the hint!

PR #283, opened 2019-05-04
fix nodeIsActive()

mariux:
this fixes isActive.image() mentioned in #225

nodeIsActive was only checking parent nodes and missed the node
currently selected.


philippkuehn:
Thanks marius!

PR #285, opened 2019-05-06
fix pasteRules and InputRules for italic mark

### LINKED ISSUE #284: transformation <em> when pasting or typing has unexpected results
**Describe the bug**
transformation to `<em>` (italic) is broken when pasting 
 - `test_test_test` it will be transformed to `tes<em>test</em>test` (note the missing `t` of the first `test`
- `_test*` it will be transformed to `<em>test</em>`
- `*test_` it will be transformed to `<em>test</em>`

- paste `test_test_test`, see result.
- paste `_test*` , see result
- paste `*test_` , see result

**Expected behavior**
- pasting `test_test_test` should result in `test<em>test</em>test`
- pasting `*test_` should not be transformed
- pasting `_test*` should not be transformed

mariux:
fixes markPasteRules and markInputRules for italic and bold marks.. fixes #284

### DIFF
--- a/packages/tiptap-commands/src/commands/markInputRule.js
+++ b/packages/tiptap-commands/src/commands/markInputRule.js
@@ -3,23 +3,28 @@ import { InputRule } from 'prosemirror-inputrules'
 export default function (regexp, markType, getAttrs) {
   return new InputRule(regexp, (state, match, start, end) => {
     const attrs = getAttrs instanceof Function ? getAttrs(match) : getAttrs
-        const { tr } = state
-        let markEnd = end
+    const { tr } = state
+    const m = match.length - 1
+    let markEnd = end
+    let markStart = start
 
-    if (match[1]) {
-      const startSpaces = match[0].search(/\S/)
-      const textStart = start + match[0].indexOf(match[1])
-      const textEnd = textStart + match[1].length
-      if (textEnd < end) {
-        tr.delete(textEnd, end)
+    if (match[m]) {
+      const matchStart = start + match[0].indexOf(match[m - 1])
+      const matchEnd = matchStart + match[m - 1].length - 1
+      const textStart = matchStart + match[m - 1].lastIndexOf(match[m])
+      const textEnd = textStart + match[m].length
+
+      if (textEnd < matchEnd) {
+        tr.delete(textEnd, matchEnd)
       }
-      if (textStart > start) {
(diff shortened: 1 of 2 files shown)


philippkuehn:
This will break bold inputrules :(
May-07-2019 07-48-03

mariux:
true true... i will come up with something new..

mariux:
removed the input rule from the pull request so only fixes this on paste for now

mariux:
improved markInputRule to support 2nd match group so we can simulate a lookbehind behaviour which will be supported out of the box by browsers in the future anyway

mariux:
should be fine now even for all edge cases i had think of.. ;)

philippkuehn:
Sorry 😅
May-07-2019 22-26-46

mariux:
;)

packages/tiptap-commands/src/commands/markInputRule.js
```diff
@@ -3,23 +3,28 @@ import { InputRule } from 'prosemirror-inputrules'
...
 
-    if (match[1]) {
-      const startSpaces = match[0].search(/\S/)
-      const textStart = start + match[0].indexOf(match[1])
-      const textEnd = textStart + match[1].length
-      if (textEnd < end) {
-        tr.delete(textEnd, end)
+    if (match[m]) {
+      const matchStart = start + match[0].indexOf(match[m - 1])
+      const matchEnd = matchStart + match[m - 1].length
```
mariux: @philippkuehn actually the issue is here. but i don't get why... because if you add -1 (which i actually had before) when deleting the prefix it is range 4,6 (three positions) and when deleting the suffix it is 8,9 (just two positions) which is weird in my eyes but maybe you can come up   with an explanation i will just force push the -1 again which also had no side effects for my tests....

mariux: okay... the match seems to be handled before the last input character actually reaches the transaction thats why there is this index weirdness at the end of the input...

mariux:
shoudl be fixed by removing 1 char less at the end... see comments in code - the rule is triggered before the last printed char enters the transaction (at least this seems to be the only explanation and also finds some proof in the code..) ... please test again...

Alecyrus:
Seems work now.  [deploy preview](https://deploy-preview-285--scrumpy-tiptap.netlify.com/)

Issue #289, opened 2019-05-06
BubbleMenu and FloatingMenu do not work now.


Alecyrus:
**Describe the bug**
BubbleMenu and FloatingMenu does not work now. I do NOT know what's going on ? When I rewrite my editor, I notice that the basic plugins does not Work. I add some `console.log()` in its `onUpdate()` like the following:
```javascript
// EditorMenuBubble.js 
 return editor.registerPlugin(MenuBubble({
                            element: this.$el,
                            keepInBounds: this.keepInBounds,
                            onUpdate: menu => {
                                console.log("on Update", menu);
                                // the second check ensures event is fired only once
                                if (menu.isActive && this.menu.isActive === false) {
                                    this.$emit('show', menu)
                                } else if (!menu.isActive && this.menu.isActive === true) {
                                    this.$emit('hide', menu)
                                }
                                this.menu = menu
                            },
                        }));
```
However, nothing happens in the output, selecting text do not trigger the menu to pop up. I try many times and can not figure it out. And further,  I found the bubblemenu in  [live demo](https://tiptap.scrumpy.io) does not work too and the floating menu as well. 

**Steps to Reproduce / Codesandbox Example**
Just open [live examples](https://tiptap.scrumpy.io/menu-bubble), and select some text.

**Screenshots**
![1222](https://user-images.githubusercontent.com/13865581/57221544-3f0cf400-7032-11e9-858e-d487f874029f.gif)
![May-06-2019 19-06-47](https://user-images.githubusercontent.com/13865581/57221545-4207e480-7032-11e9-984c-1c6c4b0419b7.gif)

**Environment**
 - OS: macOS Mojave 10.14.3
 - Browse:
     - Chrome Version  73.0.3683.103 && 74.0.3729.131
     - Safari 12.0.3 (14606.4.5)
 - Version: Latest
 - Mobile / Desktop:  Desktop

philippkuehn:
oh no! looking into it.

philippkuehn:
Hmm, this seems to be a bug with ProseMirror :(
You can follow up [here](https://discuss.prosemirror.net/t/state-reconfigure-do-not-apply-new-plugins/1931).

Alecyrus:
Em, and I try this [tooltip](http://prosemirror.net/examples/tooltip/), not work too.

philippkuehn:
Yeah state.reconfigure is not working anymore for view updates. This method is used for adding plugins after initializing the editor.

[...]

PR #290, opened 2019-05-06
fix potential problem that may cause plugins not work.

Alecyrus:
#289


philippkuehn:
hmm, [deploy preview](https://deploy-preview-290--scrumpy-tiptap.netlify.com/menu-bubble) is still broken :(

PR #297, opened 2019-05-07
change deprecated `slot-scope` to `v-slot`

jjangga0214:
Directive `slot-scope` is deprecated as of Vue v2.6.0. See [here](https://vuejs.org/v2/guide/components-slots.html) for details. Currently tiptap depends on Vue ^2.6.10. So this commit introduces new directive `v-slot` on README.md and examples.

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,8 +117,8 @@ The `<editor-menu-bar />` component is renderless and will receive some properti
 
 ```vue
 <template>
-  <editor-menu-bar :editor="editor">
-    <div slot-scope="{ commands, isActive }">
+  <editor-menu-bar :editor="editor" v-slot="{ commands, isActive }">
+    <div>
       <button :class="{ 'is-active': isActive.bold() }" @click="commands.bold">
         Bold
       </button>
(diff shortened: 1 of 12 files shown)


hybridwebdev:
You don't have to change a single thing to be able to use the new v-slot syntax. Just change slot-scope to v-slot and pass in the same exact code. These changes aren't necessary.

jjangga0214:
@hybridwebdev Why? Would you explain the reason you don't want any little single extra change? I believe this change overall improves the code. For example, `slot-scope` was on the root tag of slot before, but I moved `v-slot` to component. It's more cleaner and intuitive. That's why **abbreviated syntax for lone default slot** is made. You can read the official description [here](https://vuejs.org/v2/guide/components-slots.html#Abbreviated-Syntax-for-Lone-Default-Slots).

philippkuehn:
I agree with @jjangga0214. The new syntax is more intuitive. Will merge this.

hybridwebdev:
Changes made add nothing whatsoever.

PR #298, opened 2019-05-07
refactor src/Utils/ExtensionManager.js

jjangga0214:
- remove duplication of logic
- reduce code lines by half (not the entire module, but the part this refactor is applied to)

If preferred, change the name of new functions `apply` and `handle` as you please.


philippkuehn:
Thank you!

PR #300, opened 2019-05-08
refactor src/Utils/Emitter.js

jjangga0214:
- Make flow of logic more compact. It reduces 12 code lines.
- Use `filter` for simple element removal. Unless there are lots of elements, breaking a loop at the middle of iteration doesn't actually have practical benefit, so replaced it with `filter`.
- More flexible. It would be easier to refactor even when argument is changed
to option pack (json).


philippkuehn:
Thank you!

PR #314, opened 2019-05-15
adding option to disable resizing of tables

### LINKED ISSUE #310: Tables documentation
The documentation currently don't cover tables.
I'm trying to disable resizing, but find it unclear if it should be done in TipTap og the Prosemirror-tables plugin? Great work btw! 😍

Chrissi2812:
closes #310


philippkuehn:
Thanks again, Chris!
Now, if there is an option for resizing I would set the default value to `false` (And set it to `true` for the demo). Will merge it after that! 👍

PR #317, opened 2019-05-17
toggle between unordered_list and ordered_list

### LINKED ISSUE #4: toggle between bullet list and ordered list not working
Find a way to toggle bullet lists and ordered lists (and maybe todolists)
aug -27-2018 15-27-29

Chrissi2812:
I'm not 100% sure that this is the way to go 🙈

fixes #4

### DIFF
--- a/packages/tiptap-commands/package.json
+++ b/packages/tiptap-commands/package.json
@@ -25,6 +25,7 @@
     "prosemirror-model": "^1.7.0",
     "prosemirror-schema-list": "^1.0.3",
     "prosemirror-state": "^1.2.3",
+    "prosemirror-utils": "^0.8.1",
     "tiptap-utils": "^1.5.2"
   }
 }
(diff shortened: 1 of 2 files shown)


philippkuehn:
Tough challenge! :D Sometimes there is an issue when using with nested lists :(

May-17-2019 16-08-19

Chrissi2812:
Always good that you find these "edge cases" 
- If empty selection the current list type will change instead of current list_item beeing wrapped by other list type. For nested lists TAB or Command must be used.
- Unwrapping is now fixed (Hopefully in any case, this time).

togglelist

Chrissi2812:
Also if selection contains nested lists only the top level list type will change.

philippkuehn:
Thank you sooo much! But there is another edge case 😅

You can't do anything if there is a list within a blockquote 👀

May-20-2019 13-30-12

Chrissi2812:
Okay have to do more work on this. Currently Todo Lists are broken....

Chrissi2812:
Now you can even have lists within blockqoutes within lists within blockqoutes (Inception 😅)

image

philippkuehn:
Awesome! Looks good so far – can't find any other edge case 😅

PR #324, opened 2019-05-23
add text search and replace

### LINKED ISSUE #224: Highlight search results
It would be great to have an extension that allows to highlight search results or other plain text based features using Decorations, like this example:

https://gist.github.com/marijnh/cee01b30b1890a5947c0

Chrissi2812:
Working on a search and replace plugin:
- [x] search
- [x] clear search
- [x] replace (only next found match will be replaced)
- [x] replace all

Known Bugs:
- [x] ~search hightlight doesn't get updated on text changes~ (fixed in 753fc7632418dd7066062488ef045cf7dc4902c6)
- [x] ~search across marks is not possible (@philippkuehn do you know how that can be achived?)~ (fixed in 331ba1c36b738ece3eb3b03e31747afdb3c71a59)
- [x] ~currently on pressing spacebar the click handler of search button is triggerd (why? 😅)~ (fixed in 036d853378ba8323532d7ed9df5749301f5b4365)
- [x] ~Some searchterms break the editor. (`.*` crashes, `.` breaks emojis) (maybe disable regex at all 🤔)~ (fixed by 0476d355995bacd96a1d2c72a39145148feb5b6c)
- [x] ~Emojis are still broken with some RegExps ...~ (fixed by  (3246b84229ea8492b8f596f216410543d57dbede))

This will close #224

### DIFF
--- a/packages/tiptap-extensions/src/extensions/Search.js
+++ b/packages/tiptap-extensions/src/extensions/Search.js
@@ -25,6 +25,7 @@ export default class Search extends Extension {
       searching: false,
       caseSensitive: false,
       disableRegex: true,
+      alwaysSearch: false,
     }
   }
 
@@ -121,7 +122,7 @@ export default class Search extends Extension {
         state: {
           init() { return DecorationSet.empty },
           apply: (tr, old) => {
-            if (this.options.searching) {
+            if (this.options.searching || (tr.docChanged && this.options.alwaysSearch)) {
               return this.createDeco(tr.doc)
             }
             if (tr.docChanged) {
--- a/packages/tiptap-extensions/src/extensions/Search.js
+++ b/packages/tiptap-extensions/src/extensions/Search.js
@@ -115,26 +115,29 @@ export default class Search extends Extension {
     }
   }
 
-  rebaseNextResult(replace, index) {
+  rebaseNextResult(replace, index, lastOffset = 0) {
     const nextIndex = index + 1
-    if (!this.results[nextIndex]) return
+    if (!this.results[nextIndex]) return null
 
-    const nextStep = this.results[nextIndex]
-    const { from, to } = nextStep
-    const offset = (to - from - replace.length) * nextIndex
+    const { from: currentFrom, to: currentTo } = this.results[index]
+    const offset = (currentTo - currentFrom - replace.length) + lastOffset
 
+    const { from, to } = this.results[nextIndex]
     this.results[nextIndex] = {
       to: to - offset,
       from: from - offset,
     }
+
+    return offset
   }
 
   replaceAll(replace) {
     return ({ tr }, dispatch) => {
+      let offset
       this.results.forEach(({ from, to }, index) => {
         tr.insertText(replace, from, to)
 
-        this.rebaseNextResult(replace, index)
+        offset = this.rebaseNextResult(replace, index, offset)
       })
 
       dispatch(tr)
--- a/packages/tiptap-extensions/src/extensions/Search.js
+++ b/packages/tiptap-extensions/src/extensions/Search.js
@@ -94,7 +94,8 @@ export default class Search extends Extension {
     mergedTextNodes.forEach(({ text, pos }) => {
       const search = this.findRegExp
       let m
-      while (m = search.exec(text)) {
+      // eslint-disable-next-line no-cond-assign
+      while ((m = search.exec(text))) {
         if (m[0] === '') {
           break
         }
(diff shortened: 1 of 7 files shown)


packages/tiptap-extensions/src/extensions/Search.js
```diff
@@ -0,0 +1,109 @@
...
+    return this.decorations ? DecorationSet.create(doc, this.decorations) : []
+  }
+
+  get plugins() {
+    return [
+      new Plugin({
+        state: {
+          init: (_, { doc }) => this.createDeco(doc),
+          apply: (tr, old) => (
+            (tr.docChanged || this.options.searching) ? this.createDeco(tr.doc) : old
```
philippkuehn: this should do the trick for mapping the decorations on text changes:

```
(tr.docChanged || this.options.searching) ? this.createDeco(tr.doc) : old.map(tr.mapping, tr.doc)
```

philippkuehn:
Have you seen this plugin? https://github.com/mattberkowitz/prosemirror-find-replace
I think it's outdated but maybe it will help you with searching across marks?

philippkuehn:
Something like this seems to work for me

```js
const results = []
const search = RegExp('search string', 'g')
const mergedTextNodes = []
let index = 0

doc.descendants((node, pos) => {
  if (node.isText) {
    if (mergedTextNodes[index]) {
      mergedTextNodes[index] = {
        text: mergedTextNodes[index].text + node.text,
        pos: mergedTextNodes[index].pos,
      }
    } else {
      mergedTextNodes[index] = {
        text: node.text,
        pos,
      }
    }
  } else {
    index += 1
  }
})

mergedTextNodes.forEach(({ text, pos }) => {
  let m
  while (m = search.exec(text)) {
    results.push({
      from: pos + m.index,
      to: pos + m.index + m[0].length,
    })
  }
})
```

packages/tiptap-extensions/src/extensions/Search.js
Chrissi2812: Thanks for that hint.

I've implemented it in 753fc7632418dd7066062488ef045cf7dc4902c6.

Chrissi2812:
> Have you seen this plugin? https://github.com/mattberkowitz/prosemirror-find-replace
> I think it's outdated but maybe it will help you with searching across marks?

Yeah that was my first inspiration for this. But it's for prosemirror 0.11 or something I think (Decorations were later introduced) so I used the highlight-example as the base.

> Something like this seems to work for me

Yes that seems to works well for marks.

Chrissi2812:
Emojis now work as of  3246b84229ea8492b8f596f216410543d57dbede 🎉

Before:
image

After: 
image

philippkuehn:
nice! 👍

Chrissi2812:
Okay I now try to implement the replace and replaceAll functions.

This is what I've tried, but this only works if the search and replace term have the same length.
```js
  replaceAll(replace) {
    return ({ tr }, dispatch) => {
      this.results.forEach(({ from, to }) => {
        tr.insertText(replace, from, to)
      })

      dispatch(tr)
    }
  }
```

Calculating the offset manually by subtracting the length diff, seems hacky..

@philippkuehn Can you give me a hint what i can do? (Transforms, Steps, Manually?)

philippkuehn:
@Chrissi2812 I think you have to map all results to its new position after each `insertText`. Like mapping the decorations on each change. It's also what the [deprecated plugin](https://github.com/mattberkowitz/prosemirror-find-replace/blob/master/src/index.js#L327) does.

Chrissi2812:
Tada 🎉
search_replace

As I didn't know how mappings and steps work, I update the results position with an offset. 
If there is a better way please tell me.

Chrissi2812:
ReplaceAll and regexes are not working correctly at the moment if the length of matches is different.
replace_regex_bug

philippkuehn:
This is huge! Thanks again, Chris! Tell me when you are in Berlin. I should buy you a beer.

philippkuehn:
BTW: In the release I removed the `toggleSearch` command because I think this should be solved on the client side.

Chrissi2812:
> This is huge! Thanks again, Chris! Tell me when you are in Berlin. I should buy you a beer.

Thanks for your appreciation 😊 

> BTW: In the release I removed the `toggleSearch` command because I think this should be solved on the client side.

Already noticed,, wasn't sure either how to sepearte it. Was in my mind something like a toggle so the search is updating results while the search bar is open and the document changes. But the `alwaysSearch` can be toggled to so it was kind of redundant.

PR #336, opened 2019-06-01
Update index.vue

iliyaZelenko:
According to this line https://github.com/scrumpy/tiptap/blob/master/packages/tiptap-extensions/src/extensions/Placeholder.js#L46


philippkuehn:
Thanks!

PR #341, opened 2019-06-05
prevent resizehandles in IE 11 on todo_items

### LINKED ISSUE #340: [IE 11] Todo List resize-handles
**Describe the bug**
In IE 11 `todo_items` are resizable. Looks pretty funny though 😂

ie11_todo

**Expected behavior**
No resize handles should be shown

**Environment**
 - OS: Windows
 - Browser Internet Explorer
 - Version 11
 - Mobile / Desktop: Desktop

Chrissi2812:
Accoridng to this [answer](https://stackoverflow.com/a/26101367/6370595) from stackoverflow it helps to apply `contenteditable="false"` to the parent element. 

So this is what this PR does.

fixes #340

### DIFF
--- a/packages/tiptap-extensions/src/nodes/TodoItem.js
+++ b/packages/tiptap-extensions/src/nodes/TodoItem.js
@@ -24,7 +24,7 @@ export default class TodoItem extends Node {
         },
       },
       template: `
-        <li :data-type="node.type.name" :data-done="node.attrs.done.toString()">
+        <li :data-type="node.type.name" :data-done="node.attrs.done.toString()" contenteditable="false">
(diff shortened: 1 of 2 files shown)


philippkuehn:
Weird, pressing enter isn't working anymore 👀

Jun-05-2019 11-45-38

Chrissi2812:
Didn't thought that this could break so much 😨

IE 11 current Enter key behaviour is totally broken. 
ie11_todo_more_bugs

IE11 Also has currently problems with the `dropcursor` plugin , search and replace, suggestions, and the placeholder example. Most of them simply need some polyfills.

*Edge Chromium will hopefully rescue us all, from the folks that are still using IE 11*

#### As I don't know any other solution and this won't work, I close this PR

PR #376, opened 2019-07-05
Handle composition state in MenuBubble

murata0705:
Thank you for great software tiptap!!
Menu Bubble show menu when I select text but I think showing is not needed when I'm selecting Japanese character and I think using browser composing function user agree with me.

bad
bad

good
good

I add small composing handler.
Please check them when you have time.
Thank you.


philippkuehn:
Thank you, Junichi!

PR #382, opened 2019-07-10
Switch check from !isMac to isWindows for specificity.

Aaron3:
Only Windows uses Mod+y, this avoids adding the shortcut to other operating systems (such as Linux or iOS).

This adds better support for spoofed user-agents when testing mobile sites on desktop devices.

You could also add `|| /Pocket/.test(navigator.platform)` but it really is optional


philippkuehn:
Thank you, Aaron!

PR #399, opened 2019-07-23
Fix some overflow's with long words.

afwn90cj93201nixr2e1re:
With this fix:
изображение

Without this fix:
изображение


afwn90cj93201nixr2e1re:
@philippkuehn review.

philippkuehn:
i don't want to add stylingf or selected nodes in this demo for now. the overflow is fixed.

PR #406, opened 2019-07-24
Don't refocus the editor after inserting a mention

### LINKED ISSUE #404: Cursor position is lost after inserting a mention
**Describe the bug**
In the official example, cursor position is reset back to the start after inserting a mention from the suggestions menu.

**Steps to Reproduce**
1. Open the official example: https://tiptap.scrumpy.io/suggestions
1. Place cursor somewhere in the text, eg. at the end of the first paragraph.
2. Type `@` then press Enter to select the first suggestion.
3. Observe that the cursor is moved to the very start of the editor, before the "Suggestions" heading.

**Expected behavior**
The cursor should be placed directly after the newly-inserted mention.

**Environment**
Reproducible on Firefox 68.0.1 Mac and Safari 12.

tobyzerner:
Fixes #404

### DIFF
--- a/examples/Components/Routes/Suggestions/index.vue
+++ b/examples/Components/Routes/Suggestions/index.vue
@@ -207,7 +207,6 @@ export default {
           label: user.name,
         },
       })
-      this.editor.focus()
     },
 
     // renders a popup with suggestions


philippkuehn:
Thanks for your PR but this focus trigger is needed for clicking a dropdown item with your mouse, because then your editor will loose focus. But I fixed #404 :)

PR #419, opened 2019-08-01
Allow Extensions to redefine the behaviour of setSelection

rmckayfleming:
I've been implementing CodeMirror within ProseMirror per the [ProseMirror example](https://prosemirror.net/examples/codemirror/). However, the example redefines the behaviour of setSelection on the NodeView so that it focuses the CodeMirror instance. This PR allows Extensions to define a new setSelection function.

### DIFF
--- a/packages/tiptap/src/Utils/ComponentView.js
+++ b/packages/tiptap/src/Utils/ComponentView.js
@@ -41,6 +41,10 @@ export default class ComponentView {
       updateAttrs: attrs => this.updateAttrs(attrs),
     }
 
+    if (typeof this.extension.setSelection === 'function') {
+      this.setSelection = this.extension.setSelection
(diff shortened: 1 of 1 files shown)


philippkuehn:
thank you!

BrianHung:
@rmckayfleming I'm looking through that ProseMirror implementation of a CodeMirror block. How did you manage to call focus on `this.cm` from `setSelection`? Did you choose to initialize it within `get view()` or within the Node constructor? Thanks.

rmckayfleming:
@BrianHung this.cm is initialized from within the custom vue component.

BrianHung:
To access the CodeMirror instance in the custom Vue component from setSelection, did you end up using `this.vm.cm` then?

PR #459, opened 2019-09-24
MenuBubbles below the selection

rmckayfleming:
This adds a top property to the menu so that you can have the MenuBubble render below the selection instead of above it. Set `top: ${menu.top}` in the MenuBubble styles to render below.


philippkuehn:
ah yeah. great idea. thank you, ryan!

PR #474, opened 2019-10-02
set selection to first cell after table insert

### LINKED ISSUE #447: Selection is not correct after creating a table
**Describe the bug**
After creating a table, the selection is focused into the last cell, rather than the first. Even worse, when there is content after the table, the selection is at the position. This bug is causing issues for me trying to call `findTable` on the selection after inserting a table and returning null since the selection is not in the table.

**Steps to Reproduce**
1. Add some text to the editor.
2. Add a new line above text.
3. Insert a table in the new line above text.
4. Selection is on the line of text below the table.

**Expected behavior**
For the selection to be in the first cell of the newly created table and `findTable` to work on the selection after `createTable` has been called.

**Screenshots**
Screenshot

**Environment**
- OS: Win 10
- Browser: Chrome 76
- Version: 1.25.0
- Mobile / Desktop: Desktop

Chrissi2812:
fixes #447

### DIFF
--- a/packages/tiptap-extensions/src/nodes/Table.js
+++ b/packages/tiptap-extensions/src/nodes/Table.js
@@ -19,6 +19,7 @@ import {
   fixTables,
 } from 'prosemirror-tables'
 import { createTable } from 'prosemirror-utils'
+import { TextSelection } from 'prosemirror-state'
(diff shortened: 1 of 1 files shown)


philippkuehn:
thank you!

PR #475, opened 2019-10-02
add support for empty selection to removeMark and updateMark

### LINKED ISSUE #449: Remove link under cursor when no text is selected
Using the menu bubble component, I am trying to build a UI for editing/remove links similar to Google Docs. The desired behaviour is that when you place the cursor inside a link, the menu bubble shows with the link's URL, and controls to edit/remove the link.

link

I can get halfway there.

First, I overrode the Link extension to remove the ProseMirror plugin that causes the URL to open when you click on a link mark (I think that is not a great choice for the default behaviour, at the very least it should be configurable, or maybe require a modifier key like Ctrl to be pressed).

Second, rather than using `menu.isActive` to show or hide the menu bubble, if I just use `isActive.link()` I can get the menu bubble to display when the cursor is inside a link, even without a text selection:

```html
<editor-menu-bubble :editor="editor" v-slot="{ commands, isActive, getMarkAttrs, menu }">
  <div
    class="tooltip richtext-link"
    :class="{ 'is-active': isActive.link() }"
    :style="{ left: `${menu.left}px`, bottom: `${menu.bottom}px` }"
  >
    <a :href="getMarkAttrs('link').href" target="_blank">{{ getMarkAttrs('link').href }}</a>
    <button type="button" class="button tooltip-button" @click="commands.link({})">Remove</button>
  </div>
</editor-menu-bubble>
```

However, clicking the Remove button without a text selection, it does nothing.

If the link command is executed with `{ attr: null }` without a text selection, I would like any link marks that the cursor is within to be removed.

Chrissi2812:
This will allow us to change marks under cursor without to select it first.
fixes #449

Only downside as far as I can tell is that without selection and running a command like `commands.link()` will link the currently nearest node.

### DIFF
--- a/packages/tiptap-commands/src/commands/removeMark.js
+++ b/packages/tiptap-commands/src/commands/removeMark.js
@@ -1,6 +1,17 @@
 export default function (type) {
   return (state, dispatch) => {
-    const { from, to } = state.selection
+    const { empty, $cursor } = state.selection
+    let { from, to } = state.selection
+
+    if (empty && $cursor) {
+      const { parent, pos, textOffset } = $cursor
+      const parentOffset = pos - textOffset
+      const { node: { nodeSize } } = parent.childAfter(parentOffset)
+
+      from = parentOffset
+      to = from + nodeSize
+    }
+
     return dispatch(state.tr.removeMark(from, to, type))
   }
 }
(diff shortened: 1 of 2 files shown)


philippkuehn:
Hey Chris,

this is a great idea but there is already a getMarkRange function for this, so I used it instead. https://github.com/scrumpy/tiptap/commit/08aa7e88bc546eb9789d9d8df321f990fd04a576
There is also no problem with running a command without a selection anymore.

Oct-04-2019 14-14-36

PR #476, opened 2019-10-02
Allows to disable the click handler and modify a modifier key as function

Chrissi2812:
Im not 100% sure if the modifier should really be a function. It accepts either a function or a boolean `false` to disable it entirely.

*P.S Sorry for spamming but I didn't want to include it in #475 as it not directly related*

Related to #449


philippkuehn:
Thank you for this, Chris!

I don't want to change the default behavior so the the default is still `openOnClick: true`. I also renamed `openClick` to `openOnClick`. A modifier key is too custom so I removed it.

PR #478, opened 2019-10-03
fix(menu-bubble-hide): only send hide update, target not child of editor

### LINKED ISSUE #473: relatedTarget cancels blur
**Describe the bug**
Using the `EditorMenuBubble` when I select some text and focus outside of the editor the menu bubble remains visible 

1. Go to https://codesandbox.io/embed/vue-issue-template-xi4ef
2. highlight editor text
3. click on input above editpr
4. menu bar still shown

Codesandbox
https://codesandbox.io/embed/vue-issue-template-xi4ef

**Expected behavior**
the menu bubble hide

**Environment**
 - OS: Mac
 - Browser Brave
 - Desktop

**Additional context**
seems to be this bit of code which checks for relatedTarget (anything focusable) and bails out if it exists
https://github.com/scrumpy/tiptap/blob/24f61e02be0ed091f4735266df03433ceeda119c/packages/tiptap/src/Plugins/MenuBubble.js#L64-L71

chrisjbrown:
closes #473

possible solution for menu bubble not hiding when you focus away to a control element (input, etc.)

check the parent node of the menu bubble.
usually a container that contains both the menu and the editor-content
if the relatedTarget (where we are focusing to) is a child of the parent, we don't hide
if it is not we can hide the bubble menu

### DIFF
--- a/packages/tiptap/src/Plugins/MenuBubble.js
+++ b/packages/tiptap/src/Plugins/MenuBubble.js
@@ -143,7 +143,7 @@ class Menu {
   }
 
   hide(event) {
-    if (event && event.relatedTarget) {
+    if (event && event.relatedTarget && this.options.element.parentNode.contains(event.relatedTarget)) {
       return
     }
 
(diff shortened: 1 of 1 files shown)


philippkuehn:
thank you, chris!

PR #502, opened 2019-10-23
Fixed InputRule deletes the next unrelated char that should not be deleted.

Alecyrus:


philippkuehn:
Can I reproduce the bug somehow? (before your fix)

Alecyrus:
Dec-03-2019 10-33-21

PR #511, opened 2019-11-01
Add support for historical paths

fearclear:
I realized that tiptap does not have support for historical paths, and Prosemirror has implemented this feature.Used to control the state of the button.
image


iliyaZelenko:
@fearclear that is, it will be possible to know whether to make the button `disabled` or not? (if there is no further history)

fearclear:
@iliyaZelenko  yes, we can use `commands.undoDepth()` to confirm whether the back button is available

PR #572, opened 2019-12-31
Trim suggestion matches to cursor position

snickbit:
Otherwise it will match to the end of the node, possibly matching much more than needed.


hanspagel:

hanspagel:
Thanks for your work on this @snickbit! I fixed the linting errors and did have a look at it, but I can’t reproduce the issue you’re describing, so I can’t validate the fix.

Anyway, we’ll rewrite the extension for tiptap v2, so I made sure we have this issue on our list to test against. Thank you! 🙌

PR #576, opened 2020-01-06
render(h) function support for TodoItem

ryanbliss:
This is in reference to https://github.com/scrumpy/tiptap/issues/69, which is closed but is still relevant for any project that requires a strict CSP (content security policy), such as a Chrome Extension or many enterprise apps.

Having both a render function and the template is required, as having only a render(h) causes errors with converting the node to JSON. It also makes the render(h) function more legible, as you have something to reference when reading the render(h) function. 

In that file, I added a comment that should further document why the render(h) function is embedded.

### DIFF
--- a/packages/tiptap-extensions/src/nodes/TodoItem.js
+++ b/packages/tiptap-extensions/src/nodes/TodoItem.js
@@ -60,7 +60,7 @@ export default class TodoItem extends Node {
             },
             ref: 'content',
           }),
-        ]);
+        ])
       },
     }
   }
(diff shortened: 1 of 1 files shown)


SanterreJo:
This PR will allow us to only include vue.runtime in our build. Saving about 30kb. Thank you 👍

hanspagel:
Thank you, that’s a great addition!

hanspagel:
Oh, had to remove the code again, as it duplicates the template. Sorry. 

We’ll extract the code as .vue files, so they will be rendered on build process already.

hanspagel:
Add an Issue to do this #766

PR #591, opened 2020-01-25
improvement(History extension): support ru keys

iliyaZelenko:
I am 100% sure that this combination should work. This will make it possible to manage the history on the Russian keyboard.


hanspagel:
Thanks!

PR #593, opened 2020-01-27
Custom matcher of SuggestionsPlugin 

soranoba:
The SuggestionPlugin is very easy to use, but there is no other way to duplicate the Suggestion Plugin when we want to customize the match process.
For example, I want to use a matcher that matches even if there is no space immediately before, like this:

```js
function triggerCharacter({ char }) {
  return $position => {
    if ($position.depth <= 0) {
      return false
    }

    const escapedChar = `\\${char}`
    const text = $position.doc.textBetween($position.before(), $position.end(), '\0', '\0')
    const match = new RegExp(`${escapedChar}[^\\s${escapedChar}]*$`, 'gm').exec(text)
    if (match === null) {
      return null
    }

    const from = match.index + $position.start()
    var hasNode = false
    $position.doc.nodesBetween(from, $position.end(), (node) => {
      if (!(node.isTextblock || node.isText)) {
        hasNode = true
      }
    })

    if (hasNode) {
      return null
    }

    return {
      range: {
        from: from,
        to: $position.end(),
      },
      query: match[0].slice(char.length),
      text: match[0],
    }
  }
}
```

This PR allows us specify a function as a matcher from outside.

### DIFF
--- a/packages/tiptap-extensions/src/plugins/Suggestions.js
+++ b/packages/tiptap-extensions/src/plugins/Suggestions.js
@@ -68,6 +68,7 @@ function triggerCharacter({
 }
 
 export default function SuggestionsPlugin({
+  // matcher allow a function
   matcher = {
     char: '@',
     allowSpaces: false,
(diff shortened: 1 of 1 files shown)


hanspagel:

hanspagel:
Oh, this is a great feature! Thanks for the PR @soranoba!

We don’t plan to add features to tiptap v1 as [tiptap v2 is around the corner](https://blog.ueber.io/post/our-plan-for-tiptap-2/). Anyway, I think we’ll add this feature to the suggestion extension, which we are about to rewrite. So thank you very much! 💖

PR #601, opened 2020-02-05
feat(mention): add class attr to mentions multiple mention classes

chrisjbrown:
this PR adds a class attribute to the mention node

this allows developers to pass a class when inserting a suggestion
giving the mention element a custom class if desired 

otherwise fallback to the class passed into the options object

example use  
```js
    selectSuggestion(suggestion) {
      this.insertSuggestion({
        range: this.suggestionRange,
        attrs: {
          id: suggestion.key,
          label: `@${suggestion.label}`,
          class: condition ? this.oneclass : this.otherclass
        }
      })
    },
```

### DIFF
--- a/packages/tiptap-extensions/src/nodes/Mention.js
+++ b/packages/tiptap-extensions/src/nodes/Mention.js
@@ -25,7 +25,9 @@ export default class Mention extends Node {
       attrs: {
         id: {},
         label: {},
-        class: '',
+        class: {
+          default: this.options.mentionClass,
+        },
       },
       group: 'inline',
       inline: true,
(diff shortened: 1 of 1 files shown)


chrisjbrown:
Any interest in this? :D
This change is backwards compatible

hanspagel:

hanspagel:
Thanks for your suggestion @chrisjbrown! We don’t plan to add new features to tiptap v1, as v2 is already around the corner. We’re currently working on rewriting the suggestion extension for v2 and likely consider your suggestion to make it to the core in v2. So thank you very much! 🙌

PR #614, opened 2020-02-25
Allow to manually set the active state on the EditorMenuBubble

juliushaertl:
This change allows to manually overwrite the active state of the EditorMenuBubble component, so that it can be shown based on conditions provided by the user.

A common use case is shown in the Link example, where you don't open a link directly on click, but show the menu bubble as a hint to show the URL before actually opening the link.

I've updated the link example to use that since I was unsure if it makes sense to actually add another extra example page.

image

### DIFF
--- a/examples/Components/Routes/Links/index.vue
+++ b/examples/Components/Routes/Links/index.vue
@@ -1,6 +1,6 @@
 <template>
   <div class="editor">
-    <editor-menu-bubble class="menububble" :editor="editor" @hide="hideLinkMenu" v-slot="{ commands, isActive, getMarkAttrs, menu }">
+    <editor-menu-bubble class="menububble" :editor="editor" @hide="hideLinkMenu" :is-active-callback="({ isActive }) => isActive.link()" v-slot="{ commands, isActive, getMarkAttrs, menu }">
       <div
         class="menububble"
         :class="{ 'is-active': menu.isActive }"
@@ -15,6 +15,13 @@
         </form>
 
         <template v-else>
+          <a
+            class="menububble__link"
+            v-if="isActive.link()"
+            :href="getMarkAttrs('link').href"
+          >
+            {{ getMarkAttrs('link').href }}
+          </a>
           <button
             class="menububble__button"
             @click="showLinkMenu(getMarkAttrs('link'))"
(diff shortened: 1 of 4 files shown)


kfirba:
@juliushaertl Your solution would also always show the bubble menu when the user just uses the arrow keys to navigate. When the cursor is within a link range, it would show. I think that ideally, it should only be when the user clicks the link as it shows a user intent.

I've actually implemented this in my own project by simply overriding the default `MenuBubble.js` file. I needed more fixes regarding the positioning of the bubble (my editor is RTL so needed it to open from the right side of the editor) and wanted the bubble not to overflow the window (and actually correct its position). I've also added a "tail" for the bubble menu (which fixes its position when the bubble menu should've overflowed the window).

Anyways, just to fix the "show only on click" issue, you can easily do that by adding a `mousedown` and `mouseup` event handlers in the plugin itself. The way it works is that `mousedown` fires first, then `update` and only then `mouseup` so you can set a flag on `mousedown`, check for the flag in `update` and reset the flag in `mouseup`:

```js
export default function(options) {
  return new Plugin({
    key: new PluginKey('menuBubble'),
    view(editorView) {
      return new Menu({editorView, options});
    },
    props: {
      handleDOMEvents: {
        mousedown: (view) => {
          justClicked = true;
        },
        mouseup: (view) => {
          justClicked = false;
        },
      },
    },
  });
}
```

Just my opinion, tho

juliushaertl:
Thanks a lot for that hint, that seems like a nice solution as well. I'll give that a try.

hanspagel:

anna-holmes:
Any update on this? This would be super useful for my team as well.

juliushaertl:
I didn't have time to look into this again yet but still on my list of things to revisit.

hanspagel:
Thanks @juliushaertl for your work on this! We’re about to rewrite all interface components for the upcoming [tiptap v2](https://blog.ueber.io/post/our-plan-for-tiptap-2/), so I’m closing this here. ✌️

juliushaertl:
Yes, sorry this apparently lost priority on my side ;)

I'm looking forward to what is coming with tiptap v2. Thanks for all your hard work :tada:

PR #658, opened 2020-04-08
Change registerPlugin to add plugin after Extensions plugins

BrianHung:
This commit changes `registerPlugin` by 1. adding the new plugin to `this.plugins` and 2. updating `this.plugins` within `this.state.plugins` by splicing with the new `this.plugins` length. Previously, new plugins were simply added to the end of `this.state.plugins` and were not appended to `this.plugins`.

Screen Shot 2020-04-08 at 3 27 19 AM

By placing registered plugins at this new location within `this.state.plugins`, we prioritize registered plugins over default ProseMirror plugins such as `keymap(baseKeymap)`. This will allow new plugins to have precedence over props such as `handleKeyDown`. A use case of this defining a floating Editor Menu that's navigable with arrow and enter keys.

### DIFF
--- a/packages/tiptap/src/Editor.js
+++ b/packages/tiptap/src/Editor.js
@@ -487,14 +487,10 @@ export default class Editor extends Emitter {
       }), {})
   }
 
-  registerPlugin(plugin = null) {
-    if (!plugin) {
-      return
-    }
-
-    this.plugins.push(plugin);
-    this.state.plugins.splice(this.plugins.length, 0, plugin);
-    const newState = this.state.reconfigure({plugins: this.state.plugins});
+  registerPlugin(plugin = null, handlePlugins) {
+    const plugins = typeof handlePlugins === 'function'
+      ? handlePlugins(plugin, this.state.plugins) : [...plugin, this.state.plugins]
+    const newState = this.state.reconfigure({ plugins })
     this.view.updateState(newState)
   }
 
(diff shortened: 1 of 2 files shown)


packages/tiptap/src/Editor.js
```diff
@@ -492,9 +492,9 @@ export default class Editor extends Emitter {
       return
     }
 
-    const newState = this.state.reconfigure({
-      plugins: this.state.plugins.concat([plugin]),
-    })
+    this.plugins.push(plugin);
+    this.state.plugins.splice(this.plugins.length, 0, plugin);
+    const newState = this.state.reconfigure({plugins: this.state.plugins});
```
philippkuehn, line 497: Should be the same with `[plugin, ...this.state.plugins]`, right? 🤔

BrianHung: Pushing new plugins to the front of `this.state.plugins` could also work, as long as they're in-front of the default ProseMirror plugins.

An alternative to accommodate both of these solutions would be to modify parameters of `registerPlugin`, and to allow a helper function to be passed in, e.g.
```js
  registerPlugin(plugin, helper) {
    let plugins = typeof helper == 'function' ? function(plugin, this.state.plugins) : [...plugin, this.state.plugins];
    const newState = this.state.reconfigure({plugins});
    this.view.updateState(newState)
  }
```

This would allow custom behavior for where new plugins should be registered within `this.state.plugins`. An example is if plugins had to be registered in a certain order. Thoughts?

philippkuehn: This is great! Lets do this!

BrianHung: Ok, I've updated the PR to allow a function to be passed in. :)

PR #694, opened 2020-05-10
fixed issue of user object format coming from search() of fuse.js

DebaPrasad14:
fix:
When someone types something after '@' like @P then nothing shows due user object format returning from search() of fuse.js as below. And now after this fix, it is showing correctly.
@philippkuehn - please have a look at it.

fix of the issue created - https://github.com/scrumpy/tiptap/issues/693

image

### DIFF
--- a/examples/Components/Routes/Suggestions/index.vue
+++ b/examples/Components/Routes/Suggestions/index.vue
@@ -134,7 +134,12 @@ export default {
                 keys: ['name'],
               })
 
-              return fuse.search(query)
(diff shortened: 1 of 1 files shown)


oliviermills:
Fuse.js returns `[ { item : {.. }, refIndex : ...} ...]` not ` [ {...}, {...} ]` , so we need to return just the item prop from the fuse search results, for tippy to have the correct object: `return fuse.search(...).map((it)=> it.item`

philippkuehn:
no need for this because of #750

PR #705, opened 2020-05-21
Add options to enable / disable dropCursor and gapCursor

christosnc:
The title says it all :)

It's very important that tiptap **suggests**, but doesn't **enforce** plugins, so optional enable / disable parameters must exist.


hanspagel:
Thanks for the PR! 👍  Will be part of the next version.

Issue #724, opened 2020-06-04
XSS issue when importing through getHTML() function


vido4:
**Describe the bug**
When using the editor and importing data which was stored as raw HTML string (usually by getHTML() function) it is possible to inject javascript code there and create a stored XSS in the application using the editor.

The code that is causing the issue seems to be that one [Editor.js:274-276]
```
if (typeof content === 'string') {
      const element = document.createElement('div')
      element.innerHTML = content.trim()
```

Most javascript code cannot be injected through .innerHTML property, but by using events of `<img>` tag it is possible. Simplest payload that can be used is `<img src=x onerror=alert(1)>` 

The code is executed in the moment of assigning to innerHTML property and only then passed on to the parser which properly sanitizes input.

**The solution:**
The simplest one seems to be creating element not in context of a document. That solution was already proposed in #713  which is directly related to this issue and should fix both of them.

jnguyen32:
Is this project not being maintained? Surprised there is no response to this.

jnguyen32:
```
import {Editor} from 'tiptap';
import {DOMParser as ProseMirrorDOMParser} from 'prosemirror-model';

export default class MyEditor extends Editor {
    createDocument(content, parseOptions = this.options.parseOptions) {
        if (content === null) {
            return this.schema.nodeFromJSON(this.options.emptyDocument)
        }

        if (typeof content === 'object') {
            try {
                return this.schema.nodeFromJSON(content)
            } catch (error) {
                console.warn('[tiptap warn]: Invalid content.', 'Passed value:', content, 'Error:', error)
                return this.schema.nodeFromJSON(this.options.emptyDocument)
            }
        }

        if (typeof content === 'string') {
            const htmlString = `<div>${content}</div>`;
            const parser = new DOMParser;
            const element = parser.parseFromString(htmlString, "text/html").body.firstChild;
            return ProseMirrorDOMParser.fromSchema(this.schema).parse(element, parseOptions)
        }

        return false
    }
}
```

philippkuehn:
fixed in https://github.com/scrumpy/tiptap/pull/747

Issue #713, opened 2020-05-26
Stop tiptap from pre-fetching all images when initialized with HTML content


kfirba:
**Is your feature request related to a problem? Please describe.**
Given we are using tiptap as a WYSYWG editor feeding it HTML and exporting HTML out of it, there is an **major** issue with the way `tiptap` parses the initial HTML string.

Given we have the following HTML:

```html
<p>
    Text <img src="foo.jpg">
    Text <img src="bar.jpg">
</p>
<figure>
    <img src="baz.jpg">
</figure>
```

When tiptap passes this string to `ProseMirror`'s parser, it will wrap the string in a `<div>` to ensure a single root element. The wrapping is very trivial, it uses `div.innerHTML = content` to do so. However, this is where the issue lays.

For some reason (unknown to me) when we create an in-memory element and load in `img` in it, the browser attempts to fetch the image! The `div` is not in the DOM in any way, nor the `img` but the browser still attempts to load the image. Given the HTML structure we have above, when `tiptap` calling `div.innerHTML = content` in its `createDocument` method, it will actually trigger 3 network requests which is a big no-no especially when you may have over 60 images in that string (in my case it may be over 100 images...)

**Describe the solution you'd like**
The best solution I've found for now is quite interesting. We can use the browser's native [DOMParser](https://developer.mozilla.org/en-US/docs/Web/API/DOMParser) to get an HTML node without fetching any image:

```js
const htmlString = `<div>${content}</div>`;
const parser = new DOMParser;
const element = parser.parseFromString(htmlString, 'text/html').body.firstChild;
// pass the div to ProseMirror's parser at this point...
return DOMParser.fromSchema(this.schema).parse(element, parseOptions)
```

I've also opened a question about this in Stackoverflow, just in case anyone is interested: https://stackoverflow.com/questions/62019538/preventing-images-from-being-downloaded-when-they-are-inside-an-in-memory-elemen

I hope this can be fixed soon! If you are willing to fix this, I wouldn't mind creating a PR or you could just edit 3 lines in `Editor.js` at `createDocument` method.

Thanks!

philippkuehn:
fixed in https://github.com/scrumpy/tiptap/pull/747

kfirba:
@philippkuehn please see my comment on #747

thesiti92:
[...]

PR #747, opened 2020-07-01
XSS bug fix issue #724: XSS issue when importing through getHTML() function 

jnguyen32:
Referencing issue: 

https://github.com/scrumpy/tiptap/issues/724

Solution proposed here:

https://github.com/scrumpy/tiptap/issues/713

**Describe the bug**
When using the editor and importing data which was stored as raw HTML string (usually by getHTML() function) it is possible to inject javascript code there and create a stored XSS in the application using the editor.

The code that is causing the issue seems to be that one [Editor.js:274-276]

```
if (typeof content === 'string') {
      const element = document.createElement('div')
      element.innerHTML = content.trim()
```

Most javascript code cannot be injected through .innerHTML property, but by using events of  tag it is possible. Simplest payload that can be used is ``

The code is executed in the moment of assigning to innerHTML property and only then passed on to the parser which properly sanitizes input.

**The solution:**
The simplest one seems to be creating element not in context of a document. That solution was already proposed in #713 which is directly related to this issue and should fix both of them.

In this pull request, the issue is resolved by passing the HTML string through JavaScript's native DOMParser.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.history
 .DS_Store
 node_modules
 dist/
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,20570 @@
+{
+  "name": "tiptap-packages",
+  "requires": true,
+  "lockfileVersion": 1,
+  "dependencies": {
+    "@babel/code-frame": {
+      "version": "7.10.4",
+      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.10.4.tgz",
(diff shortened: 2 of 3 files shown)


kfirba:
@jnguyen32 @philippkuehn This PR actually **breaks** the editor!

This PR implements my suggestion at #713  however, it left out a crucial part where we get the parsed DOM first element as the root element.

This PR parses the HTML string using the browser's native DOMParser and passes the body as the root DOM node. However, in order to parse the HTML string, we wrapped the content in a single `div` root node. This makes Tiptap break on certain occasions when using custom extensions. I have yet to understand why it happens, but it seems like that just passing the `body.firstElementChild` as the root DOM solves this issue effortlessly.

```js
// change this line:
const element = parser.parseFromString(htmlString, "text/html").body;

// to this:
const element = parser.parseFromString(htmlString, "text/html").body.firstElementChild;
```

PR #767, opened 2020-07-20
Added comment about data-drag-handle.

### LINKED ISSUE #481: Adding data-drag-handle doesn't always work
**Describe the bug**
On a simple component, like the [drag handle example](https://tiptap.scrumpy.io/drag-handle), adding `data-drag-handle` makes dragging only work for that element.

On something more complicated, it doesn't.

1. Go to https://codesandbox.io/embed/simplified-tiptap-editor-example-5s2k3
2. Drag one of the "Set of fields" blocks from anywhere.

**Expected behavior**

Dragging should only work when dragging from the handle (the darker box on the top right with the `=`)

**Additional context**

I know that something could probably be done using `stopEvent`, but I can't work out what.
I tried something along those lines, but it would also affect events within the nested inputs.

Greenscreener:
This partially resolves\ #481.

### DIFF
--- a/examples/Components/Routes/DragHandle/DragItem.js
+++ b/examples/Components/Routes/DragHandle/DragItem.js
@@ -17,7 +17,7 @@ export default class DragItem extends Node {
       }],
     }
   }
-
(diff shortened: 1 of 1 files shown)


Greenscreener:
The code sample within this very file should work for you.

PR #772, opened 2020-07-27
Fix an issue with with parsing custom extensions

kfirba:
see details in #747:

>This PR implements my suggestion at #713 however, it left out a crucial part where we get the parsed DOM first element as the root element.

>This PR parses the HTML string using the browser's native DOMParser and passes the body as the root DOM node. However, in order to parse the HTML string, we wrapped the content in a single div root node. This makes Tiptap break on certain occasions when using custom extensions. I have yet to understand why it happens, but it seems like that just passing the body.firstElementChild as the root DOM solves this issue effortlessly.

### DIFF
--- a/packages/tiptap/src/Editor.js
+++ b/packages/tiptap/src/Editor.js
@@ -274,7 +274,7 @@ export default class Editor extends Emitter {
     if (typeof content === 'string') {
       const htmlString = `<div>${content}</div>`
       const parser = new window.DOMParser()
-      const element = parser.parseFromString(htmlString, 'text/html').body
+      const element = parser.parseFromString(htmlString, 'text/html').body.firstElementChild
       return DOMParser.fromSchema(this.schema).parse(element, parseOptions)
     }
 


hanspagel:
Thanks kfirba! Looks good to me!

kfirba:
@hanspagel Would be GREAT if this could be tagged soon as our editor is breaking in production and we have no reliable way of fixing it :)

kfirba:
@hanspagel After further investigation, I know exactly what the issue is.

We used to use CKeditor as our rich content editor. Since we wanted something more "custom", we decided to use tiptap for the job. However, we wanted to support HTML markups that had been generated up until that point as well. In order to do so, we created many custom extensions and pretty much made an override for every base plugin tiptap has. One of the things we needed to support was parsing `div` tags as simple paragraphs.

Since #747 passes the `body` as the root dom node, prosemirror parses the newly added root `div` element as a paragraph which then it invalidates our custom extensions since they are not allowed inside a paragraph.

Since the added `div` is completely useless and serves no purpose other than having a root node that "holds" the document, we can safely merge this PR. This PR introduces **no breaking changes**. In fact, #747 introduces a huge breaking change where anyone who made some custom node that uses the `div` tag will now get incorrect results.

hanspagel:
Thanks for clarification! Merged the fix and tagged tiptap 1.29.3. 🙌

kfirba:
@hanspagel Thanks

PR #860, opened 2020-10-15
fix missing username when copy paragraph contains mentions

Priestch:


hanspagel:
So cool, thank you! 🙌

hanspagel:
Tagged and published as tiptap-extensions 1.33.2.

PR #861, opened 2020-10-16
fix can not copy content of new inserted mention

Priestch:


hanspagel:
Thanks for the follow up PR, great work!

It changes the behaviour of `replaceText` and would break the app for people using it already. Any idea how to fix that without a breaking change? If not, no worries, I can have a look next week.

Priestch:
@hanspagel Maybe add another default parameter for the `replaceText` function, and create node content with attrs and the new added argument when called.

hanspagel:
Sounds good!

Priestch:
@hanspagel I will send another pr later.

Priestch:
@hanspagel I pushed a new commit to fix the compatible problem.

USasikumar:
@Priestch Which version is the issue of mentioning multiple mentions in the same paragraph fixed? I have installed tiptap through npm install 

This is my current tiptap version

 **"tiptap": "^1.30.0",
 "tiptap-extensions": "^1.33.2",**

please tell me how to fix the issue

Priestch:
@USasikumar This pr is still open, so the multiple mentions problem still there, but you can have a look the last commit above.

USasikumar:
> @USasikumar This pr is still open, so the multiple mentions problem still there, but you can have a look the last commit above.

@hanspagel  can you review and push the lastest commit if it solves the multiple mentions in the same paragraph issue

andreasvirkus:
@hanspagel could you take a look whenever possible? 🙏

hanspagel:
Thanks for the work on this and the mention here! Sorry for the long wait. Life hit me and I hope to be able to look at it in the first week of January. 😏 It’s #1 on my list for the new year.

USasikumar:
> Thanks for the work on this and the mention here! Sorry for the long wait. Life hit me and I hope to be able to look at it in the first week of January.  It’s #1 on my list for the new year.

@hanspagel 
Please look into this on priority my entire application is stuck because of this issue. We are critically behind client deadlines.

andreasvirkus:
@USasikumar you can create a local workaround with custom extensions until this is resolved. See https://github.com/ueberdosis/tiptap/issues/874#issuecomment-742236801

hanspagel:
Thanks @Priestch! Looks pretty good to me. :) I start to prepare the next release and merge it. 

Recent releases introduced too many bugs (like this one here), so I’ll take some time before publishing, but expect it in the first two weeks of January.

And thanks to all others helping out here and in the issues, great work! 👏

Issue #868, opened 2020-10-20
Nested content islands using Vue component slots and Prosemirror Sub-views


enjibby:
**Is your feature request related to a problem? Please describe.**
While the document-schema structure of ProseMirror provides excellent flexibility and Vue components allow visual and interactive customisation, there is little to support a rich Vue-component tree structure due to the limitations of prosemirror's contenteditable implementation. @marijnh articulated this in [this comment](https://github.com/ProseMirror/prosemirror/issues/745#issuecomment-372384609) when he said that ProseMirror could not support contenteditable islands.

He does, however, point to [an example](https://prosemirror.net/examples/footnote/) where content inside a descendant element of a ineditable section could be made editable using a new instance of prosemirror-view along with a dispatch function to proxy events between the parent and child view.

**Describe the solution you'd like**
I would really like if TipTap could leverage this functionality, along with Vue component slot elements, to allow Vue component editing to result in a rich, nested document indicative of a traditional SGML structure.

I have done some precursor work (which I will soon PR) to pull in the code from the aforementioned example into the `ComponentView` class to instantiate a prosemirror-view instance inside an element of the component (similar to how $refs.content currently works) but ultimately it needs to work much better.

I imagine the `ComponentView` instantiation could detect the slot elements in the new `vm` instance and hook each of them up as a special slot Node, containing a `prosemirror-view` instance with the content provided by the component node's parseDOM().

E.g. the Tiptap Node will expect the DOM input and output to be in a format:

```vue
<staff-list>
  <template v-slot:default="slotProps">
    <h3>Our Staff</h3>
    <p>We are all beautiful!</p>
  </template>
  <template v-slot:staff-members="slotProps">
    <staff-member name="Mr En Jibby" avatar="enjibby-grinning.jpg">
      <template v-slot:default="slotProps">
        <p>Chief Speculative Programming Officer<p>
      </template>
    </staff-member>
  </template>
</staff-list>
```

Perhaps the document schema would have to include:

```javascript
...
stafflist: {
  content: "defaultblockslot staffmemberslot",
  group: "block",
  selectable: true,
  draggable: true,
  parseDOM: [{ tag: "staff-list", }],
  toDOM:() => [ "staff-list"', 0 ],
},
defaultblockslot: {
  content: "block*",
  group: "slot",
  attrs: {
    "v-slot:default": { default: "slotProps", },
  },
  selectable: false,
  draggable: false,
  parseDOM: [{
[...]

PR #869, opened 2020-10-20
Use ProseMirror Sub-views to implement editability of Vue-component slot content

enjibby:
Please understand the the code in its current strate is very much a work in progress, but I will be attempting to improve this to a point where things make a little more sense.

This includes the pre-cursor work to implement the inner-view code from [this example](https://prosemirror.net/examples/footnote/).

### DIFF
--- a/examples/Components/Routes/VueSlots/StaffList/DefaultBlockSlotExtension.js
+++ b/examples/Components/Routes/VueSlots/StaffList/DefaultBlockSlotExtension.js
@@ -15,7 +15,7 @@ export default class extends Node {
       selectable: false,
       draggable: false,
       parseDOM: [{
-        tag: 'template[v-slot:default]',
+        tag: 'template[v-slot\\:default]',
         getAttrs: dom => ({
           'v-slot:default': dom.getAttribute('v-slot:default'),
         }),
--- a/examples/Components/Routes/VueSlots/StaffList/DefaultBlockSlotExtension.js
+++ b/examples/Components/Routes/VueSlots/StaffList/DefaultBlockSlotExtension.js
@@ -1,0 +1,26 @@
+import Node from '../../../../../packages/tiptap/src/Utils/Node'
+
+export default class extends Node {
+  get name() {
+    return 'stafflistdefaultblockslot'
+  }
+
+  get schema() {
+    return {
+      content: 'block*',
+      group: 'block',
+      attrs: {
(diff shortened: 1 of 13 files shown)


enjibby:
I will be working on this to implement what I can of #868

enjibby:
These changes are the first attempt at implementing the example code outlined in #868 and the beginnings of updating the ComponentView to support it. I expect that I will need to implement a ComponentSlotView for each of the editable sub-areas, but I haven't fully mapped this out yet. Fingers crossed!

examples/Components/Routes/VueSlots/StaffList/DefaultBlockSlotExtension.js
```diff
@@ -0,0 +1,26 @@
...
+      },
+      selectable: false,
+      draggable: false,
+      parseDOM: [{
+        tag: 'template[v-slot\\:default]',
+        getAttrs: dom => ({
+          'v-slot:default': dom.getAttribute('v-slot:default'),
+        }),
+      }],
+      toDOM: node => ['template', node.attrs, 0],
```
kepta, line 23: I am not familiar with Vue, but wouldn't this clash with the DOM [template](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/template).

hanspagel:
Thanks for working on this! It’s still on our list to look at subviews and at your implementation. I suppose your work will help us to wrap our heads around it.

That said, don’t expect us to merge that into tiptap 1. We try put all our time into the new version and won’t add new features to tiptap 1.

Thanks for your understanding!

enjibby:
I'd already expected that tiptap V2 was the focus here and I imagine that there would differences, so no issue there. I'm still to prove this would even work given possible incompatibilities in individual browsers.

I will continue to press on when I am able 🙂

examples/Components/Routes/VueSlots/StaffList/DefaultBlockSlotExtension.js
enjibby: I'm not familiar enough with the inner workings of vue to give you an answer on that myself, but Vue does make use of `<template>` elements within its document structure.

I have used this element type successfully in Vue apps using tiptap by injecting the resulting output of `getHTML()` (that includes a `<template>` element) into a Vue component as so:

```
<component :is="injectedDocString"></component>
```

PR #880, opened 2020-11-02
Text Color and Background extensions, and additional supporting of the "type" attribute to OrderedList 

Joost-dm:
Text Color and Background extensions, and additional supporting of the "type" attribute of <ol> to OrderedList =)

### DIFF
--- a/packages/tiptap-extensions/src/marks/BackgroundColor.js
+++ b/packages/tiptap-extensions/src/marks/BackgroundColor.js
@@ -1,0 +1,43 @@
+import { Mark } from 'tiptap'
+import { updateMark } from 'tiptap-commands'
+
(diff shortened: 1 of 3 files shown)


hanspagel:
Thanks for your work on this @Joost-dm! We’ll add support for background-color and color to [tiptap v2](https://blog.ueber.io/post/our-plan-for-tiptap-2/). 🙌 I’m not sure about the type option for OrderedLists, but it’ll definitely easier to add this with v2 for everyone who wants to do it. It’ll make a great example for the documentation. So thank you very much!

If you don’t mind: For maintainers it’s way easier to have multiple small PRs for features, than one PR with multiple features. It’s easier to look through and it’s easier to merge one by one. Have a nice day, and thanks again for your contribution! 💖

PR #883, opened 2020-11-05
SetCellAttr in prosemirror-tables needs parameters

### LINKED ISSUE #862: Table cell background color with tiptap extensions
I have made a button to change the `background-color` style attribute of a table cell. I am using `setCellAttr('background','#F100F1')` to change the color, but nothing happens.

Here is an example with my button on the top left of the page. When you select a cell and click on the button, nothing happens.

https://codesandbox.io/s/vue-issue-template-forked-z22wj?file=/src/App.vue

Am I using `setCellAttr` wrong? Thank you!

Drabble:
The `setCellAttr` command has no use if we can't pass parameters to it. My issue was setting the background color of a cell in a table.

Fixes #862.


hanspagel:
Looks good! Thanks for the PR! 👍 Will be part of the next release.

Issue #577, opened 2020-01-08
Error: Looks like multiple versions of prosemirror-model were loaded


karol-f:
**Describe the bug**
When using the editor (e.g. using ENTER key) it's not working, and we got error in the console:

```
Can not convert <> to a Fragment (looks like multiple versions of prosemirror-model were loaded)
```

Due to `prosemirror-model@1.8.2` and `prosemirror-model@1.9.0` loaded simultaneously.

`tiptap-extensions` package requires strictly version `1.8.2` and other packages `prosemirror-model@^1.0.0, prosemirror-model@^1.1.0, prosemirror-model@^1.8.1` which resolves to `1.9.0`

**Steps to Reproduce / Codesandbox Example**
Steps to reproduce the behaviour:
1. install 
```
"tiptap": "^1.27.1",
"tiptap-extensions": "^1.29.1"
```
with Yarn. This will install `prosemirror-model@1.8.2` and `prosemirror-model@1.9.1`. 

2. When using the editor (e.g. using ENTER key) it's not working, and we got error in the console

**Expected behaviour**
Editor should work, there should be only one `prosemirror-model` package installed.

### **Quick and dirty workaround** for users (it's not proper fix!) - use `resolutions` entry in `package.json`:
```
    "resolutions": {
        "prosemirror-model": "1.9.1"
    }
```
Then remove `node_modules`, `yarn.lock` and install packages again.

njoguamos:
I am experiencing this issue too. 

In the meantime, the workaround suggested by @karol-f is resolving the issue .

samiell:
I am experiencing the same issue. The suggested fix is not working either.
`RangeError: Can not convert <> to a Fragment (looks like multiple versions of prosemirror-model were loaded)`.
However, if I paste a paragrapho text, it works fine. An odd behaviour

karol-f:
@samiell Did You install packages again after adding `resolutions` to package.json?

Hyperblaster:
I think this is the same issue. I get this in the console
`Error in created hook: RangeError: Adding different instances of a keyed plugin (plugin$)`

samiell:
> @samiell Did You install packages again after adding `resolutions` to package.json?
> 
> Mentioned quick fix is working, I wonder why it's not for You.

@karol-f, it is working now with fresh yarn install

kfields:
This looks like a smoking gun ...
[...]

PR #897, opened 2020-11-16
Update package.json

Guillaume-N:
Update package.json to fix the prosemirror-model mismatch versions.
See my comment here: https://github.com/ueberdosis/tiptap/issues/577#issuecomment-728143889

https://github.com/ueberdosis/tiptap/blob/main/packages/tiptap/package.json
https://github.com/ueberdosis/tiptap/blob/main/packages/tiptap-extensions/package.json

Using the latest tiptap (1.30.0) and tiptap-extensions (1.32.2) versions.

tiptap requires `"prosemirror-model": "1.11.2"` while tiptap-extensions and others require `"prosemirror-model": "^1.11.2"`. Hence both `1.11.2` and `1.12.0` are installed, resulting in a "looks like multiple versions of prosemirror-model were loaded" error.
This should fix the issue.

### DIFF
--- a/packages/tiptap/package.json
+++ b/packages/tiptap/package.json
@@ -25,7 +25,7 @@
     "prosemirror-gapcursor": "1.1.5",
     "prosemirror-inputrules": "1.1.2",
     "prosemirror-keymap": "1.1.4",
-    "prosemirror-model": "1.11.2",
+    "prosemirror-model": "^1.11.2",
     "prosemirror-state": "1.3.3",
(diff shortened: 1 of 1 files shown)


hanspagel:
Thanks for the PR! 

I’m not sure this will really fix the "multipe versions" issue, as we tried this already a few times in the past. I started to prepare the next release and updated all dependencies.

The recent releases introduced too many bugs, so I’ll take time before publishing it, but expect it to happen in the first two weeks of January. 👍

Issue #874, opened 2020-10-26
Suggestions Bug - Can't write a second mention after adding a first


Hyperblaster:
I can write @hans, then select a person, then when i try to write @ someone else and it won't work until i create a new paragraph. I believe this used to work, any ideas?

Illustrated on the example page here:
https://tiptap.dev/suggestions

Even if you add a few paragraphs, When you write a second mention the entire dropdown goes buggy and moves to the top left of corner of the browser.

hanspagel:
Thanks for reporting! This is an issue with tippy.js, not with tiptap. So I’m not sure we can do something about that.

Hyperblaster:
@hanspagel what a pain. Funnily enough, with some mucking around, i duplicated the Suggestions Plugin and Mention extension myself and was able to get it working. But i'm not sure exactly what i did or how.

Thanks for an awesome editor

axwalker:
@hanspagel regarding the first line of this issue, I'm not sure this is an issue with tippy.js. The `onEnter`/`onChange` callbacks are not called regardless of whether tippy is used or not.

I don't have an issue with positioning when adding a second mention in a future paragraph. I'm just completely unable to add two mentions in the same paragraph using `@`.

emzet93:
@hanspagel I agree with @axwalker. I'm not using `tippy.js` in my implementation at all and handle dropdowns with Vuetify. For me `onEnter` event doesn't fire too when I already have another mention in the line.

I'm sure it was working before and problem arrived after updating `tiptap-extensions` from `1.33.0` to `1.33.2`

hanspagel:
Okay, I don’t have time to look at it now, but I’ll reopen the issue.

andreasvirkus:
As a current workaround fixating `tiptap-extensions` to `1.33.0` works 👌

greghub:
@hanspagel I can confirm that the bug doesn't appear in 1.33.0. Deleting `content: 'inline*'` from schema in `Mention.js` fixes the issue, but I can't say for sure whether that breaks anything else

Priestch:
@greghub If current `content: 'inline*'` in schema is correct, then must be some bugs in `SuggestionsPlugin`.
It can not find a valid match, because the `from`, `to` can not pass the condition below, so not tigger the suggestions popup.
```
// line 59
// If the $position is located within the matched substring, return that range
        if (from < $position.pos && to >= $position.pos) {
          position = {
            range: {
              from,
              to,
            },
[...]

PR #898, opened 2020-11-17
Remove content expression from Mention nodes

andreasvirkus:
Fixes sibling Mention creation inside the same paragraph
Fixes https://github.com/ueberdosis/tiptap/issues/874 🤞 

Tested the Suggestions example over at the [GitPod demo](https://gitpod.io/#https://github.com/ueberdosis/tiptap/) and it didn't seem to have any side effects besides allowing multiple Mention nodes in the same paragraph. If someone could validate this, that'd be top 🙏 

@Priestch maybe you can share some insight into why the `content: 'inline*'` addition was necessary - https://github.com/ueberdosis/tiptap/commit/9ccfa168f8336a018e5abf4c126c4864ad7e41c1?

### DIFF
--- a/packages/tiptap-extensions/src/nodes/Mention.js
+++ b/packages/tiptap-extensions/src/nodes/Mention.js
@@ -33,7 +33,6 @@ export default class Mention extends Node {
       },
       group: 'inline',
       inline: true,
-      content: 'inline*',
       selectable: false,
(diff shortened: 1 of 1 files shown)


Priestch:
@andreasvirkus it was added to fix copy issue in #860

PR #902, opened 2020-11-25
Allow Links to use their individual attrs.target

dbakan:
Currently there seems to be no way to use `target="value"` on some Link marks and `target="other"` on another Link mark, because in `toDOM` the value of `node.attrs.target` will always be overwritten with `defaultOptions.target`. 
I think this was introduced in #619.

With the PR, the individual values are preferred and the `defaultOptions.target` will only be used as a fallback.


hanspagel:
Thanks for your PR! Will be part of the next version. 👍

PR #968, opened 2021-02-08
More flexible matching of trigger character

epixian:
Adds an option `allowMatchWithin` to allow matching the trigger character from within text (i.e. not at the beginning of the line or preceded by a space).


byul5810 (approved):

hanspagel:
Thanks for the PR! We’re about to release the new tiptap 2 next week, including a complete rewrite of the Mention extension. We don’t plan to add new features to v1, so I’m closing this here for now.

PR #981, opened 2021-02-24
Update code highlighting when transaction encompases entire code block node

ascott18:
The code highlighting plugin has cases where highlighting will be lost. One such case is when a transaction is applied that encompasses the entire code block but does not specifically affect anything within that code block node.

This fix may not be exhaustive, but it resolves the main issue I was having which was that collab with y-prosemirror will sometimes perform transactions that impact the entire document. Such transactions were wiping out the syntax decorations. 

In my testing, this change does not cause any additional or excessive decoration recomputations in "normal" situations, such as typing outside of a code block.


hanspagel:
Thanks for the PR! 🙌 We incorporated those changes in the upcoming v2 repo, we expect a public beta release next week.

PR #987, opened 2021-03-02
Constrain `from` to non-negative numbers

### LINKED ISSUE #954: MenuBubble: Error in v-on handler: "IndexSizeError: Failed to execute 'setStart' on 'Range'
**Describe the bug**

using the editor-menu-bubble for adding a Link will result in the following error:

`Error in v-on handler: "IndexSizeError: Failed to execute 'setStart' on 'Range': The offset 4294967295 is larger than the node's length (16)."`

Browser: Chrome (Google Chrome, Vivaldi):

```    <editor-menu-bubble class="menububble" :editor="editor" @hide="hideLinkMenu" v-slot="{ commands, isActive, getMarkAttrs, menu }">
      <div
        class="menububble"
        :class="{ 'is-active': menu.isActive }"
        :style="`left: ${menu.left}px; bottom: ${menu.bottom}px;`"
      >

        <form class="menububble__form" v-if="linkMenuIsActive" @submit.prevent="setLinkUrl(commands.link, linkUrl)">
          <input class="menububble__input" type="text" v-model="linkUrl" placeholder="https://" ref="linkInput" @keydown.esc="hideLinkMenu"/>
          <button class="menububble__button" @click="setLinkUrl(commands.link, null)" type="button">
            <icon name="remove" />
          </button>
        </form>

        <template v-else>
          <button
            class="menububble__button"
            @click="showLinkMenu(getMarkAttrs('link'))"
            :class="{ 'is-active': isActive.link() }"
          >
            <span>{{ isActive.link() ? 'Update Link' : 'Add Link'}}</span>
            <icon name="link" />
          </button>
        </template>

      </div>
    </editor-menu-bubble>
```

**Screenshots**

Screen Shot 2021-01-25 at 12 56 36

MarcelloTheArcane:
Fix #954

There is an issue where `document.createRange()` is being called with negative values which throws an error.

This pull request constrains it to be always greater than 0, so the error doesn't happen any more.  I've tested this and it seems to work okay.

I'm not sure why the `from` variable is being returned as -1 sometimes, so perhaps this is an underlying issue that needs to be fixed too.

### DIFF
--- a/packages/tiptap/src/Plugins/MenuBubble.js
+++ b/packages/tiptap/src/Plugins/MenuBubble.js
@@ -3,7 +3,7 @@ import { Plugin, PluginKey } from 'prosemirror-state'
 function textRange(node, from, to) {
   const range = document.createRange()
   range.setEnd(node, to == null ? node.nodeValue.length : to)
-  range.setStart(node, from || 0)
+  range.setStart(node, Math.max(from, 0))
   return range
 }
 


MarcelloTheArcane:
@hanspagel It's to do with these lines here too. `end` must be truthy some of the times that `offset` is zero, which means it takes the first branch, subtracts 1, and throws the error.

https://github.com/ueberdosis/tiptap/blob/d292a3d55bcb40deb612580bd417ae068d9e35f8/packages/tiptap/src/Plugins/MenuBubble.js#L19-L26

It might be to do with a 'backwards' selection (starting at the rightmost part and going left), though I'm not totally convinced.

hanspagel:
Thanks for the PR! 🙌

PR #996, opened 2021-03-10
Fix tippy position incorrect after resize or scroll

Elvincth:
A fix for: When the user scrolled or resize the screen, the tooltip position didn't update


hanspagel:
Thanks for the PR! 🙌

PR #1196, opened 2021-04-22
Add install instructions

MarcelloTheArcane:
This makes it easier than having to navigate over to the `installation` page.


hanspagel:
Thanks for the PR! tiptap v1 was Vue-only, so it might be better to show instructions with the Vue package. Would you mind to change that?

MarcelloTheArcane:
@hanspagel Done! Is that correct?

hanspagel:
Amazing, thanks! 🙌

PR #1204, opened 2021-04-23
Exit code block on down arrow

techstonia:
👋

This PR adds functionality to `extension-code-block` and allows the user to exit by pressing `DownArrow`:

exit-code-block-down-arrow

### DIFF
--- a/packages/extension-code-block/src/code-block.ts
+++ b/packages/extension-code-block/src/code-block.ts
@@ -115,6 +115,24 @@ export const CodeBlock = Node.create<CodeBlockOptions>({
 
         return false
       },
+      ArrowDown: () => {
+        const state = this.editor.state
+        const { from, to } = state.selection
+
+        if (from > 1 && from === to) {
+          let inCodeBlock = false
+          state.doc.nodesBetween(from - 1, to - 1, node => {
+            if (node.type.name === 'codeBlock') inCodeBlock = true
+          })
+
+          const cursorAtEnd = to === state.doc.content.size - 1
+          if (inCodeBlock && cursorAtEnd) {
+            return this.editor.commands.exitCode()
+          }
+        }
+
+        return false
+      },
     }
   },
 


packages/extension-code-block/src/code-block.ts
```diff
@@ -113,6 +113,28 @@ export const CodeBlock = Node.create<CodeBlockOptions>({
...
+            if (node.type.name === 'codeBlock') inCodeBlock = true
+          })
+
+          let nothingOnRight = true
+          state.doc.nodesBetween(from + 1, to + 1, node => {
+            if (node) nothingOnRight = false
+          })
+
+          if (inCodeBlock && nothingOnRight) {
+            return (this.editor.commands as any).setHardBreak()
```
philippkuehn, line 134: It's not good to rely on commands of another extensions. Not everyone is using the hard break extension so this can lead to errors. There is a core command called `exitCode()`. Can you try this instead?

philippkuehn:
Hey, thanks for that PR!

But I found an issue with your implementation:

https://user-images.githubusercontent.com/2500670/115924284-d0740b80-a47f-11eb-87f1-a1ee08723169.mp4

techstonia:
Good find @philippkuehn. I improved the PR a bit. Does this make more sense? :) Thanks!

packages/extension-code-block/src/code-block.ts
techstonia: `exitCode` works nicely 👍

bishalbaaniya123 (approved):

philippkuehn:
I [pushed some code](https://github.com/ueberdosis/tiptap/commit/ffafff9e365f9636fae4a67bb825d49b7d95cb94) with a few more checks so I’m closing this. Thanks anyway! 👍 

@uniho It’s also possible to leave code blocks with triple enter!

PR #1241, opened 2021-04-28
Make Horizontal Rule compatible with Typography extension

chrisarmstrong:
The Typography extension automatically converts two hyphens `--` into an em dash `–`, which means typing three hyphens in a row results in an em dash and a hyphen `–-`, instead of triggering an hr. 

I propose adding the `–-` pattern to this extension‘s `nodeInputRule` so typing three hyphens in a row will still trigger an hr, even if 'smart dashes' are enabled via the Typography extension or otherwise.


philippkuehn:
Good idea!

PR #1245, opened 2021-04-29
feat: use composition-api to init  EditorContent

### LINKED ISSUE #1219: $attrs is readonly.
Not able to view editor after importing the basic template from the documentation. [https://www.tiptap.dev/installation/vue2](documentaion).

in the console the following errors come up:

```
[Vue warn]: $attrs is readonly.

found in

---> <EditorContent>
       <TiptapTest> at app/javascript/components/notes/tiptap_test.vue
         <Notes> at app/javascript/components/notes.vue
           <Root>
           
vue.esm.js:648 [Vue warn]: $listeners is readonly.
```

```
found in

---> <EditorContent>
       <TiptapTest> at app/javascript/components/notes/tiptap_test.vue
         <Notes> at app/javascript/components/notes.vue
           <Root>
           
[Vue warn]: Avoid mutating a prop directly since the value will be overwritten whenever the parent component re-renders. Instead, use a data or computed property based on the prop's value. Prop being mutated: "editor"
```
```
found in

---> <EditorContent>
       <TiptapTest> at app/javascript/components/notes/tiptap_test.vue
         <Notes> at app/javascript/components/notes.vue
           <Root>
```

zcuric:
### This PR
- [x] uses @vue/composition-api to initialise EditorContent component for @tiptap/vue-2 
- [x] closes #1219 

[@vue/composition-api](https://github.com/vuejs/composition-api) provides better TS support and vue@2 compatible.

### DIFF
--- a/packages/vue-2/package.json
+++ b/packages/vue-2/package.json
@@ -28,6 +28,7 @@
   "dependencies": {
     "@tiptap/extension-bubble-menu": "^2.0.0-beta.9",
     "@tiptap/extension-floating-menu": "^2.0.0-beta.6",
+    "@vue/composition-api": "^1.0.0-rc.8",
     "prosemirror-view": "^1.18.2"
   }
 }
--- a/packages/vue-2/src/BubbleMenu.ts
+++ b/packages/vue-2/src/BubbleMenu.ts
@@ -1,7 +1,10 @@
-import Vue, { PropType } from 'vue'
 import { BubbleMenuPlugin, BubbleMenuPluginKey, BubbleMenuPluginProps } from '@tiptap/extension-bubble-menu'
+import Vue from 'vue'
+import VueCompositionAPI, { defineComponent, PropType } from '@vue/composition-api'
 
-export const BubbleMenu = Vue.extend({
+Vue.use(VueCompositionAPI)
+
+export const BubbleMenu = defineComponent({
   name: 'BubbleMenu',
 
   props: {
--- a/packages/vue-2/src/EditorContent.ts
+++ b/packages/vue-2/src/EditorContent.ts
@@ -1,9 +1,11 @@
-import Vue, { PropType } from 'vue'
+import Vue from 'vue'
 import { Editor } from './Editor'
+import VueCompositionAPI, { defineComponent, PropType } from '@vue/composition-api'
 
-export const EditorContent = Vue.extend({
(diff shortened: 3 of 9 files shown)


philippkuehn:
Have you tested this on your system? This fixed your issue for sure?

We use `Vue.extend` [at some more places](https://github.com/ueberdosis/tiptap/search?q=Vue.extend) so I think we should replace them too?

zcuric:
@philippkuehn yes, I've tested it, there is no slowing down and the warnings have disappeared. And this has proper TS support now. 

I'm currently experiencing #577, but that could be because I've used `npm link` to link `@tiptap/vue-2` to my project with other tiptap packages. 

I didn't know about other Vue.extend use case, I'll update the PR!

zcuric:
@philippkuehn Done. Only one thing that I'm not sure is that should VueCompositionAPI be included in every file or somewhere in top level.

zcuric:
@philippkuehn After some reading, talking and consideration, I concluded this is not good approach. There is a problem with the combination when using `Vue.extend` with `prosemirror-view`. My guess is that, the renderer in Vue case keeps on rendering when exit and entering the prosemirror editor and it fills the memory. 

I had this similar experience when trying to build a prosemirror plugin that would inject Vue component with Vue.extend into the prosemirror editor. Renderer went crazy and devtools profiler showed that it was slowing the browser. 

To could avoid Vue.extend if use vue-class-component or you could go the other way around totally and use [vue-demi](https://github.com/vueuse/vue-demi) and you could have only one vue package if you wanted.

philippkuehn:
Now I’m confused. You say that Vue.extend is a problem and this PR is about removing it and now you closed it?

zcuric:
@philippkuehn It is the problem, but I don't like the approach of solving it 😅 . It adds another dependency which doesn't fit in overall scheme of things. I'll look into another solutions. Sorry.

philippkuehn:
Hmm, I think it’s not too bad to use the composition API if it fixes the issue 🤷‍♂️

zcuric:
The way that @vue/composition-api for vue@2 works is different and, although, it fixes the issue, the approach wrong, because in the documentation for @vue/composition-api it says: 
> You must install @vue/composition-api as a plugin via Vue.use() before you can use the Composition API to compose your component.

And from what I've seen in the code base, you don't have such a setup because of the way the things are written.

PR #1251, opened 2021-05-02
When text align is default, don't add a style attribute

robguthrie:


packages/extension-text-align/src/text-align.ts
```diff
@@ -37,9 +37,13 @@ export const TextAlign = Extension.create<TextAlignOptions>({
...
-            renderHTML: attributes => ({
-              style: `text-align: ${attributes.textAlign}`,
-            }),
+            renderHTML: attributes => {
+              if (attributes.textAlign == this.options.defaultAlignment){
+                return {}
+              } else {
+                return { style: `text-align: ${attributes.textAlign}` }
+              }
+            },
```
zcuric, line 46: Use `===` for comparison. You don't need `else`. 

```suggestion
            renderHTML: attributes => {
              if (attributes.textAlign === this.options.defaultAlignment) {
                return {};
              }
               
               return { style: `text-align: ${attributes.textAlign}` }
            },
```

zcuric (approved):

zcuric (changes requested):
Good idea, but please address these suggestions.

robguthrie:
thanks for the tips

zcuric:
@philippkuehn LGTM :)

klaasgeldof:
Was running into issues with TextAlign today. Even if you don't touch the align buttons, it always sets a style attribute of `text-align: left`. Then when you use the value in your HTML, for example like this:

```html
<div class="text-center">
    {{ $value }}
</div>
```

...the text is always left aligned.

This PR should fix this.

PR #1252, opened 2021-05-03
Fix returning true/false in can().chain().run()

Markario:
I think this was previously meant to return a function in the proxy object getter, but now it's a standalone function and should return the result of the callbacks.


philippkuehn:
Uff, good catch!

PR #1255, opened 2021-05-03
[Docs] Improve example integration with Laravel Livewire

iksaku:
The current draft works fine when integrated into "static" Livewire components, however, it doesn't work in cases where the Livewire component reloads its data, be it due to a component method, lifecycle hook or simply because the flow for the component includes an option to reset its data. Old data will be cached into TipTap and new data will not come.

I re-wrote the example to work in these cases, and also for it to be more aligned to [Alpine's example](https://www.tiptap.dev/installation/alpine), with the only difference that we'll now use Alpine as a proxy for Livewire<->TipTap communication.

If there are any questions, suggestions or jokes, feel free to let me know :smile:

### DIFF
--- a/docs/src/docPages/installation/livewire.md
+++ b/docs/src/docPages/installation/livewire.md
@@ -12,12 +12,12 @@ The following guide describes how to integrate tiptap with your [Livewire](https
 TODO
 
 ## editor.blade.php
-```html
-<!--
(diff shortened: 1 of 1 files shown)


hanspagel:
Amazing! Thanks @iksaku! 🙌

PR #1267, opened 2021-05-05
HorizontalRule is included in defaultExtensions

Duncank:
Set 'default extension' for horizontalRule to true in api-nodes documentation.


philippkuehn:
Good catch! Thanks!

PR #1301, opened 2021-05-11
feat: export mark helpers

zcuric:
### This PR
- [x] Exports more mark helpers for easier extension development, for example: `getMarkRange` for available in tiptap@1.


philippkuehn:
thanks!

PR #1305, opened 2021-05-11
docs: complete list of extensions with changed name

Deckluhm:
- Removed extensions with no name change
- Added missing extensions with name change


hanspagel:
Thanks for the PR! 🙌

Issue #1316, opened 2021-05-13
Custom Attributes for Mentions


christopher4lis:
Right now, it seems you can only use the default attribute of "id" for mentions. Previously, you could insert both an id attribute as well as an attribute for a label.

**The solution I would like**
I'd like to have the ability to use a different piece of data for the mention's id attribute, so the id attribute and actual mention label display something different.

For instance:

Rather than 
   ` <span class="mention" data-mention="8" contenteditable="false">@8</span>`

I'd like to see
    `<span class="mention" data-mention-id="8" contenteditable="false">@John Doe</span>`

Checked the source code and it doesn't seem like this is possible at the moment, but let me know if I'm wrong, thanks.

hanspagel:
Try to extend the Mention node and overwrite addAttributes and parseHTML with the code from the original extension, but replace data-mention with data-mention-id.

Does that work?

https://github.com/ueberdosis/tiptap/blob/main/packages/extension-mention/src/mention.ts#L46

https://www.tiptap.dev/guide/custom-extensions#attributes

christopher4lis:
Came up with a workaround, but I think that should do it. At some point would be nice to have the attr: {} option similar to tiptap v1 mentions, but happy to have a solution rn. Thanks for all you do and your help.

mmachatschek:
I just created a small codesandbox example how working with a UUID could look like:

https://codesandbox.io/s/bold-voice-l58oq

Adding an additional 'data-mention' attribute would be helpful if its provided from the server side to immediately display the name of a user and not having to wait for the client request to get the needed data to render the mention

philippkuehn:
This initial issue should be fixed with #1322

PR #1322, opened 2021-05-15
Mention: Add text attribute

tomhrtly:
This PR allows you to change the text that is to be rendered in the DOM when a selection is made from the suggestion.

This is so that you may want the `data-mention` attribute to use a user's ID but the user's name would be displayed instead of the ID e.g. "@JohnSmith" instead of "@1".

### DIFF
--- a/packages/extension-mention/src/mention.ts
+++ b/packages/extension-mention/src/mention.ts
@@ -62,19 +62,19 @@ export const Mention = Node.create<MentionOptions>({
           }
         },
       },
-      text: {
+      label: {
         default: null,
         parseHTML: element => {
           return {
-            text: element.getAttribute('data-text'),
+            text: element.getAttribute('data-label'),
           };
         },
         renderHTML: attributes => {
-          if (!attributes.text) {
+          if (!attributes.label) {
             return {};
           }
           return {
-            'data-text': attributes.text,
+            'data-label': attributes.label,
           };
         },
       },
@@ -93,12 +93,12 @@ export const Mention = Node.create<MentionOptions>({
     return [
       'span',
       mergeAttributes(this.options.HTMLAttributes, HTMLAttributes),
-      `${this.options.suggestion.char}${node.attrs.text ?? node.attrs.id}`,
+      `${this.options.suggestion.char}${node.attrs.label ?? node.attrs.id}`,
     ]
   },
 
   renderText({ node }) {
-    return `${this.options.suggestion.char}${node.attrs.text ?? node.attrs.id}`
+    return `${this.options.suggestion.char}${node.attrs.label ?? node.attrs.id}`
   },
 
   addKeyboardShortcuts() {
--- a/packages/extension-mention/src/mention.ts
+++ b/packages/extension-mention/src/mention.ts
@@ -77,12 +77,12 @@ export const Mention = Node.create<MentionOptions>({
     return [
       'span',
       mergeAttributes(this.options.HTMLAttributes, HTMLAttributes),
-      `${this.options.suggestion.char}${node.attrs.id}`,
+      `${this.options.suggestion.char}${node.attrs.text ?? node.attrs.id}`,
     ]
   },
 
   renderText({ node }) {
-    return `${this.options.suggestion.char}${node.attrs.id}`
+    return `${this.options.suggestion.char}${node.attrs.text ?? node.attrs.id}`
   },
 
   addKeyboardShortcuts() {
--- a/packages/extension-mention/src/mention.ts
+++ b/packages/extension-mention/src/mention.ts
@@ -66,7 +66,7 @@ export const Mention = Node.create<MentionOptions>({
         default: null,
         parseHTML: element => {
           return {
-            text: element.getAttribute('data-label'),
+            label: element.getAttribute('data-label'),
           };
         },
         renderHTML: attributes => {
--- a/packages/extension-mention/src/mention.ts
+++ b/packages/extension-mention/src/mention.ts
@@ -62,6 +62,22 @@ export const Mention = Node.create<MentionOptions>({
           }
         },
       },
+      text: {
+        default: null,
+        parseHTML: element => {
+          return {
+            text: element.getAttribute('data-text'),
+          };
+        },
+        renderHTML: attributes => {
+          if (!attributes.text) {
+            return {};
+          }
+          return {
+            'data-text': attributes.text,
+          };
+        },
+      },
     }
   },
 


nadar:
👍 looking for such an option

nadar:
@tomhrtly Do you know if there is any option to override the current mention render value? because i am returning an object as user. Basically instead of 

```
renderText({ node }) {    return `${this.options.suggestion.char}${node.attrs.id}`  },
```

i like to return 

```
renderText({ node }) {    return `${this.options.suggestion.char}${node.attrs.id.NANE}`  },
```

So its not possible to configure `renderText` from outside, right?

tomhrtly:
@nadar Can you not pass the specific value you need into the data attributes that already exist? I'm not sure exactly what you're trying to achieve

nadar:
I think i am trying to do the same as you, maintain a NAME and an ID. But then display the NAME of the user. Sorry i don't want to be off-topic in your PR.

renderText() decides what should be rendered as value inside the element, and in my case the selection returns an object with NAME (or TEXT) and ID

nadar:
i know its still off-topic, but i found the solution to customize the output when working with an object as mention response value. Maybe others are google for such a solution and are beginners like me:

Its not possible to override the extensions`renderText ` function but of course its possible to override the `renderHTML` function. Therefore use something like this:

```js
renderHTML({ node, HTMLAttributes }) {
    return [
      'span',
      mergeAttributes(this.options.HTMLAttributes, HTMLAttributes),
      `@${node.attrs.name}` // <---- 
    ]
  },
```

Where name is an attribute defined previously:

```js
addAttributes() {
    return {
      name: {
        default: null,
        parseHTML: element => {
          return {
            name: element.getAttribute('data-mention-name'),
          }
        },
        renderHTML: attributes => {
          if (!attributes.name) {
            return {}
          }
          return {
            'data-mention-name': attributes.name,
          }
        },
      },
      id: {
        default: null,
        parseHTML: element => {
          return {
            id: element.getAttribute('data-mention-id'),
          }
        },
        renderHTML: attributes => {
          if (!attributes.id) {
            return {}
          }
          return {
            'data-mention-id': attributes.id,
          }
        },
      },
    }
  },
})
```

As people often need context, here is the full solution

```js
const CustomMention = Mention.extend({
  renderHTML({ node, HTMLAttributes }) {
    return [
      'span',
      mergeAttributes(this.options.HTMLAttributes, HTMLAttributes),
      `@${node.attrs.name}`
    ]
  },
  addAttributes() {
    return {
      name: {
        default: null,
        parseHTML: element => {
          return {
            name: element.getAttribute('data-mention-name'),
          }
        },
        renderHTML: attributes => {
          if (!attributes.name) {
            return {}
          }
          return {
            'data-mention-name': attributes.name,
          }
        },
      },
      id: {
        default: null,
        parseHTML: element => {
          return {
            id: element.getAttribute('data-mention-id'),
          }
        },
        renderHTML: attributes => {
          if (!attributes.id) {
            return {}
          }
          return {
            'data-mention-id': attributes.id,
          }
        },
      },
    }
  },
})
```

and then define the editor with the custom mention extension

```js
new Editor({
      extensions: [
        StarterKit,
        CustomMention.configure({
          HTMLAttributes: {
            class: 'mention',
          },
          suggestion: {
            items: async function(query) {
              const response = await this.$axios('users?query=' + query)
              var values = []
              response.data.forEach(user => {
                values.push({id: userr.user_id, name: userr.user.firstname})
              })

              return values
            }.bind(this),
//.... here goes the rest from the example in the docs
```

I could send a PR for the docs, if you are interested in such extensive guide examples @hanspagel

philippkuehn:
BTW: `renderText` is only used for the text representation when copying mentions into plain text fields/editors. by default atom nodes would be rendered as an empty string.

tomhrtly:
@philippkuehn What changes do you need me to make to get this PR approved? What's the best way forward?

mmachatschek:
Issue #1316 also mentions this

philippkuehn (commented):
Hey, I think I would like to merge that, however I would want to rename "text" to "label". Can you change this?

tomhrtly:
@philippkuehn All done

philippkuehn:
Great!

Another thing I’m not quite sure about: Currently the id is stored via the attribute `data-mention`. Does it make sense to change this to `data-id`? `data-mention` would still be necessary to parse the node.

tomhrtly:
@philippkuehn In my use case it would, but other developers using this may not be passing the ID but a username instead so `data-mention` covers different types of data that could be sent (username, ID, full name)

hanspagel:
id = identifier = the thing that identifies it, so +1 für data-id from me 🙃

tomhrtly:
Happy to change this if others think the same @philippkuehn

shadow-light:
Just a suggestion but would it be worth making the label dynamic by not storing it as an attribute? If the above use cases are similar to mine, I prefer to have the label get updated when the data does. By storing it as an attribute it is similar to hard-coding it with the value it first had when chosen. Where as if a user's name changes, this won't be reflected in the HTML.

Instead I'm using a method where I don't add a new attribute at all, and just override `renderHTML` so that it adds text content by looking up the latest value with the given `id`. I'm currently supplying the values via a custom configuration option, though you could also do it via the existing `items` function, but it isn't ideal since that's for querying the items.

If going down this route, perhaps it would be best to just supply a new `label` config option that takes a function and the `id` as an argument, and let the user lookup the value.

shadow-light:
```typescript
Mention.configure({
    label: id => MY_LABELS[id].label,
}).extend({
    renderHTML({node, HTMLAttributes}){
        return [
            'span',
            mergeAttributes(this.options.HTMLAttributes, HTMLAttributes),
            this.options.label(node.attrs.id),
        ]
    },
}),
```

philippkuehn:
@tomhrtly I’ve added `data-id` 👍

philippkuehn:
@shadow-light Yeah you are right. I’ve learned that rendering mentions is very individual. But I like your idea and added a `renderLabel` option to overwrite the default way (https://github.com/ueberdosis/tiptap/commit/3b78af44a054fa5d0101d8f7d836409b0b68547e).

So right now there are some ways to render mentions.
1. use id
2. use id + label
3. use id + add a custom node view to render names asynchronously (fetched by API for example)
4. use id + custom `renderLabel` method if names are available synchronously anywhere in your app

HeebeLee:
> @shadow-light Yeah you are right. I’ve learned that rendering mentions is very individual. But I like your idea and added a `renderLabel` option to overwrite the default way ([3b78af4](https://github.com/ueberdosis/tiptap/commit/3b78af44a054fa5d0101d8f7d836409b0b68547e)).
> 
> So right now there are some ways to render mentions.
> 
> 1. use id
> 2. use id + label
> 3. use id + add a custom node view to render names asynchronously (fetched by API for example)
> 4. use id + custom `renderLabel` method if names are available synchronously anywhere in your app

Why didn't I get the right variables?

```
suggestion: {
            items: function (query) {
              return [
                { id: 1, label: "aaa" },
                { id: 11, label: "aaa1" },
                { id: 12, label: "aaa2" },
                { id: 13, label: "aaa3" },
              ];
            },
```
But here's the variable:
```
{id: Proxy, label: null}
  id: Proxy
    [[Handler]]: Object
    [[Target]]: Object
    [[IsRevoked]]: false
  label: null
```
Everything else is default.

  "name": "@tiptap/extension-mention",
  "description": "mention extension for tiptap",
  "version": "2.0.0-beta.61",

@nadar

villageseo:
@HeebeLee did you find out how to win this ?

HeebeLee:
I think there is a bug here, there is a mention of plug-ins.  Its input
must be String or {id:{id:’any’, label:’any’}}, so I copied and modified
the plug-in to meet my format.

You can choose a way, although the current data structure is not
good-looking, but it can still be satisfied.

villageseo:
@HeebeLee huge thanks! Yes, you're right, I'm also not so happy about the data format.. But ok for now :)

villageseo:
@HeebeLee I've won this by minor changing in our custom MentionList.vue component: 

` selectItem(index) {
      const item = this.items[index];
      if (item) {
        this.command({ id: item.id, label: item.label, avatar: item.avatar });
      }
    },`
    
Then it works like a charm. It wasn't very obvious at first..

HeebeLee:
> @HeebeLee I've won this by minor changing in our custom MentionList.vue component:
> 
> ` selectItem(index) { const item = this.items[index]; if (item) { this.command({ id: item.id, label: item.label, avatar: item.avatar }); } },`
> 
> Then it works like a charm. It wasn't very obvious at first..

```
selectItem(index) {
      const item = this.items[index];
      // object
      if (item && item instanceof Object) {
        this.command(item);
      } else {
        // string or number
        this.command({ id: item });
      }
    },
```

```
<template>
  <div class="items">
    <button
      class="item"
      :class="{ 'is-selected': index === selectedIndex }"
      v-for="(item, index) in items"
      :key="index"
      @click="selectItem(index)"
    >
      {{ item.label ?? item }}
    </button>
  </div>
</template>
```

This is our code.

PR #1323, opened 2021-05-16
Extend Non-Empty Marks

stallion-supreme:
### Motivation

Screen Shot 2021-05-15 at 6 18 27 PM

I'm trying to add the ability to edit a link's `href` attribute whenever any part of the link is selected. Currently, the commands for updating links only update the selected portion of the mark, as expected, and I wanted to use `extendMarkRange` to make sure I could update the whole mark, but it nooped on non-empty selections..

### Changes
- Removes the requirement for a selection to be empty in order to extend the selection to contain the whole mark.
- Adds to the `Link` example to show how to update an entire link (though you could do this for any mark, now)
- Adds `extendMarkedRange` documentation

### Testing
- Manually validated that it works as expected (using the updated Link use case)
- Manually validated that when this is done for adjacent marks (ie. one link directly next to another) that the extend only encompasses the selected mark. @philippkuehn I wasn't able to introduce the edge-case you mentioned in discord about two adjacent links being grabbed, and if the selection contains two link marks, only the first is selected.

### DIFF
--- a/docs/src/demos/Marks/Link/index.vue
+++ b/docs/src/demos/Marks/Link/index.vue
@@ -6,6 +6,9 @@
     <button @click="editor.chain().focus().unsetLink().run()" v-if="editor.isActive('link')">
       remove
     </button>
+    <button @click="editor.chain().focus().extendMarkRange('link').unsetLink().run()" v-if="editor.isActive('link')">
+      remove entire link
+    </button>
     <editor-content :editor="editor" />
   </div>
 </template>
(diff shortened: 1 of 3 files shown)


philippkuehn:
Thanks for your work. There are some edge cases that your solution don’t capture. So I’m closing this. 

But I’ve added a version that should work as expected for your needs. (https://github.com/ueberdosis/tiptap/commit/ff7dd9b919c7b5e0f845369c744116045bab5028)

stallion-supreme:
Thanks @philippkuehn. I appreciate how fast you did that, and that commit is very informative as well.

PR #1337, opened 2021-05-18
FIX: add empty string arg to DOMImplementation.createHTMLDocument Method

kysaeed:
I used transpiled tiptap editor in IE11.
It's worked with this change.

reference:
https://developer.mozilla.org/en-US/docs/Web/API/DOMImplementation/createHTMLDocument


philippkuehn:
tiptap 2 is working in IE11!?

I’m not sure about this PR because we don’t want to support a dead browser anymore.

kysaeed:
I understand you don't want to spoil your codes by old specifications.
and I'll publish build environment in my repository for rare use case like this.


https://kysaeed.github.io/tiptap/
tiptap2 is working in this page with IE11.

PR #1345, opened 2021-05-19
Add getAttributes command to docs

Nick-Mazuk:
This PR adds the getAttributes (as mentioned in this discussion https://github.com/ueberdosis/tiptap/discussions/1342) command to the API docs.


philippkuehn:
getAttributes is a helper method and not a command. A command is a method that changes the editor state.

Issue #1365, opened 2021-05-23
Allow width and height attributes for image


Nick-Mazuk:
**The problem I am facing**

It's good practice to include `height` and `width` attributes on images to prevent layout shift as they load. The current plugin does not allow you to set these attributes.

Soon, metrics like cumulative layout shift are going to be a factor in Google search rankings, which means this is about to become critical.

**The solution I would like**

To be able to do

```ts
editor.chain().focus().setImage({ src, width, height }).run()
```

**Alternatives I have considered**

- Unlike attributes like `loading="lazy"` which can be set with `HTMLAttributes`, width and height are specific to each image and need to be set with `setImage`.
- Create my own version of the plugin

philippkuehn:
You can extend the node for that:


```js
import Image from '@tiptap/extension-image'

export const CustomImage = Image.extend({
  addAttributes() {
    return {
      ...this.parent?.(),
      width: {
        default: null,
      },
      height: {
        default: null,
      },
    }
  },
})

```

PR #1366, opened 2021-05-23
add width/height to image extension

Nick-Mazuk:
fixes https://github.com/ueberdosis/tiptap/issues/1365

Allows users to also add the width/height attributes to images to prevent layout shift as images load.

### DIFF
--- a/packages/extension-image/src/image.ts
+++ b/packages/extension-image/src/image.ts
@@ -16,7 +16,7 @@ declare module '@tiptap/core' {
       /**
        * Add an image
        */
-      setImage: (options: { src: string, alt?: string, title?: string }) => Command,
+      setImage: (options: { src: string, alt?: string, title?: string, width?: number, height?: number }) => Command,
     }
   }
 }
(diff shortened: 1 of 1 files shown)


philippkuehn:
I don’t merge that because you don’t always want to explicitly set the dimensions. And as you can see this takes only some lines of code so it’s easy to create a custom extended version of the image extension. We also plan to provide a resizable image extension somewhere in the future so maybe you can use this extension at some time.

tommy4st:
I don't think this change implies that you have to explicitly set the dimensions, but allows you to do so if you want. As for the "alt" or "title" attribute, if it is not set, it is simply not present. Please reconsider your rejection, as I would find this change really useful as well.

Nick-Mazuk:
Additionally, while you can extend the JavaScript, I was not able to extend the TypeScript definitions. This means if you try to extend the image as mentioned in https://github.com/ueberdosis/tiptap/issues/1365#issuecomment-853799239, you'll get a type-error every time you attempt to add an image with width and height attributes.

philippkuehn:
that’s why you can update the typescript types aswell https://www.tiptap.dev/guide/typescript#command-type

PR #1376, opened 2021-05-25
Allow passing of DependencyList to useEditor

YousefED:
in case the options depend on props of the consuming component, it would be nice to be able to pass a dependencylist to `useEditor`. Unless there's a better way to handle dependency updates


philippkuehn:
Thanks!

PR #1381, opened 2021-05-26
Make HTML in docs valid

MoPaMo:
You were missing an `<html>`-starter tag,  so users who copy-pasted the code would have experienced errors - not a very good first impression…
(Furthermore, I changed `<!doctype html>` to `<!DOCTYPE html>`since that's just more common)


philippkuehn:
thank you!

PR #1399, opened 2021-05-31
Prevent tiptap from creating duplicate style tags when injecting css

mmachatschek:
Code sandbox example [here](https://codesandbox.io/s/objective-glitter-mj9t0?file=/src/App.vue)

When using multiple Editor instances on the same page with `injectCSS` (which defaults to true), you get duplicate style that apply the same styles in the head.

This prevents those duplicate entries

image


philippkuehn:
Thanks!

PR #1416, opened 2021-06-04
Include children in placeholder plugin

nickdbush:
Adds an option to the placeholder plugin to include children in the search for empty nodes.

Consider the use case in which there is a `header` node with `title` and `subtitle` nodes. Currently, the header node will be the only one that receives the `.is-empty` class. Therefore, typing something into the title node makes the subtitle placeholder disappear, which is confusing to users. With the modifications in this PR, the search for empty nodes can be continued to children. Therefore, the subtitle node would retain its placeholder until modified.

I'd like to get some feedback on this before updating the docs and such.


philippkuehn:
Oh sorry for that delay. Looks good!

PR #1429, opened 2021-06-07
Fix name of FloatingMenu

shadow-light:
This seems like a mistake but please let me know if I'm missing something...


philippkuehn:
ah thank you!

PR #1436, opened 2021-06-09
Exclude superscript from subscript, and vice versa.

BrianHung:


philippkuehn:
ah yes. makes sense!

hanspagel:
Throws an error when the other isn’t loaded. #1438 😅

philippkuehn:
reverted. not sure about a workaround for now.

umar-ahmed:
> reverted. not sure about a workaround for now.

@philippkuehn Maybe add to the docs for both plugins that the user can extend the schema to make them mutually exclusive:

```js
import Superscript from '@tiptap/extension-superscript'
import Subscript from '@tiptap/extension-subscript'

const CustomSuperscript = Superscript.extend({
  excludes: 'subscript',
})

const CustomSubscript = Subscript.extend({
  excludes: 'superscript',
})
```

That approach seems to make the least assumptions about the user's editor configuration

PR #1444, opened 2021-06-09
Make drop cursor default to regular caret color

shadow-light:
Currently drop cursor defaults to black rather than the text color, which is what normal carets default to. Using `currentcolor` should provide a better default value.


philippkuehn:
Thanks!

PR #1465, opened 2021-06-14
Fix removal of textStyle mark when any style resets

bttger:
I built some extensions based on the textStyle extension but whenever I reset a style/attribute, the whole textStyle mark gets removed when there are still other styles defined in the mark.

### DIFF
--- a/packages/extension-text-style/src/text-style.ts
+++ b/packages/extension-text-style/src/text-style.ts
@@ -51,7 +51,7 @@ export const TextStyle = Mark.create<TextStyleOptions>({
     return {
       removeEmptyTextStyle: () => ({ state, commands }) => {
         const attributes = getMarkAttributes(state, this.type)
-        const hasStyles = Object.entries(attributes).every(([, value]) => !!value)
+        const hasStyles = Object.entries(attributes).some(([, value]) => !!value)
 
         if (hasStyles) {
           return true


philippkuehn:
`removeEmptyTextStyle` should prevent rendering an empty style tag (`style=""`) so I can’t merge this.

But I can’t understand the issue right now. Could you please create an issue + codesandbox for it?

bttger:
> removeEmptyTextStyle should prevent rendering an empty style tag (style="") so I can’t merge this.

Yes, my fix does exactly that. I also used the fontFamily extension from the tiptap/core and when calling the `unsetFontFamily()` method, it removes the styles tag even though there are still other extensions (built exactly like fontFamily) that have set an property on the styles object.

This is an example style object:
`{fontFamily: "Ubuntu, sans-serif", fontSize: "xx-large", fontColor: "#FFFFFFFF", lineHeight: "1", underline: "underline"}`

When you unset one of these attributes (e.g. to `null` like in the fontFamily extension), the `Object.entries(attributes).every(([, value]) => !!value)` operation returns false:
```
const hasStyles = Object.entries(attributes).every(([, value]) => !!value)
if (hasStyles) {
    return true
}
```

I believe that this is not intentional and that we should use the `some()` method to check if there is at least one style set on the mark.

philippkuehn:
Ah sorry, you are absolutely right!

PR #1469, opened 2021-06-14
export createTable

YousefED:
I think this can be useful for consumers as well


hanspagel:
Yes, I think so too! Thanks @YousefED! 🙌

PR #1470, opened 2021-06-14
use contentWrapperRef for react nodeviews

YousefED:
This PR refactors React NodeViews to get the contentWrapper element via a `ref`. This fixes two things:

1. We were previously relying on `update` being called to move the contentDOMElement. React and Prosemirror's lifecycles don't interop nicely; while React renders asynchronously, Prosemirror instantly needs a result for `contentDOM` and `dom`. When Prosemirror calls `contentDOM`, we call `maybeMoveContentDOM`, but we can't know for sure that the React tree has already been rendered and the element with `[data-node-view-content]` is readily available. I've encountered cases where this is breaking because ProseMirror didn't call `contentDOM` anymore afterwards (only before it was available).
2. It removes the need to query for `[data-node-view-content]`, which arguably is not a very nice pattern.

(Ideally, I would have gotten rid of the `contentDOMElement` wrapper entirely, but I didn't get that to work as we'd need to force prosemirror to rerender. Maybe this is possible by adding a decoration to force a rerender of the nodeView.)

Unfortunately this is a breaking change, as consumers now need to use the following:

```
<NodeViewContent ref={props.contentWrapperRef} />
```

Looking forward to your feedback,

PS: 
1) I can give a demo of the related issue in our codebase. As it only occured with nested react nodeviews, it's a bit cumbersome to replicate.
2) I can update the docs / tests if you'd be ok with the suggested change

### DIFF
--- a/packages/react/src/NodeViewContent.tsx
+++ b/packages/react/src/NodeViewContent.tsx
@@ -12,7 +12,6 @@ export const NodeViewContent: React.FC<NodeViewContentProps> = React.forwardRef(
     <Tag
       {...props}
       ref={ref}
-      data-node-view-content=""
       style={{
         ...props.style,
         whiteSpace: 'pre-wrap',
--- a/packages/react/src/ReactNodeViewRenderer.tsx
+++ b/packages/react/src/ReactNodeViewRenderer.tsx
@@ -16,6 +16,10 @@ interface ReactNodeViewRendererOptions {
   update: ((node: ProseMirrorNode, decorations: Decoration[]) => boolean) | null,
 }
 
+type ReactNodeViewProps = NodeViewProps & {
+  contentWrapperRef: (element: HTMLElement | null) => void
+}
+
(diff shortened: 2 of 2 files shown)


philippkuehn:
Sorry for that late response. There is an [official issue](https://github.com/ueberdosis/tiptap/issues/1747) now :D and I can reproduce this.

I’m not sure what this means for node views at all because `<NodeViewContent>` would be obsolete and I liked the similarity to the implementation in Vue. Must think about it!

philippkuehn:
ok, I [found a fix](https://github.com/ueberdosis/tiptap/commit/b15a8a8683bafbc9fd57dc64d60329f9ab08ecfd) in the current implementation so I'm closing this for now.

YousefED:
Thanks for keeping us posted @philippkuehn. Do note that using a `ref` is a more idiomatic React way of doing things. An attribute such as `[data-node-view-content]` really feels like a "hack" for me as React developer. I understand however that a breaking change like this is not ideal, but wanted to point out the above nevertheless

PR #1471, opened 2021-06-15
Fix parsing of mention nodes

shadow-light:
See here: https://github.com/ueberdosis/tiptap/commit/c64761a9e5f2c9cb7c90668abcdb8856c46547f8#commitcomment-52163786

I assume this is necessary. Please confirm this fixes it.


philippkuehn:
oops. you are absolutely right!

Issue #1518, opened 2021-06-28
Can't update `editorProps` after initialization. 


robertvanhoesel:
**Description**
I haven't found a way to add or modify `editorProps` after initialization – calling `editor.setOptions({ editorProps: { ... }` does not work, neither does modifying `editor.options.editorProps.*` directly.

**Steps to reproduce the bug**
```ts

const transformPastedHTML = (html: string) => {
    window.alert('transformPastedHTML called')
	return html
};

// A. Constructing the editor with editorProps works
const editor = new Editor({
	extensions: [ StarterKit ]
	editorProps: { transformPastedHTML },
});

// B. Constructing the editor and setting editorProps later does not work
const editor = new Editor({
	extensions: [ StarterKit ]
});

// Either of the below options do not have any effect
editor.setOptions({ editorProps: { transformPastedHTML } });
editor.options.editorProps.transformPastedHTML = transformPastedHTML;

```

**CodeSandbox**
I created a [CodeSandbox](https://codesandbox.io/s/tiptap-issue-template-forked-8m7my?file=/src/components/Tiptap.vue) to help you debug the issue.

**Expected behavior**
EditorProps are modified, i.e. in the example my custom transformPastedHTML is being called

**Environment?**
 - operating system: Mac OS X
 - browser: Chrome
 - mobile/desktop: Both  
 - tiptap version: 2.0.0-beta.85

**Additional context**
I've tried several, but not all, other editorProps to validate is present on more than one prop.

hanspagel:
The StarterKit comes with a CodeBlock extension already, disable it like this:

```js
StarterKit({ codeBlock: false })
```

Does that help?

moonrise-tk:
I agree, this is a pretty big problem. This is `setOptions`:
```jsx
public setOptions(options: Partial<EditorOptions> = {}): void {
    this.options = { ...this.options, ...options }
}
```

I'm not sure what the problem is, this should work? Unless `this.options` is set to be a const...?

robertvanhoesel:
The problem is that options.editorProps is not an actual reference to ProseMirror `view.props`. On creation the editorProps are destructured onto the ProseMirror view, and in `setOptions` it just updates the TipTap Editor' options, not actually setting the props on proseMirror. 

https://github.com/ueberdosis/tiptap/blob/939fc3d93ab67d54b7fe67b64c423290b3a40df8/packages/core/src/Editor.ts#L231-L238

[...]

PR #1540, opened 2021-07-04
Use correct reference for options.editorProps

robertvanhoesel:
This change ensures that `Editor.options.editorProps` correctly references the ProseMirror View props `this.view.props`.It enables users to read and manipulate the props as expected. Additionally calls `view.setProps(...)` when `editor.setOptions(...)` is being called with `editorProps` so they can be set and updated through this method as well.

Includes tests, closes https://github.com/ueberdosis/tiptap/issues/1518

### DIFF
--- a/packages/core/src/Editor.ts
+++ b/packages/core/src/Editor.ts
@@ -133,6 +133,12 @@ export class Editor extends EventEmitter {
    */
   public setOptions(options: Partial<EditorOptions> = {}): void {
     this.options = { ...this.options, ...options }
+
+    // Update editorProps directly on the view and store reference to configured props
+    if (this.view) {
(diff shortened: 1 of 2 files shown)


robertvanhoesel:
Any chance to get this reviewed and merged? @philippkuehn

robertvanhoesel:
@hanspagel @philippkuehn it would be a tad bit shameful community management if first time contributions stay untouched for a month. I would understand if you are too busy but a reply in any form would be appreciated.

philippkuehn:
Hey, sorry for that late response. Thanks for that PR. I made a few more changes after that: https://github.com/ueberdosis/tiptap/commit/80296aa8d0b46df2067bea16d33c30a64846c2c9

Please always use `setOptions` instead of mutating the options object directly.

PR #1555, opened 2021-07-08
doc update: onSelection → onSelectionUpdate

forresto:
PR to check my understanding, has this API changed?

Live doc: https://www.tiptap.dev/guide/custom-extensions#events

Edit: typescript doesn't like `{editor}` as the argument, though it seems to work in practice.

```
Type '({ editor }: { editor: any; }) => void' is not assignable to type '(this: { name: string; options: unknown; editor: Editor; parent: ... | null | undefined; }) => void'.ts(2322)
(property) onSelectionUpdate?: ((this: {
    name: string;
    options: unknown;
    editor: Editor;
    parent: ... | null | undefined;
}) => void) | null | undefined
```

### DIFF
--- a/docs/src/docPages/guide/custom-extensions.md
+++ b/docs/src/docPages/guide/custom-extensions.md
@@ -404,7 +404,7 @@ const CustomExtension = Extension.create({
   onUpdate() {
     // The content has changed.
   },
-  onSelection() {
+  onSelectionUpdate({editor}) {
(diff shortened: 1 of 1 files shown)


hanspagel:
Oh yes, good find! 🙌

hanspagel:
Could you create an issue for the TypeScript error?

PR #1556, opened 2021-07-08
Hide the BubbleMenu when the editor loses focus / the user is selecting (optional)

SamDuvall:
I've come across a few use cases where the BubbleMenu is displayed, where I don't think it should be.

### Editor loses focus

When the user selects a block of text
And clicks a button in the BubbleMenu that puts focus on an input outside of the editor
Then the user should not see the BubbleMenu

### User is selecting text

When the user is selecting text
Then the user should not be able to put focus on the BubbleMenu

The PR gives the developer an option to hide the BubbleMenu via `hideWhenSelecting` while the user is selecting text (like Dropbox Paper), so the BubbleMenu isn't in the user's way when they are selecting text.

Todo
- [ ] Add `hideWhenSelecting` as an option in the `vue-2` and `vue-3` packages

### DIFF
--- a/packages/extension-bubble-menu/src/bubble-menu-plugin.ts
+++ b/packages/extension-bubble-menu/src/bubble-menu-plugin.ts
@@ -11,6 +11,7 @@ import tippy, { Instance, Props } from 'tippy.js'
 export interface BubbleMenuPluginProps {
   editor: Editor,
   element: HTMLElement,
+  hideWhenSelecting?: boolean
   tippyOptions?: Partial<Props>,
 }
 
@@ -25,21 +26,29 @@ export class BubbleMenuView {
 
   public view: EditorView
 
+  public hideWhenSelecting: boolean
+
   public preventHide = false
+
+  public viewSelecting = false
 
   public tippy!: Instance
 
   constructor({
     editor,
     element,
+    hideWhenSelecting = false,
     view,
     tippyOptions,
   }: BubbleMenuViewProps) {
     this.editor = editor
     this.element = element
+    this.hideWhenSelecting = hideWhenSelecting
     this.view = view
     this.element.addEventListener('mousedown', this.mousedownHandler, { capture: true })
     this.view.dom.addEventListener('dragstart', this.dragstartHandler)
+    this.view.dom.addEventListener('mousedown', this.viewMousedownHandler)
+    this.view.dom.addEventListener('mouseup', this.viewMouseupHandler)
     this.editor.on('focus', this.focusHandler)
     this.editor.on('blur', this.blurHandler)
     this.createTooltip(tippyOptions)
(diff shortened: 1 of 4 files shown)


packages/extension-bubble-menu/src/bubble-menu-plugin.ts
```diff
@@ -25,21 +26,29 @@ export class BubbleMenuView {
...
     tippyOptions,
   }: BubbleMenuViewProps) {
     this.editor = editor
     this.element = element
+    this.hideWhenSelecting = hideWhenSelecting
     this.view = view
     this.element.addEventListener('mousedown', this.mousedownHandler, { capture: true })
     this.view.dom.addEventListener('dragstart', this.dragstartHandler)
+    this.view.dom.addEventListener('mousedown', this.viewMousedownHandler)
+    this.view.dom.addEventListener('mouseup', this.viewMouseupHandler)
```
forresto, line 51: We wanted the bubble menu to also hide while a selection was in progress with the keyboard. This might be lower priority, as the menu doesn't get in the way of the selection.

If you mouse up outside of the view (or window), the selection should also be considered done. A pattern for this is to **not** add the `mouseup` listener here....

packages/extension-bubble-menu/src/bubble-menu-plugin.ts
```diff
@@ -76,6 +85,16 @@ export class BubbleMenuView {
     this.hide()
   }
 
+  viewMousedownHandler = () => {
+    this.viewSelecting = true
```
forresto, line 89: ... rather, add the `mouseup` listener here, in the `mousedown` handler. By adding the listener to `window`, it will also work if the mouseup happens outside of the window.

```js
window.addEventListener('mouseup', this.viewMouseupHandler)
```

packages/extension-bubble-menu/src/bubble-menu-plugin.ts
```diff
@@ -76,6 +85,16 @@ export class BubbleMenuView {
...
+  }
+
+  viewMouseupHandler = () => {
+    this.viewSelecting = false
+    // we use `setTimeout` to make sure `selection` is already updated
+    setTimeout(() => this.update(this.editor.view));
```
forresto, line 95: Add this here to clean up. 🧹 

```js
window.removeEventListener('mouseup', this.viewMouseupHandler)
```

forresto (commented):
Some notes from my own exploration around this issue:

getreworked:
Any progress on getting this merged? Can I help in any way?

packages/extension-bubble-menu/src/bubble-menu-plugin.ts
```diff
@@ -90,10 +109,17 @@ export class BubbleMenuView {
...
 
   update(view: EditorView, oldState?: EditorState) {
+    const { viewSelecting } = this
     const { state, composing } = view
     const { doc, selection } = state
-    const isSame = oldState && oldState.doc.eq(doc) && oldState.selection.eq(selection)
 
+    const hideSelecting = this.hideWhenSelecting && viewSelecting
+    if (!view.hasFocus() || hideSelecting) {
+        this.hide()
```
forresto, line 118: `!view.hasFocus()` ... does this break the ability to tab to the bubble menu? Eg, a bubble menu item is focused via keyboard navigation, but this then hides it, because the editor view doesn't have focus.

SamDuvall: That is a good question and I'm fairly certain this would break the ability to tab to the bubble menu. I don't have that requirement (now), but I'm sure there are those that do, especially for accessibility. Making any UI plugins is so tricky to handle all the use cases. Any suggestions on how we might solve that issue?

Issue #1566, opened 2021-07-09
TaskItem does not update data-checked attribute


giorgiopogliani:
**Description**
Checking and unchecking a task item does not update `data-checked` attribute to true/false in real time

**Steps to reproduce the bug**
Create a taskItem and with the dev tools the attribute does not change in real time

**CodeSandbox**
The docs have this problem too: https://www.tiptap.dev/api/nodes/task-item/#taskitem

**Expected behavior**
I would like to style the text of a marked `taskItem`, it works if the content is refreshed but does not work in real time.

giorgiopogliani:
I still have this issue also the docs still doesn't work.  In the console I get:
```
Uncaught RangeError: Index 8 out of range for
```

philippkuehn:
@giorgiopogliani The demo is working for me. Maybe a browser extension is causing this issue. See #2372

giorgiopogliani:
Thanks for the quick reply! 

You are right! It was the extension for my password keychain.. 😶

PR #1567, opened 2021-07-09
update task-item.ts nodeview to update data-checked

BrianHung:
#1566 

Another alternative would be to remove the label wrapper element so the nodeview output is

```html
<li data-checked="true" class="">
  <input type="checkbox">
  <div><p>A list item</p></div>
</li>
```

This would allow people to style the content of the todo-item with a sibling css selector:

```css
ul[data-type=taskList] > li > input[type=checkbox] + div { 
  text-decoration: line-through;
}
```


hanspagel:
Thanks for the fix @BrianHung! 🙌

PR #1590, opened 2021-07-14
Replace node-sass with dart sass and fix deprecation warning

mmachatschek:
The deprecation warning shown in the console is triggered by using the "old" way to divide to values.

https://sass-lang.com/documentation/breaking-changes/slash-div

`node-sass` was replace by `sass` at it's recommended to upgrade to the maintained version. The [readme](https://github.com/sass/node-sass) of the `node-sass` says:

> Warning: LibSass and Node Sass are deprecated. While they will continue to receive maintenance releases indefinitely, there are no plans to add additional features or compatibility with any new CSS or Sass features. Projects that still use it should move onto Dart Sass.

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -44,7 +44,7 @@
     "@babel/preset-env": "^7.14.5",
     "@babel/preset-react": "^7.14.5",
     "html-loader": "^1.3.2",
-    "node-sass": "^5.0.0",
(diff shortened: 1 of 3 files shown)


hanspagel:
OMG, this is amazing. This also makes it work with Node 16. Thanks for the PR!

PR #1592, opened 2021-07-15
fix: export text align extension options

amursky:
Text align extension`s options were not exported as for other extensions.


hanspagel:
Thanks! ✌️

PR #1593, opened 2021-07-15
fix: export starter kit extension options

amursky:
Starter kit extension options were not exported.
Without this change, we have to import types from the dist directory:
```typescript
import { StarterKitOptions } from "@tiptap/starter-kit/dist/packages/starter-kit/src/starter-kit";
```


hanspagel:
Thaaanks! ✨

Issue #1588, opened 2021-07-14
Pasting into empty editor with autofocus adds newlines


mmachatschek:
**Description**
When initializing an empty editor instance with the option `autofocs: true` and pasting text after the initialization, the editor adds the pasted text and adds a newline at the end of the editor.

**Steps to reproduce the bug**
1. Go to the sandbox
2. Copy a text from somewhere
3. After loading the sandbox, paste the text without clicking outside the editor
4. A new line is added to the content

**CodeSandbox**
I created a CodeSandbox to help you debug the issue: https://codesandbox.io/s/funny-tdd-e3w0n

**Expected behavior**
No newline in the editor after pasting with autofocus set.

**Environment?**
 - tiptap version: 2

PR #1595, opened 2021-07-15
Fix selection with autofocus when editor content is empty on startup

mmachatschek:
Fixes https://github.com/ueberdosis/tiptap/issues/1588

It seams that the selection of the anchor and head is not zero based.

When initializing an empty editor and setting the focus by clicking into the editor. The `editor.state.selection` will have:

```
selection: Object
- type: "text"
- anchor: 1
- head: 1
```

image

### DIFF
--- a/packages/core/src/commands/focus.ts
+++ b/packages/core/src/commands/focus.ts
@@ -59,8 +59,8 @@ export const focus: RawCommands['focus'] = (position = null) => ({
 
   const { from, to } = resolveSelection(editor.state, position) || editor.state.selection
   const { doc, storedMarks } = tr
-  const resolvedFrom = minMax(from, 0, doc.content.size)
-  const resolvedEnd = minMax(to, 0, doc.content.size)
+  const resolvedFrom = minMax(from, 1, doc.content.size)
+  const resolvedEnd = minMax(to, 1, doc.content.size)
   const selection = TextSelection.create(doc, resolvedFrom, resolvedEnd)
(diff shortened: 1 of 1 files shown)


mmachatschek:
@hanspagel I couldn't find an appropriate test for this where I can put this, as the only error that happens without the fix is the one I describe in my issue.

hanspagel:
Thanks! I think it was 0 for a reason, but I’m not sure. I need to check this before merging.

BrianHung:
I was digging through `prosemirror-state`, and the way [selection is instantiated by default](https://github.com/ProseMirror/prosemirror-state/blob/36473f3aade8d84de78d56b1bd34167b55b3ff44/src/state.js#L24-L27) is by using [`Selection.atStart`](https://github.com/ProseMirror/prosemirror-state/blob/36473f3aade8d84de78d56b1bd34167b55b3ff44/src/selection.js#L154-L163). 

Using that function should give you a valid position or selection independent of the `topNode` (default doc might be doc -> paragraph, which gives offset of 1 but another doc could be doc -> blockquote -> paragraph which gives offset of 2).

philippkuehn:
Thanks for that hint @BrianHung! I was able to fix it with [that patch](https://github.com/ueberdosis/tiptap/commit/9425e7249a15e0fb4e8c714545c09f367d216536) so I’m closing this.

mmachatschek:
Thanks @BrianHung and @philippkuehn for fixing and looking into this 🙌

PR #1599, opened 2021-07-15
Doc fix: Remove braces from isEmpty & isEditable

WilliamIPark:
Hi! I've updated the docs to make it more clear that `editor.isEmpty` and `editor.isEditable` are bools rather than functions that need called. 

This is a quick fix and I'd recommend these maybe be moved out of the "methods" section of the docs, but as of right now there's no obvious section to move them to, and I don't want to make assumptions about what these should be called.


hanspagel:
Oh, yes, thanks!

PR #1625, opened 2021-07-22
Allow a rule to be skipped from the getAttributes callback

joevallender:
For example: you are matching URLs with a Regex but want to make additional checks on the match array before deciding if you want to apply the mark or not. Returning null or false won't attempt to apply the mark. 

Currently the return value of getAttributes is applied unconditionally. So this is somewhere between a feature request and a bug fix.


philippkuehn:
makes sense. thank you!

PR #1650, opened 2021-07-27
Only getting empty mark range if forced for link on unsetMark

isubasti:
Toggling mark around an empty selection makes sense for things like links. But it kinda has an unexpected behaviour for formatting marks like bold, strikethrough and others of the same kind. This PR adds a way to differentiate between the two types of marks


hanspagel:
Thanks! What’s that mentioned unexpected behaviour?

philippkuehn:
@isubasti Thanks for that PR. I’m closing this in favor of https://github.com/ueberdosis/tiptap/pull/1859 (I’ve renamed and moved some things).

PR #1671, opened 2021-08-02
Apply the correct regex in markPasteRule

joevallender:
Bug fix 👍 

`markPasteRule` has been using the `inputRegex` which is terminated with $ so it's only being applied to when the pasted text is _only_ code, not when it _contains_ code.


philippkuehn:
Thanks!

Issue #1040, opened 2021-04-09
ReactRenderer for React.FunctionComponent with ref


zofiag:
**Is your feature request related to a problem? Please describe.**
I couldn't find a way to run a method from the function component, which is wrapped into `RectRenderer`. From what I can see `RectRenderer` pass a ref as a prop only for class components
```javascript
if (isClassComponent(Component)) {
  props.ref = (ref: React.Component) => {
    this.ref = ref
  }
}
```

**Describe the solution you’d like**
A way to be able to run methods from the function components wrapped into `ReactRenderer`.

Example:
For the Mention extension `import Mention from "@tiptap/extension-mention";` I'd like to be able to configure it with my own `MentionList` component. `MentionList` is a function component with `forwardRef` and `useImperativeHandle` hook.
```javascript
export const MentionList = forwardRef<HTMLDivElement, SuggestionProps>(
  ({ command, items }, ref) => {

    useImperativeHandle(ref, () => ({
      onKeyDown: ({ event }: { event: KeyboardEvent }) => {
        if (event.key === "ArrowUp") {
          upHandler();
          return true;
        }

        if (event.key === "ArrowDown") {
          downHandler();
          return true;
        }

        if (event.key === "Enter") {
          enterHandler();
          return true;
        }

        return false;
      },
    }));

...
});
```
then I'd expect to able to call `reactRenderer.ref?.current.onKeyDown(props)` in `onKeyDown(props)` of `Mention.configure` under `suggestion` key.

Please let me know if there's already a way to make it work with function components, that I'm not aware of! If not, would love to hear your thoughts on the suggested approach.

aizaiz:
@philippkuehn @zofiag remove this line and it should work with function component
`if (isClassComponent(Component)) {`

softmarshmallow:
Need better react support!

philippkuehn:
always happy for some react contributions!

philippkuehn:
functional components are now supported. I also migrated the [mention demo](https://www.tiptap.dev/examples/suggestions). source: https://github.com/ueberdosis/tiptap/blob/b95507797d2fbf9147396de7f753e45e83736687/demos/src/Examples/Community/React/MentionList.jsx

PR #1690, opened 2021-08-05
✨ Follow Ref on ForwardRef component in reactRenderer

castroCrea:
- Allow us to pass Ref to a function ForwardRef component in reactRenderer (useful for mention)

### DIFF
--- a/packages/react/src/ReactRenderer.tsx
+++ b/packages/react/src/ReactRenderer.tsx
@@ -9,11 +9,20 @@ function isClassComponent(Component: any) {
   )
 }
 
+function isForwardRefComponent(Component: any) {
+  return !!(typeof Component === 'object' && Component.$$typeof?.toString() === 'Symbol(react.forward_ref)')
+}
+
 export interface ReactRendererOptions {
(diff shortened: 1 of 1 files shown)


philippkuehn:
Hey @castroCrea, can you please explain to me what this is helpful for?

castroCrea:
It is helpful if we use functional component `React.forwardRef` instead of class component.

For example, in Mention, you pass onKepPress to the ref but it only works for Class component. Will be nice for this to work with functional component also.

philippkuehn:
Ah great! So this would be a fix for https://github.com/ueberdosis/tiptap/issues/1040?

Since I don’t really have experience with `forwardRef`: What would a functional component have to look like to be able to call `onKeyDown` from outside there as well?

castroCrea:
Yes exactly

philippkuehn:
Perfect! Thanks!

philippkuehn:
The [mentions demo](https://www.tiptap.dev/examples/suggestions) is now using a functional component 🎉

Issue #1480, opened 2021-06-16
Request: optional shouldShow parameter for FloatingMenu and BubbleMenu


thatsjonsense:
**The problem I am facing**
In both the FloatingMenu and BubbleMenu, I wish I could customize the logic for when the menu displays.

Two specific use cases:
- In BubbleMenu, I want to trigger even when the selection is empty if you click inside a link
- In FloatingMenu, I want to trigger at depth 2 instead of depth 1 because my schema is a document with multiple sections

**The solution I would like**
An optional `shouldShow` function I can pass in the options to these menus where I can write my own logic. This would be consumed in the `update` method of each plugin. If you don't provide your own, it would use the default. Ideally, you could also access the default method from your custom one to do an OR.

**Alternatives I have considered**
Right now I've forked both menus just to change 1 line of logic in each one, which means it's hard to stay up to date with future improvements.

bubble-menu-plugin:
```
    if (empty || isEmptyTextBlock) {
      // OUR CUSTOM LOGIC STARTS HERE
      const isLink = selection.$anchor
        .marks()
        .some((mark) => mark.type.name === 'link')
      if (!isLink) {
        this.hide()
        return
      }
      // END OUR CUSTOM LOGIC
    }
```

floating-menu-plugin:
```
const isRootDepth = $anchor.depth === 2 // Was 1
```

philippkuehn:
Yes, this is something I want to add! Additionally it would be great to add multiple instances of the menus with different configuration.

Issue #1313, opened 2021-05-13
Bubble menu doesn't position correctly if image has alignment


zcuric:
**Description**
When using bubble menu with image nodes and if there is image alignment functionality present which is achieved with `float`, the bubble menu above image is not positioned correctly. 

**Steps to reproduce the bug**
1. Go to https://codesandbox.io/s/tiptap-issue-template-forked-p5sny
2. Click on image and see how the bubble menu is position in the middle of the editor
3. Click on alignment button and change to left and see how the bubble menu is outside of editor
4. Click on text nodes, bubble menu works as expetced

**CodeSandbox**
I created a CodeSandbox to help you debug the issue:
https://codesandbox.io/s/tiptap-issue-template-forked-p5sny

**Expected behavior**
This should behave same as on text nodes.

**Additional context**
I did some investigation on the bubble menu and how it calculates position. It looks like the `float` changes the start and the end positioning of the node.

joevallender:
I also get inconsistent coordinates from [coordsAtPos](https://github.com/ueberdosis/tiptap/blob/main/packages/core/src/helpers/coordsAtPos.ts) even when the image aren't floated when the editor alignment is anything other than left (where I believe it is returning the correct left pos by happy accident).

I cloned the repo and logged/traced it, and just got a bit lost seeing exactly what was happening in coordsAtPos (first time I'm digging into this repo so I'm lacking some context for what all the bits being passed in are). While debugging, `side` was coming out as left for one image node, and then right for another identically sized/positioned/etc. image.

It appears to be a rework of the [ProseMirror function](https://github.com/ProseMirror/prosemirror-view/blob/master/src/domcoords.js) and although the tiptap one has been updated more recently time-wise, it might have been based off an older version of the original? I found a few bubble menu position issues from the ProseMirror code while hunting this. 

So, no actual help there - sorry! 🙈

michelson:
Hello, I have a guess here:
- it seems that the coords are retrieved from a selection, but in a non editable context, like image blocks the selection does not exist, but at least we have an active element, the coords in this context will contain width == 0px, so the tippy is relocated to the left at px 0 .
- the solution to this is "detect" the non editable element and call the `.getBoundingClientRect()` on it, this value will be enough for tippy to display properly.

### a proof of concept to fix this:
[...]

PR #1710, opened 2021-08-11
Add shouldShow callback to bubble menu

georgeboot:
Relates to #1480 
Relates to #1313 (see https://github.com/ueberdosis/tiptap/issues/1313#issuecomment-842071420)

This PR adds a `shouldOpen` callback for the bubble menu. This allows users to control whether the bubbleMenu should open or not. By default, the new callback will always return `true` so there isn't an issue with bc.

### DIFF
--- a/docs/src/docPages/api/extensions/bubble-menu.md
+++ b/docs/src/docPages/api/extensions/bubble-menu.md
@@ -35,9 +35,9 @@ new Editor({
   extensions: [
     BubbleMenu.configure({
       element: document.querySelector('.menu'),
-      shouldShow: (view) => {
-        // your logic to control whether the menu should show
-        return true
+      shouldShow: ({ editor }) => {
+        // only show the buble menu for images and links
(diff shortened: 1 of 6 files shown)


hanspagel:
I think that’s a great PR! @philippkuehn What do you think?

@georgeboot Any chance you could add a usage example to the docs?

georgeboot:
@hanspagel I've made a few tweaks and added a real usage example to the docs

philippkuehn:
Hey, thanks for that but this is a really bad timing because I’ve done this in a [separate branch](https://github.com/ueberdosis/tiptap/tree/feature/advanced-menus) yesterday 😬 
In this branch I’m adding some more features for menus so I’m closing this here.

hanspagel:
@philippkuehn Oh! Can you make sure to also keep a usage example in the docs?

philippkuehn:
Jep, on it!

Issue #1716, opened 2021-08-11
focus command seems to be broken in React


philippkuehn:
It seems that the focus command leads to some problems in react or is not working at all. I think it has to do with the fact that React renders asynchronously. Multiple issues are affected:

- https://github.com/ueberdosis/tiptap/issues/1608
- https://github.com/ueberdosis/tiptap/issues/1520

Every attempt to fix this has failed. Any help is appreciated!

PR #1717, opened 2021-08-11
Added toggle to react editable content demo

SamDuvall:
I slightly altered the React content editable example to be a defining block, which I _think_ would be closer to a real world example. I added a toggle button to allow users to toggle the wrap, like the blockquote demo. The reason I did all of this was to highlight a selection issue that I'm running into. When I `wrapIn` the reactComponent, the cursor is jumping to the end of the reactComponent, but the selection state looks to modified correctly by the `wrapIn` (see attached video). Any ideas what could be causing this/how to fix it? 

https://user-images.githubusercontent.com/498226/129076954-2fccc449-b013-4277-a1dd-f68517946c04.mov

Also, I wasn't able to reproduce this issue when I did a vanilla javascript NodeView.

### DIFF
--- a/docs/src/demos/Guide/NodeViews/ReactComponentContent/Extension.js
+++ b/docs/src/demos/Guide/NodeViews/ReactComponentContent/Extension.js
@@ -7,7 +7,9 @@ export default Node.create({
 
   group: 'block',
 
-  content: 'inline*',
+  content: 'block*',
+
+  defining: true,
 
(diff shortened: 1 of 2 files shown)


philippkuehn:
Hmm, I think this is related to #1716.

philippkuehn:
I don’t want to complicate the demos but I can reproduce your issue. follow up here #1854

Issue #531, opened 2019-11-27
Disable undo/redo button if no history is available:  Accessing the history  steps


JanineLourens:
**Is your feature request related to a problem? Please describe.**

I've been trying to disable the undo/redo buttons if the user is at the last  or first step === no redo undo actions are available to step to.  

From a user perspective _the undo / redo buttons are clickable even if there is nothing to undo._ 
I've been scouring the Editor object for a while for the HISTORY but I can't seem to find where it is exposed.

**Describe the solution you'd like**

- I would like to be able to access where the history is stored. 
- I'd like to give the user a good UX by indicating that there's nothing to undo/redo,

**Describe alternatives you've considered**
I've considered manually keeping track of every action accrued and storing them in an array and then keeping track of it with a pointer. but this seems to be double work as the history has to be stored somewhere that I could access

fearclear:
It has been resolved in #511

domnantas:
With tiptap v2, `undoDepth` and `redoDepth` [seems to be removed](https://github.com/ueberdosis/tiptap/blob/main/packages/extension-history/src/history.ts). Is there any way to read history depth in the current release? Why was it removed?

philippkuehn:
@domnantas In tiptap 2 you can check for available actions with `editor.can().undo()` and `editor.can().redo()`. Does that help?

PR #1721, opened 2021-08-12
Docs: Disable history demo buttons when commands are not available

domnantas:
Related https://github.com/ueberdosis/tiptap/issues/531

I think this is a common usecase and also a good demo of `can()`


philippkuehn:
Oh thank you very much for this! 👍

PR #1754, opened 2021-08-19
add docs on attrs

jamilabreu:


philippkuehn:
Can you look at this @hanspagel? I think it’s not the right place for that code example.

hanspagel:
Hm … thanks for the PR! I expect most people to use `getHTML()` first, and then it should obvious where to manipulate the JSON. I feel your example doesn’t have a good enough context right now, and I’m not sure we want to provide that context so it makes sense to new users.

I’ll keep it in mind, and if that comes up more than once, I’ll consider to write a dedicated section about the JSON, the manipulation and so on. ✌️ Thanks!

jamilabreu:
Are people storing HTML on the server? I was under the impression most stored as JSON for more flexibility. Either way, all good!

hanspagel:
Oh, I wanted to write getJSON()!

Issue #1292, opened 2021-05-08
Placeholder for heading nodes 2 -> 6 not working, due to isNodeEmpty method


smlparry:
**Description**
I have discovered and issue which I _believe_ is actually a bug in TipTap. I am using the Placeholder extension with headings, and trying to display a placeholder. Placeholder works perfectly fine with Paragraph nodes as well as H1 nodes (Heading w/ level 1)

However, I could not get the placeholder to render at all on Headings level 2 -> 6. I have found the issue, and been able to fix locally, but I would like to raise this here.

**Steps to reproduce the bug**
See the CodeSandbox, basically its just trying to configure the placeholder based on the heading level

**CodeSandbox**
I created a CodeSandbox to help you debug the issue:
https://codesandbox.io/s/tiptap-issue-template-forked-lzflw?file=/src/components/Tiptap.vue

**Expected behavior**
I expect to see the placeholder for headings 2 -> 6

**Screenshot, video, or GIF**
See codesandbox

**Suggested Fix**
I've been able to track down and fix the issue locally, just not sure if my solution will cause any issues elsewhere in TipTap

The issue lies here: https://github.com/ueberdosis/tiptap/blob/main/packages/core/src/helpers/isNodeEmpty.ts#L4

The call to `const defaultContent = node.type.createAndFill()?.toJSON()` creates a default Heading node with `level: 1`:
![Screen Shot 2021-05-08 at 12 14 43 pm](https://user-images.githubusercontent.com/7920702/117522486-058f6a80-aff7-11eb-82b2-34766ca114f6.png)

[Line 5](https://github.com/ueberdosis/tiptap/blob/main/packages/core/src/helpers/isNodeEmpty.ts#L5) for headings 2 -> 6 get evaluated as expected:
![Screen Shot 2021-05-08 at 12 16 51 pm](https://user-images.githubusercontent.com/7920702/117522516-48e9d900-aff7-11eb-9192-ddce47b67402.png)

But obviously [Line 7](https://github.com/ueberdosis/tiptap/blob/main/packages/core/src/helpers/isNodeEmpty.ts#L7) evaluates to `false`.

I have been able to fix by updating [Line 4](https://github.com/ueberdosis/tiptap/blob/main/packages/core/src/helpers/isNodeEmpty.ts#L4) to the following:

```
const defaultContent = node.type.createAndFill({ ...node.attrs })?.toJSON()
```

[...]

PR #1777, opened 2021-08-23
Fix example with placeholder

DannyFeliz:


hanspagel:
What issue does this solve?

DannyFeliz:
@hanspagel In the usage you don't necessary show how to use the placeholder in the template https://www.tiptap.dev/api/extensions/placeholder/#usage

philippkuehn:
This is not how the placeholder extension works.

goamn:
Then how is it supposed to be used? The example is very unclear. It looks like we can add a string into the css rule to make it change. But how do we change it via JS ... via the settings "placeholder" value?

DannyFeliz:
@philippkuehn so, how do you use this extension then? I couldn't understand by only reading the documentation; that's why I created the PR

goamn:
@DannyFeliz I have figured it out. The css rule is necessary. To change the placeholder it will be like this:
```
const editor = useEditor({
    extensions: [
      Placeholder.configure({
        placeholder: 'Enter some details here...',
      }),
      StarterKit,
    ],
  });
```
Also regarding the CSS rule, you can remove it on focus by adding ":not(:focus)" to the rule:
```
.ProseMirror:not(:focus) p.is-editor-empty:first-child::before {
  content: attr(data-placeholder);
  float: left;
  color: #ced4da;
  pointer-events: none;
  height: 0;
}
```

DannyFeliz:
I see, thanks

philippkuehn:
dynamic placeholders are also possible: https://github.com/ueberdosis/tiptap/issues/1292#issuecomment-867557863

goamn:
Thanks @philippkuehn . Would be nice to have this part in the docs: `Placeholder.configure({placeholder: 'xyz'})`

PR #1790, opened 2021-08-24
docs: add color picker to color extension demo

domnantas:
A color picker is commonly used to change text color. This PR adds such example to the docs. This is also a good example of `getAttributes('textStyle')`.

Added tests for color extension


hanspagel:
Wow, that’s cool! ✨ 

Let’s deploy this ...

PR #1820, opened 2021-08-31
Fix: Don’t initialize tippy on requestAnimationFrame to avoid race conditions

enriquecastl:
This PR defers the initialization of the tippy library in the bubble and floating menu extensions to the moment when the `update` hook is first invoked by ProseMirror. The goal of this change is initializing the tippy library only when the Editor’s element is attached to the document. The `update` hook provides a convenient place to check whether the aforementioned condition is met. 

### Why is this change necessary 

We are implementing a WYSIWYG Editor using Tiptap’s Vue 2 extension. We ran into a problem where the `requestAnimationFrame` call is executed without delay right after the bubble menu plugin is initialized. This behavior causes a race condition where tippy is initialized before the Vue component is fully mounted. 

### How to reproduce the bug?

I can’t explain why `requestAnimationFrame` is invoked differently from the demos you provide in this repository. I couldn’t create an isolated sandbox where the problem can be reproduced either. However, you can reproduce the problem if you try to edit a Project Wiki in https://gitlab.com.

This approach ensures that a race condition can’t exist because the menu is only initialized when the editor’s state changes which is a sign that Tiptap’s Editor is fully mounted.

### DIFF
--- a/packages/extension-bubble-menu/src/bubble-menu-plugin.ts
+++ b/packages/extension-bubble-menu/src/bubble-menu-plugin.ts
@@ -38,6 +38,8 @@ export class BubbleMenuView {
 
   public tippy: Instance | undefined
 
+  public tippyOptions?: Partial<Props>
+
   public shouldShow: Exclude<BubbleMenuPluginProps['shouldShow'], null> = ({ state, from, to }) => {
     const { doc, selection } = state
     const { empty } = selection
@@ -59,7 +61,7 @@ export class BubbleMenuView {
     editor,
     element,
     view,
-    tippyOptions,
+    tippyOptions = {},
     shouldShow,
   }: BubbleMenuViewProps) {
     this.editor = editor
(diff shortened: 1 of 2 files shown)


packages/extension-bubble-menu/src/bubble-menu-plugin.ts
```diff
@@ -74,13 +76,10 @@ export class BubbleMenuView {
     this.view.dom.addEventListener('dragstart', this.dragstartHandler)
     this.editor.on('focus', this.focusHandler)
     this.editor.on('blur', this.blurHandler)
+    this.tippyOptions = tippyOptions
+    // Detaches menu content from its current parent
+    this.element.remove()
```
enriquecastl, line 81: **note**: I considered safe to detach the menu element because tippy also detaches the element on mount.

philippkuehn:
That looks great. Thank you!

PR #1824, opened 2021-09-01
extended suggestion startOfLine option

cameronmbell:
Sometimes suggestions should be contextual


cameronmbell:
This straight up doesn't work sorry. [Defer to this](https://github.com/ueberdosis/tiptap/pull/1826)

Issue #1384, opened 2021-05-27
`findSuggestionMatch`， There should be a configuration to configure whether to match the prefix


Akimotorakiyu:
`findSuggestionMatch`， There should be a configuration to configure whether to match the prefix:

below function should have a switch configuration:

https://github.com/ueberdosis/tiptap/blob/6990012a7b4ca20f44d8bbbe3293ffbf366a65fe/packages/suggestion/src/findSuggestionMatch.ts#L28

https://github.com/ueberdosis/tiptap/blob/6990012a7b4ca20f44d8bbbe3293ffbf366a65fe/packages/suggestion/src/findSuggestionMatch.ts#L45-L51

way-jm:
I also think it well, if i want use mention ,  I must  input a space before '@', it looks not elegant

Issue #329, opened 2019-05-27
Ability to insert variables, similar to suggestions


moltar:
**Is your feature request related to a problem? Please describe.**

Ability to insert pre-defined "variables" into the text.

E.g. user could type "today is {{ today }}" or "today is $today".

The editor would provide a list of suggested variables that are pre-defined.

The editor would also decorate these variables somehow once they are inserted.

The editor would allow to backspace and erase them.

**Describe the solution you'd like**

Similar to how suggestions work. But ability to use multiple characters, I guess. Like "{{" to begin suggestions.

**Describe alternatives you've considered**

N/A

**Additional context**

Integromat has a similar looking feature, but does not provide a suggest dropdown when typing. Can only use a mouse. http://take.ms/HSf3l

Zapier also has a similar one http://take.ms/NhODH

---

Thanks!

moltar:
I think I have figured out a work-around:

https://codesandbox.io/embed/vue-template-y0o5z

Notably:

```scss
.mention::after {
  content: "}}";
}
```

and

```js
          new Mention({
            matcher: {
              char: "{{"
            },
```

Seems to work!

If you have any better suggestions, please let me know.

If you think this is the best way, then please feel free to close this issue.

Thanks!

asseti6:
@moltar this is completely possible similiar to how you are using the Mention class already by defining your own custom node and plugin or use the suggestions plugin already available.  As a working example your node would look something like:

```
import { Node } from "tiptap";
import { replaceText } from "tiptap-commands";
import VariableSuggestionPlugin from "../plugins/CustomSuggestions";

export default class Variables extends Node {
  get name() {
    return "variable";
  }

  get defaultOptions() {
    return {
      matcher: {
        char: "%%",
        allowSpaces: false,
        startOfLine: false
      },
      variableClass: "variable",
      suggestionClass: "variable-suggestion"
    };
  }

  get schema() {
    return {
      attrs: {
        id: {},
        label: {}
      },
      group: "inline",
      inline: true,
      selectable: false,
      atom: true,
      toDOM: node => [
        "span",
        {
          class: this.options.variableClass,
          "data-variable-id": node.attrs.id
        },
        `${this.options.matcher.char}${node.attrs.label}`
      ],
      parseDOM: [
        {
          tag: "span[data-variable-id]",
          getAttrs: dom => {
            const id = dom.getAttribute("data-variable-id");
            const label = dom.innerText
              .split(this.options.matcher.char)
              .join("");
[...]

PR #1826, opened 2021-09-01
Allow triggering suggestions without prefix space

jkosir:
Currently `suggestion` extension only allows triggering the suggestions dropdown if the suggestion character is prefixed by space. This of course makes sense for @mentions but for some other use cases it is desirable to also trigger suggestions dropdown if suggestions character is not prefixed by space (e.g. with variables inside the text).

Already requested in:
* https://github.com/ueberdosis/tiptap/issues/1384
* https://github.com/ueberdosis/tiptap/issues/329

### DIFF
--- a/packages/suggestion/src/findSuggestionMatch.ts
+++ b/packages/suggestion/src/findSuggestionMatch.ts
@@ -4,6 +4,7 @@ import { ResolvedPos } from 'prosemirror-model'
 export interface Trigger {
   char: string,
   allowSpaces: boolean,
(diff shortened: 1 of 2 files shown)


jkosir:
Any progress on reviewing this? Anything I should change or is there a reason this wouldn't be accepted at all?

philippkuehn:
Thanks! Sorry for that delay.

jkosir:
Thanks for the quick release also! 👍

so2liu:
how should user use this if no document?

Issue #1816, opened 2021-08-30
It doesn't seem possible to add twice the mention extension on the editor 


GitHubish:
Until today I added to Tiptap twice the Mention extension configured differently (@ and #, the first with spaces allowed, not for the second). But since the update of the package ("@tiptap/core": "^2.0.0-beta.102")  when I add the plugin twice, I get this message: 
`ERROR RangeError: Adding different instances of a keyed plugin (mention$)`

How is this possible? How else?
Thanks

GitHubish:
Here is an example with the Angular Framework implementation (by @sibiraj-s ) where the problem occurs :  [stackblitz example](https://stackblitz.com/edit/ngx-tiptap-mentions-owzctg?file=src%2Fapp%2Fapp.component.ts)

hanspagel:
You can add `pluginKey` to the suggestion utility to separate those two.

https://prosemirror.net/docs/ref/#state.PluginKey

GitHubish:
@hanspagel Ah great thanks indeed it was just that. I didn't see that I could change it. Thanks for your responsiveness for this really cool editor!

Issue #11, opened 2018-09-01
Add support for mentions, tags & commands


philippkuehn:
Everyone needs `@mentions`, `#tags` and `/commands`. This should be possible with one extension. Rendering should be outside of tiptap.

codepotato:
+1 for this Philip! 👍

janiskelemen:
Hi @philippkuehn, do you have any news about this? I just looked around for a prosemirror extension and found that you have already opened an issue on prosemirror-suggestions. Did you have a chance to try this plugin? I'm willing to dig into it but wanted to check with you before.

philippkuehn:
Hey @janiskelemen, it's the number one of my todo list. Currently I'm on vacation so I have not *that* much time to work on tiptap. `prosemirror-suggestions` is great and I already created a branch for integrating that but it's still much work to do.

philippkuehn:
I'm making progress :)
![sep-29-2018 08-51-51](https://user-images.githubusercontent.com/2500670/46242242-91b06e80-c3c6-11e8-9db9-97b926bb1d5b.gif)

philippkuehn:
There is a [demo](https://tiptap.scrumpy.io/suggestions) for suggestions now.

AveryBibeau:
Thanks for putting this together. I had to modify `get name` in `nodes/Mention.js` to differentiate between mention instances since I need both @ and # as separate use cases.

Since `get name` was returning the same identifier for both instances the `this.name` in `command: ({ range, attrs, schema }) => replaceText(range, schema.nodes[this.name], attrs),` confused the two. I modified `nodes/Mention.js` like this:

```
get name() {
  return `mention-${this.options.matcher.char}`
}
```

as a quick fix and it seems to be working. Interested if anyone has used a different workaround for this.

andreasvirkus:
@AveryBibeau Could you elaborate how you made both `@` and `#` work as triggers for the Mentions popup? Also need this 🙏

mattwr18:
hey @AveryBibeau, I was looking at your fork and at `nodes/Mention.js` and I didn't see the code you mentioned :point_up: 
sorry if I am missing something obvious, but I didn't see how you got it working with `#` as well
We have been able to get it working with a suggestion list, but have not worked out how to get it working if there are no hashtags with the same name. We would like for the user to be able to create the hashtag if it doesn't exist already. 
Does anyone have any open-source code/demo of hashtags working with tiptap?

[...]

PR #1840, opened 2021-09-04
Multiple mention chars ['@', '#',...]

ep1kt3t0s:
It is common in social media platforms to use `@` and `#` characters for mentions.
Implementing multiple mention characters in Tiptap.

This PR is a draft PR as I did not poke around enough to how to build and test the code.
I appreciate if someone can take over to confirm it works and make any necessary changes then we can merge to `main`.

### DIFF
--- a/packages/suggestion/src/findSuggestionMatch.ts
+++ b/packages/suggestion/src/findSuggestionMatch.ts
@@ -2,7 +2,7 @@ import { Range } from '@tiptap/core'
 import { ResolvedPos } from 'prosemirror-model'
 
 export interface Trigger {
-  char: string,
+  char: string | string[],
   allowSpaces: boolean,
   startOfLine: boolean,
   $position: ResolvedPos,
@@ -12,6 +12,7 @@ export type SuggestionMatch = {
   range: Range,
   query: string,
   text: string,
+  usedChar: string,
 } | null
 
 export function findSuggestionMatch(config: Trigger): SuggestionMatch {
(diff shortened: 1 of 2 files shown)


cadars:
The use of `#` might clash with the planned(?) hashtag extension that could be used for tagging: https://www.tiptap.dev/api/nodes/hashtag

philippkuehn:
@ep1kt3t0s Do you have any example app/servive that uses both (@ & #) for the same mention functionality?

ep1kt3t0s:
@cadars why hashtag extension is separate? It is only a mention/suggestion system with a different character.
@philippkuehn Facebook and Twitter both has @user and #topic systems. Why would we think of them as separate functionalities?
Also, Quill supports exactly the feature I tried to make with this PR: https://github.com/afry/quill-mention#example

ep1kt3t0s:
Plus, we'd like to provide functionality to people to create more suggestion/mention characters, one might want to use `+something`  as well.

philippkuehn:
Yeah but for that case it’s better to create multiple instances of the mention extension (or a copy of it) to create different nodes.

ep1kt3t0s:
@philippkuehn I think I tried multiple instances but I don't think I could get it working. Does it work? I can give it another shot if so.

ep1kt3t0s:
@philippkuehn, cannot add two Mention instances:
```
Uncaught RangeError: Adding different instances of a keyed plugin (mention$)
    at index.es.js:761
    at Array.forEach (<anonymous>)
    at new Configuration (index.es.js:759)
    at EditorState.reconfigure (index.es.js:917)
    at Editor.createView (tiptap-core.esm.js:2725)
    at new Editor (tiptap-core.esm.js:2574)
    at new Editor (tiptap-react.esm.js:28)
    at tiptap-react.esm.js:59
    at invokePassiveEffectCreate (react-dom.development.js:23487)
```

I've also found https://github.com/ueberdosis/tiptap/issues/11 but it was also not helpful.

So do you want to go ahead with this PR (after testing it) or do you have another solution to get `@` and `#` and possibly more characters like `/` and `+` working together?

hanspagel:
Thanks for the PR, but yes, we would recommend use two instances of the Mention extension. Add a pluginKey to the configuration, like that:

pluginKey: new PluginKey('hashtag')

Red more: https://prosemirror.net/docs/ref/#state.PluginKey
See also: #1816

Thanks for your work on this! I think we’ll keep that way and just make it more clear in the docs.

ep1kt3t0s:
@hanspagel thank you for the pointer, just tried different PluginKeys with multiple Mention instances but when an item is inserted into the editor it always gets the second `char` e.g. `@word` works but becomes `#word` after selected.
Am I missing something else I wonder? It doesn't look like both mention instances share any stateful object.

Example: https://codesandbox.io/s/tiptap-mention-plugin-multiple-mentions-wuohj?file=/src/Editor.jsx:791-2281

philippkuehn:
Ah, you are right. There is a limitation about rendering/parsing different mention nodes because they share the same `parseHTML` and `renderHTML` methods. https://github.com/ueberdosis/tiptap/blob/73a278a5d3853e8742560939453cbd5b4d6f7ad4/packages/extension-mention/src/mention.ts#L107-L113

We could improve that by making these methods dynamic.

```js
parseHTML() {
  return [
    {
      tag: `span[data-${this.name}]`,
    },
  ]
}
// ...
```

Later we could create different extensions based on the mention extension.

```js
const UserMention = Mention
  .extend({
    name: 'user',
  })
  .configure({
    suggestion: {
      char: '@',
      pluginKey: new PluginKey('user-mention'),
    },
  })

const HashtagMention = Mention
  .extend({
    name: 'hashtag',
  })
  .configure({
    suggestion: {
      char: '#',
      pluginKey: new PluginKey('hashtag-mention'),
    },
  })
```

But I’m not sure about this change. I’ve always seen the mention extension as a user mention extension. For a hashtag node, for example, I still wanted to build it so that it doesn’t render as an atom node, but as text that changes dynamically as you type (a mention node is currently always rendered as a "block"). So maybe it’s best to create a copy of the mention extension for now. What do you think @hanspagel?

ep1kt3t0s:
Maybe we can leave the system less opinioated and not fixed to one or two (@ and #) characters. Is there any gain having assumptions on which mention (suggestion) character correlates to users/PRs/issues/commands/topics/notes/groups/teams/channels..?

philippkuehn:
The logic for this is in the [@tiptap/suggestion](https://github.com/ueberdosis/tiptap/tree/main/packages/suggestion) package, which you can use for any of your nodes. The [@tiptap/extension-mention](url) is only one example (probably the most common one) for using suggestions together with user mentions.

If you want to use suggestions for other things like PRs/issues/commands/topics/notes/groups/teams/channels/wikis this is fine. There is already one [experiment for using it with commands](https://www.tiptap.dev/experiments/commands).

ep1kt3t0s:
@philippkuehn I saw that example page but I did not see an example for React.
Is there any documentation on how to use it with React?

PR #1845, opened 2021-09-06
fix code-block language attribute value

TooBug:
If paste a piece of html with highlight, the HTML may be like this:

```html
<pre><code class="hljs language-javascript">
...
</code></pre>
```

Then, this extension will try to get the language from className, which results in `hljs language-javascript`, but actually it should be `javascript`.

And then, when render this code block, the className becomes `language-hljs language-javascript`, which leads to a very strange result.

This PR get class list via DOM API `.classList`, and see if a className starts with `languageClassPrefix`, if it is, then try to get the language value.

### DIFF
--- a/packages/extension-code-block/src/code-block.ts
+++ b/packages/extension-code-block/src/code-block.ts
@@ -47,17 +47,22 @@ export const CodeBlock = Node.create<CodeBlockOptions>({
       language: {
         default: null,
         parseHTML: element => {
-          const classAttribute = element.firstElementChild?.getAttribute('class')
-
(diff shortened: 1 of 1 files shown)


philippkuehn:
Hey, thanks for that bugreport/PR. Since there were still some problems with some linting rules/code style, I fixed that in the following commit: https://github.com/ueberdosis/tiptap/commit/472f2139c52b9e0211ccd735e937486df2cb3d30

TooBug:
@philippkuehn nice work!

PR #1859, opened 2021-09-08
feat: add extendEmptyMarkRange option to mark commands

philippkuehn:
This PR adds an option `extendEmptyMarkRange` to the `unsetMark` and `toggleMark` commands. This defaults to `false` now (until now it was `true`). This options is useful for the link mark, to remove a link even if there is an empty selection but it’s rather confusing for other marks.

Thanks for @isubasti for the main idea (#1650).

### DIFF
--- a/packages/core/src/commands/toggleMark.ts
+++ b/packages/core/src/commands/toggleMark.ts
@@ -9,17 +9,27 @@ declare module '@tiptap/core' {
       /**
        * Toggle a mark on and off.
        */
-      toggleMark: (typeOrName: string | MarkType, attributes?: Record<string, any>) => ReturnType,
+      toggleMark: (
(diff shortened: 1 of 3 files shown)


isubasti:
Hi @philippkuehn thanks for this. Though there is one more corner case on this one but haven't got a chance to think about it. it's for things like two link marks that are next to each other with different `attrs` then if the empty selection is on the first one then only the first one should be removed instead of removing mark for both.

philippkuehn:
ah you are right! will look at this later! (we just need to pass the current mark attributes to `getMarkRange`)

philippkuehn:
should be fixed with [this patch](https://github.com/ueberdosis/tiptap/commit/a5dca087eee3da81818e370de88cd29a93c3f67b).

PR #1875, opened 2021-09-09
feat: add getText() and generateText() methods (fix #1428)

### LINKED ISSUE #1428: Create function for transforming JSON state to text representation
**The problem I am facing**
I let my users send newsletters to their audience (think Substack). 

We store their custom rich text body as JSON and then when we want to send the email, we need to convert the message into HTML and text. 

**The solution I would like**
I would like a function `generateText` that converts a JSON state to text. 

This could be done in two ways:

- the function `generateText` looks at nodes and converts them to text
- we add a `toText` option on every Node / Mark in the schema and call that in order to convert a node to text

**Alternatives I have considered**

I am not using TipTap yet, but I wrote my own wrapper around ProseMirror and created this file to serialize our JSON state to HTML / text — https://gist.github.com/vpontis/e5e373ea9950b413c18c867a5aedf352

philippkuehn:
This PR adds two methods `getText()` and `generateText()`. You can define text rendering of nodes with the `renderText` config. This is mostly useful for atom nodes like hard breaks, mentions etc.

```js
const HardBreak = Node.create({
  name: 'hardBreak',

  renderText() {
    return '\n'
  },

  // ...
})
```

### usage of `getText()`

```js
// basic usage
const text = editor.getText()

// with some options
const text = editor.getText({
  // define a custom block separator if you want to
  blockSeparator: '\n',
  // overwrite `renderText` method of nodes
  textSerializers: {
    hardBreak: () => '\n',
  },
})
```

### usage of `generateText()`

```js
import { generateText } from '@tiptap/core'
import HardBreak from '@tiptap/extension-hard-break'

const json = {} // ProseMirror JSON

// basic usage
const text = generateText(json, [
  Document,
  Paragraph,
  Text,
  HardBreak,
  // other extensions …
])

// with some options
const text = generateText(json, [
  Document,
  Paragraph,
  Text,
  HardBreak,
  // other extensions …
], {
  // define a custom block separator if you want to
  blockSeparator: '\n',
  // overwrite `renderText` method of nodes
  textSerializers: {
    hardBreak: () => '\n',
  },
})
```


lostdesign:
Oh damn 🙌 Very nice addition. Comes in handy as I won't need the tokeniser for flexsearch anymore!

BrianHung:
@philippkuehn I think the [`clipboardTextSerializer`](https://prosemirror.net/docs/ref/#view.EditorProps.clipboardTextSerializer) should only be serializing the given slice as opposed to the entire doc.

philippkuehn:
@BrianHung But I serialize only a part of the doc based on the current selection. Is there any issue with the current solution?

BrianHung:
> But I serialize only a part of the doc based on the current selection. Is there any issue with the current solution?

Ah nevermind; I see the `range = {from, to}`  as opposed to the `slice`.

tsl1127:
Is there any way to make the positions of the text output by getText() the same as the positons in the editor? Especially the text in the table

MahirAlam:
Is there any way to limit text generation?
like generate 100 words then return?

nperez0111:
nope @MahirAlam

PR #1882, opened 2021-09-12
Fix "destory" method in view plugins.

KaneCohen:
Update `destory` methods on BubbleMenu and FloatingMenu plugins to correctly remove event listeners.

### DIFF
--- a/packages/extension-bubble-menu/src/bubble-menu-plugin.ts
+++ b/packages/extension-bubble-menu/src/bubble-menu-plugin.ts
@@ -190,7 +190,7 @@ export class BubbleMenuView {
 
   destroy() {
     this.tippy?.destroy()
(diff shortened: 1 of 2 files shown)


philippkuehn:
what do you mean by `correctly remove event listeners`? is there any issue that you can explain? thanks!

KaneCohen:
Mousedown event is added with a "capture" flag:

```ts
 this.element.addEventListener('mousedown', this.mousedownHandler, { capture: true })
```

In order to remove event listener same "signature" of the event should be used, including capture flag. Currently this is not the case:
```ts
this.element.removeEventListener('mousedown', this.mousedownHandler)
```

Which leads to mousedown event not being removed.

multiple events

philippkuehn:
ah great! thanks!

PR #1910, opened 2021-09-18
feat(Suggestion): Pass additional props to suggestion render fns

### LINKED ISSUE #1909: Suggestion Extension - Pass additional data to the render methods
**The problem I am facing**
We're using the @tiptap/suggestion extension to power a floating menu (Following the example from the docs for @mentions https://tiptap.dev/api/nodes/mention)

Were noticing that whenever another doc editor types above the current user, the tippy instance is recreated due to both `onExit` and then immediately after `onStart` being called: 
https://github.com/ueberdosis/tiptap/blob/main/packages/suggestion/src/suggestion.ts#L135

### Collaborative Editing

It's a little weird that `onExit` gets called anytime someone else types above you earlier in the doc. There may be a better way to handle this case, but I'm not sure what that is.

The proposed change here will allow consumers implementing `onExit` to check if the new state of the plugin is still `active` (and decide to no-op in that case)

**The solution I would like**
Additional props passed to the suggestion render methods (`SuggestionProps`) to provide more insight into the current state of the plugin instance.

**Alternatives**

With the addition of the PluginKey option for Suggestion, I realized I can work around this by providing my own key and looking up the state myself. I think the suggested method of passing the props would still be preferred though.

**_Note that `onStart` lazily creates the popup/renderer_**

```ts
  onExit(props: SuggestionProps) {
    const next = MenuKey.getState(props.editor.view.state)
    if (next.active !== true) {
      popup?.destroy()
      reactRenderer?.destroy()

      popup = null
      reactRenderer = null
    }
  },
```

jamesopti:
# Changes

- Pass the following additional props to the `@tiptap/suggestion` render fns:
  - `active: boolean`
  - `moved: boolean`
  - `started: boolean`
  - `stopped: boolean`
  - `changed: boolean`

This enables a higher fidelity of decision making in consumers' implementation of `onStart`, `onUpdate` and `onExit`

# Context

Fixes #1909

### DIFF
--- a/packages/suggestion/src/suggestion.ts
+++ b/packages/suggestion/src/suggestion.ts
@@ -35,6 +35,11 @@ export interface SuggestionProps {
   range: Range,
   query: string,
   text: string,
+  active: boolean,
+  moved: boolean,
+  started: boolean,
+  stopped: boolean,
+  changed: boolean,
   items: any[],
   command: (props: any) => void,
   decorationNode: Element | null,
@@ -98,6 +103,11 @@ export function Suggestion({
             range: state.range,
             query: state.query,
             text: state.text,
+            active: next.active,
+            moved,
+            started,
+            stopped,
+            changed,
             items: (handleChange || handleStart)
               ? await items(state.query)
               : [],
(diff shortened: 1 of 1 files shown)


packages/suggestion/src/suggestion.ts
```diff
@@ -98,6 +103,11 @@ export function Suggestion({
             range: state.range,
             query: state.query,
             text: state.text,
+            active: state.active,
```
jamesopti, line 106: I actually think it might be more correct to always use `next.active` here

PR #1917, opened 2021-09-20
Fix: code-block-lowlight child extensions do not highlight code

enriquecastl:
This MR fixes a bug in the Code Block Lowlight extension that doesn’t allow extending the extension further. When we extend `CodeBlockLowlight`, the code in the child extension is not highlighted. We discovered that the root cause of this problem is a hard coded extension named provided to the `LowlightPlugin`. Passing the extension name fixes this problem. 

This PR also contains a failing test.


philippkuehn:
great! thanks!

PR #1922, opened 2021-09-21
Use pasteRegex in addPasteRules

jvissers:
Seems like inputRegex was used in addPasteRules by mistake (?)


philippkuehn:
oh great! thanks!

Issue #1107, opened 2021-04-21
Input Rules not recognized for code blocks


itsnwa:
I'm trying to type ``` to create a code block, but it doesn't get recognized, while creating code blocks with the keyboard shortcut works as expected.

Using Vue 3 with Composition API.

<img width="397" alt="CleanShot 2021-03-11 at 11 03 24@2x" src="https://user-images.githubusercontent.com/19958806/110770272-6ef23780-8259-11eb-8d6a-2de82937429a.png">

fengzilong:
@itsnwa I just send a PR that will solve your problem( support `enter`, heading to the PR link above )

PR #1940, opened 2021-09-25
feat: trigger inputrule on enter for code block

fengzilong:
```
```{space|enter}
```

Support using {space} or {enter} to trigger inputrule

related issue: https://github.com/ueberdosis/tiptap/issues/1107

### DIFF
--- a/packages/extension-code-block/src/plugin.ts
+++ b/packages/extension-code-block/src/plugin.ts
@@ -1,0 +1,45 @@
+// Modified from
+// https://github.com/ProseMirror/prosemirror-inputrules/blob/master/src/inputrules.js
+// license: MIT
+
+import { InputRule } from 'prosemirror-inputrules'
+import { Plugin as ProseMirrorPlugin } from 'prosemirror-state'
+import type { Plugin, TextSelection } from 'prosemirror-state'
+import type { EditorView } from 'prosemirror-view'
+
+const MAX_MATCH = 500
+
+function run(view: EditorView, from: number, to: number, text: string, rules: any[], plugin: Plugin) {
+  if (view.composing) return false
+  const state = view.state; const
+    $from = state.doc.resolve(from)
+  if ($from.parent.type.spec.code) return false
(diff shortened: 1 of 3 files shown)


packages/extension-code-block/src/plugin.ts
```diff
@@ -0,0 +1,45 @@
...
+    }))
+    return true
+  }
+  return false
+}
+
+export default function (options: any = {}) {
+  const rules: InputRule[] = options.rules || []
+  const plugin: Plugin = new ProseMirrorPlugin({
+    props: {
```
BrianHung, line 34: This is missing the `compositionend` event in the original [`inputrule`](https://github.com/ProseMirror/prosemirror-inputrules/blob/master/src/inputrules.js).

fengzilong: https://github.com/ProseMirror/prosemirror-inputrules/pull/6

Because marijnh said {enter} is not a kind of inputrule, so I made a standalone plugin

fengzilong: The origin inputrule plugin is not removed, IMO there is no necessary to add `compositionend` here

BrianHung: > I'm not sure whether `the origin plugin state` is needed here

The plugin state is kept there so that the `undoInputRule` function (which is binded to `backspace`) can work using the `setMeta` and `getMeta` fields.

https://prosemirror.net/docs/ref/#state.Transaction.setMeta
https://prosemirror.net/docs/ref/#state.Transaction.getMeta

> Because marijnh said {enter} is not a kind of inputrule, so I made a standalone plugin

I think you could argue enter is a type of inputrule, but where the trigger isn't an explicit key being entered, but an implicit key being pressed; plus, you're still using the `InputRule` abstraction, just handling / running it differently.

philippkuehn:
I currently working on a re-write of Inputrules and Pasterules. I’ve already implemented this feature in that branch [based on this solution](https://discuss.prosemirror.net/t/trigger-inputrule-on-enter/1118/5) (thanks @BrianHung 😀). So I don’t want to merge this. Thank you anyway!

PR #1954, opened 2021-09-28
demos: fix export json content

cncolder:
Fix the incorrect json content ~[object Object]~

https://tiptap.dev/guide/output#option-1-json

Screenshot 2021-09-29 01 48 31


philippkuehn:
Oh, didn’t know that this was broken. We moved from Vue 2 to Vue 3 for the demos und Vue 3 handles this case a bit different. I fixed it [a bit cleaner](https://github.com/ueberdosis/tiptap/commit/50b6e6c1c6d5780b19743aa4bd32cb159369d79b) I think. But thanks anyway!

PR #1959, opened 2021-09-29
Added better types for event emitter

### LINKED ISSUE #1898: Add better types for event emitter
`editor.on()` and `editor.off()` need some better types, so it will show the correct event names and props instead of a generic `string` and `Function`.

HuiiBuh:
Fixes #1898

+ Made the `EventEmitter` generic to be able to specifically type the events
+ Created an interface `EditorEvents` for the event types
+ Updated the types in `EditorOptions` to use the types provided by the new `EditorEvents` interface

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,5 @@ yarn-error.log*
 
 tests/cypress/videos
 /tests/cypress/screenshots
+# Ignore intellij project files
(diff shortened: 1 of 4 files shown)


philippkuehn:
great! thank you very much! ✌️

Issue #1747, opened 2021-08-18
(React) Content of nested NodeViews will not render until focus


WilliamIPark:
**Description**
When nesting NodeViews, the content of the deeper node view will not render until the editor gains focus.

**Steps to reproduce the bug**
1. Go to https://codesandbox.io/s/tiptap-v2-beta-rendering-bug-yfftj
2. Notice no text content in the blue box (field node)
3. Click into the blue box.
4. Content will appear.

**CodeSandbox**
I created a CodeSandbox to help you debug the issue:
* React: https://codesandbox.io/s/tiptap-v2-beta-rendering-bug-yfftj

**Expected behavior**
The text content should be visible on load.

**Environment?**
 - operating system: MacOS
 - browser: Chrome / FireFox / Safari all reproduce the issue.
 - tiptap version: 2.0.0-beta.99

WilliamIPark:
Some additional details here, including how I worked around the issue in certain cases, but I've run into a case where the workaround won't suit:
https://github.com/ueberdosis/tiptap/discussions/1675

philippkuehn:
I can confirm this issue. It seems to be fixed with this [PR](https://github.com/ueberdosis/tiptap/pull/1470). But since that would be a breaking change, I'll have to take another look.

philippkuehn:
It’s fixed in the latest version of `@tiptap/react`.

PR #1960, opened 2021-09-29
Use ref to move contentDOM

### LINKED ISSUE #1942: Issue with ReactNodeViewRenderer and tippyjs-react caused by extension-collaboration
**Description**

When using collaboration with a custom node and tippy-react, the following error occurs
`Cannot read properties of undefined (reading 'element')`

1. Create a custom node rendered by ReactNodeViewRenderer 
2. Make sure node contains **NodeViewContent**
3. Add a basic @tippyjs/react component
4. Add custom extension to the editor
5. Add collaboration to editor

**CodeSandbox**

https://codesandbox.io/s/tiptap-react-forked-ewun2?file=/src/custom-node/Component.tsx:703-726

**Environment?**
 - operating system:
 - browser:
 - mobile/desktop:
 - tiptap version: 2.0 beta 66

**Additional context**
If I conditionally render the tippy-react component, it seems to work. Refer custom-node/Component.tsx for an example.

Thank you so much for this wonderful editor, this is beyond amazing on what we can build with it.

philippkuehn:
Instead of calling `maybeMoveContentDOM` manually, we are using a `ref`.

fixes #1942, which was broken because of[ this patch](https://github.com/ueberdosis/tiptap/commit/b15a8a8683bafbc9fd57dc64d60329f9ab08ecfd) (which was a fix for #1747). #1747 is still fixed with this PR.

@YousefED I had to remove `forwardRef` for `NodeViewContent` (which was a [merged PR from you](https://github.com/ueberdosis/tiptap/pull/1452)) for that so I would like to hear you thoughts about this change. I’m still not quite sure what you needed this ref for. Can you give me an example?

### DIFF
--- a/packages/react/src/NodeViewContent.tsx
+++ b/packages/react/src/NodeViewContent.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect } from 'react'
+import React from 'react'
 import { useReactNodeView } from './useReactNodeView'
 
 export interface NodeViewContentProps {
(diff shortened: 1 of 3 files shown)


YousefED:
The `ref` was used in combination with https://github.com/ueberdosis/tiptap/pull/1470/files, which would refactor the React nodeviews to avoid usage of `this.dom.querySelector('[data-node-view-content]')`.

If I understand this PR correctly, you've come to a similar approach (remove the hacky queryselector) as https://github.com/ueberdosis/tiptap/pull/1470/files, but use Context as a way to pass the ref, instead of forcing consumers to pass along the ref. Your approach has the advantage that it doesn't change the API, at the cost of a little less easy to grasp pattern using Context. I think that's a fair trade-off so if this has been tested well, I'd suggest to merge it!

philippkuehn:
@YousefED thanks! :)

PR #1990, opened 2021-10-07
Add extensions for settings font size and line height of spans.

### LINKED ISSUE #388: Font Size, Font Family, Text Color
Hi, this editor really lacks of very important features of rich text editor like Font Size, Font Family and Text Color. There is also no examples of how this can be implemented. Currently I'm trying to find a way to do this, but still have no luck. Could you at least show some working examples of how this can be achieved?

### LINKED ISSUE #887: Option for paragraph line height
Implementing a button menu bar that would have options to line height would solve my problem.
[check this out](https://images.app.goo.gl/ohWJxuVDRJmipUvr7)
[and this](https://images.app.goo.gl/KJAjewsUj2crQSJc9)

mjamro:
I've added two extensions - one for setting `font-size` and one for `line-height`. They're working in the same way as `extension-font-family` or `extension-text-align`.

```
  {
    "type": "text",
    "marks": [
      {
        "type": "textStyle",
        "attrs": {
          "fontFamily": "Constantia",
          "fontSize": "5px",
          "lineHeight": "1"
        }
      }
    ],
    "text": "k;k;l"
}
```

Resolves #388, resolves #887

### DIFF
--- a/packages/extension-font-size/CHANGELOG.md
+++ b/packages/extension-font-size/CHANGELOG.md
@@ -1,0 +1,8 @@
+# Change Log
+
+All notable changes to this project will be documented in this file.
+See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
+
+# [2.0.0-beta.1](https://github.com/ueberdosis/tiptap/compare/@tiptap/extension-font-size@2.0.0-beta.1...@tiptap/extension-font-size@2.0.0-beta.1) (2021-10-07)
+
+Initial version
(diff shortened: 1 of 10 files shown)


BroJenuel:
any update on this?

Daniel-Sudhindaran (approved):

Daniel-Sudhindaran:
I agree with @svkale. Can we just add fontSize for now and handle the rest in an other PR?

gregveres:
> Thanks for the PR! I think the LineHeight should work like TextAlign and register an attribute with nodes, not with the TextStyle extension. Also docs and tests are missing.
> 
> But don’t update the PR, it’ll probably take a while until we work on that one and we can keep it from there then. ✌️

I have created a gist for a line-height extension based on TextAlign as suggested. You can find it here:
https://gist.github.com/gregveres/8757756d56becc2c053c46540cb6b314

I had already created gist for font-size based on the existing font-family extension
https://gist.github.com/gregveres/64ec1d8a733feb735b7dd4c46331abae

PR #1997, opened 2021-10-08
Integrate input rules and paste rules into the core

### LINKED ISSUE #1872: Pasting links with comma / Use real link parser
**Description**

#881 #824 seems to have appeared again in tiptap 2

1. Go to https://www.tiptap.dev/api/marks/link/ demo
2. Paste ```https://example.com/test_(123)?a=b,c``` or ```https://example.com/test?a=b,c``` ```https://example.com/test_(123)```
3. Link is highlighted only partially

Adding comma to the regexps in link.ts seems to work for this, but breaks pasting text with commas after links, like
```https://test.ru, text```

Is it possible to integrate a real link parser like https://soapbox.github.io/linkifyjs/ instead of these regexps? +5kb seems not too much for accurate link paste.

Seems to work fine for me like this (in link plugin, only relevant parts):
```
import { find } from 'linkifyjs'

const linkifyHandler = (type, fragment, parent) => {
  const nodes = []

  fragment.forEach(child => {
    if (child.isText && child.text) {
      const { text } = child

      const links = find(text).filter((lnk) => lnk.isLink)

      if (links.length > 0) {
        let pos = 0

        links.forEach((link) => {

          if (link.start > 0) {
            nodes.push(child.cut(pos, link.start))
          }

          const attrs = {
            target: "_blank",
            href: link.href,
          }

          // adding the markdown part to nodes
          nodes.push(child
            .cut(link.start, link.end)
            .mark(type.create(attrs).addToSet(child.marks)))

          pos = link.end
        })

        // adding rest of text to nodes
        if (pos < text.length) {
          nodes.push(child.cut(pos))
        }
        return true
      }
    }

    nodes.push(child.copy(linkifyHandler(type, child.content, child)))
  })

  return Fragment.fromArray(nodes)
}

export const Link = Mark.create({
  name: 'link',
  ....
  addPasteRules() {
    return [
      new Plugin({
        key: new PluginKey('markPasteRule'),
        props: {
          transformPasted: slice => {
            return new Slice(linkifyHandler(this.type, slice.content), slice.openStart, slice.openEnd)
          },
        },
      })
    ]
  },
}
```

### LINKED ISSUE #1889: Don't apply typography when inside the Code mark
When I'm inside a `<code>` mark, I usually don't want to replace `--` with a dash, or have curly quotes, etc. It would be nice to disable typography there.

The CodeBlock node already features this behavior.

philippkuehn:
With this PR we are removing the `prosemirror-inputrules` dependency and adding a custom implementation into the core.

## Features
- instead of a regex, we can now add a custom handler to find a match. This is useful for using complex libraries like [linkified](https://github.com/Hypercontext/linkifyjs) for link detection where a simple regex is not enough (fixes #1872)
- paste rules are applied after the content is pasted instead of before. This has the advantage that you have access to the state instead of just a slice. Also, it is possible that input rules and paste rules share logic.
- input rules are not executed within code marks (fixes #1889)
- input rules can now also be executed after enter. This is useful for creating code blocks with ```{enter}. For using that you have to put `[\s\n]$` at the end of your input rule regex.
- all input rules of `prosemirror-inputrules` are exported by `@tiptap/core` now. There are also some custom input rules and paste rules. You can use:
  - `markInputRule`
  - `nodeInputRule`
  - `textblockTypeInputRule`
  - `textInputRule`
  - `wrappingInputRule`
  - `markPasteRule`
  - `textPasteRule`

## Breaking changes

- New syntax. Instead of multiple parameters all input rules and paste rules expects an object for better readability:
  ```js
  // old
  markInputRule(/(?:^|\s)((?:\*\*)((?:[^*]+))(?:\*\*))$/gm, this.type)

  // new
  markInputRule({
    find: /(?:^|\s)((?:\*\*)((?:[^*]+))(?:\*\*))$/gm,
    type: this.type,
  })
  ```
- incompatibility to `prosemirror-inputrules`. Because we provide these new feature, you can’t put input rules of `prosemirror-inputrules` into `addInputRules()` in your extensions. If you still want to use these input rules you have to add them via the `inputRules` plugin exported by `prosemirror-inputrules`.


wengtytt:
> * If you still want to use these input rules you have to add them via the `inputRules` plugin exported by `prosemirror-inputrules`

Hi there, is there any example how can we add old prosemirror-inputrules via the inputRules plugin exported by prosemirror-inputrules to the custom extension?

PR #2005, opened 2021-10-08
Fix savvy example using new textInputRule

jakedolan:
The Savvy example needs to be updated due to the breaking changes in pull request #1997.


jakedolan:
Closing/resubmitting to clean up commit tree. Also want to look into that linting error.

PR #2006, opened 2021-10-09
Fix savvy example using new textInputRule

jakedolan:
The Savvy example needs to be updated due to the breaking changes in pull request #1997.


jakedolan:
My apologies for the spam. It is the end of the week and apparently I should have done this through a local repo and not the GitHub site tool. I should have done that to start, but I was initially thinking it was just a documentation update.

I don't want to further spam this so will close due to lint error (I think this is related to the fork->branch). You can see the code that needs to be updated. Happily will submit again after the weekend from a local repo if needs that follows the [contributing instructions](https://tiptap.dev/overview/contributing/) more closely.

PR #2011, opened 2021-10-10
Improve `ReactRenderer` types

rfgamaral:
This is an attempt to convert the "Mentions" community example to TypeScript. I'm still new to tiptap's codebase, so forgive me if I'm not doing things properly.

Unfortunately I wasn't able to fix every error, so this PR is a bit incomplete. I'm hoping someone can help me figure out the remaining bits:

image
image

### DIFF
--- a/packages/react/src/ReactRenderer.tsx
+++ b/packages/react/src/ReactRenderer.tsx
@@ -23,12 +23,12 @@ export interface ReactRendererOptions {
   as?: string,
 }
 
-type ComponentType =
+type ComponentType<R> =
   | React.ComponentClass
   | React.FunctionComponent
-  | React.ForwardRefExoticComponent<{ items: any[], command: any } & React.RefAttributes<unknown>>
+  | React.ForwardRefExoticComponent<{ items: any[], command: any } & React.RefAttributes<R>>
 
-export class ReactRenderer {
+export class ReactRenderer<R = unknown> {
   id: string
 
   editor: ExtendedEditor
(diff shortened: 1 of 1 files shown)


hanspagel:
Thanks for the effort! We decided to not use TypeScript in the examples though. My impression is that only 10 % or less use TypeScript, and they are confused by something they don’t know, but TS users are able to add types on their own.

@philippkuehn What do you think?

rfgamaral:
> Thanks for the effort! We decided to not use TypeScript in the examples though. My impression is that only 10 % or less use TypeScript, and they are confused by something they don’t know,

That's understandable, I can revert the example to JS, no problem.

However, this PR also fixes a few types in `ReactRenderer` to be able to properly type our client apps. Could you review that, please? I'm not 100% I'm doing it right.

Also...

> but TS users are able to add types on their own.

Almost there 😅, as I have no idea how to fix the errors above. Can you help with that? Maybe further changes to the source types are required? 🤔

philippkuehn:
> @philippkuehn What do you think?

Same!

> However, this PR also fixes a few types in ReactRenderer, necessary to properly type our client apps. Could you review that, please? I'm not 100% sure about these changes.

Looks good to me. Could you please update the PR to contain only these changes or create a new one?

rfgamaral:
> > However, this PR also fixes a few types in ReactRenderer, necessary to properly type our client apps. Could you review that, please? I'm not 100% sure about these changes.
> 
> Looks good to me. Could you please update the PR to contain only these changes or create a new one?

Yes, I can update the PR, but those errors need to be fixed first. They show that the changes I'm making here are not finished/correct.

PR #2018, opened 2021-10-11
Add `editor` dependency when registering `BubbleMenuPlugin`

ValentaTomas:
When we are initializing editor via the `useEditor` hook with dependencies the `BubbleMenu` component is only registered the first time the editor is initialized.

Adding editor to the dependency array registering/unregistering the `BubbleMenuPlugin` fixes this. (I tested exactly this code in our project.)

I also added a check that ensures that the menu element referenced via the `useRef` is defined when registering the plugin - otherwise, there is no point in registering the plugin.


philippkuehn:
thanks!

PR #2069, opened 2021-10-21
Add extension storage

philippkuehn:
At some point you probably want to save some data within your extension instance or want to provide some methods that are accessible outside of the extension. With this PR it’s possible now.

The stored data is mutable. You can access it within the extension under `this.storage`.

```js
import { Extension } from '@tiptap/core'

const CustomExtension = Extension.create({
  name: 'customExtension',

  addStorage() {
    return {
      awesomeness: 100,
    }
  },

  onUpdate() {
    this.storage.awesomeness += 1
  },
})
```

When using Vue or React, this storage is reactive within your templates.

Outside the extension you have access via `editor.storage`. Make sure that each extension has a unique name.

```js
const editor = new Editor({
  extensions: [
    CustomExtension,
  ],
})

const awesomeness = editor.storage.customExtension.awesomeness
```

When using TypeScript you can define a type for your storage.

```ts
import { Extension } from '@tiptap/core'

export interface CustomExtensionStorage {
  awesomeness: number,
}

const CustomExtension = Extension.create<{}, CustomExtensionStorage>({
  addStorage() {
    return {
      awesomeness: 100,
    }
  },
})
```

Global storage types can’t be inferred so you have to give a hint:

```ts
const storage = editor.storage.customExtension as CustomExtensionStorage
```

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -31,7 +31,7 @@ module.exports = {
       extends: [
         'plugin:@typescript-eslint/eslint-recommended',
         'plugin:@typescript-eslint/recommended',
-        'plugin:vue/strongly-recommended',
+        'plugin:vue/vue3-strongly-recommended',
         'airbnb-base',
       ],
       rules: {


packages/react/src/useEditor.ts
```diff
@@ -17,7 +17,13 @@ export const useEditor = (options: Partial<EditorOptions> = {}, deps: Dependency
 
     setEditor(instance)
 
-    instance.on('transaction', forceUpdate)
+    instance.on('transaction', () => {
+      requestAnimationFrame(() => {
+        requestAnimationFrame(() => {
```
cusxio, line 22: @philippkuehn is there a reason why we need to call `requestAnimationFrame` within a `requestAnimationFrame` ? Why is 1 `requestAnimationFrame` not sufficient?

philippkuehn: I always do a double RAF because of [this](https://stackoverflow.com/questions/44145740/how-does-double-requestanimationframe-work). Maybe it’s not needed here. Is there any issue with that?

PR #2119, opened 2021-11-02
feat: Allow array of extensions for `enableInputRules` and `enablePasteRules`

aguingand:
### Context
In multiple project I wan't to enable only input / paste rules I need but there is no easy way to do it. (e.g I only want to paste links)

### Feature
Allow to pass an array of extension or extension name for `enableInputRules` and `enablePasteRules` editor options:

```js
new Editor({
  extensions: [
    StarterKit,
    Link,
  ],
  enablePasteRules: [Link],
  enableInputRules: ['horizontalRule'],
})
```

### Comment
I will update docs if it's ok for you 👍


philippkuehn:
Thanks!

PR #2153, opened 2021-11-09
update getJSON return type to JSONContent

lecstor:
Just a small typescript tweak to the `getJSON` return type.
If there is a reason it wasn't this way, feel free to close and I'll know not to type my onUpdate handlers with it. 😄 

Screen Shot 2021-11-10 at 8 38 26 am


philippkuehn:
yeah let’s do this! thanks!

PR #2155, opened 2021-11-10
update zeed dom

floriankrueger:
As @holtwick did [fix the underlying issue](https://github.com/holtwick/zeed-dom/issues/2#issuecomment-964329350) in zeed dom I thought it would take some weight off of your shoulders if I would prepare the update of that dependency.


philippkuehn:
Great. Thank you!

PR #2166, opened 2021-11-15
update ReactRenderer to include CSS class

kuka:
Includes extension name to root `react-renderer` node so that styling can be done directly.

### DIFF
--- a/packages/react/src/ReactRenderer.tsx
+++ b/packages/react/src/ReactRenderer.tsx
@@ -50,6 +50,9 @@ export class ReactRenderer<R = unknown> {
     this.props = props
     this.element = document.createElement(as)
     this.element.classList.add('react-renderer')
+    if (props.extension.name) {
+      this.element.classList.add(props.extension.name)
+    }
     this.render()
   }
 


philippkuehn:
Not sure about this one. You can add a class to your component within ReactRenderer, right?

kuka:
Thanks! 

Yeah, @philippkuehn we can do that but the tree looks like `.react-renderer > .custom-component`. So the node that has `.react-renderer` can’t be modified via CSS. My use case is that I have two `.react-renderer` components and the visuals should be conditionally modified with:

```css
.first-component + .next-component {
  margin-top: -3rem;
}
```

in this structure:

```html
<div class="react-renderer first-component">
  …
</div>
<div class="react-renderer next-component">
  …
</div>
```

This CSS above doesn’t work right now because going up the tree in CSS is not possible so this CSS below doesn’t have access to the `.next-component`:

```css
.react-renderer > .first-component + ?
```

as structure is:

```html
<div class="react-renderer">
  <div class="first-component"></div>
</div>
<div class="react-renderer">
  <div class="first-component"></div>
</div>
```

---

It’s possible to do it programmatically but not ideal, thus this PR.

Let me know if this makes sense.

hanspagel:
Thanks for the PR!

The solution feels somehow wrong to me. If we’ll add support for it, it should be configureable and maybe even off by default. 

Another nitpick: We use lowerCamelCase for extension names, so the class names would end up like `react-renderer taskItem`, which I couldn’t take. 😅

kuka:
Ok, that’s fair @hanspagel. What about adding extension name as a `data-extension` attr then?

thatsjonsense:
We're currently using a decoration to achieve this as a workaround, but I think it would be easier and more performant if Tiptap did this out of the box. Here's our custom plugin for reference:

```
Extension.create({
  addProseMirrorPlugins() {
    return [
      new Plugin({
        props: {
          decorations: ({ doc }) => {
            const decorations: Decoration[] = []
            const decorate = (node, pos) => {
              decorations.push(
                Decoration.node(pos, pos + node.nodeSize, {
                  class: `node-${node.type.name}`,
                })
              )
            }
            doc.descendants(decorate)
            return DecorationSet.create(doc, decorations)
          },
        },
      }),
    ]
  },
})
```

philippkuehn:
Thanks for this PR but instead of adding the extension name to the `ReactRenderer` I added a [className option](https://github.com/ueberdosis/tiptap/commit/c9dc1e1ec3295fe3dce7ed7d314c5a7e4140af83). Within `ReactNodeViewRenderer` I use this option to add a class name based on the extension name.

kuka:
Awesome, thanks @philippkuehn

Issue #2068, opened 2021-10-21
Optionally sanitize links rendered by the links extension


joelreske:
### What problem are you facing?

XSS (cross-site scripting) is a vulnerability where user input can hijack the intended functionality of a page via scripting.

Working with TipTap, we sometimes have untrusted users use the editor. Those users could create a link with an href like `javascript:alert(123)`. Then, an unsuspecting user on the other end viewing the document could click the link and see the alert.

As a best practice then, when rendering links, they should be sanitized (assuming you don't want users to be able to inject JS).

### What’s the solution you would like to see?

I'd like to provide an option to the Link mark called `sanitizeHref` or something similar which either:
- is a boolean which, if true, protects from these types of attacks using the `@braintree/sanitize-url` package
- is a function which, if provided, does sanitation using whatever library the end user would like

Happy to follow guidance based on whatever the maintainers prefer! Not sure of the sensitivity to adding new packages, but it probably makes it easier to opt in if it's just a boolean.

I would also ideally like this functionality to be enabled by default (possible with the boolean, package-included solution)! I think many users are probably not thinking of this vulnerability, and would have to have a pretty specific use case to disable.

### What alternatives did you consider?

Currently, I'm implementing this downstream by extending the `Link` mark and overriding the render properties. This works pretty well and is not too difficult! But ideally this would be available upstream.

### Anything to add? (optional)

I'm happy to contribute and put up a PR! This library has been hugely helpful to us over at @Ashbyhq , we have a few more feature ideas we'll be happy to contribute upstream.

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

PR #2169, opened 2021-11-16
Optionally sanitize links rendered by the links extension

joelreske:
closes https://github.com/ueberdosis/tiptap/issues/2068

- Allows a sanitize function to be provided to the links extension, with a quick explanation of why it's important
- Adds sanitization with a common library for the demo

### DIFF
--- a/demos/package.json
+++ b/demos/package.json
@@ -8,6 +8,7 @@
     "ts": "tsc --project tsconfig.base.json --noEmit && tsc --project tsconfig.react.json --noEmit && tsc --project tsconfig.vue-2.json --noEmit && tsc --project tsconfig.vue-3.json --noEmit"
   },
   "dependencies": {
+    "@braintree/sanitize-url": "^5.0.2",
     "d3": "^7.1.1",
     "fast-glob": "^3.2.7",
     "remixicon": "^2.5.0",
--- a/demos/src/Marks/Link/React/index.jsx
+++ b/demos/src/Marks/Link/React/index.jsx
@@ -5,6 +5,7 @@ import Paragraph from '@tiptap/extension-paragraph'
 import Text from '@tiptap/extension-text'
 import Link from '@tiptap/extension-link'
 import Code from '@tiptap/extension-code'
+import { sanitizeUrl } from '@braintree/sanitize-url'
 import './styles.scss'
 
 export default () => {
@@ -15,6 +16,7 @@ export default () => {
       Text,
       Link.configure({
         openOnClick: false,
+        sanitizeUrl,
       }),
       Code,
     ],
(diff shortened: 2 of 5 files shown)


hanspagel:
Thanks! Naive question: Is it a good idea to do this on every re-render? Or should this be done initially and then as part of the commands?

joelreske:
Hi! A bit of a tradeoff - I think having it in the render step is the safest. There would be no way to get around it, whereas I think with commands if a new one was registered they would all have to take sanitization into account? Definitely could be wrong there though.

In terms of performance, I'd be interested in a caching option (could save the key and resolved value on the mark object I think?).

These sanitize functions are really meant to be very simple string operations[ (example here) ](https://github.com/braintree/sanitize-url/blob/main/src/index.ts) so I'm not very concerned bit it might be good.

philippkuehn:
Hmm, alternatively it would be possible to track every inserted link (as a plugin – not within commands) so we can sanitize it once if needed. With your solution a malicious link would still be present in the document forever. What do think?

joelreske:
> Hmm, alternatively it would be possible to track every inserted link (as a plugin – not within commands) so we can sanitize it once if needed. With your solution a malicious link would still be present in the document forever. What do think?

An interesting idea! I do think that we could solve a similar problem by having a way of generally sanitizing stored documents, potentially on edit, but I do think of this solution being a bit differently.

This PR is really meant to be a 'last line of defense' for harmful links, the user of the option shouldn't have to worry about whether a sanitization step was done before the document being rendered was stored. The concern with XSS is not so much that we always guarantee sanitized data in our DB, but rather that we are protected from malicious data that might find its way somehow into our system when rendering.

Does that distinction make sense for what you were thinking? I think that we could theoretically add both and we'd be addressing two separate but important things.

franciscolourenco:
@joelreske thank you for the PR?
Any ETA to adopt this PR? And Shouldn't TipTap be safe agains XSS by default, instead of providing this protection as an option which is off by default?

franciscolourenco:
Sanitization on link insertion wouldn't solve this issue, if sanitization is not done server side. Another option would be to pass the whole html by a library like [DOMPurify](https://github.com/cure53/DOMPurify), before passing the content to tiptap. 

Any ETA for the adoption of this PR or an equivalent solution? Thanks!

joelreske:
Hi @franciscolourenco , I need to put together another version of this which caches so that it's not computing every render.

I do think that TipTap should provide this behavior by default, but I think users should have the option of disabling or replacing the sanitization function.

In terms of server-side sanitization, note from earlier:
> The concern with XSS is not so much that we always guarantee sanitized data in our DB, but rather that we are protected from malicious data that might find its way somehow into our system when rendering.

franciscolourenco:
As a workaround, we are currently using [`DOMPurify.sanitize(html, { USE_PROFILES: { html: true } })`](https://github.com/cure53/DOMPurify) before passing the html to TipTap.

As a word of caution, @braintree/sanitize-url has disclosed [security vulnerabilities](https://github.com/braintree/sanitize-url/issues/41), and I wouldn't rely on it to sanitize urls. Moreover, for TipTap's use case, I don't see a reason not to use the **simpler** and **safer** alternative, which is to use a protocol whitelist (`http://|https://|`).

janthurau:
fixed by #4602 (Thanks @joelreske for raising this two years ago!)

Issue #1971, opened 2021-10-03
Loading the editor on iOS (when isiOS=true) causes content to be wiped


jamesopti:
### What’s the bug you are facing?

All our content was disappearing when it was loaded on any device where the [isiOS](https://github.com/ueberdosis/tiptap/blob/main/packages/core/src/utilities/isiOS.ts#L1) function returned `true`

We tracked it down to the core [NodeView `ignoreMutation` function](https://github.com/ueberdosis/tiptap/blob/main/packages/core/src/NodeView.ts#L199-L214)

### How can we reproduce the bug on our side?

For us this was instantly reproducible in the chrome debugger by just choosing any device that matches the `isiOS` check (like iPad)

I haven't been able to reproduce in an isolated sandbox but I'll keep trying and update here if I can. I suspect it may only be an issue with nested NodeViews

### What did you expect to happen?

Content would not be erased and load as normal.

### Anything to add? (optional)

We worked around this with a fork of tiptap react that overrides the `ignoreMutation` logic above to check if the editor is focused:

```ts
if (
      this.dom.contains(mutation.target) &&
      mutation.type === 'childList' &&
      isiOS() &&
      this.editor.isFocused // --- GAMMA CHANGE HERE
    ) {
...
```

### Did you update your dependencies?

- [X] Yes, I’ve updated my dependencies to use the latest version of all packages.

### Are you sponsoring us?

- [X] Yes, I’m a sponsor. 💖

philippkuehn:
Is this still happening with the latest version of `@tiptap/react`? Sounds a bit like #1747 which already has been fixed.

jamesopti:
> Is this still happening with the latest version of `@tiptap/react`? Sounds a bit like #1747 which already has been fixed.

Well I feel silly for not verifying this on the current version before reporting.

I can't reproduce it anymore! Thanks for double checking I did the obvious @philippkuehn  (because I didnt 🤦 )

I'll close this now and reopen if were able to trigger it again.

Issue #1214, opened 2021-04-24
Extra new line on "return" click in iOS


nucleartux:
**Description**

**Steps to reproduce the bug**
1. Go to https://www.tiptap.dev/guide/node-views/react/#adding-a-content-editable
2. Press "Return"
4. See extra new line inside box

In Vue example extra new whole interactive box appears.

**Environment?**
 - operating system: iOS
 - browser: Safari

https://user-images.githubusercontent.com/199706/115960859-62c9ed00-a51c-11eb-9268-9120d6d098f0.mp4

https://user-images.githubusercontent.com/199706/115960873-76755380-a51c-11eb-9545-e2e2a51d4f0e.mp4

philippkuehn:
This seems to be an issue with ProseMirror. I created an issue: https://github.com/ProseMirror/prosemirror/issues/1162

philippkuehn:
Okay this seems to be fixed in the latest release. But there is still an issue with iOS as described here (https://github.com/ProseMirror/prosemirror/issues/1162#issuecomment-827681792): node views will be re-created when pressing enter. There has been an idea for a while to build a node view cache that could probably catch this case but for now I’m closing this.

PR #2170, opened 2021-11-17
Ignore iOS mutations when unfocused

thatsjonsense:
We're running into a weird bug where NodeViews get stuck an infinite loop of mounting and unmounting when they contain an iframe. This causes a React [maximum update depth exceeded](https://reactjs.org/docs/error-decoder.html/?invariant=185) crash, and also causes the content of some of those nodeviews to be wiped out. 

To reproduce, try visiting https://9hgxv.csb.app/ on an iPhone. The same code works fine on a desktop.

We believe the issue is related to the default `ignoreMutation` method, which returns false when contenteditable nodes are added on iOS. I don't understand all the context of this function, but we found that patching this function to add a check for `this.editor.isFocused` (to narrow the if statement to only apply when triggered by the user's keyboard) fixed it in our fork.

We previously raised this issue in https://github.com/ueberdosis/tiptap/issues/1971 but thought it was fixed. It turns out it was only partially fixed. Instead of happening with all nodeviews in iOS, it now only seems to happen in the narrower case when the iframe is present. But the fix described in #1971 still works.

Related issues: https://github.com/ueberdosis/tiptap/issues/1214, https://github.com/ProseMirror/prosemirror/issues/1162#issuecomment-827681792

### DIFF
--- a/packages/core/src/NodeView.ts
+++ b/packages/core/src/NodeView.ts
@@ -202,7 +202,7 @@ export class NodeView<
     // this is because ProseMirror can’t preventDispatch on enter
     // this will lead to a re-render of the node view on enter
     // see: https://github.com/ueberdosis/tiptap/issues/1214
-    if (this.dom.contains(mutation.target) && mutation.type === 'childList' && isiOS()) {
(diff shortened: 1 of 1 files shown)


philippkuehn:
iOS safari is extremely buggy so let’s try this! thank you!

PR #2190, opened 2021-11-23
feat: add CodeMirror6

ocavue:
This PR is not 100% ready so I set it as a draft.  

TODO:
- [ ] options
- [ ] input rules


philippkuehn:
Hey! Unfortunately, we don’t accept PRs for such complex extensions, as we still need to maintain them. Thanks anyway!

PR #2199, opened 2021-11-25
Add `setEditable` to the Editor documentation

floriankrueger:
I was looking for an option to toggle edit ability on the editor and only found `setOptions` in the docs. Was surprised by code completion with the option of setting `editable` directly this way.


hanspagel:
Thanks! ✌️

PR #2204, opened 2021-11-27
extension-link: add support for title attribute

Spone:
In some use cases, it could be useful to specify the `title` of the link, for accessibility or SEO purposes.

This PR add support for this attribute.

### DIFF
--- a/packages/extension-link/src/link.ts
+++ b/packages/extension-link/src/link.ts
@@ -27,11 +27,11 @@ declare module '@tiptap/core' {
       /**
        * Set a link mark
        */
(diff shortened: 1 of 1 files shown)


philippkuehn:
Hey, thanks for the PR but I don’t think a `title` is absolutely necessary. If it is needed, you can extend the extension very easily.

Spone:
Thanks for the review @philippkuehn. How would you proceed to extend the extension? Should I fork it, or would you suggest to require it in my project and inherit from it?

philippkuehn:
There are some examples in the docs: https://tiptap.dev/guide/custom-extensions#attributes

PR #2212, opened 2021-11-30
initialize autofocus selection in `createView`

BrianHung:
Right now, the initial selection because it is undefined in `EditorState.create` is equal to [`Selection.atStart`](https://github.com/ProseMirror/prosemirror-state/blob/61bb964783b20800f6abc18dad855bd3bbf8cdd7/src/state.js#L25). 
Expected behavior is that if `autofocus` is defined, it is equal to the selection created on first focus. 

Example use-case is if `autofocus='end'`, then within a ProseMirror plugin `init` method: `state.doc.selection` should equal `Selection.atEnd` instead of `Selection.atStart`.


philippkuehn:
Great, but would be great if you can fix these linting errors! ✌️

philippkuehn:
thanks!

PR #2213, opened 2021-12-01
feat(ReactNodeViewRenderer): Add `as` option and pass through to ReactRenderer

jessicalc:
Right now, `ReactNodeViewRenderer` renders as a `<div class="react-renderer">` wrapper. We'd like to be able to pass an `as` prop to specify what kind of top-level HTMLElement corresponds with our node content. This could give people are a more granular control over how their elements get added (i.e. instead of just `div` or `span`, we could let people render an `img` or `li` directly).

At Gamma, our main use case is to allow us to have `li` elements adjacent to each other (rather than nested inside the div produced by ReactNodeViewRenderer). 

Additionally, (and for a separate PR), we like being able to pass the top-level `<li>` an `indent` attribute to dynamically do counter-resets. Right now there's no flexibility to add arbitrary attributes to the outermost wrapper when you use `ReactNodeViewRenderer`, so we're working on a way to do this in our own fork. We'll probably attempt this in a separate PR.

### DIFF
--- a/packages/react/src/ReactNodeViewRenderer.tsx
+++ b/packages/react/src/ReactNodeViewRenderer.tsx
@@ -20,7 +20,7 @@ export interface ReactNodeViewRendererOptions extends NodeViewRendererOptions {
     newDecorations: Decoration[],
     updateProps: () => void,
   }) => boolean) | null,
-  as?: React.Element,
+  as?: React.ElementType,
 }
 
 class ReactNodeView extends NodeView<React.FunctionComponent, Editor, ReactNodeViewRendererOptions> {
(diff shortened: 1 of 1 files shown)


hanspagel:
Thanks for the PR! The build is failing though:

> [!] (plugin rpt2) Error:/packages/react/src/ReactNodeViewRenderer.tsx(23,14): semantic error TS2724: 'React' has no exported member named 'Element'. Did you mean 'CElement'? 

Would you mind to look into that? You can test it locally with `yarn build:packages`.

philippkuehn:
It’s probably `React.ElementType`

https://github.com/ueberdosis/tiptap/blob/531d9e67143686b80bb7346c5da80aa08b2e4011/packages/react/src/NodeViewWrapper.tsx#L4-L7

philippkuehn:
thanks!

PR #2218, opened 2021-12-01
fix: update zeed-dom from 0.9.16 to 0.9.17 to fix #2217

### LINKED ISSUE #2217: generateHTML in @tiptap/html doesn't escape html properly

When the input in the editor starts with `<`, the `generateHTML` funtion in `@tiptap/html` won't escape the content.

For example, when typing:
```html
this is <strong>hello</strong>
```
the output of `generateHTML` is:
```plain
<p>this is &lt;strong&gt;hello&lt;/strong&gt;</p>
```
which is what we expect.

However, when typing:
```html
<strong>hello</strong>
```
the output of `generateHTML` is:
```plain
<p><strong>hello</strong></p>
```

That is, the function does not escapte the html at all.

Install `@tiptap/html@2.0.0-beta.142`, which depends on `zeed-dom@0.9.16`. I believe the problem is with `zeed-dom@0.9.16`.

https://codesandbox.io/s/tiptap-generate-html-issue-l6mk2

`generateHTML` should escape all html entities.

The problem lies in `zeed-dom@0.9.16`.

It has already been fixed here: https://github.com/holtwick/zeed-dom/commit/e5e1c400167b2c39aa8d02818d9c7dc464fefb2a#diff-86fe84b457508e0156e89b33424a861a2e689cae804dd33b13f663eb7d3af41aL111

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

aliasliao:
`zeed-dom@0.9.16` has issue that doesn't escape html perperly (see #2217), which has been fix in 0.9.17.

### DIFF
--- a/packages/html/package.json
+++ b/packages/html/package.json
@@ -23,7 +23,7 @@
   "dependencies": {
     "@tiptap/core": "^2.0.0-beta.143",
     "prosemirror-model": "^1.15.0",
-    "zeed-dom": "^0.9.16"
+    "zeed-dom": "^0.9.17"
   },
   "repository": {
     "type": "git",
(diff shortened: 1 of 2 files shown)


aliasliao:
https://github.com/ueberdosis/tiptap/issues/2217#issuecomment-983983769

PR #2225, opened 2021-12-02
Add support to have multiple rules for the same text input

isubasti:
Right now, it is assumed that input rules can only be skipped if it doesn't have any transaction step. This makes a few things harder to do as sometimes it is required to do some tr steps to figure out if all steps needed for the input rules can be done.
Prosemirror's official `prosemirror-inputrules` plugin does this by returning `null` if the input rules can't apply all the steps.

This changes introduce the same behaviour of `prosemirror-inputrules` to `@tiptap/core`'s InputRule plugin

### DIFF
--- a/packages/core/src/InputRule.ts
+++ b/packages/core/src/InputRule.ts
@@ -1,4 +1,4 @@
-import { EditorState, Plugin, TextSelection } from 'prosemirror-state'
+import { EditorState, Plugin, TextSelection, Transaction } from 'prosemirror-state'
 import { Editor } from './Editor'
 import CommandManager from './CommandManager'
 import createChainableState from './helpers/createChainableState'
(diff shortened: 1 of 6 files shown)


philippkuehn:
Good point. Makes sense to me. My only problem is that now input rules and paste rules behave differently. Input rules are executed one after the other, that’s not a problem. Paste rules are executed within a single `appendTransaction` hook. That means that the `tr` is passed through all paste rules. Thus it’s not possible to undo a paste rule that has already made changes to the transaction (if `null` is returned).

isubasti:
@philippkuehn fair point, although I don't know how else this can be tackled.

Having input rules behaviour change makes it much simpler to tackle things such as mixed style list (e.g ordered list with bullet list inside). At the moment, list(bullet/ordered) have a `wrappingInputRule` which works perfectly fine if it is a paragraph block at the top level. But if it's inside a list item of a list of a different type, the `wrappingInputRule` will fail and this needs to be handled by another input rule that handles changing the grandparent of that item to the right list type.

For paste rule. Do you think that it will makes more sense to use `transformPasted` instead and works with Slices rather than `tr` ?

philippkuehn:
closing in favor of #2368

PR #2226, opened 2021-12-02
Add support for autolink

### LINKED ISSUE #1209: Autolink
It would be great to have some kind of autolinking support for the link component. In other words typing `google.com` automatically turns it into a url.

You can see how Remirror does that here:
https://github.com/remirror/remirror/blob/beta/packages/remirror__extension-link/src/link-extension.ts#L356

Regexwise, I suggest using this package which has hard-coded TLDs.
https://www.npmjs.com/package/url-regex-safe

philippkuehn:
This PR adds an option `autolink` to the link mark (enabled by default). It lets you automatically add or remove links as you type.

Features:

- simply typed links like `tiptap.dev` get a correct url → `http://tiptap.dev`
- an email address like `humans@tiptap.dev` get a correct url → `mailto:humans@tiptap.dev`
- a link will be removed if it’s not a valid url anymore, e.g. if you insert a space `tiptap .dev`

fixes #1209

preview: https://deploy-preview-2226--tiptap-embed.netlify.app/preview/Marks/Link


rfgamaral:
>     * simply typed links like `tiptap.dev` get a correct url → `http://tiptap.dev`

I think the default should be `https` and not `http`, other than that, this looks good to me (feature-wise, didn't look through the code).

philippkuehn:
@rfgamaral Yes definitely! I’m using linkifyjs and there is an [open issue](https://github.com/Hypercontext/linkifyjs/issues/357) for that. But it’s already fixed and just needs to be released. 👍

hanspagel:
Amazing work @philippkuehn!

Not sure about the https thing! Servers typically redirect http to https, but sites without https can’t redirect to http. So http is probably a better default.

Any feedback on the implementation @maccman?

philippkuehn:
@hanspagel good point!

rfgamaral:
> So http is probably a better default.

A _better_ default, but one that enables a less secure Web.

@philippkuehn I know it's not possible right now (without linkifyjs v4), but once it's possible, how easy would it be to allow the link extension to have the `defaultProtocol` option from linkifyjs?

philippkuehn:
Yes, that should be possible.

ArunGovil:
Is the autolink feature still present in latest releases? 

I'm unable to create links while typing unless theres a whitespace followed by the url.

Is this an expected behaviour?

@philippkuehn

PR #2236, opened 2021-12-06
Fix export for CommandManager

isubasti:
`CommandManager` is exported but it only has export default on `CommandManager.ts` hence it is not properly exported.


philippkuehn:
Thanks. But instead of fixing this one single export I created a PR that replaces all default exports with named exports 👍 
https://github.com/ueberdosis/tiptap/pull/2238

PR #2240, opened 2021-12-06
Svelte Example: remove wrong `type="context"` tag

kahverengi001:
This example doesn't compile because in Svelte 3 there is no `type="context"`, there is only `context="module"`. Even if you use `context="module"` it is supposed to give an error because there is no `onMount` and `onDestroy` in module context.  Removing this tag from the Script element will fix the issue.


philippkuehn:
thanks!

PR #2256, opened 2021-12-08
A brand new `CharacterCount` extension

### LINKED ISSUE #1049: CharacterCount counts non-existing characters
**What happens?**
CharacterCount extension counts non-existing characters. 

An empty editor with CharacterCount enabled will show +2 for every newline, and + 4 for every list; even though there are no characters. When adding characters, the count increments on top of those non-existing counted characters.

I haven't tested "Code" and "CodeBlock", but I expect a similar result.

```
1. Go to the editor on https://next.tiptap.dev/api/extensions/character-count/
2. On a newline, type "Enter".
3. See count + 2, where it should be + 0.
```

```
1. Create a default editor with "Ordered List" or "Unordered List".
2. Add "CharacterCount" extension.
2. On a newline, create a list.
3. See count + 4, where it should be + 0.
```

Create a new Codesandbox replicating your error
Try in the documentation editor: [https://next.tiptap.dev/api/extensions/character-count/](https://next.tiptap.dev/api/extensions/character-count/) 

**What behavior did you expect?**
Count +0 on newlines and lists.

**What’s your environment?**
 - macOS
 - Chrome
 - Desktop

**Anything else?**
I haven't tested "Code" and "CodeBlock", but I expect a similar result.

### LINKED ISSUE #1550: Table crashes when combined with CharacterCount
**Description**
The app crashes as soon as the set limit for character count is reached. But this behavior only happens when Table is used with CharacterCount.

1. Clone my test repo https://github.com/mark-lumbao/react-rich-text-editor
2. Install dependencies then run **npm run start:dev** 
3. Open the app in the browser
4. Insert a table then continue adding table content till the limit is reached

**Github Repo**
I created a Github repo to help you debug the issue:

* React: https://github.com/mark-lumbao/react-rich-text-editor

**Expected behavior**
As seen in the attached video, the app crashes when the limit is reached.

**Screenshot, video, or GIF**

https://user-images.githubusercontent.com/9671486/124768967-ace33c00-df6b-11eb-88c8-b26818ed3c69.mp4

**Environment?**
 - operating system: Linux Manjaro
 - browser: Brave Browser
 - mobile/desktop: Desktop
 - tiptap version: 2 beta

### LINKED ISSUE #1839: `CharacterCount.limit` does not match `getCharacterCount()` or `state.doc.content.size`
**Description**
When the content of the editor contains something like a list the set limit for `CharacterCount` does not match any properties returned from the editor.

For example:
image

The limit is set to 20, but `getCharacterCount()` returns 22 and `state.doc.content.size` returns 24.

While I would expect the actual count to be 15 (chars + new line), I think at least the CharacterCount extension should match the limit to `getCharacterCount`. I've looked into the source code and it seems like the extension uses `state.doc.content.size` but for whatever reason the result is even more off.

**CodeSandbox**

* Vue: https://codesandbox.io/s/tiptap-issue-template-forked-hm0vx?file=/src/components/Tiptap.vue

### LINKED ISSUE #2245: CharacterCount config limit doesn't work correctly

I have issue when I use extension CharacterCount

1. When I press Enter, the function editor.getCharacterCount() count character is 2.
2. When I type and reach the limit, then I set the caret in the middle of the text. If I type, the new text will insert and remove the last text

Issue 1: 
1. Import extension CharacterCount
2. Use editor.getCharacterCount() below the editor to show the character have count like Usage in document https://tiptap.dev/api/extensions/character-count
3. Check charecter

Issue 2:
1. Import extension CharacterCount and set config limit
2. Type text reach to the limit
3. Set caret in the middle of the text
4. Type some text and watch the last text

https://codesandbox.io/s/character-count-error-qe7yz?file=/src/App.vue

I expect in
Issue 1: When I press Enter then character just count 1 instead of count 2
Issue 2: Can not type in the middle of the text if reach the limit

_No response_

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

philippkuehn:
With this PR we introduce the redesigned `CharacterCount` extension. It’s a full re-write and fixes some bugs.

### Add `mode` option
You can now choose between `textSize` and `nodeSize` for calculating the character size. Defaults to `textSize`.

```js
CharacterCount.configure({
  mode: 'nodeSize',
})
```

### Add `characters()` storage method

Get the number of characters for the current document with this new storage method. This deprecates `editor.getCharacterCount()`.

```js
editor.storage.characterCount.characters()

// Get the size of a specific node.
editor.storage.characterCount.characters({ node: someCustomNode })

// Overwrite the default `mode`.
editor.storage.characterCount.characters({ mode: 'nodeSize' })
```

### Add `words()` storage method

Get the number of words for the current document with this new storage method.

```js
editor.storage.characterCount.words()

// Get the number of words for a specific node.
editor.storage.characterCount.words({ node: someCustomNode })
```

## Preview

https://deploy-preview-2256--tiptap-embed.netlify.app/preview/Extensions/CharacterCount

Fixes #1049
Fixes #1550
Fixes #1839
Fixes #2245

### DIFF
--- a/demos/src/Examples/Community/React/index.jsx
+++ b/demos/src/Examples/Community/React/index.jsx
@@ -34,14 +34,14 @@ export default () => {
   })
 
   const percentage = editor
-    ? Math.round((100 / limit) * editor.getCharacterCount())
+    ? Math.round((100 / limit) * editor.storage.characterCount.characters())
     : 0
 
   return (
     <div>
       <EditorContent editor={editor} />
       {editor
-        && <div className={`character-count ${editor.getCharacterCount() === limit ? 'character-count--warning' : ''}`}>
+        && <div className={`character-count ${editor.storage.characterCount.characters() === limit ? 'character-count--warning' : ''}`}>
           <svg
             height="20"
             width="20"
@@ -73,7 +73,7 @@ export default () => {
           </svg>
 
           <div className="character-count__text">
-            {editor.getCharacterCount()}/{limit} characters
+            {editor.storage.characterCount.characters()}/{limit} characters
           </div>
         </div>
       }
--- a/demos/src/Examples/Community/Vue/index.vue
+++ b/demos/src/Examples/Community/Vue/index.vue
@@ -2,7 +2,7 @@
   <div>
     <editor-content :editor="editor" />
 
-    <div v-if="editor" :class="{'character-count': true, 'character-count--warning': editor.getCharacterCount() === limit}">
+    <div v-if="editor" :class="{'character-count': true, 'character-count--warning': editor.storage.characterCount.characters() === limit}">
       <svg
         height="20"
         width="20"
@@ -34,7 +34,7 @@
       </svg>
 
       <div class="character-count__text">
-        {{ editor.getCharacterCount() }}/{{ limit }} characters
+        {{ editor.storage.characterCount.characters() }}/{{ limit }} characters
       </div>
     </div>
   </div>
@@ -87,7 +87,7 @@ export default {
 
   computed: {
     percentage() {
-      return Math.round((100 / this.limit) * this.editor.getCharacterCount())
+      return Math.round((100 / this.limit) * this.editor.storage.characterCount.characters())
     },
   },
 
--- a/demos/src/Extensions/CharacterCount/React/index.jsx
+++ b/demos/src/Extensions/CharacterCount/React/index.jsx
@@ -35,6 +35,8 @@ export default () => {
 
       <div className="character-count">
         {editor.storage.characterCount.characters()}/{limit} characters
+        <br />
+        {editor.storage.characterCount.words()} words
       </div>
     </div>
   )
--- a/demos/src/Extensions/CharacterCount/React/index.jsx
+++ b/demos/src/Extensions/CharacterCount/React/index.jsx
@@ -34,7 +34,7 @@ export default () => {
       <EditorContent editor={editor} />
 
       <div className="character-count">
-        {editor.getCharacterCount()}/{limit} characters
+        {editor.storage.characterCount.characters()}/{limit} characters
       </div>
     </div>
   )
--- a/demos/src/Extensions/CharacterCount/Vue/index.vue
+++ b/demos/src/Extensions/CharacterCount/Vue/index.vue
@@ -3,7 +3,7 @@
     <editor-content :editor="editor" />
 
     <div class="character-count" v-if="editor">
-      {{ editor.getCharacterCount() }}/{{ limit }} characters
+      {{ editor.storage.characterCount.characters() }}/{{ limit }} characters
     </div>
   </div>
 </template>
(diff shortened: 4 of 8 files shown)


hanspagel:
So cool! Thanks for putting so much effort into it!

caroillemann:
Hi :) I just tested your deploy-preview url and it seems that the first word on a new line isn't added to the total number of words. As you can see here the word 'And' is actually word number 4 and 'more' is number 5 - but the total count is only 4.

Screenshot 2021-12-09 at 12 55 17

philippkuehn:
@caroillemann Oh you are right. Should be fixed with [this patch](https://github.com/ueberdosis/tiptap/commit/0879a2f71684e28e6de5a392b4ea675e11327d1f).

rfgamaral:
@philippkuehn Just noticed `editor.getCharacterCount` was deprecated, and arrived here after trying to implement the new options. I'm using this with TypeScript, and I'm not sure about some decisions, here are my suggestions:

1. Although the docs have `editor.storage.characterCount.characters()` as an example, the `characters` function is expecting a required `options` object:
   https://github.com/ueberdosis/tiptap/blob/49a99a0410352d8f70eb7107371bff864cfeda06/packages/extension-character-count/src/character-count.ts#L20-L23
   I think `options` should be made optional so that you can actually call `characters()` without arguments.
2. I'm wondering if we can avoid having `characters` and `words` being optional by moving the functions from `onBeforeCreate` to `addStorage` instead (that way they wouldn't be `undefined`) and calling them `characters?.()` wouldn't be needed.

To finish, I have a question about this issue:

image

The only I can think of to work around this is:

```ts
const count = (editor.storage.characterCount as CharacterCountStorage).characters?.({}) || 0
```

And I'm wondering if there's a better way without casting?

philippkuehn:
Oh, the first bug should be fixed with [this patch](https://github.com/ueberdosis/tiptap/commit/d1daf9cca06b4ac5855485533255b1275391b524).

We can’t move these methods to `addStorage` because there is no `editor` instance at this moment. That’s why it’s technically correct that these methods are `optional`. But I agree that it can be a bit confusing. I already thought about making the storage type required, even if it wouldn't be technically correct. Probably the DX would be a bit better anyway.

rfgamaral:
> We can’t move these methods to `addStorage` because there is no `editor` instance at this moment.

Granted that you have a lot more context than me, what if you check for `this.editor` first, and if it's not available, return `0`? Wouldn't that work? 🤔

philippkuehn:
@rfgamaral There is no `this.editor` context within `addStorage` at all so this won’t work.

rfgamaral:
@philippkuehn Got it! What about having dummy functions that just return `0` instead of being set as `undefined`? Once `onBeforeCreate` is called, these dummy functions would be replaced.

philippkuehn:
@rfgamaral yeah, I like that 👍 [Fixed](https://github.com/ueberdosis/tiptap/commit/ed7f93a2b84234b365f74ff39424f72ca5cf42ea).

rfgamaral:
Awesome, thanks!

PR #2258, opened 2021-12-09
fix: Replace `data-placeholder` with `aria-placeholder`

rfgamaral:
A small step in improving Tiptap accessibility.

### DIFF
--- a/demos/src/Examples/CustomDocument/React/styles.scss
+++ b/demos/src/Examples/CustomDocument/React/styles.scss
@@ -7,7 +7,7 @@
 
 /* Placeholder (at the top) */
 /*.ProseMirror p.is-editor-empty:first-child::before {
-  content: attr(data-placeholder);
+  content: attr(aria-placeholder);
   float: left;
   color: #ced4da;
   pointer-events: none;
@@ -16,7 +16,7 @@
 
 /* Placeholder (on every new line) */
 .ProseMirror .is-empty::before {
-  content: attr(data-placeholder);
+  content: attr(aria-placeholder);
   float: left;
   color: #ced4da;
   pointer-events: none;
--- a/demos/src/Examples/CustomDocument/Vue/index.vue
+++ b/demos/src/Examples/CustomDocument/Vue/index.vue
@@ -3,14 +3,14 @@
 </template>
 
 <script>
-import { Editor, EditorContent } from '@tiptap/vue-3'
-import Document from '@tiptap/extension-document'
-import Placeholder from '@tiptap/extension-placeholder'
-import StarterKit from '@tiptap/starter-kit'
+import { Editor, EditorContent } from "@tiptap/vue-3";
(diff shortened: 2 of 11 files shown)


rfgamaral:
Not sure why `build / lint` broke... What did I do wrong?👀

hanspagel:
The broken pipeline is an issue with GitHub actions I hadn’t time to look into. :/

hanspagel:
Ah, you did change a lot of code styling related things though. Double quotes instead of single quotes, semicolons …

rfgamaral:
> Ah, you did change a lot of code styling related things though. Double quotes instead of single quotes, semicolons …
That was not meant to happen, sorry. Probably caused by my VS Code extensions.

> That said, it’s not very likely that we’ll merge it like this as it would break a lot of people’s code. But it’s good that you brought this issue to our attention.

Yeah, forgot about that 😅

Out of curiosity, how do you intend to fix this issue, so it doesn't introduce a breaking change? Any kind of solution I could help with?

hanspagel:
Thanks for asking! Tiptap has more “issues” than we can tackle. That means, it’s important to keep track of everything, and see what issues we need to tackle to be as effective as possible (= helping as much people as possible). A11y is definitely on that list, but not that specific issue alone. 

I see a lot of potential in regards of a11y in general. But working only on specific issues now and then leads to us losing oversight and eventually an inconsistent API.

For us, it makes more sense to focus on the whole topic for a few weeks. Unfortunately, there’s no company/organization willing to pay for the work, but we’re in talks with an open source fund who might be able to help. Once that’s set up, the brought up placeholder issue will be one of the things we’ll tackle.

That’ll probably introduce multiple breaking changes, but the community is probably more willing to accept a few breaking changes for improving the accessibility in general, instead of breaking changes with every release for "tiny" specific issues.

Does that make sense?

rfgamaral:
Thanks for taking the time to clarify things for me.

> there’s no company/organization willing to pay for the work

Can you clarify exactly what you mean with this? Are you referring to accessibility work, or something different? What would be your expectation for a company "willing to pay for the work"?

> That’ll probably introduce multiple breaking changes, but the community is probably more willing to accept a few breaking changes for improving the accessibility in general, instead of breaking changes with every release for "tiny" specific issues.

It's understandable, but then again, Tiptap is still in "beta", right? 😅 So breaking changes should be expected 🙃

hanspagel:
All the sponsorships pay for the maintenance and support work, but to take on bigger projects (like working on a11y for a few weeks), we need additional financial contributions. Sometimes a company/organization eager to have a new feature/extension and is willing to pay for the development. For a11y I couldn’t find someone (yet). But as I said, I’m on it.

Yeah, it’s in beta, but it has a decent amount of downloads/month already and for every breaking change we get a bunch of issues and support requests, sometimes written by people in a bad mood. :)

PR #2260, opened 2021-12-10
join lists on toggleList

### LINKED ISSUE #1329: Lists not being joined on toggleList
When you use toggleList, the new list is not joined with lists above / below.

Bug demonstration:
list bug

I'd expect it to behave similar to when I type the "- " shortcut to create a list:
list ok

The WrappingInputRule takes this scenario into account (https://github.com/ProseMirror/prosemirror-inputrules/blob/master/src/rulebuilders.js), but toggleList does not.

I think this should happen automatically (join with lists above / below), or be an option for the toggleList command.

philippkuehn:
With this PR we try to join lists if possible on `toggleList`.

### Before

https://user-images.githubusercontent.com/2500670/145576288-26314d76-49fc-4b70-b2a6-c0aebee3294b.mp4

### After

https://user-images.githubusercontent.com/2500670/145576322-0bae810b-68bd-4c08-875e-10cc4bc2a906.mp4

fixes #1329

### DIFF
--- a/packages/core/src/commands/toggleList.ts
+++ b/packages/core/src/commands/toggleList.ts
@@ -1,8 +1,52 @@
 import { NodeType } from 'prosemirror-model'
+import { Transaction } from 'prosemirror-state'
+import { canJoin } from 'prosemirror-transform'
 import { RawCommands } from '../types'
 import { getNodeType } from '../helpers/getNodeType'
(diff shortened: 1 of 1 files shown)


YousefED:
Great stuff! Thanks guys!

PR #2275, opened 2021-12-14
fix: show FloatingMenu by default only if focused

jaulz:
Right now, if any extension updates the editor state the `FloatingMenu` will become visible. In my opinion the `FloatingMenu` should only be visible when the user actually focuses the editor. I know that I can pass a custom `shouldShow` function to change the behaviour but I think it should also be the default behaviour.


philippkuehn:
Good point. Thanks.

Issue #2040, opened 2021-10-14
How to handle flicker on mount


kylesuss:
### What’s the bug you are facing?

Hi! We are working to integrate TipTap into our app & are pretty early in the process. Thanks for the hard work on the lib.

When first rendering the editor, we notice that there is a period of time where nothing is rendered. Taking the [CodeSandbox](https://codesandbox.io/s/brave-thunder-7h41n?file=/src/App.js) (also linked below), I can toggle the editor & see that when the editor is meant to be shown, the first render cycle is empty.

Is this expected? Can it be fixed?

**Before toggle**
----
![data_image_jpg;base… (1)](https://user-images.githubusercontent.com/3035355/137371183-9bfb3515-f7b0-4f20-a69e-d0a01715c86e.jpeg)

**After toggle**
----
![data_image_jpg;base… (2)](https://user-images.githubusercontent.com/3035355/137371185-c0d32bc0-9c49-4754-9629-9b24281db487.jpeg)

**Editor now visible ("test")**
----
![data_image_jpg;base… (3)](https://user-images.githubusercontent.com/3035355/137371186-5937d074-52e6-4551-902d-a319bbb4b485.jpeg)

### How can we reproduce the bug on our side?

I was using the Chrome devtools Performance tab to record the behavior and capture screenshots.

### Can you provide a CodeSandbox?

https://codesandbox.io/s/brave-thunder-7h41n?file=/src/App.js

### What did you expect to happen?

I would expect there to be a way to avoid this flicker situation & have the editor visible on mount, but perhaps I am missing something in my setup?

### Anything to add? (optional)

I am using the React lib & `useEditor`.

### Did you update your dependencies?

- [X] Yes, I’ve updated my dependencies to use the latest version of all packages.

### Are you sponsoring us?

- [X] Yes, I’m a sponsor. 💖

Miloshinjo:
Having the same issue in React with `useEditor`. Editor shows empty before showing text, causing the screen flicker.

philippkuehn:
I had to revert the last change. If you want to prevent that flicker you can use a [custom `useEditor`](https://gist.github.com/ryanto/4a431d822a98770c4ca7905d9b7b07da) hook.

ospfranco:
![Screen Recording 2022-01-18 at 19 46 49](https://user-images.githubusercontent.com/1634213/150000015-272a3077-e560-4712-893f-a99da52d60d6.gif)

[...]

Issue #2182, opened 2021-11-21
useEditor returning null


Neesh774:
### What’s the bug you are facing?

I'm trying to use `useEditor`, but it returns null at first, and then returns an Editor object. I create it and log the variable I set it to immediately, and it prints out null first, and then it prints out the Editor object. I am using create-react-app in React.

### How can we reproduce the bug on our side?

1. Create a create-react-app project in React
2. Put the following code in `App.js` before the return statement:
```
const editorUse = useEditor({
    extensions: [StarterKit],
    content: JSON.parse(localStorage.getItem("text")) ?? "",
  });
  console.log(editorUse);
```
3. Run the project and check the console.

If the issue persists, it will print null first, and then the Editor object afterwards.

### Can you provide a CodeSandbox?

[Here you go](https://codesandbox.io/s/zealous-darwin-iswww?file=/src/App.js)

### What did you expect to happen?

There shouldn't be a null, the Editor object should be the only thing being printed

### Anything to add? (optional)

This issue makes it difficult when I want to use the editor object in other parts of my UI, like using its `getText()` method.

### Did you update your dependencies?

- [X] Yes, I’ve updated my dependencies to use the latest version of all packages.

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

philippkuehn:
This is intended because the editor instance is created after mounting your component (which is needed because tiptap needs access to the DOM).

https://github.com/ueberdosis/tiptap/blob/bc242fe3424300f934b623477158aa8f837cf1c7/packages/react/src/useEditor.ts#L15-L31

If you need the editor instance in your template you have to check if editor is not null. After mounting, your component should be re-rendered so this shouldn’t an issue at all.

ryanto:
I have a PR that makes sure the editor is available on the first render: https://github.com/ueberdosis/tiptap/pull/2282

Issue #2087, opened 2021-10-26
Roadmap to stable v2


philippkuehn:
Hey, it’s amazing how far tiptap v2 beta has already made it. Thanks a lot for your support. 🙌

I'm trying to make a list here of features we still want to implement until stable v2.
This list is not final. If you have any wishes, please comment below.

### Features
- [x] Autolink on type #1209
- [x] Find a better implementation of `getCharacterCount`, #1839

### Breaking Changes
- [x] Add custom implementation of `InputRule`, #1997
- [x] Replace `defaultOptions` with `addOptions()` for extensions, #2088 
- [ ] Refactor core commands (possibly simplify, merge, find clearer names, etc.)
- [x] Use the [new storage feature](https://github.com/ueberdosis/tiptap/pull/2069) in `CharacterCount`
- [x] Use the [new storage feature](https://github.com/ueberdosis/tiptap/pull/2069) in `CollaborationCursor`

### QA
- [x] Add at least one test for every example

### Docs
- [x] Add docs for missing [commands](https://tiptap.dev/api/commands)

### Misc
- [x] Switch setup from yarn to npm

yellow1912:
I think you need to document these BreakingChanges and what to do as well. I had to track down the issues to understand what went wrong and the changes needed.

emma-campbell:
Hey, any plans on implementing #763? I am looking at a use case where I would need to force a title, subtitle & text (paragraph) "blocks". Would be a great feature to have!

tylerforesthauser:
Perhaps this would work for you?

https://tiptap.dev/examples/custom-document

cadars:
> Do you know where these options may be documented? 

Have a look at the Schema docs, this part in particular: https://tiptap.dev/api/schema#the-node-schema

philippkuehn:
> > Perhaps this would work for you?
> > https://tiptap.dev/examples/custom-document
> 
> I tried this example, and was able to get the title but have struck out with subtitle as I had trouble finding documentation on what a valid string for the content option would be, besides `block+` or `heading block*`, to support a subtitle. Do you know where these options may be documented? I saw the issue I previously mentioned had been closed with a possibility of being implemented for v2, but couldn't seem to find anything about whether or not it would be implemented.

you can create a `title` and `subtitle` node. after that overwrite the document content expression with `title subtitle block+`.

emma-campbell:
[...]

Issue #1334, opened 2021-05-17
SSR with more complex components


kennethacohen:
**The problem I am facing**
I would like to render more components (in my case with React) while continuing to support SSR for editor content. As of now, I can either a) use the `generateHTML` method, which only allows for basic HTML markup rendering, and b) render the full editor, with `editable=false`, which means none of the editor content is able to be server-side rendered.

This isn't my use case, but it may be a helpful example: let's say you have a post and want to embed a stock price chart. You want the main body of the post to support SSR, and you want to fetch the stock price chart along with various interactive elements at the time of render - you just show a loading state for the chart when fetching this remote data. When composing the post, you would insert a custom `StockChart` extension of some kind, which stores the stock ticker or ID in the node attributes. Given that stock prices change over time, you wouldn't store any actual price data in the node attrs.

**The solution I would like**
To be honest I don't feel equipped to propose any particular solution, but hopefully the example above articulates the goal!

**Alternatives I have considered**
1. Render the editor with `editable=false`: None of the editor content is available for SSR (I'm using NextJS)
2. Render editor content with `generateHTML` method: cannot fetch remote data and render more complex components in the `renderHTML` method.

**Additional context**
Example of a post that achieves something similar - you'll see that the body of the post is server side rendered, but the price data shows a loading state and fetches remote data:

https://notboring.mirror.xyz/tw9hW87Sd1YXz_IqcvJXIS9pCxSSYBFVtZ2umBOdng0

PR #2282, opened 2021-12-15
Make sure editor is available on first render

ryanto:
This is a fix for #2182 - it makes sure that the editor is available on the first render of `useEditor` in the react package.

The reason for this fix is that the first render never has an editor and that causes the page's content to jump and flash. First render has no editor, so nothing is rendered. Then the second render has the editor and the WYSIWYG editor is rendered causing all the content on the page to shift down.

The effect will also re-create the editor if anything in the deps list changes.

cc @philippkuehn

### DIFF
--- a/packages/react/src/useEditor.ts
+++ b/packages/react/src/useEditor.ts
@@ -9,13 +9,18 @@ function useForceUpdate() {
 }
 
 export const useEditor = (options: Partial<EditorOptions> = {}, deps: DependencyList = []) => {
-  const [editor, setEditor] = useState<Editor | null>(null)
+  const [editor, setEditor] = useState<Editor>(() => new Editor(options))
   const forceUpdate = useForceUpdate()
 
   useEffect(() => {
-    const instance = new Editor(options)
+    let instance: Editor
 
-    setEditor(instance)
+    if (editor.isDestroyed) {
+      instance = new Editor(options)
+      setEditor(instance)
+    } else {
+      instance = editor
+    }
 
     instance.on('transaction', () => {
       requestAnimationFrame(() => {


philippkuehn:
Oh, didn't think that was possible. Great! And it seems like this also fixes #2040

vpontis:
Should we update the types:

```
export declare const useEditor: (options?: Partial<EditorOptions>, deps?: DependencyList) => Editor | null;
```

to not specify the `null` option?

vpontis:
Actually could we revert this change?

CleanShot 2021-12-16 at 10 48 16@2x

It totally breaks any Server Side Rendering with Next.js because Tiptap can't be rendered in Node. Or it should at least be an option...

philippkuehn:
Oh no. That’s bad. Revert seems like the best option until we found another fix for it. I’m on the way right now. Will do it later.

ryanto:
@vpontis Nice catch.

That's a bummer! What do you guys think of having some sort of API for opting into having the editor available during the first render. Just throwing this out there but something like...

```jsx
useEditor({
  availableOnFirstRender: true
});
```

When `true` it would init the editor right away and it would always be available. When `false` it would init the editor inside of an effect (like before) so the page can be generated on the server. We would default it to `false`, since that's how it worked before this PR and users can opt into `true` when they need it.

Trying to think of a way we can get best of both worlds. Allow folks writing client side apps to avoid the page jumping, but also allowing this hook to work seamless on the server.

vpontis:
@ryanto yea, that makes sense to me.

philippkuehn:
@vpontis I always thought that tiptap doesn’t work with SSR at all (#1334) Do you do anything special? You probably don’t use React node views, right?

@ryanto I must say that I’m not *that* familiar with React. Can you explain to me why this error happens with your change? If there is no other option, I would be ok with an `availableOnFirstRender` option. Could you create a PR for it? 😬🙏

vpontis:
> @vpontis I always thought that tiptap doesn’t work with SSR at all (#1334) Do you do anything special? You probably don’t use React node views, right?

Hmm, we do use React Node views and it does work with SSR. 

We do some smart stuff like not render toolbars if `editor == null`

https://user-images.githubusercontent.com/1319079/146414736-7ca570b6-69d1-4068-8d88-8ae7140cdb10.mp4

philippkuehn:
> Should we update the types:
> 
> ```ts
> export declare const useEditor: (options?: Partial<EditorOptions>, deps?: DependencyList) => Editor | null;
> ```
> 
> to not specify the `null` option?

@vpontis Strange. I don’t see this `null`. The types are already updated correctly.

> Hmm, we do use React Node views and it does work with SSR.

Good to know. Need to test this.

vpontis:
Oh @philippkuehn I could have been looking at an old version of the types. 

Lmk if there's anything w/ SSR that I can help w/ since we rely on it and it works well!

ryanto:
Yup, I'll absolutely PR. Will have it to you in an hour or so.

So tiptap doesn't render on the server since it references `document`, which isn't defined in Node.js (SSR uses Node to do server side rendering). When node encounters a `document` it thinks your code is referencing an undefined variable and throws throws an error. The screenshot @vpontis posted is the classic example of this :)

In React, `useEffect` does not run on the server. So previously when the editor was being created inside a useEffect it was never running on the server. The editor was null on the server, null on the first render, and then the useEffect would create it and it would become available on the second render.

Here's some pseudo code to help explain.

```
function Component() {
  let [name, setName] = useState("");
  useEffect(() => {
    setName("Alice");
  });

  return <p>hi {name}!</p>
});
```

1. Server render: "hi !"
2. Initial browser render "hi !"
3. useEffects now run
4. useEffect sets name and triggers a re-render
5. Second render: "hi Alice!"

One rule of react is that the server render must match the first browser render, so that react can re-hydrate the page's HTML correctly.

Let me know if that clears it up! Happy to jump on a chat too, React+SSR+Initial render can absolute be confusing.

ryanto:
Oh also, I think long term making tiptap run on the server by generating html, but not setting up any of the browser APIs/events/conenteditable (as pointed out in #1334) is a great idea! That way you can do something like....

```jsx
function Component() {
  // if were on the server opt of editing APIs
  let [isEditable, setIsEditable] = useState(typeof window !== "undefined");
  let [editor, setEditor] = useState(() => new Editor());

  useEffect(() => {
    setIsEditable(true);
  });

  // Editor can render an HTML only version, or WYSIWYG editor with all the browser APIs based on isEditable.
  return <Editor editor={editor} isEditable={isEditable} />
});
```

However, removing all of the calls to `document` and having a SSR compatible version of `Editor()` seems like it could be a bigger task. I don't know enough about tiptap to say how easy/hard it would be.

philippkuehn:
@ryanto oh thanks, got it! 👍

philippkuehn:
> Oh also, I think long term making tiptap run on the server by generating html, but not setting up any of the browser APIs/events/conenteditable (as pointed out in #1334) is a great idea! That way you can do something like....
> 
> ```js
> function Component() {
>   // if were on the server, or the first render, opt of editing APIs on the first render
>   let [isEditable, setIsEditable] = useState(false);
>   let [editor, setEditor] = useState(() => new Editor());
> 
>   useEffect(() => {
>     // this will only run in the browser, so we can safely now opt into editing
>     setIsEditable(true);
>   });
> 
>   // Editor can render an HTML only version or WYSIWYG editor with all the browser APIs based on isEditable.
>   return <Editor editor={editor} isEditable={isEditable} />
> });
> ```
> 
> However, removing all of the calls to `document` and having a SSR compatible version of `Editor()` seems like it could be a bigger task. I don't know enough about tiptap to say how easy/hard it would be.

Sounds great. But I think we should try to tackle that after [stable 2.0](https://github.com/ueberdosis/tiptap/issues/2087)!

ryanto:
Yes absolutely, I'll do a quickfix PR for now.

ryanto:
Hey @philippkuehn Almost ready for a PR! Got everything working, but I'd like to write a test for this just to make sure it renders.

I'm familiar with Cypress, is there anything I need to know about the testing setup for tiptap? Do I need to create a new React app under `demos/src/???`. Let me know how you would test this and I'll get it all setup :)

Thanks!

philippkuehn:
Great! 👍

You can add it here for now:
`/demos/src/Examples/Default/React/index.spec.js`

Or somewhere under `/tests/`.

I can also move this around later. We need to find a better structure for tests at some time 🙃

ryanto:
https://github.com/ueberdosis/tiptap/pull/2287 is ready if you guys want to review.

cusxio:
@vpontis ,

> Lmk if there's anything w/ SSR that I can help w/ since we rely on it and it works well!

How did you get SSR to work with React for TipTap? Currently I'm running `generateHTML` on the server side, and  mount TipTap on the client side.

Did you achieve SSR by extending the extension such as `Document`, `Paragraph`, and `Text` with `addNodeView` to render React components?

vpontis:
@cusxio oh gosh, I think I may be confused. I [tested it here](https://luma-1jpgxd1k1.luma-dev.com/style-guide/editor?encodedDoc=5LkXdomaF5Ay9wQEY2fzkCunwgTeCjRSFjqQy8q7UJ8Umf7tS3Hs4qoa3TfFk3NU2HY22wLiCCFyApX8XkdtFcscinYFxanfdwJLWRbjRVepYSdwfBGzGPd4yd9VFAxH2hayB4QG1R3WhggXq3xYF4GuPCUeKouD9S1PH9ZydKjtFM8Q3D6anYwJEEEXXRBrjvojJKBFTiJBCspqnEhCMH6KWugybFfFCesaFU2rBSagTGF2GSYF954kXuCDH7SdB4ZmLp7DFKAtEHBfHnYqqP4DNiSptCcoRcz1nnREWVKDnjDeC) and you are right that SSR isn't working. 

But SSR seems to be working locally. I'm not sure why...

cusxio:
@vpontis, no worries! I don't think SSR will ever work with `TipTap` because the internals relies on [`createElement("div")`](https://github.com/ueberdosis/tiptap/blob/f24c7828c2d968b1f0c6652298a3160a0b47b558/packages/core/src/Editor.ts#L57).

hanspagel:
@cusxio it does, it’s just not part of the core: https://tiptap.dev/api/utilities/html

cusxio:
This PR can actually work with a slight adjustment to handle the server,

```diff
- const [editor, setEditor] = useState<Editor>(() => new Editor(options))
+ const [editor, setEditor] = useState<Editor | null>(
+   typeof window === "undefined" ? null : () => new Editor(options)
+ );
```

hanspagel:
According to this comment it wouldn’t:

https://github.com/ueberdosis/tiptap/pull/2287#issuecomment-996255073

cusxio:
@hanspagel, but we are not solving for SSR in this PR.

We are solving first render issue.

The first render issue currently persist in the latest `tiptap` build. 

This PR technically fixes that, but this PR does not work on the server (because there's no `document` on the server). Hence it was reverted.

The fix I propose [above](https://github.com/ueberdosis/tiptap/pull/2282#issuecomment-998383503), makes it so that this PR does not crash on the server, but can still work on the client.

rfgamaral:
Just wanted to provide feedback that I would love to see @cusxio proposal back into Tiptap, to avoid having to have a custom `useEditor` on our side.

philippkuehn:
@ryanto any thoughts about this?

rfgamaral:
Actually, looking more closely, I don't think @cusxio is the right solution because the return type still includes `null`, but the instance returned by `useEditor` on a browser would never return null with the code above.

I haven't put much thought into this, but maybe 2 React hooks would be the most optional solution, one for each scenario. I don't know, just thinking out loud.

ryanto:
@philippkuehn I believe it will cause the client/server mismatch warning. Everything will render fine, but there will be an error/warning in the console. However, I think seeing a proof-of-concept to verify this would be good here.

On that note, it might be good to get a test suite setup to test all the scenarios we're uncovering in this PR.

There's lots were discussing here with SSR, CSR first render, client/server matching html, new Editor on server, etc. I think codifying each of the behaviors we want into tests so we can show which of these suggestions will and won't work is best. I'm more than happy to help with that after the holidays, I might need some guidance from y'all because I'm not too familiar with vite.

Have a great holiday everyone! :D

benjie:
Sorry to necro this; not sure where else to comment. It seems to me that this isn't doing quite the right thing for React/TipTap integration - should we really be recreating a new editor every time deps changes (e.g. an extension gets passed a new callback)? Also expecting the user to provide a deps array when the "rules of hooks" ESLint plugin isn't automatically picking up their mistakes seems a recipe for you having a lot of support tickets :wink: 

I wonder if this might be a better approach - remove deps completely, and use `editor.setOptions` rather than recreating the editor each time:

```ts
const useTipTapEditor = (options: Partial<EditorOptions> = {}): Editor | null => {
  const editorRef = useRef<Editor | null>(null);
  const forceUpdate = useForceUpdate();

  if (!editorRef.current && typeof window !== "undefined") {
    // Created editor on initial browser render
    editorRef.current = new Editor(options);
    editorRef.current.on("transaction", () => {
      requestAnimationFrame(() => {
        requestAnimationFrame(() => {
          forceUpdate();
        });
      });
    });
  } else if (!editorRef.current) {
    // Server; ignore
  } else if (editorRef.current.isDestroyed) {
    // Attempted to update options after editor was destroyed; this shouldn't occur.
  } else {
    editorRef.current.setOptions(options);
  }

  // Destroy editor on unmount
  useEffect(() => {
    return () => {
      editorRef.current?.destroy();
    };
  }, []);

  return editorRef.current;
};
```

This _will_ result in a different result on first render on browser versus SSR, but it doesn't _render_ something different, so that can be handled elsewhere.

I am extremely unfamiliar with tiptap so I don't know if the use of `setOptions` like this is equivalent?

benjie:
Oh, also your useForceUpdate should memoize the callback so it doesn't differ on every call:

```ts
function useForceUpdate() {
  const [, setValue] = useState(0);
  return useCallback(() => setValue((value) => value + 1), []);
}
```

rfgamaral:
@benjie Are you using that `useForceUpdate` hook with the `useCallback` yourself? What, exactly, is the difference supposed to be? 

I did some testing, and I can't see any difference with or without it. Just trying to understand if there's a real benefit in adding that `useCallback` or if this is one of those cases where we shouldn't actually be using it ([ref](https://kentcdodds.com/blog/usememo-and-usecallback)).

benjie:
It's beyond my knowledge if that's required or not, it's just a copy from the existing implementation. I imagine it is probably required to synchronise editor state with React state, without it the editor would still work but React would "lag behind" state-wise.

dilizarov:
Hmm. I think it is important to denote two separate issues here.

1. SSR
2. First render within the browser

While both are often related, they're not always. In my case, I'm focused on at least having the first render work kindly.

I tried leveraging `generateHTML(initialContent, extensions)` as the docs [here](https://tiptap.dev/api/utilities/html) suggest, but I don't think that function actually works on the server. It seems to require window. Am I doing something wrong there?

I've gotten to something that resembles this:

```ts
const Editor: React.FC = () => {
  const [initialRender, setInitialRender] = useState<string>(HTML_STRING)

  const editor = useEditor(...);

  return editor ? (
  ) : (
    <div className="ProseMirror" dangerouslySetInnerHTML={{ __html: initialRender }} />
  )
}
```

As you can imagine... the important tidbit is getting `HTML_STRING`.

Now, of course `useState(generateHTML(initialContent, extensions))` would be ideal, but again on the server this won't work because we don't have `window`.

I tried leveraging `useLayoutEffect` like:

```ts
useLayoutEffect(() => {
  if (initialContent) {
    // use generateHTML here
  }
}, [])
```

That works, but of course defeats the purpose because I still see the jumpy render situation.

I'm curious if it doesn't matter how quickly I get the `window` object... we'll always have this jumpy issue.

That leads me to my current solution:

```ts
useEditor({
  onUpdate({ editor }) {
    // persist editor.getJSON AND editor.getHTML to db on server.
  },
  ...
})
```

At least this way in my case the only thing I now have to fake is an empty editor if there is no data to preload into it, which is as simple as setting up a div with appropriate height, etc.

dilizarov:
Actually, my solution might not work for custom made nodes.

For example, if you have a node named "enhancedImage", and the tag it renders is enhanced-image, then via `generateHTML` you get something like "...<enhanced-image>" which only the Editor can understand...

Back to the drawing board.

stevecastaneda:
@dilizarov Your solution works, and you're right about custom components not working as intended.  I think I found a work around.

When creating your custom extension, instead of giving it a non-standard HTML tag like `react-component`, give it: `div[data-react-component]`.  Then, when you `renderHTML`, add that as an attribute so that it can find the component and parse it.

Now here's the trick.  If you want animation libraries to work, you'll need to manually set the height on these components as they're not actually _rendered_ , but the HTML will carry over with a style tag explicitly defining the height.  This works for my use case as the custom component will always be 28px.

```
parseHTML() {
    return [
      {
        tag: "div[data-react-component]",
      },
    ];
  },
    return ["div", mergeAttributes(HTMLAttributes, { "data-react-component": "", "style": "height: 28px;" })];
  },
  ```

This, along with using `dangerouslySetInnerHTML` allows the component to have the correct height on the first render, which libraries like framer motion rely on to animate correctly.
  
Can anyone identify any major issues with this approach?

RealAlphabet:
Does this work for Vue ? Because useEditor still return `undefined`.

PR #2316, opened 2021-12-25
Prevent eslint error import/no-named-as-default

LastxTemplar:
Prevent lint error, detailed discussion here:
https://stackoverflow.com/questions/44437203/how-do-i-resolve-eslint-import-no-named-as-default


philippkuehn:
We use default imports all over the codebase. We don’t have plans to change that for now.

PR #2319, opened 2021-12-27
Using vue 2 and 3 when passing props to VueRenderer in Mention plugin

domstrueboy:
A comment clarifying the difference between using vue 2 and 3 when passing props to VueRenderer in Mention plugin
A reason - it tooks me hours to realize that I need not only use this:
```
parent: this,
propsData: props,
```
 for vue 2 but remove that as well:
```
props,
editor: props.editor,
```
It led to a performance issue - a tab freezed in my case when adding some mentions


hanspagel:
Thanks! 🙌

PR #2327, opened 2021-12-29
change reactrenderer component type definition

lukesmurray:
This PR modifies the ReactRenderer types so that they are no longer tied to the `mention` example.

#2011 makes the `ComponentType` require `items[]` and a `command` in the props, but the ReactRenderer does not make the same assumption. 

I think this should also make it possible to change the type of `props` in the `ReactRenderer` class to `P`, and similarly in the function definition for `ReactRenderer.updateProps`. But I wanted this to be a minimal change so I just fixed the issue in the `ComponentType` type definition.

### DIFF
--- a/packages/react/src/ReactRenderer.tsx
+++ b/packages/react/src/ReactRenderer.tsx
@@ -24,12 +24,12 @@ export interface ReactRendererOptions {
   className?: string,
 }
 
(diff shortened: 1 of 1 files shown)


philippkuehn:
Looks good. Thanks!

PR #2350, opened 2022-01-04
Vue3 CLI doesn't have "npm run dev"

Aiyush-G:
npm run dev --> npm run serve

"dev" used to be in the CLI for Vue 2 but has since been removed with Vue 3

see: https://stackoverflow.com/questions/51664680/how-to-solve-npm-err-missing-script-dev/52694160


hanspagel:
Cool!

PR #2354, opened 2022-01-05
fix: export type `Level` for external use

webri:
Export type `Level` to externally work with levels that are passed to the `setHeading` and `toggleHeading` commands, instead of only being able to pass literal numbers.

Example:
```
// where `selectedHeadingLevel` shall be of type `Level` and changes as part of other logic
this.editor.commands.toggleHeading({ level: this.selectedHeadingLevel })
```


philippkuehn:
Yeah. Let's do this.

PR #2361, opened 2022-01-07
Mark `@tiptap/react` and `@tiptap/core` as side effect free

dcastil:
Hi Tiptap people! 👋

I created a PR directly instead of an issue because the diff is quite small. Feel free to close this if undesired. 😊

As far as I understand, Tiptap doesn't rely on global side effects anywhere within the packages `@tiptap/core` and `@tiptap/react` (I didn't check the other packages). I added `sideEffects: false` to both package.jsons to indicate to app bundlers (primarily webpack and rollup) that unused exports within those packages can be removed safely from the final bundle. That's especially useful for large packages like `@tiptap/react` and `@tiptap/core`.

More info: https://webpack.js.org/guides/tree-shaking

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -43,5 +43,6 @@
     "type": "git",
     "url": "https://github.com/ueberdosis/tiptap",
     "directory": "packages/core"
-  }
+  },
+  "sideEffects": false
 }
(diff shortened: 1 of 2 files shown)


philippkuehn:
Thanks! Yeah I think we can do this.

dcastil:
@philippkuehn Let me know whether I should check the other packages as well and add the change there, too. 😊

philippkuehn:
I’ve also added it to the Vue packages ✌️

PR #2376, opened 2022-01-11
fix check taskItem

### LINKED ISSUE #2372: Error when check taskItem

An error occurs when trying to check/uncheck a task item : 
`Uncaught RangeError: Index 1 out of range for <taskList(taskItem(paragraph("A list item")), taskItem(paragraph("And another one")))>`
getPos() return undefined in [task-item.ts#L123](https://github.com/ueberdosis/tiptap/blob/main/packages/extension-task-item/src/task-item.ts#L123)

Try this [doc example](https://tiptap.dev/api/nodes/task-item#usage).

_No response_

Being able to check / uncheck a task.

_No response_

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

Morgbn:
It seems that the `getPos` function must be called immediately.
Resolves #2372

### DIFF
--- a/packages/extension-task-item/src/task-item.ts
+++ b/packages/extension-task-item/src/task-item.ts
@@ -82,6 +82,7 @@ export const TaskItem = Node.create<TaskItemOptions>({
       const checkboxStyler = document.createElement('span')
       const checkbox = document.createElement('input')
       const content = document.createElement('div')
+      const nodePosition = getPos()
 
       checkboxWrapper.contentEditable = 'false'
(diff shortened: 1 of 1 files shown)


Morgbn:
Okay sorry, it's the first time I dive in the tiptap code (by the way, an amazing library), I don't find another way to prevent getPos() to return undefined. And I don't even know why getPos return undefined in my Chrome browser (from tiptap.dev example or from my code or from demos folder build in local) and not in private nav or Firefox. Sorry for the time taken..

PR #2410, opened 2022-01-20
Nuxt specific corrections

gsqrt2:
Minor adjustments so that the example doesn't need tinkering to run in latest default Nuxt installation.

Line 38: Nuxt does not use a src folder by default & convert name to multi-word to avoid naming convention errors.
Line 83: Removed id attribute from page container div. Nuxt adds its own id="app" to the app root container.
Line 85:  Convert name to multi-word to avoid naming convention errors.
Line 89: Import component before use, component autoload occasionally fails.
Line 91: Fix typo client-only

### DIFF
--- a/docs/installation/nuxt.md
+++ b/docs/installation/nuxt.md
@@ -35,7 +35,7 @@ npm install @tiptap/vue-2 @tiptap/starter-kit
 If you followed step 1 and 2, you can now start your project with `npm run serve`, and open [http://localhost:8080/](http://localhost:8080/) in your favorite browser. This might be different, if you’re working with an existing project.
 
 ## 3. Create a new component
-To actually start using Tiptap, you’ll need to add a new component to your app. Let’s call it `Tiptap` and put the following example code in `src/components/Tiptap.vue`.
+To actually start using Tiptap, you’ll need to add a new component to your app. Let’s call it `TipTap` and put the following example code in `~/components/TipTap.vue`.
 
 This is the fastest way to get Tiptap up and running with Vue. It will give you a very basic version of Tiptap, without any buttons. No worries, you will be able to add more functionality soon.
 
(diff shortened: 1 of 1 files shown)


docs/installation/nuxt.md
```diff
@@ -80,15 +80,23 @@ Now, let’s replace the content of `pages/index.vue` with the following example
 
 ```html
 <template>
-  <div id="app">
+  <div>
```
hanspagel, line 83: Do we need the `<div>` then?

docs/installation/nuxt.md
```diff
@@ -35,7 +35,7 @@ npm install @tiptap/vue-2 @tiptap/starter-kit
 If you followed step 1 and 2, you can now start your project with `npm run serve`, and open [http://localhost:8080/](http://localhost:8080/) in your favorite browser. This might be different, if you’re working with an existing project.
 
 ## 3. Create a new component
-To actually start using Tiptap, you’ll need to add a new component to your app. Let’s call it `Tiptap` and put the following example code in `src/components/Tiptap.vue`.
+To actually start using Tiptap, you’ll need to add a new component to your app. Let’s call it `TipTap` and put the following example code in `~/components/TipTap.vue`.
```
hanspagel, line 38: `~` is an alias for the user home directory, so if someone is created the file from the CLI it would end up in the wrong location.

Can you just write `components/TiptapEditor` here - without the `~`?

hanspagel (changes requested):
Thanks for the PR! Tiptap should be one word. Can you use `Tiptap`, `TiptapEditor` and `tiptap-editor` in all occurrences?

docs/installation/nuxt.md
gsqrt2: In a Nuxt project configuration `~` refers to the root folder of the project. In contrast to a vue project created with vue-cli project, the official Nuxt cli project creator (create-nuxt-app) uses this notation out of the box.

docs/installation/nuxt.md
gsqrt2: We do need it, because of the `<client-only>` tag afterwards. If we do not keep the div the page will have no actual wrapper when rendered in the server side.

gsqrt2:
> Thanks for the PR! Tiptap should be one word. Can you use `Tiptap`, `TiptapEditor` and `tiptap-editor` in all occurrences?

Glad I can be of use!

So, when using one word components, default and strongly-recommended eslint rule [vue/multi-word-component-names](https://eslint.vuejs.org/rules/multi-word-component-names.html) fails:

> This rule require component names to be always multi-word, except for root App components, and built-in components provided by Vue, such as <transition> or <component>. This prevents conflicts with existing and future HTML elements, since all HTML elements are a single word.

We could use TiptapEditor, as you suggest, in both the simple and v-model cases,  which would also bridge the component name difference that users face when transitioning from `4. Add it to your app` to `5. Use v-model (optional)`. 
Will submit such changes for review.

gsqrt2 (commented):
Component name changed for this document only. The name was not changed in the v-model example because it is a pure vue demo embed that is reused in the Vue 3 installation example. Refactoring it could cause unnecessary complications.

docs/installation/nuxt.md
hanspagel: Yeah, I know. But if some says "create a file", which you normally do on your file system, editor, the bash ... in that context it stands for a different folder, the user home.

If we would say "import that component" (which then would happen in the Nuxt context) that would be a different thing, and would make sense.

gsqrt2: I see, didn't realize it was a wording problem. Then `components/TiptapEditor.vue` should do just fine.

hanspagel:
Awesome! Thanks for your work on this. 🙏

Issue #548, opened 2019-12-16
Apply toggleHeaderRow always to the first row of a table


Sakessonholm:
**Is your feature request related to a problem? Please describe.**
I want to be able to toggle the first row of a table to be a header row or not. The current implementation seems to be that you can toggle the _currently selected_ row/column/cell as a header.

**Describe the solution you'd like**
I want my users to be able to create a table and then choose to "enable a header row" which should be the first row of the table. For that I need a command to toggle the header row and it shouldn't matter which row I'm currently in. If they click it once it sets the first row as a header row, if they click it again it makes the first row a normal row again.

**Describe alternatives you've considered**
The current commands only toggles the _currently selected_ row/column/cell as a header.

**Additional context**
In my opinion this is a quite common behaviour of a lot of editors dealing with tables, to be able to decide if the top row is a header row or not.

Sakessonholm:
I'm still trying to understand how plugins work with TipTap and ProseMirror, and I _think_ this could be achieved by a plugin that finds the first row of a table and then runs the _**toggleHeaderRow**_ command on it. If my assumption is correct I would appreciate any help on how such a plugin would look.

Duncank:
I'm using v2 and am looking for something like this as well. In my case, knowing if the selected cell is on the first row would suffice.

philippkuehn:
Thanks for these code snippets. Closing it for now.

philippkuehn:
This has been fixed in #2412 🎉

codemzy:
Nice work on #2412 - could I ask if `toggleHeaderCell` will remain with its current functionality? Being able to turn any cell in a table into a `th`, rather than just first row or first column. I believe that that is [valid html](https://validator.w3.org/#validate_by_input). The pull request asks for feedback, so I just wanted to mention that is useful to have, as I'm using this feature in the editor, to get table layouts like this...

![Screenshot 2022-01-26 at 11 23 47](https://user-images.githubusercontent.com/13764272/151154942-83fd0c98-95f2-4bb7-8fd8-c5baf445c04d.png)

philippkuehn:
@codemzy `toggleHeaderCell` is still available 👍

PR #2412, opened 2022-01-20
fix: use toggleHeader from prosemirror-tables

jpobley:
This PR will update the tables extension to use the `toggleHeader` command instead of the deprecated `toggleHeaderColumn` and `toggleHeaderRow` commands from `prosemirror-tables`. This is a **breaking change** because the commands will now only change the first column or row, respectively, instead of the row(s) or column(s) in the current selection. See the attached videos for examples.

The exception to the changes is `toggleHeaderCell`. I'm not sure what we'd like to be done with it and am looking for some feedback. Currently, it toggles any cell(s) in the selection to be "header cells". This is the deprecated behavior in `prosemirror-tables`. If we change it to use `toggleHeader`, then it will only toggle cells that are within the first row or column and, even then, it will only _remove_ a header cell, it doesn't put one back. This is confusing which is why I haven't included that change here.

### Old Behavior
https://user-images.githubusercontent.com/2583256/150428805-21660e4f-9a11-48f9-91ac-c34ba3a5bd65.mp4

### New Behavior
https://user-images.githubusercontent.com/2583256/150428804-6b54d70c-169e-47de-aa81-bea25737c4d9.mp4

### DIFF
--- a/packages/extension-table/src/table.ts
+++ b/packages/extension-table/src/table.ts
@@ -18,8 +18,7 @@ import {
   deleteTable,
   mergeCells,
   splitCell,
-  toggleHeaderColumn,
-  toggleHeaderRow,
+  toggleHeader,
   toggleHeaderCell,
   setCellAttr,
   fixTables,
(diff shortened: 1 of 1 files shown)


hanspagel:
Thanks for the PR! ☀️

philippkuehn:
Great, this is also a fix for #548.

codemzy:
@jpobley nice work!! I think it makes sense that toggleHeaderRow applies to the first row. I'd like to mention that it can be useful to have `th` in other places, I currently use tiptap for creating forms etc and `toggleHeaderCell` to create tables like this...

Screenshot 2022-01-26 at 11 23 47

So if that functionality can be maintained (since you asked for feedback!) it would certainly be useful for some use cases.

PR #2425, opened 2022-01-24
extension-link: fix ignored disabled linkOnPaste

farin:
paste handler was added twice
once as pasteRule, second as ProseMirror plugin

paster rule was not guarded by option
ProseMirror plugin guarded, but seem to be not working at all


philippkuehn:
No, the `pasteRule` and the `pasteHandler` plugin are two different things.

The pasteRule transforms pasted content from this:

```
text google.com text
```
to this:
```html
text <a href="http://google.com">google.com</a> text
```

The `pasteHandler` checks whether there is exactly one link in the clipboard when pasting and then adds this to the selected text.

farin:
@philippkuehn ok, anyway it seems there is still a bug, 
pasterRule is declared regardless linkOnPaste option

in other words it makes link from pasterd text even if linkOnPaste is set to false

PR #2427, opened 2022-01-24
extension-link: do not apply paste handler when linkOnPaste is false

farin:
when set to false link was still linkified


philippkuehn:
The `linkOnPaste` option is specifically meant for the `pasteHandler` plugin. 

There is a [global setting](https://tiptap.dev/api/editor#enable-paste-rules) to disable paste rules at all. If you want to disable specific paste rules you can use `.extend()` like:

```js
Link.extend({ addPasteRules: [] })
```

Otherwise you would have to add an option for each extension to deactivate paste rules.

PR #2449, opened 2022-01-27
Allow individual Typography rules to be disabled

rfgamaral:
The Typography extension is a great extension, but it's not very customizable. It's all or nothing for the included input rules, and this doesn't really work for us. We want some rules enabled, but all of them. I thought it would be nice if the extension could be fully configurable, allowing one to disable specific rules if one desires.

Hopefully this is something you agree with including in the official extension, otherwise we'll have to maintain our own version of this. Which is not a big deal, but it would be awesome if this could be integrated 😀

### DIFF
--- a/packages/extension-typography/src/typography.ts
+++ b/packages/extension-typography/src/typography.ts
@@ -1,5 +1,29 @@
 import { Extension, textInputRule } from '@tiptap/core'
 
+export interface TypographyOptions {
+  emDash: false,
+  ellipsis: false,
+  openDoubleQuote: false,
+  closeDoubleQuote: false,
+  openSingleQuote: false,
+  closeSingleQuote: false,
+  leftArrow: false,
+  rightArrow: false,
+  copyright: false,
+  trademark: false,
+  registeredTrademark: false,
+  oneHalf: false,
+  plusMinus: false,
+  notEqual: false,
+  laquo: false,
+  raquo: false,
+  multiplication: false,
+  superscriptTwo: false,
+  superscriptThree: false,
+  oneQuarter: false,
+  threeQuarters: false,
+}
+
 export const emDash = textInputRule({
   find: /--$/,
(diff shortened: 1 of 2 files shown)


packages/extension-typography/src/typography.ts
```diff
@@ -1,5 +1,29 @@
...
+  notEqual: false,
+  laquo: false,
+  raquo: false,
+  multiplication: false,
+  superscriptTwo: false,
+  superscriptThree: false,
+  oneQuarter: false,
+  threeQuarters: false,
+}
+
```
philippkuehn, line 26: These should be `boolean`, right?

packages/extension-typography/src/typography.ts
```diff
@@ -105,32 +129,97 @@ export const threeQuarters = textInputRule({
   replace: '¾',
 })
 
-export const Typography = Extension.create({
+export const Typography = Extension.create<TypographyOptions>({
   name: 'typography',
 
   addInputRules() {
```
philippkuehn, line 135: You should also define `addOptions()` with its default values here.

packages/extension-typography/src/typography.ts
rfgamaral: I followed the same approach used in `StarterKit`, where all rules are included by default and you can only disable the ones you don't want.

I can change this to boolean if you think that makes more sense. Let me know, and I'll make the change.

packages/extension-typography/src/typography.ts
rfgamaral: If we keep the current approach where the options can only be set to `false`, I believe we don't need `addOptions`. Again, followed the exact same approach as `StarterKit` to keep things simple. But I can change to `boolean` and add `addOptions` with default set to `true`.

philippkuehn:
Thanks!

PR #2462, opened 2022-01-31
Messaging example inspired by slack

montapro:
A Slack inspired messaging example with basic functionality build with tiptap for examples overview on the tiptap website.

Screenshot 2022-01-31 at 18 36 08


demos/src/Examples/Messaging/React/index.jsx
```diff
@@ -0,0 +1,128 @@
...
+import Italic from '@tiptap/extension-italic'
+import Strike from '@tiptap/extension-strike'
+import BulletList from '@tiptap/extension-bullet-list'
+import OrderedList from '@tiptap/extension-ordered-list'
+import ListItem from '@tiptap/extension-list-item'
+import Blockquote from '@tiptap/extension-blockquote'
+import Code from '@tiptap/extension-code'
+import CodeBlock from '@tiptap/extension-code-block'
+import Image from '@tiptap/extension-image'
+import Dropcursor from '@tiptap/extension-dropcursor'
```
philippkuehn, line 17: To simplify the demo we should use the starter kit instead.
```js
```

demos/src/Examples/Messaging/React/TextformattingMenu.jsx
```diff
@@ -0,0 +1,77 @@
...
+      <button
+        onClick={() => editor.chain().focus().toggleStrike().run()}
+        className={editor.isActive('strike') ? 'is-active' : ''}
+      >
+        strike
+      </button>
+      <div className="divider"></div>
+      <button onClick={addImage}>
+          add image
+      </button>
```
philippkuehn, line 39: I would not use the image extension at all in this demo. This only overcomplicates things here.

demos/src/Examples/Messaging/React/index.jsx
```diff
@@ -0,0 +1,128 @@
...
+      return null
+    }
+
+    setMessages([{
+      value: `Message ${messages.length} ${JSON.stringify(json, null, 2)}`,
+    }, ...messages])
+
+    editor.commands.clearContent()
+    document.getElementsByClassName('conversation')[0].scrollTo(0, 0)
+  }
```
philippkuehn, line 98: I think it would be also nice to focus the editor again after sending the message.
```js
editor.commands.focus()
```
You can also create a chain together with `clearContent`
```js
editor.chain().clearContent().focus().run()
```

demos/src/Examples/Messaging/React/index.jsx
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    // Get the initial content …
+    setJson(editor.getJSON())
+
+    // … and get the content after every change.
+    editor.on('update', () => {
+      setJson(editor.getJSON())
+    })
+  }, [editor])
```
philippkuehn, line 84: I don’t get this part. Instead of storing the JSON on every change, we could get it once when adding a message, right?

```js
const json = editor.getJSON()

setMessages([{
  value: `Message ${messages.length} ${JSON.stringify(json, null, 2)}`,
}, ...messages])
```

demos/src/Examples/Messaging/React/index.jsx
```diff
@@ -0,0 +1,128 @@
...
+      Image,
+      Dropcursor,
+      Mention.configure({
+        HTMLAttributes: {
+          class: 'mention',
+        },
+        suggestion: mentionSuggestion,
+      }),
+    ],
+    content,
```
philippkuehn, line 61: I would add the standard content here directly instead of putting it in a separate file.

hanspagel:
Moving to another repository, thanks for your work @montapro!

Issue #2144, opened 2021-11-08
Allow task lists items to be checked/unchcked in read-only mode


rfgamaral:
### What problem are you facing?

Task lists on GitHub allow one to check/uncheck task items without opening a comment editor first, it would be nice if the Tiptap's read-only mode allowed for the same thing.

### What’s the solution you would like to see?

While in read-only mode (i.e., `editable: false`), allow task items to be checked/unchecked. All related events should be fired, such as `onUpdate`, so that one knows that the content changed, and we can save it to persistent storage.

### What alternatives did you consider?

Implementing a custom rendering component that doesn't allow editing of anything besides task items, and use that instead of relying on the `editable` property.

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

PR #2474, opened 2022-02-02
Add option to allow task items to be checkable (uncontrolled)

kaspnilsson:
Useful for drafting interactive documents. Could see this expanding to potentially also support passing a callback?

Will happily update documentation if this approach is accepted!


philippkuehn:
Hey, thanks!

I’m happy to accept a `checkable` callback method to implement custom behavior here. But this should also control if the command is executed to update the `checked` attribute so this could be a fix for https://github.com/ueberdosis/tiptap/issues/2144 too.

An update for the docs would be great too ✌️

bkempe:
I think this could be achieved via:
```ts
        if (!editor.isEditable && onReadOnlyChecked) {
          onReadOnlyChecked(node.attrs.nodeId, checked);
        }
```
after
```ts
if (editor.isEditable && typeof getPos === 'function') { ... }
```

kaspnilsson:
Think I did what y'all requested, let me know if this works @philippkuehn @bkempe!

kaspnilsson:
Ping @philippkuehn @bkempe

bkempe:
@kaspnilsson It's not up to me to approve, but it looks great :)

bkempe:
@kaspnilsson One thing that might be nice: revert the change _if_ `onReadOnlyChecked` returns false?

kaspnilsson:
@bdbch please take a look! :)

bdbch:
Hey @kaspnilsson - looks nice! I think we have some linting issues. Could you resolve them? I'll make sure to take a look again when you're done so we can get your work merged! 👍

kaspnilsson:
@bdbch done!

packages/extension-task-item/package.json
```diff
@@ -23,6 +23,9 @@
   "peerDependencies": {
     "@tiptap/core": "^2.0.0-beta.1"
   },
+  "dependencies": {
+    "prosemirror-model": "^1.16.1"
```
bdbch, line 27: Is this dependency needed or should this be a peerDependency?

bdbch (commented):
Just a question about the dependencies. Otherwise looks good.

packages/extension-task-item/package.json
kaspnilsson: Sure, just needed for the types

bdbch: alright, thanks for the quick answer!

Issue #2426, opened 2022-01-24
Keyboard shortcuts with active caps lock


mejo-:
### What problem are you facing?

With caps lock active, keyboard shortcuts witch character keys like <kbd>Ctrl</kbd>-<kbd>B</kbd> don't work.

### What’s the solution you would like to see?

Shortcuts with letters should work when caps lock is active.

This should be as easy as adding e.g. `'Mod-B': toggleMark(type),` to the object returned by `keys({ type })` in the relevant marks. I can prepare a pull request if you agree that this is desirable.

Unfortunately, due to the way prosemirror catches keyboard shortcuts, this also implies that <kbd>Ctrl</kbd>-<kbd>Shift</kbd>-<kbd>b</kbd> will toggle bold - which is an acceptable tradeoff in my eyes.

I also checked other online editors (notion.so, google docs, ...) and for most of them, keyboard shortcuts with active caps lock work.

### What alternatives did you consider?

We got a bug report against the Nextcloud Text app that asks for this feature: https://github.com/nextcloud/text/issues/2050

I wouldn't mind implementing the extra needed shortcuts there, but thought it might be interesting for Tiptap in general :blush: 

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

philippkuehn:
Yeah, some shortcuts have already been modified in the [default ProseMirror demos](https://github.com/ProseMirror/prosemirror-example-setup/blob/06111da08f4aa58b96e52ad4bfba32b1d3ad6fbe/src/keymap.js#L59-L66). I agree with this in general. However, there are also special cases, such as [the history extension](https://github.com/ueberdosis/tiptap/blob/16bec4e9d0c99feded855b261edb6e0d3f0bad21/packages/extension-history/src/history.ts#L53-L55), where both `Mod-z` (undo) and `Shift-Mod-z` (redo) are already registered.

PR #2478, opened 2022-02-03
Add key bindings for uppercase letters for bold, italic and underline

mejo-:
This way, key bindings 'Mod-B', 'Mod-I' and 'Mod-U' with active caps
lock have the same effect as their lowercase siblings.

Prosemirror examples did the same, see ProseMirror/prosemirror#895

Fixes: #2426

Signed-off-by: Jonas <jonas@freesources.org>


mejo-:
Seems like the failing Cypress tests in CI are unrelated - at least they also fail on latest master :thinking:

philippkuehn:
Thanks!

(Yeah, tests are broken right now 🥴)

PR #2527, opened 2022-02-11
Use correct CSS name for attribute

Flamenco:


Flamenco:
2 different files, one vue one react

Flamenco:
You can squash em if thats your fancy.

philippkuehn:
Hey, there is both `.hljs-attribute` and `.hljs-attr` but in this case it’s intended that `.hljs-attr` has no styling.

Flamenco:
No styling is still the wrong tag...  I found the error because I used the template as the foundation for a project.

Flamenco:
I also left the old tag as to not break existing code.

philippkuehn:
These are two different things. `hljs-attr` is for XML attributes and `hljs-attribute` is for object attributes.

The current theme matches our documentation style but feel free to use any of the prebuild [highlight.js themes](https://github.com/highlightjs/highlight.js/tree/main/src/styles) or a custom one.

Flamenco:
Ah, that is good to know! Ty.

PR #2528, opened 2022-02-11
Use correct CSS name for attribute

Flamenco:


andreaselia:
It looks like you accidentally created the same PR twice? #2527

Flamenco:
The other pr is a different file.

philippkuehn:
See https://github.com/ueberdosis/tiptap/pull/2527#issuecomment-1036613062

PR #2601, opened 2022-03-07
add support for CSP nonces in createStyleTag

fekle:
Hi,

I noticed that tiptap injects a style element into the page's `head`. This breaks when using a Content Security Policy that disallows `unsafe-inline`, as there is currently no way to add a `nonce` to the dynamic style element.
As the injected CSS is in a typescript file and not a CSS file, there is no easy way to include it in the application bundle, and it differs from the default `prosemirror.css`. 

To resolve this, I added an option called `injectNonce` to tiptap which, when set, adds a `nonce` attribute to the dynamic style element.
Another solution would of course be to refactor the style into a CSS file which can be imported alongside tiptap, avoiding the dynamic element entirely, but that would probably require more code changes. 
Additionally, the `injectNonce` option could be handy if any other dynamic injection of styles or scripts is required in the future.

I'm open to discussion, but it would be great if we could use this amazing editor with a CSP!

### DIFF
--- a/docs/api/editor.md
+++ b/docs/api/editor.md
@@ -349,6 +349,22 @@ new Editor({
 })
 ```
 
+### injectNonce
+When you use a [Content-Security-Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy) with `nonce`, you can specify a `nonce` to be added to dynamically created elements. Here is an example:
+
+```js
+import { Editor } from '@tiptap/core'
+import StarterKit from '@tiptap/starter-kit'
+
+new Editor({
+  extensions: [
+    StarterKit,
+  ],
+  injectCSS: true,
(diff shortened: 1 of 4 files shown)


cadars:
Isn’t you issue solved with `injectCSS: false`?
https://tiptap.dev/guide/configuration#configure-the-editor

fekle:
@cadars Well that fixes the CSP issue, but then I have to manually copy the styles from tiptap's style typescript file into my stylesheet, and keep it up-to-date with every tiptap update. As the styles are in said typescript file and not a css file, they can't be directly included in the application bundle. Adding the CSP nonce would allow users to keep using the latest tiptap CSS automatically.

bdbch (approved):

bdbch:
Thanks, looks good to me!

PR #2622, opened 2022-03-15
Reduce bundle size of @tiptap/extension-table package

enriquecastl:
This PR declares the `prosemirror-model`, and `prosemirror-state` packages as dependencies of the `@tiptap/extension-table` package. Doing this indicates to rollup that it should not build these dependencies into the distribution files of the `@tiptap/extension-table` package.

In turn, this change dramatically reduces the build size of the `@tiptap/extension-table` package. 

### How did I find this problem?

I noticed that the esm build of the table extension is 214kb, and most of the code is contributed by the prosemirror dependencies. You can see here: https://unpkg.com/browse/@tiptap/extension-table@2.0.0-beta.48/dist/tiptap-extension-table.esm.js

After applying this change, the esm build is only 11kb 🎉

### DIFF
--- a/packages/extension-table/package.json
+++ b/packages/extension-table/package.json
@@ -24,8 +24,12 @@
     "@tiptap/core": "^2.0.0-beta.1"
   },
   "dependencies": {
+    "@types/prosemirror-model": "^1.16.0",
+    "@types/prosemirror-state": "^1.2.8",
     "prosemirror-tables": "^1.1.1",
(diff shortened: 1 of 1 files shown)


enriquecastl:
hi @bdbch 👋🏼 

I’d like to bring this PR to your attention because it is very simple, but it has a significant impact on the bundle size of applications using tiptap.

bdbch:
Thank you @enriquecastl for your PR! merged your changes.

PR #2625, opened 2022-03-15
Declare lowlight as a peerDependency in `@tiptap/extension-code-block-lowlight`

enriquecastl:
This PR declares the `lowlight` package as a peerDependency and removes the direct import statement in the `@tiptap/extension-code-block-lowlight` package. 

### Why is this change necessary? 

The client application can’t control which version of `lowlight` is bundled into the final product even when the client application provides its own lowlight instance. A consequence of this situation is that using a newer major version of lowlight ends up forcing the client application to bundle two versions of lowlight/highlight.js/fault in the final distribution build. That‘s a lot of dead JavaScript code. 

The main purpose of this change is reducing the bundle size of this extension by removing the duplication of lowlight versions in the codebase.

### DIFF
--- a/demos/package.json
+++ b/demos/package.json
@@ -15,7 +15,8 @@
     "shiki": "^0.10.0",
     "simplify-js": "^1.2.4",
     "y-webrtc": "^10.2.2",
-    "yjs": "^13.5.26"
+    "yjs": "^13.5.26",
+    "lowlight": "^1.20.0"
   },
   "devDependencies": {
     "@types/uuid": "^8.3.4",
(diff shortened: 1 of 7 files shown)


bdbch:
Thanks @enriquecastl! Do you think we should include this change in some kind of documentation for the extension or should we just keep it with the console.warning you implemented?

enriquecastl:
> Thanks @enriquecastl! Do you think we should include this change in some kind of documentation for the extension or should we just keep it with the console.warning you implemented?

Hi @bdbch! Thanks for the great suggestion. There’s value in documenting how this dependency depends and interact with lowlight. I’ve pushed a documentation update. What do you think?

PR #2626, opened 2022-03-16
feat: add EditorProvider and hooks to provide granular updates

umar-ahmed:
**Note:** this solution is somewhat similar to #2562, but _I think_ it has a smaller API surface.

The current `useEditor` hook triggers a re-render for the entire React component subtree, even if it doesn't depend on editor state.

With this PR, we can reduce the amount of re-rendering to only the components that call the new `useCurrentEditor` hook.

Developers would consume `@tiptap/react` like this:

```jsx
import {  EditorProvider, EditorContent, useEditor, useCurrentEditor } from "@tiptap/react"

const MyToolbar = () => {
  const editor = useCurrentEditor()
  const bold = editor?.isActive('bold')
  // do something with editor
}

const ComponentThatDoesNotCallUseCurrentEditor = () => <p>hello world</p>

export const Editor = () => {
  const editor = useEditor({
    // ... editorOptions
  })
  return (
    <EditorProvider value={editor}>
      <EditorContent />
      <MyToolbar />
      <ComponentThatDoesNotCallUseCurrentEditor />
    </EditorProvider>
  )
}
```

In the above code `<ComponentThatDoesNotCallUseCurrentEditor />` will not re-render on every keystroke because it doesn't depend on the `useCurrentEditor()` hook.

Also notice that the `editor` prop on `<EditorContent />` is no longer required because it retrieves the editor from React context.

To give you an idea of what this looks like, here's a comparison of before and after this change in an app I'm working on:

### Before

- Notice how the entire component is re-rendering on every keystroke

https://user-images.githubusercontent.com/8302959/158673763-e5afc28f-2cd4-4027-89a0-981d3b7d411b.mov

### After

- Now, only the ones calling `useCurrentEditor` and their descendants are re-rendering

https://user-images.githubusercontent.com/8302959/158672139-efcdaec5-c08a-433b-b18d-9052fb5b9de0.mov


BrianHung:
```
const MyToolbar = () => {
  const editor = useCurrentEditor()
  const bold = editor?.isActive('bold')
  // do something with editor
}
```

> With this PR, we can reduce the amount of re-rendering to only the components that call the new useCurrentEditor hook.

I've also been thinking about this for a while but never came up with a good solution with React. Is there a way to combine React Context with https://github.com/dai-shi/use-context-selector such that a toolbar or button would only change if `editor.isActive('bold')` changes too?

Changing based on a query rather than editorState would be more efficient for re-rendering since each transaction changes state but not necessarily "queries".

umar-ahmed:
> ```
> const MyToolbar = () => {
>   const editor = useCurrentEditor()
>   const bold = editor?.isActive('bold')
>   // do something with editor
> }
> ```
> 
> > With this PR, we can reduce the amount of re-rendering to only the components that call the new useCurrentEditor hook.
> 
> I've also been thinking about this for a while but never came up with a good solution with React. Is there a way to combine React Context with https://github.com/dai-shi/use-context-selector such that a toolbar or button would only change if `editor.isActive('bold')` changes too?
> 
> Changing based on a query rather than editorState would be more efficient for re-rendering since each transaction changes state but not necessarily "queries".

@BrianHung Yeah, I think that should be possible. 

For my internal implementation, I originally had this feature when I was using Draft.js, but I haven't ported that over to TipTap just yet.

It looked something like this:

```jsx
const isBoldActive = useCurrentEditor(editor => editor.isActive('bold'))
```

React 18 makes this even easier with the `useSyncExternalStore()` hook: https://github.com/reactwg/react-18/discussions/86

BrianHung:
There's a nice [`use-sync-external-store`](https://www.npmjs.com/package/use-sync-external-store) package now that can shim for React 18 and < React 18. Example of how another state management library integrated it https://github.com/pmndrs/valtio/pull/234

umar-ahmed:
> There's a nice [`use-sync-external-store`](https://www.npmjs.com/package/use-sync-external-store) package now that can shim for React 18 and < React 18. Example of how another state management library integrated it [pmndrs/valtio#234](https://github.com/pmndrs/valtio/pull/234)

So I tried using this actually with the shim and it seems to work well in my project. Barely anything renders while typing 😍 . Will try to update this PR with the improvements soon

https://user-images.githubusercontent.com/8302959/161626815-2a8e80d3-d471-497f-99c5-15926bcc7589.mov

BrianHung:
@umar-ahmed Could you possibly post of a gist of integrating `use-sync-external-store` with `Editor`? There's not much documentation / examples on using `use-sync-external-store` surprisingly. 😅

umar-ahmed:
> @umar-ahmed Could you possibly post of a gist of integrating `use-sync-external-store` with `Editor`? There's not much documentation / examples on using `use-sync-external-store` surprisingly. 😅

@BrianHung something like this worked for me:

```js
import { useCallback, useContext } from 'react'
import { useSyncExternalStore } from 'use-sync-external-store/shim'

const defaultSelector = (editor) => editor

export const useEditorSelector = (
    selector = defaultSelector,
) => {
    const editor = useContext(EditorContext)

    invariant(
        editor !== null,
        'EditorContext is not defined. Did you forget to wrap your component in a EditorProvider?',
    )

    const selection = useSyncExternalStore(
        useCallback(
            (notify) => {
                editor.on('transaction', notify)
                return () => editor.off('transaction', notify)
            },
            [editor],
        ),
        useCallback(() => selector(editor), [selector, editor]),
    )

    return selection
}
```

BrianHung:
Thanks for the helpful pointer!

Created a quick gist of providing the same functionality but tied to ProseMirror directly: instead of listening to `editor.on('transaction')`, we create a native prosemirror-view plugin which plays the role of an event emitter / sync. Also uses `useSyncExternalStoreWithSelector` for automatic selector optimization.

https://gist.github.com/BrianHung/f2b9b5de6b05de222cbacaf13745eb3c

bdbch:
Hey guys. This looks promising and kind of what I thought about the implementation via React for a few days now. I'll need some time to take a deeper look at this PR. Thanks already for the great idea.

umar-ahmed:
@bdbch was wondering if you had some time to take a look at this PR since your last message? Happy to clean things up to bring some of these improvements to fruition

bdbch:
Hey @umar-ahmed

I'd have time to look into this PR now - could you maybe create a demo in demos/ that woulds how the usage and result of this change?

umar-ahmed:
Sure, I can add a demo.

bdbch:
I'll close this PR for now. If you want to continue working on it let me know or reopen yourself.

PR #2648, opened 2022-03-24
fix: Support inline nodes with content in @tiptap/suggestion

thatsjonsense:
Problem: @tiptap/suggestion's `findSuggestionMatch` only works when inline nodes are leaves. When you have an inline node with content, in our case a footnote modeled off of https://prosemirror.net/examples/footnote/, suggestions after that node are broken.

This happens because the current logic relies on `textBetween` returning a string whose length exactly matches the absolute position range of the node. This is true with text nodes, as well as leaf nodes like a mention, but not a content node like footnote because its opening/close take up position spaces.

Solution: I don't think textBetween on the entire parent node is actually needed here. Just looking at the text node immediately before the suggestion fixes the problem and seems simpler as well.

### DIFF
--- a/packages/suggestion/src/findSuggestionMatch.ts
+++ b/packages/suggestion/src/findSuggestionMatch.ts
@@ -32,7 +32,11 @@ export function findSuggestionMatch(config: Trigger): SuggestionMatch {
     : new RegExp(`${prefix}(?:^)?${escapedChar}[^\\s${escapedChar}]*`, 'gm')
 
   const text = $position.nodeBefore?.isText && $position.nodeBefore.text
-  if (!text) return null
+
+  if (!text) {
+    return null
+  }
+  
   const textFrom = $position.pos - text.length
   const match = Array.from(text.matchAll(regexp)).pop()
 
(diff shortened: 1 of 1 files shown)


bdbch:
Hey @thatsjonsense, thanks for your pull request. Do you have a demo where you could show the previous/after behaviour so we can see the issue and how the fix is resolving that?

As far as I understood the problem is that a footnote contains more content in the popover which is not respected via `textBetween`?

I'll take a look into this PR again if you can help me review this PR so we can get your fix merged ASAP! 👍

thatsjonsense:
> Hey @thatsjonsense, thanks for your pull request. Do you have a demo where you could show the previous/after behaviour so we can see the issue and how the fix is resolving that?
> 
> As far as I understood the problem is that a footnote contains more content in the popover which is not respected via `textBetween`?
> 
> I'll take a look into this PR again if you can help me review this PR so we can get your fix merged ASAP! 👍

Hi @bdbch thanks for offering to review. Here is a CSB demonstrating the issue and the fix: https://codesandbox.io/s/tiptap-footnote-suggestion-fix-0ef5c9?file=/src/App.js

Just go to `mention/mention.ts` and uncomment line 5 to see the fix in action.

PR #2665, opened 2022-03-30
add-empty-editor-class-to-root-div

BrianHung:
Cover case where doc is empty but no descendants (custom doc that mimics an input). Add `emptyEditorClass` to the ProseMirror dom element instead.

Also not required to call `this.editor.isEmpty` repeatedly in descendants, or call it multiple times. Instead, cache empty top node and use ProseMirror `diffStart` to do a quick compare.


bdbch:
Could you resolve the linting issues @BrianHung ?

PR #2668, opened 2022-03-31
fix: Allow tippyOptions.getReferenceClientRect in bubble menu to be overridden

fleon:
Problem: The bubble menu's position currently depends on prosemirror selection, which works well for inline elements. But for block level elements, like code blocks, if we want to show a bubble menu to do things like change the language of the code block or to delete the code block, it would be ideal to have the bubble menu stick to the top left of the code block, for example. 

Solution: Allowing overriding `getReferenceClientRect` along with other `tippyOptions` solves this problem and allows customizing the position of tippy outside of the plugin.

### DIFF
--- a/packages/extension-bubble-menu/src/bubble-menu-plugin.ts
+++ b/packages/extension-bubble-menu/src/bubble-menu-plugin.ts
@@ -180,7 +180,7 @@ export class BubbleMenuView {
     }
 
     this.tippy?.setProps({
-      getReferenceClientRect: () => {
+      getReferenceClientRect: this.tippyOptions.getReferenceClientRect || () => {
         if (isNodeSelection(state.selection)) {
           const node = view.nodeDOM(from) as HTMLElement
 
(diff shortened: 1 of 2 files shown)


bdbch:
Nice idea @fleon. Can you please add your change to the documentation as well so users understand how to use this option?

https://github.com/ueberdosis/tiptap/blob/main/docs/api/extensions/bubble-menu.md

fleon:
@bdbch I don't think this needs to be added to the documentation. The docs already have `tippyOptions` described, where this option could've gone, but due to the current implementation passing `getReferenceClientRect` in there would not have worked. After this PR is merged, it will work. So its more of a bug fix.

bdbch:
Thats true! Thanks for correcting me there.

bdbch:
I'll merge this PR even though the build/test task failed - it seems to be an unrelated timeout issue on the Savvy demo.

PR #2677, opened 2022-04-04
fix: Mark the bubble/floating menu extensions as side effect free

rfgamaral:
This is very similar to https://github.com/ueberdosis/tiptap/pull/2361, but also adds `sideEffects: false` to `@tiptap/extension-bubble-menu` and `@tiptap/extension-floating-menu`, which are `@tiptap/react` dependencies.

Without this change, importing anything from the `@tiptap/react` pulls and bundles the whole package, even if we are not using the `BubbleMenu` or the `FloatingMenu` components.

For reference: https://webpack.js.org/guides/tree-shaking

### DIFF
--- a/packages/extension-bubble-menu/package.json
+++ b/packages/extension-bubble-menu/package.json
@@ -32,5 +32,6 @@
     "type": "git",
     "url": "https://github.com/ueberdosis/tiptap",
     "directory": "packages/extension-bubble-menu"
-  }
+  },
+  "sideEffects": false
 }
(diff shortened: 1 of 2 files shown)


bdbch (approved):
Great one. Thanks for your pull request.💪

qminds:
@bdbch Shouldn't `"sideEffects": false` also go into the `@tiptap/starter-kit` (and probably its dependencies) and `@tiptap/extension-text-align` packages? Seems that they also cause the bloated build without a proper tree-shaking.

qminds:
@bdbch Any updates on my question? I would be really grateful for your (or someone other's) reply. Because it seems strange that only part of libs have sideEffects set to false. Tree shaking is not my strong suit, unfortunately.

Issue #2685, opened 2022-04-05
Home/End buttons don't move the cursor to the start/end of the text line


rfgamaral:
### What’s the bug you are facing?

A GIF is worth a thousand words...

Here's the correct behavior on ProseMirror when I press the Home/End buttons:
![PowerToys_nLzUakpn1o](https://user-images.githubusercontent.com/96476/161796409-6c3a5d0d-ca30-4f2a-ae29-82512fcc772c.gif)

### How can we reproduce the bug on our side?

If you try the same thing on Tiptap's default example, Home/End sends you to the start/end of the paragraph node, not the text like like ProseMirror does:
![PowerToys_B535p4Lsa9](https://user-images.githubusercontent.com/96476/161796860-7e8f8aa6-8c01-43bf-a249-7394255a7318.gif)

### What did you expect to happen?

I expect Home/End buttons to work like in the ProseMirror demo.

That's how the browser `<textarea>` works, that's how Microsoft Word / Google Docs / Dropbox Paper works, and probably many other editors like Tiptap (I tried Slate, and it worked as expected).

### Did you update your dependencies?

- [X] Yes, I’ve updated my dependencies to use the latest version of all packages.

### Are you sponsoring us?

- [X] Yes, I’m a sponsor. 💖 _(on behalf of Doist)_

BrianHung:
https://github.com/ProseMirror/prosemirror-commands/commit/20371c58b83945a7504aa04c05b4a55b9b004af7

Relatively simple fix: just remove Home and End keybindings in the pcKeymap as ProseMirror did.

https://github.com/ueberdosis/tiptap/blob/ab4a0e2507b4b92c46d293a0bb06bb00a04af6e0/packages/core/src/extensions/keymap.ts#L61-L65

rfgamaral:
Hum... Now I'm wondering why this behavior was introduced in Tiptap? It's not the standard behavior anywhere...

However, those keybindings are built-in, what's the best way to remove them from my side? 🤔

BrianHung:
> However, those keybindings are built-in, what's the best way to remove them from my side? 🤔

Copy paste https://github.com/ueberdosis/tiptap/blob/main/packages/core/src/extensions/keymap.ts but remove those two lines, then include that file in the `Extensions` prop. Set `enableCoreExtensions=false` and include the other core extensions.

rfgamaral:
[...]

PR #2691, opened 2022-04-08
fix: don't override behaviour of Home / End in pc keymap

scottsidwell:
# Description
Fixes https://github.com/ueberdosis/tiptap/issues/2685 by removing the Home/End keybindings in the PC keymap.

This was similarly removed in Prosemirror at https://github.com/ProseMirror/prosemirror-commands/commit/20371c58, to favour default browser / OS behaviour.

Note: This behaviour continues to be preserved on Mac under the `Ctrl-a` and `Ctrl-e` keys

For users who wish to preserve this behaviour, they can achieve this via a custom extension:

```javascript
const CustomKeyboardBehaviour = Extension.create({
  addKeyboardShortcuts() {
    return {
      ['Home']: () => this.editor.commands.selectTextblockStart(),
      ['End']: () => this.editor.commands.selectTextblockEnd(),
    }
  }
})

const editor = new Editor({
  extensions: [
    // Register your custom extension with the editor.
    CustomKeyboardBehaviour,
  ]
});
```

If we haven't reached consensus on this change, then feel free to ignore. Just raising the PR in case this is all that's missing to resolve the issue.

### DIFF
--- a/packages/core/src/extensions/keymap.ts
+++ b/packages/core/src/extensions/keymap.ts
@@ -60,8 +60,6 @@ export const Keymap = Extension.create({
 
     const pcKeymap = {
       ...baseKeymap,
-      Home: () => this.editor.commands.selectTextblockStart(),
-      End: () => this.editor.commands.selectTextblockEnd(),
     }
 
     const macKeymap = {


bdbch:
Looks good to me, thank your for your contribution. I agree - the default browser should always be mimicking the default OS behaviour.

rfgamaral:
@bdbch Are you able to build and publish new releases? Would love to get a version with this in...

bdbch:
We're looking for releasing a new version very soon! :)

Issue #2680, opened 2022-04-04
Suggestion menu appears when clicking on potential mention


grantmagdanz:
### What’s the bug you are facing?

Clicking an incomplete mention causes the suggestion menu to appear.

### How can we reproduce the bug on our side?

1. Go to the mention example on the tiptap website: https://tiptap.dev/examples/suggestions
2. Type `@win` and then `esc`. Do not complete the extention.
3. Type something else after
4. Click on `@win` (may need to click twice)
5. Mention menu will appear

https://user-images.githubusercontent.com/9100467/161542187-153c6d08-ee8f-4364-abc4-d808a1cb268a.mp4

### What did you expect to happen?

After a user indicates they do not want to mention (e.g. they hit `esc`) the mention should act as plain text.

### Anything to add? (optional)

This is particularly confusing because it happens when the editor is not in `editable` mode. 

### Did you update your dependencies?

- [X] Yes, I’ve updated my dependencies to use the latest version of all packages.

### Are you sponsoring us?

- [X] Yes, I’m a sponsor. 💖

scottsidwell:
I had some time to look into this today (separate to the maintainers) and have some comments to add.

Firstly, you've actually mentioned two bugs, the second one being that suggestions should not be triggered when `editor.isEditable === false`. Since this was fairly straightforward, I raised a fix for this in https://github.com/ueberdosis/tiptap/pull/2692.

Back to the main issue you've raised here, the root cause appears to be an underlying design decision of how the suggestions plugin works. My understanding of how it works, is that (when you have an empty selection) it looks backwards in the current text block to see if the text that precedes your cursor matches the rough semantics of `${char}.*`. Since this is reactive, i.e. it waits for change in the editor state, and then tries to decide whether to display the suggestion UI, there's no mechanism for it to **not** show the suggestion UI when you select that text again in the future.

I'll let the maintainers weigh in here on how they see suggestions evolving, since any change to this behaviour will likely have flow on impacts to other parts of the UX here. e.g. this would cease to work

https://user-images.githubusercontent.com/38820281/162371171-3858950b-65d0-4137-8df2-54b28d426074.mov

grantmagdanz:
@scottsidwell Thank you for the reply and the fix! It's very much appreciated. 

Huh, interesting. Thanks for looking into it and that makes sense. I'll see if the maintainers will weigh in. Getting it to not show up when not editable is already a big UX improvement.

jessicalc:
[...]

PR #2692, opened 2022-04-08
fix: prevent suggestions from being active when editor is readonly

scottsidwell:
# Problem
The suggestions extension will continue to react to selection changes even when the editable is "readonly", which can lead to it showing a pop-up for text that start with the trigger character. This was discovered in https://github.com/ueberdosis/tiptap/issues/2680.

See the demo below:

https://user-images.githubusercontent.com/38820281/162368119-7b98076b-af09-47e6-a301-fffdbbb11570.mov

# Fix
We need to ensure we don't attempt to detect a suggestion at the current selection when isEditable is false. The attached commit will fail-fast and cleanup any active suggestion state. Demo:

https://user-images.githubusercontent.com/38820281/162368748-50c49488-8749-485b-98e6-eab2ffe37da7.mov

### DIFF
--- a/packages/suggestion/src/suggestion.ts
+++ b/packages/suggestion/src/suggestion.ts
@@ -169,6 +169,7 @@ export function Suggestion({
 
       // Apply changes to the plugin state from a view transaction.
       apply(transaction, prev, oldState, state) {
+        const { isEditable } = editor
         const { composing } = editor.view
         const { selection } = transaction
         const { empty, from } = selection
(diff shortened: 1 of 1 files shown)


bdbch:
Looks good to me. Thanks for your submission ❤️

bdbch:
It seems like there are linting issues in your code.

```
tiptap/packages/suggestion/src/suggestion.ts
Error:   172:38  error  Extra semicolon  semi
```

could you resolve them @scottsidwell?

scottsidwell:
@bdbch thanks for flagging, fixed :)

bdbch:
Thanks! Merged :)

sjdemartini:
@bdbch Is there any chance a new version of `@tiptap/suggestion` could be published that contains this merged fix? I'm also encountering this bug and would really appreciate taking advantage of scottswidwell's change. Thanks!

PR #2717, opened 2022-04-17
fix: extendMarkRange doesn't work when cursor is at end of mark, despite isActive() returning true for that mark

### LINKED ISSUE #2535: `extendMarkRange` doesn't work when cursor is at end of mark, despite `isActive()` returning true for that mark

`extendMarkRange` (and the underlying `getMarkRange`) seem to be "off by one" in position compared to `isActive`, which leads to inconsistent/unexpected behavior. 

For example, when a cursor is placed at the end of a link (blinking just after the last character), `editor.isActive('link')` will return `true`, but calling `editor.commands.extendMarkRange('link')` will have no effect. And conversely, placing the cursor just before the first character of a link will give `editor.isActive('link')` as `false` but calling `extendMarkRange('link')` _will_ extend to encompass the whole link. The "mark range" seems to be treating the range as inclusive of the start and exclusive of the end when there is no selection, and `isActive` appears to be treating the range as exclusive of the start and inclusive of the end, confusingly.

I wanted to have a callback set up so that whenever someone clicks on a link, I would extend the selection to encompass the whole link, and take different action depending on whether a link `isActive` or not after click. But due to this bug, it doesn't work as expected, and I can't reliably know that if a link is "active" I'd also have extended the selection (or vice versa).

(The same inconsistency is true of the [docs' suggested "get current href" approach](https://tiptap.dev/api/marks/link#get-the-current-value), since `editor.getAttributes('link').href` will return a result equivalent to when `isActive` is `true` including after the last char, but `extendMarkRange` won't "match".)

Include a link `<a>` element in the editor content, and add a click-handler to compare `isActive('link')` and `extendMarkRange('link')` behavior. See CodeSandbox for demo.

https://codesandbox.io/s/elegant-vaughan-hl2ve?file=/src/App.js

Consistent behavior between a mark being "active" and extending that mark.

_No response_

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

JDinABox:
Fixes #2535


bdbch (approved):

bdbch:
Thanks for your PR!

PR #2753, opened 2022-05-04
Remove nested div/span in ReactNodeViewRenderer

apptekstudios:
The current implementation of ReactNodeViewRenderer causes issues when trying to use node views for things like lists/tables where the html elements should be directly nested without intermediate divs (eg ul > li).

This pull request removes an intermediate element that was being added, instead directly accessing the 'nodeViewContentRef' element.


abettke:
@apptekstudios Any reason why this PR was closed? I'd be interested in these changes to be able to integrate a custom NodeView to the tiptap table extension.

apptekstudios:
I closed it because it caused a few unexpected bugs. It’s definitely a feature that I’d like to see

Issue #72, opened 2018-10-27
How to set link attributes


fmay:
Using your example for link insertion ...

```
    setLinkUrl(url, type, focus) {
      type.command({ href: url, rel: '', target: '__blank', class:'my-class' })
      this.hideLinkMenu()
      focus()
    }
```

What is the correct approach for managing attributes such as target, class, rel etc?

Thanks

fmay:
Sorry for answering my own questions!! I got off my ass and looked at the source and see how the nodes/marks are created and so it seems clear I should just write my own nodes and marks, which was next on my list anyway.

The more I see, the more I like.

antimodern:
@fmay hello

> I should just write my own nodes and marks

Can you please explain in two words what did you mean? Also want `target="_blank"` for the links...

Ken-vdE:
> Sorry for answering my own questions!! I got off my ass and looked at the source and see how the nodes/marks are created and so it seems clear I should just write my own nodes and marks, which was next on my list anyway.
> 
> The more I see, the more I like.

@fmay Could you show us a code example? I want to do something like this: 
```
this.editor.chain().focus().extendMarkRange('link').setLink({href: url, class: 'some-class'}).run()
```
However the class attribute can not be added dynamically like this unfortunately...

PR #2758, opened 2022-05-05
Allow class attribute through setLink()

Ken-vdE:
While you can set a default value for the Link extensions class attribute using 
```javascript
Link.configure({
  HTMLAttributes: {
    class: 'my-custom-class',
  },
})
```
It is currently not possible to dynamically set the class attribute when calling setLink() e.g. 
```javascript
this.editor.chain().focus().extendMarkRange('link').setLink({href: url, class: 'this class should be added'}).run();
```

This change allows for that by default, without needing to extend the Link extension. E.g. https://github.com/ueberdosis/tiptap/issues/72#issuecomment-1118287785


bdbch (approved):
looks good to me. Thanks!

PR #2768, opened 2022-05-09
fix: properly calculate setDragImage position

dilizarov:
NodeView's onDragStart assumes the event is a DragEvent.

In React, the event is usually a SyntheticBaseEvent, which doesn't have offsetX/offsetY properties.

To get the offsetX/offsetY properties, have to use event.nativeEvent, which is the DragEvent.

Without this fix, x and y become NaN in React.

### DIFF
--- a/packages/core/src/NodeView.ts
+++ b/packages/core/src/NodeView.ts
@@ -82,8 +82,12 @@ export class NodeView<
       const domBox = this.dom.getBoundingClientRect()
       const handleBox = dragHandle.getBoundingClientRect()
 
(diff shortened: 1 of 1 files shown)


dilizarov:
I chose to do `(event as any)` over properly typing `event: DragEvent | SyntheticBaseEvent` because this is the core package and I didn't want to introduce dependencies on React types into it.

bdbch (approved):
looks good to me, thank you!

dilizarov:
Let me know if you need anything else from me.

bdbch:
Great! Thanks

PR #2781, opened 2022-05-13
add validate option to link extension

### LINKED ISSUE #2779: Expose Linkify validate method in Link extension

The autolink attribute on the Link extension is a nice idea, but in practice it's less useful due to the long list of TLDs included by Linkify. We regularly get feedback asking why our editor turns certain word combinations into links when a user forgets to enter a "space" after a period.

I recognize hamster.vodka is a valid gTLD, but we're a little tired of explaining that 😂 

I also realize this behavior is powered by Linkify. Would it be possible to amend the Link extension and expose the `validate` method it provides?
https://linkify.js.org/docs/options.html#validate

That should allow us to write our own regex and just look for com|net|org.

The alternative is disabling autolink, which we'll probably do in the meantime. We'd like to keep it for some of the more common gTLDs though.

_No response_

### Are you sponsoring us?

- [X] Yes, I’m a sponsor. 💖

bdbch:
As requested in #2779 this PR implements a way to validate entered urls via a `validate` option.

Since we don't use the full linkify functionality but only the `.find` function to filter out links, I decided to only use one simple function for that.

closes #2779

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,7 @@
 {
-  "typescript.tsdk": "node_modules/typescript/lib"
+  "typescript.tsdk": "node_modules/typescript/lib",
+  "conventionalCommits.scopes": [
+    "extension-link",
+    "settings"
+  ]
 }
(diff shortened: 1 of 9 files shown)


svenadlung (dismissed):
Great, thx a lot!

PR #2794, opened 2022-05-16
Add `className` option and pass through to ReactRenderer

anton-liubushkin:


packages/react/src/ReactNodeViewRenderer.tsx
```diff
@@ -87,12 +88,14 @@ class ReactNodeView extends NodeView<React.FunctionComponent, Editor, ReactNodeV
     if (this.options.as) {
       as = this.options.as
     }
+    
+    const className = this.options.className ? ` ${this.options.className}` : ''
```
bdbch, line 92: Couldn't we simplify this line via

```js
const { className = '' } = this.options
```

packages/react/src/ReactNodeViewRenderer.tsx
```diff
@@ -87,12 +88,14 @@ class ReactNodeView extends NodeView<React.FunctionComponent, Editor, ReactNodeV
...
 
     this.renderer = new ReactRenderer(ReactNodeViewProvider, {
       editor: this.editor,
       props,
       as,
-      className: `node-${this.node.type.name}`,
+      className: `node-${this.node.type.name}` + className,
```
bdbch, line 98: Since this is already a template literal we can simply just do

```js
`node-${this.node.type.name} ${className}`.trim()
```

`.trim()` gets rid of the optional whitespace of the classname string.

bdbch (changes requested):
Hey @anton-liubushkin thanks for your PR.

I added a few comments which could be improved.

packages/react/src/ReactNodeViewRenderer.tsx
anton-liubushkin: Great idea! Thank you

packages/react/src/ReactNodeViewRenderer.tsx
anton-liubushkin: yeah, it looks better

bdbch (approved):
Looks good, thanks!

bdbch:
@anton-liubushkin could you fix the linting issue? I'll take a look at the test! 👍

anton-liubushkin:
@bdbch Hmm. I'm not sure which trailing spaces it's talking about. It seems that it doesn't like the `.trim()` option.

PR #2807, opened 2022-05-19
Fix InputRule regex matcher ignoring non-text leaflets in textBefore

### LINKED ISSUE #2806: Block level markdown shortcuts removes previous mention nodes

Imagine the following content: `[mentionNode] *|`, where `|` represents the text cursor. Pressing `<space>`, will transform the whole paragraph into a bullet list:

firefox_8OQMmF1zHG

All of them. No, just `StarterKit` and `Mention`.

Follow the steps in the GIF above with CodeSandbox below.

https://codesandbox.io/s/tiptap-issue-2806-oef3p6

Block level markdown shortcuts, like the bullet list, should be ignored if there's something like a mention preceding it.

### Are you sponsoring us?

- [X] Yes, I’m a sponsor. 💖 _(on behalf of Doist)_

bdbch:
This PR should fix #2806 

Currently non-text leaflets (which for example the Mention example is) this matcher is inserting space (` `) which would fit into the regex of the extension regex match strings.

I added a placeholder for this check resulting in non-text leafs being replaced with `%leaf%` which will stop those regex from returning true when a non-text leaf is used in front.

Since this function is not used for anything else but this check I think it's okay to use this %leaf% placeholder but I'm open for feedback or ideas how to solve this differently.


bdbch:
I forced push a different solution proposed by @rfgamaral (thanks!).

We now include the textContent of non-text leafs & fallback to a `%leaf%` placeholder when no `renderText` function is provided.

packages/core/src/helpers/getTextContentFromNodes.ts
```diff
@@ -0,0 +1,17 @@
+import { ResolvedPos } from 'prosemirror-model'
+
+export const getTextContentFromNodes = ($from: ResolvedPos<any>, maxMatch = 500) => {
```
rfgamaral, line 3: Maybe export this in the `index.ts` file like all other helpers?

bdbch: thanks, added

PR #2813, opened 2022-05-20
fix(suggestion): :bug: make clientrect prop optional

### LINKED ISSUE #2795: currentDecorationNode could be null in suggestion/src/suggestion.ts

Assume you have many nodes like such repeated for emojis (uses suggestion):

> :A :B :C :D :E :F :G :H :I :J :K :L :M :N :O :P

If you hold backspace to delete all of these characters 1-by-1, it is possible to land on this error:

Screen Shot 2022-05-16 at 3 53 08 PM

This happens in this code on line 120 of `suggestion/src/suggestion.ts`:

```ts
clientRect: decorationNode
              ? () => {
                // because of `items` can be asynchrounous we’ll search for the current docoration node
                const { decorationId } = this.key?.getState(editor.state)
                const currentDecorationNode = document.querySelector(`[data-decoration-id="${decorationId}"]`)

                // @ts-ignore-error
                return currentDecorationNode.getBoundingClientRect()
              }
              : null,
```

I see `// @ts-ignore-error` is used to ignore the fact that `currentDecorationNode` can be null, which is unfortunate... because that is exactly what happens.

I imagine this is a bit of a race condition (depending on how much processing is happening/how fast your browser is), but if you slow things down with lots of console.logs its easy to reproduce this.

I'm not sure what the best fix is here.

I tried changing it to `return currentDecorationNode?.getBoundingClientRect()`, but then the function's signature becomes `() => DOMRect` to `() => DOMRect | null` which causes larger issues.

Perhaps the better solution is to detect that we're deleting this and prevent this from running.

Latest Google Chrome

1. Use the emoji extension and make a ton of them in a row, like one-hundred+ (:A :B :C...) and then hold backspace. Preferably get console.logs going on every keyDown event to help slow the browser down.

_No response_

I expect there to be no error.

_No response_

### Are you sponsoring us?

- [X] Yes, I’m a sponsor. 💖

bdbch:
This commit makes the clientRect prop optional - this means that this value can be null. This allows developers using the suggestion extension to know that they have to implement a check for the clientRect before using it.

I think this is the correct way to handle this issue. Instead of building around it, we let the developer know that a clientRect **could** be potentially not defined, specially when the rendering is to slow and the DOM element for example does not exist.

See #2795 for more information.

### DIFF
--- a/demos/src/Nodes/Mention/React/suggestion.js
+++ b/demos/src/Nodes/Mention/React/suggestion.js
@@ -1,5 +1,7 @@
-import { ReactRenderer } from '@tiptap/react'
 import tippy from 'tippy.js'
+
+import { ReactRenderer } from '@tiptap/react'
+
 import MentionList from './MentionList.jsx'
 
 export default {
@@ -46,6 +48,10 @@ export default {
           editor: props.editor,
         })
 
+        if (!props.clientRect) {
+          return
+        }
+
         popup = tippy('body', {
           getReferenceClientRect: props.clientRect,
           appendTo: () => document.body,
@@ -59,6 +65,10 @@ export default {
 
       onUpdate(props) {
         component.updateProps(props)
+
+        if (!props.clientRect) {
+          return
+        }
 
         popup[0].setProps({
           getReferenceClientRect: props.clientRect,
--- a/demos/src/Examples/Community/React/suggestion.js
+++ b/demos/src/Examples/Community/React/suggestion.js
@@ -1,5 +1,7 @@
-import { ReactRenderer } from '@tiptap/react'
 import tippy from 'tippy.js'
+
+import { ReactRenderer } from '@tiptap/react'
+
 import { MentionList } from './MentionList'
 
 export default {
@@ -20,6 +22,10 @@ export default {
           editor: props.editor,
         })
 
+        if (!props.clientRect) {
+          return
+        }
+
         popup = tippy('body', {
           getReferenceClientRect: props.clientRect,
           appendTo: () => document.body,
@@ -33,6 +39,10 @@ export default {
 
       onUpdate(props) {
         reactRenderer.updateProps(props)
+
+        if (!props.clientRect) {
+          return
+        }
 
         popup[0].setProps({
           getReferenceClientRect: props.clientRect,
--- a/demos/src/Examples/Community/Vue/suggestion.js
+++ b/demos/src/Examples/Community/Vue/suggestion.js
@@ -1,5 +1,7 @@
-import { VueRenderer } from '@tiptap/vue-3'
 import tippy from 'tippy.js'
+
+import { VueRenderer } from '@tiptap/vue-3'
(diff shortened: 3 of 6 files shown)


demos/src/Nodes/Mention/React/suggestion.js
```diff
@@ -60,6 +66,10 @@ export default {
       onUpdate(props) {
         component.updateProps(props)
 
+        if (!props.clientRect) {
+          return
+        }
```
dilizarov, line 71: While this is nice, I don't think this actually works.

Down below, you see:

`getReferenceClientRect: props.clientRect`

When you look at Tippy's typings, it goes something like this:

```ts
export interface GetReferenceClientRect {
  (): ClientRect | DOMRect;
  contextElement?: Element;
}

...
export interface Props extends LifecycleHooks, RenderProps {
  getReferenceClientRect: null | GetReferenceClientRect;
  ...
}
```

Tippy could handle if `props.clientRect` is null, but in practice I've seen it be the function. Tippy expects `() => ClientRect | DOMRect` whereas we've got `() => ClientRect | DOMRect | null`.

It is kinda problematic.

On the one hand, you could run the function yourself, see if it is `null` and then pass it in if it is not, but that doesn't really work because it could be `null` by the time Tippy runs it.

dilizarov:
Perhaps integrating with Tippy is the tough nugget here. While it is a 3rd-party I think it is important because all examples leverage it and so naturally TipTap users will gravitate towards it.

I need to see if there is a way for us to pass a clientRect to Tippy without having them handle the function calling, etc.

demos/src/Nodes/Mention/React/suggestion.js
bdbch: I see. Tricky issue since I wasn't able to reproduce it even by blowing up my machine with 10k console.logs per editor state change. 🤡 

I'll think about this a bit and search for a better and cleaner solution.

The worst solution but also the worst case could be that we pass out a fake clientRect object but that feels pretty messy.

dilizarov:
@bdbch check out this:

https://github.com/ueberdosis/tiptap/blob/13a03425b36dedb1697caba396203d68d22554e1/packages/suggestion/src/suggestion.ts#L254-L268

So, basically this function seems to be responsible for whether the span decorator is there or not...

Perhaps there is a solution where you're basically storing data on the suggestion that tells you if you're rendering any decoration.

Something like:

```ts
  if (!active) {
    this.decoratorExists = false;
    return null;
  }

  this.decoratorExists = true;
```

Then, this is information that can be used to aid in the clientRect information.

Just thinking out loud. 

Basically, the problem is when we query the document, we're expecting the decorator there, but due to possible race conditions I've pointed out, it *doesn't* have to be there.

While querying the document is useful for DOMRect calculations, perhaps before querying we could have a flag stored that tells us if there is a decorator or not and that is probably immune to race conditions and we can build a solution around that.

PR #2830, opened 2022-05-26
fix: Only trigger image input rule at the start or with a preceding space

rfgamaral:
This only allows the image input rule to be triggered:

* At the start of a paragraph/line
* If it has a preceding space

This prevents the rule from triggering when one is typing the image Markdown syntax inside an inline code mark.

### Before

https://user-images.githubusercontent.com/96476/170526204-0fb7681d-a416-484e-9d9b-69982cf6c39a.mp4

### After

https://user-images.githubusercontent.com/96476/170525738-02364be4-3823-4228-8ad5-52fe12b6b73d.mp4

### DIFF
--- a/packages/extension-image/src/image.ts
+++ b/packages/extension-image/src/image.ts
@@ -21,7 +21,7 @@ declare module '@tiptap/core' {
   }
 }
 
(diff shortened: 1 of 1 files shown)


bdbch:
Awesome, thanks!

PR #2832, opened 2022-05-29
Add support for custom protocols in extension-link

shaunabanana:
This pull request adds custom protocols to extension-link. Previously, only urls beginning with http & https are recognized. Now it is possible to do the following:

```javascript
Link.configure({
    protocols: ['ftp']
})
```

And ftp:// links will be automatically recognized and turned into hyperlinks.

The registration of custom protocols uses the builtin function `registerCustomProtocol` of linkifyjs, which is already used by the extension. 

Docs are updated as well.

### DIFF
--- a/docs/api/marks/link.md
+++ b/docs/api/marks/link.md
@@ -20,6 +20,17 @@ npm install @tiptap/extension-link
 
 ## Settings
 
+### protocols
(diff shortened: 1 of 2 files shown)


bdbch (approved):
Awesome, good idea!

PR #2857, opened 2022-06-07
Fix state update after component unmounted

SavKS:
image


packages/react/src/useEditor.ts
```diff
@@ -13,20 +13,25 @@ export const useEditor = (options: Partial<EditorOptions> = {}, deps: Dependency
   const forceUpdate = useForceUpdate()
 
   useEffect(() => {
+    let isMounted = true;
```
bdbch, line 16: Could you resolve this linting issue? eslint should do this on it's own by now.

bdbch (approved):
Approved - just needs a linting fix.

Issue #1621, opened 2021-07-20
Rtl Support dynamically when user type first char 


pooooriya:
**The problem I am facing**
when I want to type persian like "سلام"
I Cant change direction 
**The solution I would like**
dynamically change dir and added to classname list of that element 
because when I saved my work and use dangerouslyhtml then without any side effects I show my persian language with dir="rtl"

**Alternatives I have considered**
nothing

**Additional context**
nothing and thanks for this cool package

cadars:
See: https://github.com/ueberdosis/tiptap/issues/116

Still, it would be nice to have `dir="auto"` as a simple setting.

cadars:
I came up with a quick and dirty extension, it works(?) for paragraphs and headings, not sure how to make it happen for lists.

```javascript
import { Extension } from '@tiptap/core'

const AutoDir = Extension.create({
  addGlobalAttributes() {
    return [   
      {
        types: [
          'heading',
          'paragraph',
        ],
        attributes: {
          autoDir: {
            renderHTML: attributes => ({
              dir: 'auto',
            }),
            parseHTML: element => ({
              autoDir: element.dir.autoDir || 'auto',
            }),
          },
        },
      },
    ]
  },
})
```

cadars:
Update: to make blockquotes and lists work, `dir="auto"` should only be added to parent nodes… I’m not sure how to make that happen in the extension.

Here is a test for this issue: https://codepen.io/cadars/pen/poeXMXZ?editors=1100

cadars:
Updated to work with the `parseHTML` [method change](https://github.com/ueberdosis/tiptap/issues/1863):

```js
import { Extension } from '@tiptap/core'

const AutoDir = Extension.create({
  name: 'AutoDir',
  addGlobalAttributes() {
    return [   
      {
        types: [
          'heading',
          'paragraph',
          'bulletList',
          'orderedList',
          'blockquote',
        ],
        attributes: {
          autoDir: {
            renderHTML: attributes => ({
              dir: 'auto',
            }),
            parseHTML: element => element.dir || 'auto',
          },
        },
      },
    ]
  },
})
```

philippkuehn:
For now, [adding a custom extension](https://github.com/ueberdosis/tiptap/issues/1621#issuecomment-918990408) is the preferred way.

amirhhashemi:
[...]

Issue #116, opened 2018-12-03
RTL Support


GabrielKol:
Is there a way to inject the **dir="auto"** attribute to all **p** tags?
It would be nice if each paragraph's direction would automatically change based on the language in it.

philippkuehn:
Hey @GabrielKol,
I do not needed RTL support for my latest projects. Isn't that something you can do with CSS?

GabrielKol:
I would like the direction to automatically change based on the language I'm using.
For example, if I start typing in Hebrew or Arabic, it would be presented from right to left,
but If I would press enter and on the next paragraph would type in English, then it would be presented from left to right.

It should automatically switch the direction based on the text's language.
It behaves this way in Gmail and Iv'e seen the same behavior in other editors such as Draft.js and Slate.js.

The only way I know of to implement this, is using the **dir** attribute on an html tag and setting it to "auto". So if I could give each p tag the dir="auto" attribute it would behave as I expect it to.
But I cannot figure out how to change the p tags in the editor.

Example from Gmail:
![gmail_editor](https://user-images.githubusercontent.com/31170834/49530819-af67eb80-f8c1-11e8-9707-3204b4307758.gif)

philippkuehn:
Okay, I'm not sure about implementing this by default. What you can do is to use a custom paragraph extension. Since it's a default extension you have to overwrite it.

```js
import { setBlockType } from 'tiptap-commands'
import { Node } from 'tiptap'

export default class Paragraph extends Node {

  get name() {
    return 'paragraph'
  }

  get schema() {
    return {
      content: 'inline*',
      group: 'block',
      draggable: false,
      parseDOM: [{
        tag: 'p',
      }],
      toDOM: () => ['p', { dir: 'auto' }, 0],
    }
  }

  commands({ type }) {
    return () => setBlockType(type)
  }

}

```

There is an option to disable the default extensions (`useBuiltInExtensions`). But you have to import `Doc` and `Text`.

```js
import { Doc, Text } from 'tiptap'
import Paragraph from './CustomParagraph.js'

const editor = new Editor({
  useBuiltInExtensions: false,
  extensions: [
    new Doc(), // default extension
    new Text(), // default extension
    new Paragraph(), // your custom extension
    ... // all other extensions
  ],
})

```

GabrielKol:
Thank you @philippkuehn 
It works perfectly! :)

mrg0lden:
[...]

PR #2889, opened 2022-06-16
init

ahhshm:
**This is a work in progress**

This plugin automatically adds `dir="ltr"` or `dir="rtl"` to specified nodes based on their content. I stole the direction detection logic from [lexical](https://github.com/facebook/lexical)

I'm still new to ProseMirror and this plugin definitely needs to be polished. I appreciate your suggestions and help.

### Demo

https://user-images.githubusercontent.com/87268103/174011640-7badf040-9ae9-4971-936b-8e51d96b5913.mp4

HTML (from devtools):

image

### TODO
- [ ] add `setTextDirection` and `unsetTextDirection` commands to manually specify the node's direction. These should have more precedence over the automatically detected direction so we can force a node to have a specific direction.
- [ ]  add a global attribute for direction so `dir` would be an actual attribute and will be in the HTML output. I still haven't figured out how to update a tiptap node's attributes within a ProseMirror plugin.
- [ ] improve performance - only update direction if needed
- [ ] add docs and update README

### Why not `dir="auto"`?
more control

### Why not a custom paragraph node?
headings and other text nodes also need it

### Related issues  
https://github.com/ueberdosis/tiptap/issues/116
https://github.com/ueberdosis/tiptap/issues/1621

### DIFF
--- a/packages/extension-text-direction/src/text-direction-plugin.ts
+++ b/packages/extension-text-direction/src/text-direction-plugin.ts
@@ -1,0 +1,75 @@
+import { findChildren } from "@tiptap/core";
+import { Plugin, Transaction, EditorState, PluginKey } from "prosemirror-state";
+import { Node as ProsemirrorNode } from "prosemirror-model";
+import { Decoration, DecorationSet } from "prosemirror-view";
+
+const RTL = "\u0591-\u07FF\uFB1D-\uFDFD\uFE70-\uFEFC";
+const LTR =
+  "A-Za-z\u00C0-\u00D6\u00D8-\u00F6" +
+  "\u00F8-\u02B8\u0300-\u0590\u0800-\u1FFF\u200E\u2C00-\uFB1C" +
+  "\uFE00-\uFE6F\uFEFD-\uFFFF";
+
+export const RTL_REGEX = new RegExp("^[^" + LTR + "]*[" + RTL + "]");
+export const LTR_REGEX = new RegExp("^[^" + RTL + "]*[" + LTR + "]");
+
+export function getTextDirection(text: string): "ltr" | "rtl" | null {
+  if (RTL_REGEX.test(text)) {
+    return "rtl";
+  }
+  if (LTR_REGEX.test(text)) {
+    return "ltr";
+  }
+  return null;
+}
+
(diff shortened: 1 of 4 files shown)


bdbch:
Hey @ahhshm thanks for your submission. Let me know when this PR is ready to be reviewed.

packages/extension-text-direction/src/text-direction-plugin.ts
```diff
@@ -0,0 +1,69 @@
...
+
+export type Direction = 'ltr' | 'rtl';
+
+const RTL = '\u0591-\u07FF\uFB1D-\uFDFD\uFE70-\uFEFC'
+
+const LTR = 'A-Za-z\u00C0-\u00D6\u00D8-\u00F6'
+  + '\u00F8-\u02B8\u0300-\u0590\u0800-\u1FFF\u200E\u2C00-\uFB1C'
+  + '\uFE00-\uFE6F\uFEFD-\uFFFF'
+
+// eslint-disable-next-line no-misleading-character-class
```
ahhshm, line 12: Notice I disabled this rule. I'm not sure about its impact

ahhshm:
There is a problem when bullet list items have different directions:

image

This pease of css resolves it: 
```css
  .ProseMirror li[dir="rtl"] {
    text-align: right;
  }

  .ProseMirror li[dir="ltr"] {
    text-align: left;
  }
```
image

But I'm not sure how to deal with it. should I add it by default (as a decoration) or leave it to the users (since the only job of this plugin is to add the `dir` attribute)?

bdbch:
> But I'm not sure how to deal with it. should I add it by default (as a decoration) or leave it to the users (since the only job of this plugin is to add the dir attribute)?

I think we should just add the css into the example so users of this package can see how to solve the issue.

docs/api/extensions/text-direction.md
```diff
@@ -0,0 +1,77 @@
...
+
+### defaultDirection
+
+The default direction.
+
+:::warning
+If you specify a defaultDirection, the `dir` attribute will be excluded in nodes that have the same direction as the defaultDirection.
+:::
+
+Default: `null`
```
ahhshm, line 53: Hmm, maybe `ignore` is a better name for that.

PR #2896, opened 2022-06-17
feat: Allow multiple prefix characters to trigger a suggestion

rfgamaral:
This PR replaces the `prefixSpace` option in the `Suggestion` utility with a new, more flexible option, `allowedPrefixes`. This option accepts multiple characters in a string, and allows each one of those characters as a prefix when triggering suggestions.

The current `prefixSpace` option is not very flexible, only allowing spaces as the prefix character, or everything else. But we are in the need of something more flexible, where we only allow a certain set of characters to be prefixed before the suggestion trigger character (i.e. `@` for the mention extension).

The idea here is that we can configure this option to work more like GitHub. For instance, setting `allowedPrefixes: ' (['`, suggestions can be triggered if the prefix character is ` `, `(` or `[`. The default is ` `, which works exactly like the current default. Setting `allowedPrefixes: null`, disables the option, and **any** character is allowed to be a prefix (same as setting `prefixSpace: false` in the current implementation).

I've also added `allowedPrefixes` to the `Suggestion` utility documentation, whereas `prefixSpace` was currently missing.


rfgamaral:
@bdbch Any clue why Netlify deployment is failing?

bdbch:
@rfgamaral we had the same issue. Seems like Netlify cached something and now doesn't like new builds. I'll hit up a fresh run for this PR! :)

bdbch:
I manually triggered a build:
https://app.netlify.com/sites/tiptap-embed/deploys/62b0afc310168b6a4ba03e5b

The PR looks fine to me.

docs/api/utilities/suggestion.md
```diff
@@ -21,6 +21,11 @@ Allows or disallows spaces in suggested items.
 
 Default: `false`
 
+### allowedPrefixes
```
bdbch, line 24: Since this talks about `prefixes` I expected an array of strings - not only one string. How is this supposed to be used when I want to allow multiple prefixes? Do I enter them in one string with spaces separated?

bdbch: Since you use it in the regex - maybe we should make this clear in the docs?

bdbch (approved):
Just a quick comment about the documentation / clarity of the explanation.

Otherwise this LGTM

docs/api/utilities/suggestion.md
bdbch: Ah I see - I just wasn't sure how this would've been used as a string-only which is also why I thought about adding a note how this string is used.

I like the approach of an array of strings more to make it clear that each string in this regex is matched separated

bdbch (approved):

PR #2919, opened 2022-06-23
feat(extensions): new decorations extension

bdbch:
This PR adds a new extension to Tiptap that we planned to do for a long time. This initially started as a solution to show invisible characters in an editor but grew bigger and now can be used to add general decorations to your editor view.

By default it provides invisible characters when used without `.configure` so this also helps to show invisible chars in a document.

You can read more in the documentation of the extension.

---

You can see the new demo here:
https://deploy-preview-2919--tiptap-embed.netlify.app/preview/Extensions/Decorators


mkriegeskorte:
@bdbch Thanks’s for this extension!

Seems like the selection from the end of a line (where the line break decorator is rendered) doesn’t work properly.

This adds some really nice feature for the whole editing experience. I’ve found some incosistent naming in this PR. You’re using **decorators** (as in the demo name or the buttons in the demo) and **decorations** as the extension name or the command method. Guess it makes sense to unify that into **decorators** 💅

bdbch:
> @bdbch Thanks’s for this extension! This adds some really nice feature for the whole editing experience.
> 
> Seems like the selection from the end of a line (where the line break decorator is rendered) doesn’t work properly.
> 
> I’ve found some incosistent naming in this PR. You’re using **decorators** (as in the demo name or the buttons in the demo) and **decorations** as the extension name or the command method. Guess it makes sense to unify that into **decorators** 💅 

I'll do some update on the namings - the selection issue is a bit tricky. I was able to fix the hard break decorators but the paragraph decorators are behaving super weird because prosemirror is adding some extra elements after decorators on node ends. Will take me some more time to find a solution there.

Issue #2850, opened 2022-06-06
Editable state isn't reactive


chrispiech:
### What’s the bug you are facing?

Changing the editor state (for example):
editor.setEditable

Doesn't creative a reactive change to 
props.editor.isEditable

Which prevents Nodes from reacting to a change to the editable state

### Which browser was this experienced in? Are any special extensions installed?

all!

### How can we reproduce the bug on our side?

Create an editor
Create a custom Node 
editor.setEditable(false)
Notice that code in the Node
useEffect(()=> {
   alert('test')
}, [props.editor.isEditable])
is not fired

### What did you expect to happen?

I expected it to have a reactive state change

### Did you update your dependencies?

- [X] Yes, I’ve updated my dependencies to use the latest version of all packages.

### Are you sponsoring us?

- [X] Yes, I’m a sponsor. 💖

Dreched:
A CodeSandbox reproducing the issue is located [here](https://codesandbox.io/s/ecstatic-kare-zukv3b?file=/src/NodeTags.jsx).

bdbch:
Hey @chrispiech @Dreched 

While the isEditable value is not reactive by default, I pushed a change (see #2935) that should allow you to track the state of the editors editable state on your own.

It could look something like this:

```js
const [isEditable, setIsEditable] = useState(false)

const editor = useEditor(
  ...
  onUpdate(props) {
    setIsEditable(props.editor.isEditable)
  }
)
```

You can then use that state to make all of your components reactive (for example via context or any other state management solution).

i-am-the-slime:
Is this meant to work now or did this just get closed because it was stale for a week?

damien-schneider:
This issue shouldn't be closed imo because the problem isn't solve, as we expect the isEditable to trigger rerender as a state would do, the change help to have a workaround but it is still a workaround.

bdbch:
Imho the properties of the editor instance should **not** be reactive by nature. It's a simple object property on the Editor instance, that you can get by accessing it on the editor.

Otherwise the question would be which properties would need to be reactive as well? I'd better let the developers integrate reactivity themselves.

PR #2935, opened 2022-06-27
fix(core): make setEditable trigger onUpdate function

bdbch:
This emits the onUpdate function when the editable state of the editor is changed via `setEditable`

should help with #2850

### DIFF
--- a/packages/core/src/Editor.ts
+++ b/packages/core/src/Editor.ts
@@ -169,6 +169,7 @@ export class Editor extends EventEmitter<EditorEvents> {
    */
   public setEditable(editable: boolean): void {
     this.setOptions({ editable })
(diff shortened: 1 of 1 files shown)


Rhys-T:
I just ran into #2850, and tried to watch for the `update` event to detect changes to `isEditable`. (I hadn't yet realized that the current release didn't fire this yet.) However, since I'm trying to detect this change from each instance of a node view (to enable/disable its fields), I was trying to use `editor.on('update', myCallback);` rather than changing the `onUpdate` function in the editor's main options. Unfortunately, it looks like this change is calling the `options.onUpdate` function directly, and won't trigger callbacks added afterwards through `.on()`. If I'm reading the code correctly, the line that fires the event should be something like:

```js
this.emit('update', { editor: this, transaction: this.state.tr })
```

Issue #2937, opened 2022-06-28
Input rules seem completely broken


rfgamaral:
### What’s the bug you are facing?

I don't know what changed, but inline code marks were working before, and now they aren't:

https://user-images.githubusercontent.com/96476/176157179-7552db5d-d890-4ea6-9841-367975d9bf27.mp4

EDIT: Just noticed that this problem is not exclusive to inline code marks, but all other input rules as well. For instance, open the default editor, and try to type `**` before a word, you'll quickly realize there's an issue (and you won't be able to type `**` after the word, to turn the word bold).

### Which browser was this experienced in? Are any special extensions installed?

All of them. No special extensions.

### How can we reproduce the bug on our side?

Use the [default text editor](https://tiptap.dev/examples/default), and try to use backticks to applied inline code marks.

### What did you expect to happen?

Inline code marks with backticks should work without issues, as they did in the past.

### Did you update your dependencies?

- [X] Yes, I’ve updated my dependencies to use the latest version of all packages.

### Are you sponsoring us?

- [X] Yes, I’m a sponsor. 💖 _(on behalf of Doist)_

svenadlung:
@rfgamaral I made a PR yesterday. I tried to put you as a reviewer but I failed. Feel free to try my changes: https://github.com/ueberdosis/tiptap/pull/2941

PR #2941, opened 2022-06-30
fix(core): only respect text of node before current position (#2937)

svenadlung:
Hopefully this PR fixes https://github.com/ueberdosis/tiptap/issues/2937

Since `node.textContent` always looks at the entire text of the node, the desired behavior worked out well at the end but not once the position is somewhere else within the text.

I replaced it with [nodeBefore](https://prosemirror.net/docs/ref/#model.ResolvedPos.nodeBefore) 

> Get the node directly before the position, if any. **If the position points into a text node, only the part of that node before the position is returned.**

### DIFF
--- a/packages/core/src/helpers/getTextContentFromNodes.ts
+++ b/packages/core/src/helpers/getTextContentFromNodes.ts
@@ -9,7 +9,7 @@ export const getTextContentFromNodes = ($from: ResolvedPos, maxMatch = 500) => {
     (node, pos, parent, index) => {
       textBefore += node.type.spec.toText?.({
         node, pos, parent, index,
-      }) || node.textContent || '%leaf%'
+      }) || $from.nodeBefore?.text || '%leaf%'
     },
   )
 


gnapse:
Hey @svenadlung, I work closely with @rfgamaral, and he's going to be unavailable for a while, so I can take over his involvement in this issue. Would you be so kind to let me know what would be the best way to try these changes locally? I have very little experience with tiptap's development, unlike @rfgamaral.

gnapse:
> Hi @gnapse, I think you could try [deploy-preview-2941--tiptap-embed.netlify.app/preview](https://deploy-preview-2941--tiptap-embed.netlify.app/preview/) and see if the behaviour works like expected.
> 
> If you want to try it on your local machine you could clone the [fix/input-rule-fix](https://github.com/ueberdosis/tiptap/tree/fix/input-rule-fix?rgh-link-date=2022-07-04T08%3A39%3A45Z) branch.

Thanks for the pointer. Indeed, it seems to work as expected in [that instance](https://deploy-preview-2941--tiptap-embed.netlify.app/src/examples/default/react/):

https://user-images.githubusercontent.com/15199/177390842-e3d74b86-8f03-41d8-819a-7f1c6f69e621.mp4

As opposed to when I try the same in https://tiptap.dev/examples/default

So what are the next steps to have this become part of a new release? Can I help in any way?

bdbch (approved):
Good find! Thanks for fixing this issue.

PR #2943, opened 2022-07-01
fix(core): insert PasteRule Node at matched position (#2942)

### LINKED ISSUE #2942: nodePasteRule adds the Node at the end of the editor

When using `nodePasteRule` the Node is added at the end of the editor instead of at the `range.from` position

The issue is new in `2.0.0-beta.181` because `nodePasteRule` wasn’t exposed before.

I’ve had the issue when trying to add a rule to transform `@something` into mentions (by extending @tiptap/extension-mention) when pasting plain text.

- Extend the Mention extension with `addPasteRules`
- Add a `nodePasteRule` using
  ```js
  nodePasteRule({
    find: /@([^\s@]+)/g,
    getAttributes: ([_, id]) => ({ id }),
    type: this.type,
  })
  ```
- Paste plain text in the editor with `@someting` in the middle (e.g. `Hello @user, how are you today?`)
- The `@something` will be transformed into a mention but moved to the end (e.g. `Hello , how are you today?@user`)

Vue: https://codesandbox.io/s/tiptap-nodepasterule-dg1f04?file=/src/App.vue

If you copy/paste directly the example in the CodeSandbox, you need to use `Cmd/Ctrl+Shift+V` to trigger the transformation.

I expected the mention to be added at the same place as the matching text

Replacing `.insertContent({` with `.insertContentAt(range.from, {` [here](https://github.com/ueberdosis/tiptap/blob/main/packages/core/src/pasteRules/nodePasteRule.ts#L32) seems to resolve the issue but I might be missing something.

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

edlb:
This PR is meant to fix #2942

Using `nodePasteRule` the Node would be added at the end of the editor.
This PR replaces `insertContent` with `insertContentAt` to insert the Node at the matched position instead.

### DIFF
--- a/packages/core/src/pasteRules/nodePasteRule.ts
+++ b/packages/core/src/pasteRules/nodePasteRule.ts
@@ -29,7 +29,7 @@ export function nodePasteRule(config: {
       if (match.input) {
         chain()
           .deleteRange(range)
-          .insertContent({
+          .insertContentAt(range.from, {
             type: config.type.name,
             attrs: attributes,
           })


bdbch (approved):
lgtm, thanks!

bdbch:
@edlb could you get the latest main branch merged and check if the tests are still failing?

PR #2952, opened 2022-07-04
fix(figure): insert the figure at the current cursor position

amirhhashemi:
Two problems with the current implementation:
1. We get an error if we try to insert a figure at the beginning of the document (Position -1 out of range)
2. When we add a figure, the cursor goes one line up!

### DIFF
--- a/demos/src/Experiments/Figure/Vue/figure.ts
+++ b/demos/src/Experiments/Figure/Vue/figure.ts
@@ -105,7 +105,7 @@ export const Figure = Node.create<FigureOptions>({
           // set cursor at end of caption field
           .command(({ tr, commands }) => {
             const { doc, selection } = tr
-            const position = doc.resolve(selection.to - 2).end()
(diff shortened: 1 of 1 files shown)


demos/src/Experiments/Figure/Vue/figure.ts
```diff
@@ -105,7 +105,7 @@ export const Figure = Node.create<FigureOptions>({
           // set cursor at end of caption field
           .command(({ tr, commands }) => {
             const { doc, selection } = tr
-            const position = doc.resolve(selection.to - 2).end()
+            const position = doc.resolve(selection.to).end()
```
ahhshm, line 108: I don't know what's going on here. Both `selection.to` and `selection.to - 1` work here but with `selection.to - 1`, when I add the figure at the beginning of the document, it doesn't automatically focus on the `figcaption` but with `selection.to` it does.

PR #2955, opened 2022-07-05
feat(figure): add an option to exit on enter

ahhshm:


demos/src/Experiments/Figure/Vue/figure.ts
```diff
@@ -177,6 +181,48 @@ export const Figure = Node.create<FigureOptions>({
...
+          
+          if (dispatch) {
+            const pos = $head.after()
+            const tr = state.tr.replaceWith(pos, pos, type.createAndFill()!)
+
+            tr.setSelection(Selection.near(tr.doc.resolve(pos), 1))
+            dispatch(tr.scrollIntoView())
+          }
+          return true
+        })
```
ahhshm, line 222: This could be its own command "exitBlockNode". The same thing is used in the `code-block` extension.

PR #2963, opened 2022-07-06
feat(extensions): add text-direction extension

ahhshm:
This plugin automatically adds `dir="ltr"` or `dir="rtl"` to a specified list of nodes based on their content. Forcing a direction is also possible through the `setTextDirection` command. The direction detection logic comes from [lexical](https://github.com/facebook/lexical).

### Why?
Why not? tiptap is one of the most popular rich text editors and it has been adopted for very large-scale applications. IMO it should have first-class support for RTL content. This plugin may not solve all of the issues with supporting both RTL and LTR content but definitely makes it easier.

### Demo

https://user-images.githubusercontent.com/87268103/178113964-db7e21e4-05d9-4339-9efc-84421c0b3b3f.mp4

Example configuration:
```ts
  TextDirection.configure({
    types: [
      "paragraph",
      "heading",
      "blockquote",
      "listItem",
    ],
  }),
```
`editor.getHTML()` result:
```html
<p dir="ltr">Hello</p>
<p dir="ltr">سلام fsdlfksjodfiwer hello</p>
<ul>
    <li dir="ltr">
        <p dir="ltr">hello</p>
    </li>
    <li dir="rtl">
        <p dir="rtl">سلام</p>
    </li>
    <li dir="ltr">
        <p dir="ltr">sghl</p>
    </li>
</ul>
<h2 dir="rtl">سلام</h2>
<h2 dir="ltr">hello</h2>
```

The `dir` attribute must be rendered instead of being a decorator (like text-align extention) because it's not just an in-editor thing. displaying and especially styling RTL content could be different than LTR content so `dir` must be in the HTML output to clarify the direction for both browsers and developers. 

### Why not `dir="auto"`?
Mainly for more control. With `dir="auto"` we can't change the direction later and also styling would be harder (if RTL and LTR content have different styles). 
"Where to add `dir=" auto"`" is another challenging question. if we add it to an element that wraps the editor, the direction of the whole document would be determined by the first node. We can add it to each node but, again, we don't have any control over the direction.

### Why not a custom paragraph node?
headings and other text nodes also need it

### Related issues  
https://github.com/ueberdosis/tiptap/issues/116
https://github.com/ueberdosis/tiptap/issues/1621

### DIFF
--- a/docs/api/extensions/text-direction.md
+++ b/docs/api/extensions/text-direction.md
@@ -1,0 +1,77 @@
+---
+description: Left, right, center, whatever! Align the text however you like.
+icon: align-left
+---
+
+# TextDirection
+
+[![Version](https://img.shields.io/npm/v/@tiptap/extension-text-direction.svg?label=version)](https://www.npmjs.com/package/@tiptap/extension-text-direction)
+[![Downloads](https://img.shields.io/npm/dm/@tiptap/extension-text-direction.svg)](https://npmcharts.com/compare/@tiptap/extension-text-direction?minimal=true)
+
+This plugin automatically adds the `dir` attribute to a specified list of nodes. The direction is automatically detected based on the node's content.
+
+## Installation
+
+```bash
+npm install @tiptap/extension-text-direction
+```
+
+## Settings
+
+### types
+
+A list of nodes where the `dir` attribute should be added to. Usually something like `['heading', 'paragraph']`.
+
+Default: `[]`
+
+```js
+TextDirection.configure({
+  types: ["heading", "paragraph"],
+});
+```
+
+### directions
+
+A list of available options for the 'dir' attribute.
+
+Default: `['ltr', 'rtl']`
+
+```js
+TextDirection.configure({
+  directions: ["ltr", "rtl"],
+});
+```
+
+### defaultDirection
+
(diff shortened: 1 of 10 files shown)


ahhshm:
Sorry, I accidentally removed my fork, and all of my PRs got closed

docs/api/extensions/text-direction.md
```diff
@@ -0,0 +1,77 @@
...
+If you specify a defaultDirection, the `dir` attribute will be excluded in nodes that have the same direction as the defaultDirection.
+:::
+
+Default: `null`
+
+```js
+TextDirection.configure({
+  defaultDirection: "rtl",
+});
+```
```
ahhshm, line 59: Hmm, maybe ignore is a better name for that. The idea is that some apps are designed for a specific language, let's say Arabic, so it's not necessary to add dir="rtl" to every node because that's the default direction that is set in the html tag.

ahhshm:
@bdbch should I publish this as my own library as you discussed in https://github.com/ueberdosis/tiptap/discussions/2973 ?

bdbch:
@ahhshm feel free to make it your own package for now and post it to the discussions board! :) We'll discuss when we'll merge it.

PR #2965, opened 2022-07-06
refactor: update lowlight dependency to 2.6.1 and set fixed peer version

bdbch:
This PR declares lowlight version `2.6.1` to the minimum version and fixes imports in the codeblock demos for lowlight.

From my tests locally 2.6.1 worked fine when the imports were fixed (instead of using a default import, use a named import on lowlight).


bdbch:
I think this one is deprecated by https://github.com/ueberdosis/tiptap/pull/3002

Issue #2880, opened 2022-06-14
Remove `@types/lowlight` from `@tiptap/extension-code-block-lowlight` dependencies


EvitanRelta:
### What’s the bug you are facing?

Outdated type-declaration from `@types/lowlight` package, which is a dependency for `@tiptap/extension-code-block-lowlight`.
The `lowlight` package already declares its own types, thus, the outdated `@types/lowlight` should be removed from the extension's dependencies.

![image](https://user-images.githubusercontent.com/35413456/173502033-88665336-83ef-439f-a322-1fee0b136355.png)

### Which browser was this experienced in? Are any special extensions installed?

-

### How can we reproduce the bug on our side?

Install the `@tiptap/extension-code-block-lowlight`, which installs `@types/lowlight` as a dependency.
Then do:
```js
import { lowlight } from 'lowlight'
```
And you'll notice an error saying `Module '"lowlight"' has no exported member 'lowlight'.`

### Can you provide a CodeSandbox?

Not my codesandbox, but taken from another issue: https://github.com/ueberdosis/tiptap/issues/2116#issuecomment-957797324
https://codesandbox.io/s/trusting-lake-22hls

Notice the error on the lowlight import:
![image](https://user-images.githubusercontent.com/35413456/173503488-4eb53766-e152-45d1-bd53-dc07ab392102.png)

### What did you expect to happen?

To import the updated type-declaration from `lowlight` package instead of the outdated `@types/lowlight` package.

### Did you update your dependencies?

- [X] Yes, I’ve updated my dependencies to use the latest version of all packages.

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

autumninflab:
I have same issue

rohitchouhan8:
If anyone is coming to look at this later. I'm using `"lowlight": "^3.1.0"`. The following code worked for me:

```tsx
import { createLowlight, common } from "lowlight";
const lowlight = createLowlight(common);

...

  CodeBlockLowlight.configure({
    lowlight,
    defaultLanguage: null,
  })
```

PR #3002, opened 2022-07-19
fix(extension-code-block-lowlight): Bump lowlight to 2.7.0, remove outdated @types

tarngerine:
Closes https://github.com/ueberdosis/tiptap/issues/2880

Lowlight now has TS types in the package itself. This PR bumps to latest 2.7.0, and removes the @types package which is very out of date and actively prevents developers from using current version of lowlight (because @types always takes precedence over built in types, causing TS to throw errors, see https://github.com/ueberdosis/tiptap/issues/2880

We may also choose to leave lowlight at 1.2.0, but without @types built-in, it would add friction to devs who want to plug-and-play the CodeBlockLowlight extension. IMO updating to 2.7.0 preserves this single npm install experience.

FWIW the docs are a little confusing, the currently claim that:

> You should provide the lowlight module to this extension. Decoupling the lowlight package from the extension allows the client application to control which version of lowlight it uses and which programming language packages it needs to load.

which I would personally love, but the extension currently provides an out of date lowlight with types that prevent client apps from using the latest lowlight.

happy to also put up a version with that change (just removing lowlight from the extension)

### DIFF
--- a/demos/package.json
+++ b/demos/package.json
@@ -11,6 +11,7 @@
     "@hocuspocus/provider": "^1.0.0-alpha.29",
     "d3": "^7.3.0",
     "fast-glob": "^3.2.11",
+    "highlight.js": "^11.6.0",
     "lowlight": "^2.7.0",
     "remixicon": "^2.5.0",
     "shiki": "^0.10.0",
--- a/demos/package.json
+++ b/demos/package.json
@@ -11,7 +11,7 @@
     "@hocuspocus/provider": "^1.0.0-alpha.29",
     "d3": "^7.3.0",
     "fast-glob": "^3.2.11",
-    "lowlight": "^1.20.0",
+    "lowlight": "^2.7.0",
     "remixicon": "^2.5.0",
     "shiki": "^0.10.0",
     "simplify-js": "^1.2.4",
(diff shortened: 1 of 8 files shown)


tarngerine:
i also found that package-lock.json in /demos no longer gets used, i'm guessing from before lerna was used. it's probably worth just deleting that file, too. it made local testing very confusing.

bdbch (approved):
LGTM - will test this on the staging environment. Thanks for the PR! :)

bdbch:
It seems like code highlighting isn't working in the demo anymore. Could you check out what went wrong there?

https://deploy-preview-3002--tiptap-embed.netlify.app/src/nodes/codeblocklowlight/react/
https://deploy-preview-3002--tiptap-embed.netlify.app/src/nodes/codeblocklowlight/vue/

bdbch:
I think you now have to register the languages in the demos. Take a look here:
https://github.com/ueberdosis/tiptap/pull/2965

Maybe that's why code highlighting currently isn't working.

bdbch:
Any updates @tarngerine

tarngerine:
I’ll check it out today sorry for the delay!

bdbch:
Thank you a lot! :)

bdbch:
Seems like the test is failing. Could you take a look into this again @tarngerine ?

bdbch (approved):

Issue #1515, opened 2021-06-27
AlpineJS v3 - Range Error: Applying a mismatched transaction


reefki:
**Description**
I got `Range Error: Applying a mismatched transaction` when clicking on a menu button with AlpineJS v3.

**CodeSandbox**
I created a CodeSandbox to help you debug the issue:
https://codesandbox.io/s/tiptap-v2-alpinejs-v3-s5x2o

jelib3an:
Hi, I think this is an Alpine issue. Downgrading to Alpine 2.8.2 works. It might have something to do with the fact that `this.editor` is a Proxy object - though it's just a theory. 

I don't know if the `getUnobservedData()` still works, but you could try to use it to get access to the actual editor object. https://codewithhugo.com/alpinejs-inspect-component-data-from-js/

Hope this info helps :smile:

0l1v3r5:
Thanks but `getUnobservedData()` is no longer available.

0l1v3r5:
@reefki did you find a way?

jelib3an:
It appears to be definitely a proxy issue. I sort of got something working here by storing the editor object in a closure. 
https://codesandbox.io/s/tiptap-v2-alpinejs-v3-forked-7rn6o?file=/index.js

You lose the reactivity with this approach though but it's a start and maybe this can point you in the right direction.

0l1v3r5:
@jelib3an Thanks a lot.

KevinBatdorf:
If you add Alpine to the window scope you can use `Alpine.raw(editor)` to unwrap the proxy and it should work.

robertdrakedennis:
@KevinBatdorf apologies with pinging you, can you explain your solution more? I'm a bit of a novice when it comes down further into the technical stuff and would really like to understand how to alleviate this issue with your solution. Thanks!

jelib3an:
@KevinBatdorf Thanks for the suggestion! I never knew about `Alpine.raw()`

@robertdrakedennis Here's an example with the mentioned solution. 
https://codesandbox.io/s/tiptap-v2-alpinejs-v3-forked-2bbw8?file=/index.js

However, `Alpine.raw()` doesn't appear to be bindable for reactivity. I also am Alpine novice. Would be interested to know if anyone has any good suggestions as it would be kind of a hassle to create a separate attribute for every button we want to check activeness on.

hanspagel:
[...]

PR #3081, opened 2022-08-12
docs: update alpine installation

patrickbaber:
The docs about the installation with Alpine.js is not compatible with version 3.

Should help with https://github.com/ueberdosis/tiptap/issues/1515

Also the default [Vite port](https://vitejs.dev/config/server-options.html#server-port) changed.

Thanks to @EasterPeanut

### DIFF
--- a/docs/installation/alpine.md
+++ b/docs/installation/alpine.md
@@ -6,7 +6,7 @@ tableOfContents: true
 # Alpine.js
 
 ## Introduction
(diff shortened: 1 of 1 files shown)


patrickbaber:
This is the prefered way to integrate external libraries with Alpine.js v3 as discussed here:

https://github.com/alpinejs/alpine/discussions/1953#discussioncomment-1203157

bdbch (approved):
Awesome, thank you!

PR #3089, opened 2022-08-16
fix(core): isNodeSelection, isTextSelection not always false

### LINKED ISSUE #2979: `isTextSelection` always returns false

The method `isTextSelection` always return `false` even when object is a `TextSelection`

Chrome, Safari

1. Import `isTextSelection` from `@tiptap/core`

   ```js
   import { isTextSelection } from '@tiptap/core'
   ```

2. Provide `editor.state.selection` to it
   ```js
   const isTextSelected = isTextSelection(editor.state.selection)
   ```

3. Select a text and `isTextSelected` will be false

https://codesandbox.io/s/determined-gauss-7tnuji?file=/src/App.js

`isTextSelected` should return `true` when text is selected

I'm using `@tiptap/react v2.0.0-beta.114` 
 
Thanks for this awesome library!

### Are you sponsoring us?

- [X] Yes, I’m a sponsor. 💖

kivikakk:
Fixes #2979.  Since the ProseMirror TypeScript upgrade, these have always returned `false`, since the Selection type tree are all classes now.

I've also removed `isObject` and `isClass`, as they're unused now.

(I'm noticing in #3026 that the "Add pull requests to Tiptap project" CI job looks broken right now.)


bdbch (approved):
Looks good to me, thank you for this PR and a good find!

Regarding the GitHub CI action: Looking into that 👍

PR #3091, opened 2022-08-16
fix(core): setNodeSelection should not clamp pos by Selection.atStart/atEnd

### LINKED ISSUE #3090: setNodeSelection(non-leaf first node in doc with text) can create a NodeSelection for text

If you try to `setNodeSelection()` on the very first node in the document, and that node is non-leaf, if there's a text selection able to be made within that node, Tiptap will end up creating a `NodeSelection` for the first text node within that.

The text `NodeSelection` doesn't behave well, and causes an error to be thrown in its deselect when trying to select anything else.

Chrome 104.0.5112.79 (Official Build) (arm64), unlikely to be browser-specific.

See this CodeSandbox which reproduces this in `onCreate`: https://codesandbox.io/s/tiptap-setnodeselection-repro-mtht1t?file=/src/App.js

This is a contrived example using `paragraph`s for simplicity, but it happens with more complex custom node types which are both selectable themselves and contain editable content.

_No response_

The first node should be selected (if possible) OR no selection made at all. An invalid selection shouldn't be made, at any rate.

This happens because `setNodeSelection` clamps the `position` parameter to selectable ranges in the document:

https://github.com/ueberdosis/tiptap/blob/57a9e864f4a5f55f21caec2dc5a2382f11b63ae7/packages/core/src/commands/setNodeSelection.ts#L20-L23

This wasn't always the case; the change to clamp the parameter this way happened in 9425e7249. Before that commit, it was just to the bounds of the whole document:

https://github.com/ueberdosis/tiptap/blob/f57466ccd009985bda4c928aedbc50e2b8be0c79/packages/core/src/commands/setNodeSelection.ts#L19-L20

9425e7249 made the same change (using `Selection.atStart(doc).from` and `Selection.atEnd(doc).to`) to both `setTextSelection` and `setNodeSelection`, but I'm inclined to believe that **only the `setTextSelection` change** was necessary to address #1588, and that furthermore the `setNodeSelection` change was incorrect.

[`Selection.atStart`](https://github.com/ProseMirror/prosemirror-state/blob/3f9311ec2fc6f0803c1dbe59dfd2297f731dbd5c/src/selection.ts#L139-L145) proceeds to find the first cursor position within the node (because it's not an atom) and returns a `TextSelection` at position 1, which in turns becomes `minPos`:

```typescript
  /// Find the cursor or leaf node selection closest to the start of
  /// the given document. Will return an
  /// [`AllSelection`](#state.AllSelection) if no valid position
  /// exists.
  static atStart(doc: Node): Selection {
    return findSelectionIn(doc, doc, 0, 0, 1) || new AllSelection(doc)
  }
```

I think we should just revert `setNodeSelection` to bounding within `[0, doc.content.size]` again.

### Are you sponsoring us?

- [X] Yes, I’m a sponsor. 💖

kivikakk:
Fixes #3090.

This doesn't regress #1588, as that bug was only in the `focus` command.

### DIFF
--- a/packages/core/src/commands/setNodeSelection.ts
+++ b/packages/core/src/commands/setNodeSelection.ts
@@ -1,4 +1,4 @@
-import { NodeSelection, Selection } from 'prosemirror-state'
+import { NodeSelection } from 'prosemirror-state'
 
 import { RawCommands } from '../types'
 import { minMax } from '../utilities/minMax'
@@ -17,10 +17,8 @@ declare module '@tiptap/core' {
 export const setNodeSelection: RawCommands['setNodeSelection'] = position => ({ tr, dispatch }) => {
   if (dispatch) {
     const { doc } = tr
-    const minPos = Selection.atStart(doc).from
-    const maxPos = Selection.atEnd(doc).to
-    const resolvedPos = minMax(position, minPos, maxPos)
-    const selection = NodeSelection.create(doc, resolvedPos)
+    const from = minMax(position, 0, doc.content.size)
+    const selection = NodeSelection.create(doc, from)
 
     tr.setSelection(selection)
   }


bdbch (approved):
looks good to me, thanks

PR #3147, opened 2022-09-02
Syntax Error

ElderlyBoy:
1. SyntaxError: The requested module '/node_modules/.vite/deps/lowlight.js?v=539bea66' does not provide an export named 'default' ;
2. Uncaught (in promise) Error: You should provide an instance of lowlight to use the code-block-lowlight extension;


bdbch:
Thanks for your PR. I m just pushed a fix for this issue for all our demos including the one you're fixing here. Thank you for bringing it up.

PR #3150, opened 2022-09-02
Include bubble menu element when checking if the editor view still has focus

StefKors:
Before the bubble menu would flicker when selecting styles

Screen Recording 2022-09-02 at 16 06 06

---------

Afterwards the bubble menu stays visible when selecting styles

Screen Recording 2022-09-02 at 18 05 56


bdbch (approved):
Great one! Thanks for your contribution.

PR #3155, opened 2022-09-06
feat: Add alias condition to code-block-lowlight

dngwoodo:
Lowlight's `lowlight.listLanguages` does not support `alias`.

So, for example, if you use `alias` called `js`, it is not parsed properly.

Currently, a method called `registered` exists in the lowlight 2 version, but it does not exist in the 1 version, so this part is directly implemented.

Please check and merge.


bdbch:
We just upgraded lowlight in our demos to 2.7.0

Do you think this PR is still needed @dngwoodo

dngwoodo:
I think it is still necessary. It's the same as the low light internal implementation

The code `const nodes = language &&'(language.includes(language) || registered(language))` is still needed. Even if you upgrade lowlight, alias is still not checked.

bdbch (approved):

PR #3188, opened 2022-09-13
fix: move React `flushSync` to microtask

sampi:
To avoid seeing the `Warning: flushSync was called from inside a lifecycle method. React cannot flush when React is already rendering. Consider moving this call to a scheduler task or micro task.` error, we need to move the `flushSync()` code that avoids automatic batching to a microtask to not fire a lifecycle event `setState()` during rendering.

Fixes warning introduced in #2985

### DIFF
--- a/packages/react/src/ReactRenderer.tsx
+++ b/packages/react/src/ReactRenderer.tsx
@@ -78,15 +78,17 @@ export class ReactRenderer<R = unknown, P = unknown> {
 
     this.reactElement = <Component {...props } />
 
-    flushSync(() => {
-      if (this.editor?.contentComponent) {
-        this.editor.contentComponent.setState({
-          renderers: this.editor.contentComponent.state.renderers.set(
-            this.id,
-            this,
-          ),
-        })
-      }
+    queueMicrotask(() => {
+      flushSync(() => {
+        if (this.editor?.contentComponent) {
+          this.editor.contentComponent.setState({
+            renderers: this.editor.contentComponent.state.renderers.set(
+              this.id,
+              this,
+            ),
+          })
+        }
+      })
     })
   }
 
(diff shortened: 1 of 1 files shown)


bdbch (approved):
Thanks for the PR. Looks good to me!

jamesopti:
@bdbch @sampi - FY We just tested this change in our Tiptap React 18 application and noticed it has the same cursor position bug that this `flushSync` technique was supposed to fix in the first place.

For us, reproducing is as easy as:

1. Type a paragraph
2. Press ENTER

**_Expected_**: Cursor moves to next line (new empty paragraph)
**_Actual_**: Cursor remains on line before, prosemirror cursor state is on the next line where it should be

bdbch:
Thanks for reporting. I'll try reproduce this too.

jamesopti:
FWIW We are on `prosemirror-view@npm:1.26.5`, which is now higher than the exact version allowed by tiptap/core (`1.26.2`)

We had upgraded to this version because in the past, Tiptap allowed minor version upgrades.

Was there a specific reason that 1.26.2 was pinned?

bdbch:
We had a few issues with minor changes breaking the editor beforce which is why we fixed it.
I'll try out your version and see if we require to update the fixed dependency on our side (or maybe even remove the strict version fixing if possible).

jamesopti:
> @bdbch @sampi - FYI We just tested this change in our Tiptap React 18 application and noticed it has the same cursor position bug that this `flushSync` technique was supposed to fix in the first place.
> 
> For us, reproducing is as easy as:
> 
> 1. Type a paragraph
> 2. Press ENTER
> 
> **_Expected_**: Cursor moves to next line (new empty paragraph) **_Actual_**: Cursor remains on line before, prosemirror cursor state is on the next line where it should be

@bdbch I can confirm that with all the latest versions of Tiptap packages that this is still an issue.

kylealwyn:
Also experiencing this issue on latest beta

PR #3192, opened 2022-09-13
fix(core): implement deleteCurrentNode command & fix node joining on Delete key

### LINKED ISSUE #2924: Pressing `Delete` on an empty paragraph clears the node afterwards

This is better explained with a short demo:

firefox_4DBYnuFKvn

All of them. No special extensions.

Use the [default example](https://tiptap.dev/examples/default) editor, and replicate what you see in the GIF above.

Pressing `Delete` on an empty paragraph (or `Shift + Backspace` on macOS, I believe), should delete the empty paragraph where the text cursor is located (and not clear the node that comes afterwards).

### Are you sponsoring us?

- [X] Yes, I’m a sponsor. 💖 _(on behalf of Doist)_

bdbch:
This PR should fix #2924 by adding a new command called `deleteCurrentNode`.

This command is then used inside the keymap core extension right before the `joinForward` command to check if the current node is empty - if it is, the current node will be simply deleted instead of joining the empty node with the next node.

This way, if for example an empty paragraph node is followed by an blockquote, the empty paragraph node will be removed and the selection will be moved into the blockquote.

If the paragraph node is **not** empty, the blockquote will be transformed into a paragraph node as it used to be.


rfgamaral:
@bdbch This is awesome, thanks for taking the time to dig into this one.

However, while attempting to fix this on my side (I couldn't), I found an edge case that I'm not sure if this implementation takes care of. The deployment seems to have failed, but I'll try to find some time today to check out this branch, and test it out. I'll let you know what I find as soon as possible.

bdbch:
Thanks for your response @rfgamaral! I'm taking a look into the failed deployment right now.

bdbch:
Seems like the Netlify cache was breaking the build. It's deployed now. Feel free to test here:
https://6320b6c00a7e9271f33a732c--tiptap-embed.netlify.app/src/examples/book/react/

bdbch:
What edge case did you find?

rfgamaral:
@bdbch This one:

https://user-images.githubusercontent.com/96476/189993644-efb11f3c-503e-4860-905c-f55c2fa94cba.mp4

Based on my experience with other editors, in this particular case, the blockquote node (or other block node like a list item) should indeed be removed, and the text joined with the paragraph where the cursor was.

Short demo from Google Docs:

https://user-images.githubusercontent.com/96476/189994121-296f4ba8-caa9-4cfd-9ae7-52fd87837dc2.mp4

Hope you can fix this one as well 🤞

bdbch:
I'll take a look at this case too!

bdbch:
@rfgamaral sorry didn't find time yesterday, was quite late already haha. I'll take a look at my PR later this day and I'll see if I can find a quick solution for the issue you showed me.

rfgamaral:
@bdbch That's okay, I understand, take the time required to get it right 👍

bdbch:
@rfgamaral I worked on this a bit but couldn't figure out the "right" way to do it. What I'm on right now:

1. Get to the parent node (for example the paragraph)
2. Look for the next text node found after the parent node (for example in the next paragraph, list item, etc)
3. Copy the text to the end of the current cursor $anchor position
4. Remove the parent of the copied text node

But I'm not 100% sure if this is the expected behaviour. I'm still experimenting - just a heads up for what is going on in here.

bdbch:
@rfgamaral since this PR takes longer I'll merge it soon and release it with the next release. I created a new issue for the edgecase you reported: #3454

packages/core/src/commands/deleteCurrentNode.ts
```diff
@@ -0,0 +1,41 @@
+import { RawCommands } from '../types'
+
+declare module '@tiptap/core' {
+  interface Commands<ReturnType> {
+    deleteCurrentNode: {
+      /**
+       * Delete the node that currently has the selection anchor.
+       */
+      deleteCurrentNode: () => ReturnType,
```
aleclarson, line 9: This command should be marked `@internal` or at least have its behavior documented more clearly. I tried using it thinking it would delete the current node, but the current node was not empty, so it wasn't deleted, leading me to think Tiptap was broken.

PR #3205, opened 2022-09-16
fix(core): InputRules does not work for ranges containing multiple text nodes

### LINKED ISSUE #3071: InputRules does not work for ranges containing multiple text nodes.

When markdown tags cover multiple text nodes, they will not work. See GIF:

动画

The initial problem was #2937, fixed by #2941.

But if a paragraph contains multiple text nodes within it, and InputRule needs to span those nodes, it will not work.

I'm using the latest code from the main branch.

Chrome/103.0.0.0. No special extensions.

1. type 'A   B   C'
2. make 'B' bold
3. type '*' at the begining
4. type '*' at the end

correct result: *A **B** C*
incorrect result: \*A **B** C\*

No CodeSandbox, but there is a "Deploy Preview" provided by #2941

<https://deploy-preview-2941--tiptap-embed.netlify.app/>

InputRule should work between multiple text nodes.

_No response_

- [x] Yes, I’ve updated my dependencies to use the latest version of all packages.

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

hamflx:
try fix #3071

### DIFF
--- a/packages/core/src/helpers/getTextContentFromNodes.ts
+++ b/packages/core/src/helpers/getTextContentFromNodes.ts
@@ -3,13 +3,15 @@ import { ResolvedPos } from 'prosemirror-model'
 export const getTextContentFromNodes = ($from: ResolvedPos, maxMatch = 500) => {
   let textBefore = ''
 
+  const to = $from.parentOffset
   $from.parent.nodesBetween(
-    Math.max(0, $from.parentOffset - maxMatch),
-    $from.parentOffset,
+    Math.max(0, to - maxMatch),
+    to,
     (node, pos, parent, index) => {
-      textBefore += node.type.spec.toText?.({
+      const chunk = node.type.spec.toText?.({
         node, pos, parent, index,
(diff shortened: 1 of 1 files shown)


packages/core/src/helpers/getTextContentFromNodes.ts
```diff
@@ -3,13 +3,15 @@ import { ResolvedPos } from 'prosemirror-model'
 export const getTextContentFromNodes = ($from: ResolvedPos, maxMatch = 500) => {
   let textBefore = ''
 
+  const to = $from.parentOffset
```
bdbch, line 6: Just a quick one - since you're doing `to = $from` - would there be a more fitting variable name to explicitly tell what this is "to"? Otherwise this looks fine to me.

bdbch: That sounds better. Thank you! I'd merge this afterwards.

PR #3208, opened 2022-09-17
fix(types): fix link and table type errors

johnfraney:
A small PR to fix some typing errors in the link and table extensions,


bdbch (approved):
Looks good. Thanks for the PR!

johnfraney:
@bdbch I'm happy to help! This library has the best WYSIWYG table editor I could find, so it's the least I could do

PR #3223, opened 2022-09-22
fix(core): Can() does not work for setting marks

### LINKED ISSUE #3057: can() command working Unexpectedly

We have [can command](https://tiptap.dev/api/editor#can) in tiptap to check if a command can be executed - without actually executing it. I am using this command in my editor to check weather [toggleBold command](https://tiptap.dev/api/marks/bold#toggle-bold) can be executed or not in presence of active [code Mark](https://tiptap.dev/api/marks/code).
In source code of extension-code, [here](https://github.com/ueberdosis/tiptap/blob/a06209890a881c7a1540e1fe8080740d7ae7b9aa/packages/extension-code/src/code.ts#L43) it is shown that [code](https://tiptap.dev/api/marks/code) excludes all the other marks, so ideally commands like setBold or toggleBold can't be executed, but when I am printing result of can with these commands, it is showing as true.

**To explain my use-case:** I have built a toolbar where I am disabling buttons in case their action can't be run. Like when code / CodeBlock is active, we can't make the text bold, but can command is not working correctly.

**PS:** With CodeBlock which disables all other marks, shown in code [here](https://github.com/ueberdosis/tiptap/blob/a06209890a881c7a1540e1fe8080740d7ae7b9aa/packages/extension-code-block/src/code-block.ts#L58), can is also not working correctly.

Correct me, if there is any problem with my understanding of correct behavior of can command.

In all browsers.

Use this: https://codesandbox.io/s/cranky-johnson-78eszt?file=/src/App.js
Or can use any example of tiptap starter-kit.

https://codesandbox.io/s/cranky-johnson-78eszt?file=/src/App.js

[can command](https://tiptap.dev/api/editor#can) should return false if we test for [toggleBold command](https://tiptap.dev/api/marks/bold#toggle-bold) if Code / CodeBlock is active.

_No response_

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

C-Hess:
Previously, setting marks did no schema validation checks for dry runs (like the `.can()` command). The `setMark` raw command will now properly check if the mark is possible to be set given the editor node/mark schema.

The logic was added is similar to the private `markApplies` method used internally by ProseMirror's `toggleMark` command but with additional functionality to support stored cursor marks and mark exclusions.

fix #3057

### DIFF
--- a/demos/src/Examples/Default/React/index.jsx
+++ b/demos/src/Examples/Default/React/index.jsx
@@ -13,24 +13,52 @@ const MenuBar = ({ editor }) => {
     <>
       <button
         onClick={() => editor.chain().focus().toggleBold().run()}
+        disabled={
+          !editor.can()
+            .chain()
+            .focus()
+            .toggleBold()
+            .run()
+        }
         className={editor.isActive('bold') ? 'is-active' : ''}
       >
         bold
       </button>
       <button
         onClick={() => editor.chain().focus().toggleItalic().run()}
+        disabled={
+          !editor.can()
+            .chain()
+            .focus()
+            .toggleItalic()
+            .run()
+        }
         className={editor.isActive('italic') ? 'is-active' : ''}
       >
         italic
       </button>
       <button
         onClick={() => editor.chain().focus().toggleStrike().run()}
+        disabled={
+          !editor.can()
+            .chain()
+            .focus()
+            .toggleStrike()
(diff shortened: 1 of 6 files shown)


bdbch (approved):
Great pull request. Thanks @C-Hess

I didn't really see an issue with your implementation.

PR #3232, opened 2022-09-25
feat(extension-link): Change autolink to only apply after breaking a word

### LINKED ISSUE #3225: Autolinking links should only trigger after hitting space

The autolink feature for extension-link has a minor UX issue that is preventing us from using the feature in our production environment. Essentially, the autolink feature applies while you are typing a link. Other editors do not detect/convert URLs until **after** the space key is pressed including Google Docs, TinyMCE, CKEditor, and Confluence/Atlassian.

When you can "unlink" a link (for example if you don't want to link https://google.com), any time you make a change to the URL after the link is broken it will re-add the link. However, if you require the space key to be pressed first, you can make changes to the link without it trying to re-link it.

I'd like the auto-link feature only to apply after the space key is entered to break up the word. In addition, it's probably best not to have this be a feature behind an option, but rather a change to the base functionality).

I considered trying to make my own plugin to autolink and disable the base plugin, but I figure it's best just to try and fix the extension.

We could probably just update the autolink helper/plugin to only fire if the changed text ranges contain a space. I'd have to experiment to see if that works. I'm okay with opening a PR for this, but I figured I'd open an issue first to get any input/suggestions for this

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

C-Hess:
Made a change to the autolink plugin for extension-link to only autolink after hitting space, which is a more common behavior on WYSIWYG editors. I wanted to also remove the auto-"unlinking" feature, as this is also not a common behavior, but I figured I wouldn't change this in this PR unless requested

Fix #3225

### DIFF
--- a/demos/src/Examples/AutolinkValidation/React/index.jsx
+++ b/demos/src/Examples/AutolinkValidation/React/index.jsx
@@ -23,8 +23,48 @@ export default () => {
     },
   })
 
+  const setLink = React.useCallback(() => {
+    const previousUrl = editor.getAttributes('link').href
+    const url = window.prompt('URL', previousUrl)
+
+    // cancelled
+    if (url === null) {
+      return
+    }
+
+    // empty
+    if (url === '') {
+      editor.chain().focus().extendMarkRange('link').unsetLink()
+        .run()
+
+      return
+    }
+
+    // update link
+    editor.chain().focus().extendMarkRange('link').setLink({ href: url })
+      .run()
+  }, [editor])
+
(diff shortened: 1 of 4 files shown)


bdbch (approved):
Thanks for the PR! Looks good to me. Nice.

xiaweiss:
@C-Hess I found a bug here: https://github.com/ueberdosis/tiptap/issues/3254

I add an `if` to skip `undefined`, can you review it?

Issue #3238, opened 2022-09-26
Mention Plugin cannot display suggest list with punctuation mark case


Matrixbirds:
### What’s the bug you are facing?

I'm using the alpha version of @tiptap/suggestion and @tiptap/mention plugin.
I try to type 「dot」「at」, mention plugin doesn't show mention list.

### Which browser was this experienced in? Are any special extensions installed?

```bash
Chrome on MacOS
Version:
105.0.5195.125 (Official Build) (arm64)
Revision | 1e8949033cc650c8496b3b0af7cd54372f7ff62b-refs/branch-heads/5195@{#1107}
OS | macOS Version 12.1 (Build 21C52)
```

### How can we reproduce the bug on our side?

Gif of reproduce steps:
![reprodecure-gif](https://user-images.githubusercontent.com/9990676/192280829-b71c69c3-3768-4978-a556-b6e0b0df3721.gif)

### What did you expect to happen?

It should be available open mention list.

### Did you update your dependencies?

- [X] Yes, I’ve updated my dependencies to use the latest version of all packages.

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

PR #3239, opened 2022-09-26
fix: set default allowedPrefixes null

Matrixbirds:
fix issue #3238


akrolsmir:
Hrm, given the number of breakages, I'd say this was bad change -- imo default behavior should be `[' ']`, as it used to be.

At the very least, y'all should update https://tiptap.dev/api/utilities/suggestion to mention the new default

PR #3251, opened 2022-09-28
Ensure text blocks exist before referencing them

### LINKED ISSUE #3250: Cannot read position pos of undefined in autolink.ts

Right now, if you are using both the image extension, and the link extension, you can cause the editor to crash with an error by dragging an image into the editor. 

Browser: Chrome latest.
Relevant extensions: image and link

Have a look at the CodeSandbox. It appears as though the steps are:

- Use ProseMirror with both the link and image extensions,
- Drag an image into the editor

Here's a codesandbox https://codesandbox.io/s/small-thunder-k9wl2m?file=/src/App.vue

No error, but nothing in the editor should happen.

Happy to raise a PR. My fix would be to add a check on the `nodesInChangedRange` array in the condition below to check that the array is not empty:

https://github.com/ueberdosis/tiptap/blob/main/packages/extension-link/src/helpers/autolink.ts#L76-L79

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

ScopeyNZ:
Fixes #3250

### DIFF
--- a/packages/extension-link/src/helpers/autolink.ts
+++ b/packages/extension-link/src/helpers/autolink.ts
@@ -74,8 +74,9 @@ export function autolink(options: AutolinkOptions): Plugin {
             ' ',
           )
         } else if (
+          nodesInChangedRanges.length
           // We want to make sure to include the block seperator argument to treat hard breaks like spaces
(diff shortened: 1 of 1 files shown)


bdbch (approved):
Thanks for the quick fix!

PR #3258, opened 2022-09-30
fix(extension/link): fix last word value being undefined

bdbch:
This PR fixes the lastWordBeforeSpace value to not be passed through the autolink plugin when it is undefined. This can happen, when the filtered `wordBeforeWhitespace` array has a length of 0 since `lastWordBeforeSpace` is set to `wordsBeforeWhitespace.length - 1` which would result in an index of -1.


nqnduy:
I have problem with extension/link despite not interacting with link (using dragging image above or below another line, or resize image (customize extension))

this ERROR is:
TypeError: Cannot read properties of undefined (reading 'pos')

newState.doc.textBetween(newRange.from, newRange.to, ' ', ' ').endsWith(' ')) {
    textBlock = nodesInChangedRanges[0];
    **textBeforeWhitespace = newState.doc.textBetween(textBlock.pos, newRange.to, undefined, ' ');**
  }

PR #3269, opened 2022-10-03
Fix #3084 - remove old newline-fix for ios that freezes safari on ios

### LINKED ISSUE #3084: [iOS, React] Pressing "Enter" in an interactive node view freezes the whole page

On an iOS device, pressing "Enter" in an interactive node view makes the page unresponsive. The issue exists only for React components. Vue and plain JavaScript work fine.

- Safari on iPhone 7 (15.6)
- Safari on some iPad Pro (15.6)

1. Open https://tiptap.dev/guide/node-views/react#adding-a-content-editable on an iOS device.
2. Place the cursor inside the interactive node view and press "Enter".

Doesn't seem to be necessary.

Pressing "Enter" in an interactive node view should not freeze the whole page.

The issue is also reproducible in the latest version of tiptap.

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

albertogiunta:
Started experiencing this issue (#3084) after updating to the latest versions of TipTap. Also confirmed it's present on the official TipTap website. It took a while to triage, and I finally solved it by completely removing a fix/patch that had been previously added to avoid double new lines.

I tried the editor after removing this fix, and it now 1) doesn't freeze anymore 2) doesn't add multiple lines on a single Enter key press.

I might be missing some context on issue #1214 (the one that the patch I'm removing originally fixed), but if it was due to a now outdated versions of TipTap/ProseMirror we might be good with removing it.

### DIFF
--- a/packages/core/src/NodeView.ts
+++ b/packages/core/src/NodeView.ts
@@ -203,28 +203,6 @@ export class NodeView<
       return false
     }
 
-    // try to prevent a bug on iOS that will break node views on enter
-    // this is because ProseMirror can’t preventDispatch on enter
-    // this will lead to a re-render of the node view on enter
-    // see: https://github.com/ueberdosis/tiptap/issues/1214
-    if (
-      this.dom.contains(mutation.target)
-      && mutation.type === 'childList'
-      && isiOS()
-      && this.editor.isFocused
-    ) {
-      const changedNodes = [
-        ...Array.from(mutation.addedNodes),
-        ...Array.from(mutation.removedNodes),
-      ] as HTMLElement[]
-
-      // we’ll check if every changed node is contentEditable
-      // to make sure it’s probably mutated by ProseMirror
-      if (changedNodes.every(node => node.isContentEditable)) {
-        return false
-      }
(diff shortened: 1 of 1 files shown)


fenok:
Just a word of caution: for me, removing this fix/patch brings multiple lines on a single Enter key press back.

bdbch:
Since this is reintroducing the bug mentioned by @fenok I'd rather would like to fix this issue instead of removing the fix. Do you see any option how to fix the freezing iOS view without removing the fix entirely? @albertogiunta

bdbch:
@nperez0111 since you mentioned me here for a review:

I think we need to double verify this PR and make sure it doesn't introduce the double-line issue again (as @fenok said). I'm not sure when this issue occurs exactly.

arnaugomez:
It looks like the original issue has been fixed. Therefore, this PR is no longer linked to any open issue. Should we close it? @bdbch

bdbch:
I'll close this PR for now @albertogiunta if you still run into the described issue let me know and I'll reopen

PR #3288, opened 2022-10-08
feat(extensions): new hashtag extension

Narixius:
This PR adds hashtag support to the editor. This extension can be integrated with suggestions.

### DIFF
--- a/demos/src/Marks/Hashtag/React/HashtagList.jsx
+++ b/demos/src/Marks/Hashtag/React/HashtagList.jsx
@@ -1,0 +1,70 @@
+import './HashtagList.scss'
+
+import React, {
+  forwardRef, useEffect, useImperativeHandle,
+  useState,
+} from 'react'
+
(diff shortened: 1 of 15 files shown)


bdbch:
From a first look it looks good.

https://deploy-preview-3288--tiptap-embed.netlify.app/src/marks/hashtag/react/

While testing here, I found a few bugs, specially when the hashtag gets used at the start of a line. Could you investigate this behaviour?

piotrski:
Is there a reason why hashtag and mention do not have a common implementation? https://github.com/ueberdosis/tiptap/blob/main/packages/extension-mention/src/mention.ts

bdbch:
After coming back to this I wonder if it's actually needed. The mentions package is doing this already for you and most of the logic is a 1:1 copy of the suggestions + mention setup just with a different suggestion char I think?

nperez0111:
Agreed, closing

Issue #3310, opened 2022-10-14
Select all text in ordered-list not active button paragraph and text align


jambonn:
### What’s the bug you are facing?

When I integrate extensions paragraph, bullet, ordered-list. I change the text to the order-list then I set the text align center for the entire order-list, when I ctrl+a all the text the button center is not active.

### Which browser was this experienced in? Are any special extensions installed?

Chrome

### How can we reproduce the bug on our side?

Go to the included stackblitz, screen preview action ctrl+a text, seen button center and paragraph not active

### Can you provide a CodeSandbox?

https://stackblitz.com/edit/vitejs-vite-ncdisw?file=src%2Fcomponents%2FTiptap.vue&terminal=dev

![Alt Text](https://media.giphy.com/media/guoHzy3mPAivvNGRSv/giphy.gif)

### What did you expect to happen?

Action button when select all text in order-list

### Did you update your dependencies?

- [X] Yes, I’ve updated my dependencies to use the latest version of all packages.

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

PR #3313, opened 2022-10-15
fix: select all text order-list not active state text align

jambonn:
- I don't understand why after selecting all the text of the order-list, the range is always smaller than the selectionRange by 2 units, so I try to fix this way
- Issue: [#3310](https://github.com/ueberdosis/tiptap/issues/3310).


jambonn:
https://deploy-preview-3313--tiptap-embed.netlify.app/src/experiments/all/vue/ Work as expected

jambonn:
@bdbch I'm dealing with this. Hope you let me comment on this PR. Thank you!

nperez0111:
This is wrong, and the issue has been resolved since

PR #3331, opened 2022-10-20
Fix cursor not following to new node when using a react node view

### LINKED ISSUE #3200: Cursor does not follow new line / paragraph when configured with a node view

Following the documentation, I've tried to extend the paragraph extension with a custom node-view. When pressing the `Enter` key a new node is created, however the cursor remains in the initial paragraph.

 Desktop / Linux / MacOS / Chrome & Firefox (no extensions)

<details>
  <summary>Minimal Example Code</summary>

```typescript
import * as React from 'react';

import {
  EditorContent,
  useEditor,
  NodeViewWrapper,
  NodeViewContent,
  ReactNodeViewRenderer,
} from '@tiptap/react';

import { Document } from '@tiptap/extension-document';
import { Text } from '@tiptap/extension-text';
import { Paragraph } from '@tiptap/extension-paragraph';

const NodeView = () => {
  return (
    <NodeViewWrapper>
      <div>
        <NodeViewContent />
      </div>
    </NodeViewWrapper>
  );
};

const CustomParagraph = Paragraph.extend({
  addNodeView() {
    return ReactNodeViewRenderer(NodeView);
  },
});

export default function App() {
  const editor = useEditor({
    extensions: [Document, CustomParagraph, Text],
    content: '...',
  });
  return (
    <div>
    </div>
  );
}
```

</details>

 - Start typing
 - Press `Enter`
 - Keep typing

https://stackblitz.com/edit/react-ts-gth2ze?file=App.tsx

I expected the cursor to always jump to the new paragraph.

I would also note that sometimes the cursor _will_ jump into the new node if there is a trailing space before pressing `Enter`.

https://user-images.githubusercontent.com/2750352/190435761-3c79273f-4916-4967-8483-fdde76611849.mov

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

ruipserra:
This fixes the cursor problem described in #3200.

Node views need to be rendered immediately when they're created so that the editor can correctly position the cursor inside them. That's achieved using `flushSync` whenever a new node view renderer is added.

However, `flushSync` cannot be used from inside a React component lifecycle method. Thus #3188 moved the `flushSync` call into a microtask. This delays the `flushSync`, and so the editor can't correctly position the cursor inside the new node view, because it doesn't exist in the DOM yet.

By keeping an instance variable to determine if initialization has happened, we can avoid using `flushSync` from inside the `componentDidMount` and `componentDidUpdate` methods, and still call it whenever a new node view is created afterwards.

### How I tested this

I added the following code to the `ReactComponentContent` demo (but I didn't commit this):

```diff
diff --git a/demos/src/GuideNodeViews/ReactComponentContent/React/Extension.js b/demos/src/GuideNodeViews/ReactComponentContent/React/Extension.js
index b3cde8328..6d70f3c61 100644
--- a/demos/src/GuideNodeViews/ReactComponentContent/React/Extension.js
+++ b/demos/src/GuideNodeViews/ReactComponentContent/React/Extension.js
@@ -10,6 +10,18 @@ export default Node.create({
 
   content: 'inline*',
 
+  addKeyboardShortcuts() {
+    return {
+      Enter: ({ editor }) => {
+        if (editor.state.selection.$from.parent.type.name === this.name) {
+          return this.editor.chain().splitBlock().setNode(this.name).run()
+        }
+
+        return false
+      }
+    }
+  },
+
   parseHTML() {
     return [
       {

``` 

Then I tested hitting enter at the end of the react node:

| Before | After
| - | - |
| <video src="https://user-images.githubusercontent.com/3883777/196845068-ae4f3809-804d-4164-a342-177578b34c53.mov" /> |  <video src="https://user-images.githubusercontent.com/3883777/196845197-aa76dd82-1025-43a2-b8d5-33f00377f28a.mov" /> |

### Additional changes in this PR

* Moved the code that was updating `EditorContent`'s state in `ReactRenderer` into instance methods in `EditorContent`.
* Updated `EditorContent` to keep the renderers in a JS object instead of an ES6 Map. ES6 Maps are mutable, and React discourages mutating state.

---

Fixes #3200

### DIFF
--- a/packages/react/src/Editor.ts
+++ b/packages/react/src/Editor.ts
@@ -2,7 +2,13 @@ import { Editor as CoreEditor } from '@tiptap/core'
 import React from 'react'
 
 import { EditorContentProps, EditorContentState } from './EditorContent'
+import { ReactRenderer } from './ReactRenderer'
+
+type ContentComponent = React.Component<EditorContentProps, EditorContentState> & {
+  setRenderer(id: string, renderer: ReactRenderer): void;
+  removeRenderer(id: string): void;
+}
 
 export class Editor extends CoreEditor {
-  public contentComponent: React.Component<EditorContentProps, EditorContentState> | null = null
+  public contentComponent: ContentComponent | null = null
 }
--- a/packages/react/src/EditorContent.tsx
+++ b/packages/react/src/EditorContent.tsx
@@ -4,10 +4,10 @@ import ReactDOM, { flushSync } from 'react-dom'
 import { Editor } from './Editor'
 import { ReactRenderer } from './ReactRenderer'
 
-const Portals: React.FC<{ renderers: Map<string, ReactRenderer> }> = ({ renderers }) => {
+const Portals: React.FC<{ renderers: Record<string, ReactRenderer> }> = ({ renderers }) => {
   return (
     <>
-      {Array.from(renderers).map(([key, renderer]) => {
+      {Object.entries(renderers).map(([key, renderer]) => {
         return ReactDOM.createPortal(
           renderer.reactElement,
           renderer.element,
@@ -23,7 +23,7 @@ export interface EditorContentProps extends HTMLProps<HTMLDivElement> {
 }
 
 export interface EditorContentState {
-  renderers: Map<string, ReactRenderer>
+  renderers: Record<string, ReactRenderer>
 }
 
 export class PureEditorContent extends React.Component<EditorContentProps, EditorContentState> {
@@ -34,7 +34,7 @@ export class PureEditorContent extends React.Component<EditorContentProps, Edito
     this.editorContentRef = React.createRef()
 
     this.state = {
-      renderers: new Map(),
+      renderers: {},
     }
   }
 
@@ -71,11 +71,12 @@ export class PureEditorContent extends React.Component<EditorContentProps, Edito
   setRenderer(id: string, renderer: ReactRenderer) {
     queueMicrotask(() => {
       flushSync(() => {
-        const { renderers } = this.state
-
-        renderers.set(id, renderer)
-
-        this.setState({ renderers })
+        this.setState(({ renderers }) => ({
+          renderers: {
+            ...renderers,
+            [id]: renderer,
+          },
+        }))
       })
     })
   }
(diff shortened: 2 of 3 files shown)


stevemckenzie:
Thanks for this @ruipserra ! This fixes the problem for me but the cursor doesn't actually show until you start typing so you can't actually see that the new node is in focus.

ruipserra:
Hey @stevemckenzie glad to hear your original problem was fixed!

I tried the code from the CodeSandbox you linked in #3338 with the fix from this PR and I couldn't see any issues with the cursor. I uploaded a [branch to my fork with the code](https://github.com/ruipserra/tiptap/compare/fix/3200-react-node-view-cursor-position-after-enter...ruipserra:tiptap:custom-react-paragraph?expand=1), would you mind giving it a try?

```sh
git clone git@github.com:ruipserra/tiptap.git ruipserra/tiptap
cd ruipserra/tiptap
git checkout custom-react-paragraph
npm install
npm start
```

And then open http://localhost:3000/preview/Examples/CustomParagraph in your browser.

This is what I see btw:

https://user-images.githubusercontent.com/3883777/197435336-02d20002-bd5a-4048-b132-ffb15cdb7a11.mov

stevemckenzie:
@ruipserra well that's odd lol. I actually changed my project to temporarily use your fork already and ya for some reason the cursor just doesn't show when a new custom paragraph is created until I start to type so the focus clearly works properly.

But you are correct, your example above works as expected so I will dig more into our own setup.

Thanks for working on this!

mylesj:
Oh I didn't see a notification for this. Nice one! 
I can confirm that this build fixes the original issue for me too.

stevemckenzie:
So how do we get this merged? It's a pretty bad bug for React users.

svenadlung:
@stevemckenzie we are on it. I am pretty sure it will be in next release. Sorry for the inconvenience.

stevemckenzie:
@svenadlung no worries, thanks for the quick update!

stevemckenzie:
I was just doing some more testing around this and found another issue that I'm not sure if it is related.

I have a custom extension with the following configuration:

```
content: 'text*',

group: 'inline',

inline: true,

isTextblock: true,
```

When I press enter, nothing happens. I also can't seem to move the cursor past this element once the cursor is at the end of it. I assume this is related to it being an inline block.

I can confirm that the `splitBlock()` call is happening. Any thoughts?

bdbch (approved):
Thanks for your PR. Looks pretty good to me. I'll add this for the next release.

PR #3341, opened 2022-10-21
Optionally delete full mention chip

Tommy-Sun:
# What 
Made available the option to delete the full mention chip as well as the suggestion char. 

# Changes
* Created an configuration flag for the mention extension called: `deleteOnBackspace` and defaulted it to `false`.

# Why
Using Tiptap for a project our UX requirement was to remove the entire mention with only one backspace. This pull request also retains the original functionality by default.


nVitius:
@bdbch I worked w/ @Tommy-Sun to get this PR ready. Can you take a look and let us know if you have any comments?

eruetsch:
It would be really great if we could get this pull request merged! The functionality seems cool and useful

packages/extension-mention/src/mention.ts
```diff
@@ -9,6 +9,7 @@ export type MentionOptions = {
     options: MentionOptions,
     node: ProseMirrorNode,
   }) => string,
+  deleteOnBackspace: boolean,
```
bdbch, line 12: I guess we would need to add this new option to the documentation then? Otherwise looks good to me.

Tommy-Sun:
Documentation added! @bdbch

Tommy-Sun:
Hi, just following up on this. Just need a review on the documentation I added! Would be great to use sooner rather than later. Thanks

eruetsch:
yeah @bdbch please review the documentation so everything can be merged in and we can use this feature :)

bdbch (approved):
Approved. @svenadlung what do you think?

packages/extension-mention/src/mention.ts
rfgamaral, line 12: I want to add my thoughts to the naming, feels a bit too abstract. The default behaviour is already "delete on backspace", this option doesn't really clarify what it actually does. People reading the documentation might be confused and think this needs to be always `true` otherwise backspace might not work, which is not true.

My suggestion, rename all references to: `deleteTriggerCharWithBackspace` or something along those lines, something more obvious, more explicit...

bdbch: Sounds good! The option is a bit long. Do you think `deleteTriggerWithBackspace` would also be fitting? Is it clear that the trigger is the actual trigger char?

Usually I like to go with shorter, easier to remember options (even though Typescript should boot this argument out of the window).

Maybe even go with the full word if we don't care about option length? `deleteTriggerCharacterWithBackspace` or `includeTriggerCharacterOnDeletion`?

rfgamaral: Whatever you're most comfortable, I just think the original one is a bit misleading.

bdbch: @Tommy-Sun could we go with `deleteTriggerWithBackspace`?

bdbch:
@Tommy-Sun any updates?

Tommy-Sun:
> @Tommy-Sun any updates?

Hi, not my best showing by any means but I finally came back around to this. 

Caught up the branch and inputted a change renaming `deleteOnBackspace` to `deleteTriggerWithBackspace`

Tommy-Sun:
@bdbch This should be good to go.

packages/extension-mention/src/mention.ts
Tommy-Sun: @bdbch added 👍

bdbch (approved):

PR #3379, opened 2022-11-01
[PROPOSAL] dynamic default attributes

mylesj:
> Sorry if you'd really prefer that I raise this as an issue first - it just seemed like a simple enough idea to demonstrate in a PR. If I've missed something in the public API that would otherwise solve the same problem, I'd appreciate some input. 

## Description

Resolve a `default` attribute thunk once when a node is created, in order to compute a default value dynamically.

## Use-case

I'd like to be able assign an attribute to a node that uniquely identifies it. 

### Example usage

```javascript
Node.create({
    // ...
    addAttributes() {
        return {
            id: {
                default: () => generate_uuid(),
                parseHTML: element => element.getAttribute('id')
            }
        }
    }
})
```

### DIFF
--- a/packages/core/src/helpers/getAttributesFromExtensions.ts
+++ b/packages/core/src/helpers/getAttributesFromExtensions.ts
@@ -93,7 +93,11 @@ export function getAttributesFromExtensions(extensions: Extensions): ExtensionAt
           ...attribute,
         }
 
-        if (attribute?.isRequired && attribute?.default === undefined) {
+        if(typeof mergedAttr?.default === 'function') {
+          mergedAttr.default = mergedAttr.default()
+        }
+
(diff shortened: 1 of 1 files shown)


bdbch:
LGTM. Nice idea! @svenadlung do you have any thoughts on this?

mylesj:
Oh nice, thanks for this. 

I did have one additional thought (not required for my use-case however) but imagined it could be useful if the `editor` was passed as an argument to the thunk. I wasn't sure how problematic this would be though - given that the node definition is generally required to generate the schema _before_ the editor is instantiated.

bdbch:
Feel free to open a feature request issue for that - I like the idea but wouldn't keep track of it in this PR :)

PR #3388, opened 2022-11-06
Update mention.md

ZachMoreno:
Adding missing `renderLabel` key


bdbch:
Thanks for your PR! @ZachMoreno 

Is the arrow function important at this case? The method looked fine before.

bdbch:
I'll close this PR for now since I don't think this change is needed.

PR #3410, opened 2022-11-15
Fix #3392 - Unexpected renderText() for contentful nodes

### LINKED ISSUE #3392: Custom renderText on editable node results in duplicate content copied to clipboard

I've specified a custom `renderText` on an editable node expecting to be able to customise the serialisation that is shared with the tiptap clipboard handler, however copy-pasting the node as plain text results in the text selection being appended in addition to the output of `renderText`.

N/A

 - Define a custom node with a "hole" for text content
 - Select a range of text inclusive of the custom node
 - Paste (Shift-Paste) the plain text content to another document

https://stackblitz.com/edit/react-ts-ysehpq?file=App.tsx

I expected the output of `renderText` to be the only content copied in relation to the node.

I considered two ways of getting around this issue:
  - Extending the builtin text node in such away that I could nullify the `renderText` for text nodes contained within specific parents (this produced a number of other exotic outcomes and felt like a hack)
  - ProseMirror recently introduced a [`transformCopied`](https://prosemirror.net/docs/ref/#view.EditorProps.transformCopied) API option that lets you manipulate a slice of content before it is passed to the clipboard-serialiser. It seems that TipTap doesn't support this yet, as the [`clipboardTextSerializer` arguments are ignored](https://github.com/ueberdosis/tiptap/blob/8c6751f0c638effb22110b62b40a1632ea6867c9/packages/core/src/extensions/clipboardTextSerializer.ts#L15) in favour of a custom implementation.
  
This issue might be a feature request if supporting `transformCopied` is the better outcome.

### Video example

https://user-images.githubusercontent.com/2750352/200613826-dad46c17-79af-43b7-b4da-fb19fe528702.mov

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

mylesj:
- resolves: #3392

This issue is causing a number of downstream problems for me as it impacts multiple touch points, namely:
  - `editor.getText()` state
  - clipboard copy-paste
  - reasoning about the selection state of the editor

I had a go at patching it myself and this appears to be the simplest solution - it might raise the question of how deeply nested nodes should be handled but I think it is reasonable to assert that: ***any nodes with a custom text renderer are solely responsible for how their children are rendered***.

### DIFF
--- a/packages/core/src/helpers/getTextBetween.ts
+++ b/packages/core/src/helpers/getTextBetween.ts
@@ -33,7 +33,11 @@ export function getTextBetween(
           range,
         })
       }
-    } else if (node.isText) {
+      // do not descend into child nodes when there exists a serializer
+      return false
+    }
+
+    if (node.isText) {
       text += node?.text?.slice(Math.max(from, pos) - pos, to - pos) // eslint-disable-line
       separated = false
     } else if (node.isBlock && !separated) {
--- a/tests/cypress/integration/core/generateText.spec.ts
+++ b/tests/cypress/integration/core/generateText.spec.ts
@@ -1,0 +1,66 @@
+/// <reference types="cypress" />
+
+import { generateText, Node, NodeConfig } from '@tiptap/core'
+import Document from '@tiptap/extension-document'
+import Paragraph from '@tiptap/extension-paragraph'
+import Text from '@tiptap/extension-text'
+
+describe(generateText.name, () => {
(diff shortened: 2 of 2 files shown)


mylesj:
Hey folks, can we have some input on this one please? I've just updated my branch and this still appears to be relevant. Can it be rolled into your roadmap?

bdbch (approved):

PR #3425, opened 2022-11-17
fix(link): allow to unset target attribute

### LINKED ISSUE #3276: Link Extension, always adding target="_blank" when instatiating editor.

As stated in title, not sure why its happening, but with link extension enabled, when theres a link without target attribute its added there anyway with _blank value in it.

Chrome

open editor with:
```
<p>
<a href="/jobs/new">jobs</a>
</p>
```
Interesting enough this is not passed on with value - just rendered inside editor content element.

_No response_

It would be great to have some configuration allowing to discard this behaviour

_No response_

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

dargmuesli:
Resolves #3276.

### DIFF
--- a/packages/extension-link/src/link.ts
+++ b/packages/extension-link/src/link.ts
@@ -41,11 +41,11 @@ declare module '@tiptap/core' {
       /**
        * Set a link mark
        */
-      setLink: (attributes: { href: string, target?: string }) => ReturnType,
+      setLink: (attributes: { href: string, target?: string | null }) => ReturnType,
       /**
        * Toggle a link mark
        */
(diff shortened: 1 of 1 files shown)


bdbch (approved):

dargmuesli:
@bdbch thank you for merging this! Seems like there was [a problem](https://github.com/ueberdosis/tiptap/actions/runs/3549230394/jobs/5961362948) with the merge's build though and the fix was not released.

bdbch:
Hey @dargmuesli - we have a few problems with our tests being capped by Github Actions. We're currently looking into potential solutions.

Your merge will be released with the next release. We do releases manually so not every merged commit is automatically released right away.

PR #3436, opened 2022-11-21
Fix #3435 - CommonJS and ESM loading confusion

### LINKED ISSUE #3435: Node (16/next) imports CommonJS instead of ESM due to missing `type` and `exports` in package.json

There have been several user discussions and bug reports related to `prosemirror-model` failing with some form of `looks like multiple versions of prosemirror-model were loaded`, such as:
* https://github.com/ueberdosis/tiptap/discussions/3282
* https://github.com/ueberdosis/tiptap/issues/577#issuecomment-1219122701
* https://github.com/ueberdosis/tiptap/issues/865
* https://github.com/ueberdosis/tiptap/issues/3209
* https://github.com/ueberdosis/tiptap/issues/2836#issuecomment-1164435952
* etc.

Some of these are due to multiple versions getting loaded, e.g. CommonJS and ES Modules, at the same time and then causing runtime errors.

We recently ran into this with our codebase (all TypeScript/ESM) and build system (Rush/Heft) and tracked it down to the `package.json` files not having `"type": "module"` and the newer `exports` field set.

The TypeScript documentation regarding [ECMAScript Modules in Node.js](https://www.typescriptlang.org/docs/handbook/esm-node.html) has a good explanation and resolution.

For example, `tiptap-core` works great by changing the file extension for CommonJS to `.cjs` (and ESM to `.mjs` for consistency and explicitness) and modifying `package.json`, so that it provides the new syntax and fallbacks for older versions of Node and other build systems:
```json
  "type": "module",
  "exports": {
    ".": {
      "types": "./dist/packages/core/src/index.d.ts",
      "import": "./dist/tiptap-core.mjs",
      "require": "./dist/tiptap-core.cjs"
    }
  },
  "main": "dist/tiptap-core.cjs",
  "umd": "dist/tiptap-core.umd.js",
  "module": "dist/tiptap-core.mjs",
  "types": "dist/packages/core/src/index.d.ts",
```

but without we get a `RangeError: Can not convert (something) to a Fragment (looks like multiple versions of prosemirror-model were loaded)` when different code paths load a mix of CJS and ESM (our library, y-prosemirror, tiptap, etc.).

Note: If I'm reading Dangers of exporting both CJS and ESM in [Dual package hazard](https://nodejs.org/docs/latest-v16.x/api/packages.html#dual-commonjses-module-packages) correctly, it is still possible to run into the same issue if there is shared state, but these changes should at least clean up the code paths in many cases.

In Node, not the browser.

* Create a stub project with `npm init`
* (Optional) Add `"type": "module",` to `package.json`.
* Create `index.mjs` and add the following code to it:
```javascript
import { getSchema } from "@tiptap/core";

getSchema(null);
```
* Run `node index.mjs`
* The exception clearly displays that the CJS version is loaded:

```
TypeError: Cannot read properties of null (reading 'map')
    at Function.flatten (/some-path/tiptap-test/node_modules/@tiptap/core/dist/tiptap-core.cjs.js:953:14)
```

_No response_

The expectation is that an `import` from an ESM project uses a single ESM instance of TipTap, so that these errors don't happen. Also that CommonJS loads CJS.

I have a PR ready to go that patches all the `package.json` files for the repo, and I will submit it shortly.

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

tomi-bigpi:
* Fix TipTap getting loaded as CommonJS when the intent is to use the ES Module version.
* Use `.cjs` and `.mjs` file extensions to help resolution in Node
* `package.json` change also makes explicit exports required
* Update `core` utilities exports to include all utilities
* Update tests to use exported utilities


bdbch:
Thanks @tomi-bigpi 

since the PR is quite large I'll need some time to review it locally. I'll get back to you when I'm done!

bdbch:
I didn't find any issues while testing this locally. The code also looks fine to me and the change makes sense. I'll merge it for now to see if we'll have issues with the new script names with the next release.

Thanks for your support @tomi-bigpi

tomi-bigpi:
> I didn't find any issues while testing this locally. The code also looks fine to me and the change makes sense. I'll merge it for now to see if we'll have issues with the new script names with the next release.

My take is that the likeliest thing to break is if anyone is directly importing any files as the `exports` option also requires explicit exports of classes/functions/etc. (In addition to the issue mentioned above)

Thanks for the quick turnaround!

PR #3460, opened 2022-11-27
fix: make TextAlign defaultAlignment setting work as expected

### LINKED ISSUE #3459: Strange behavior of the defaultAlignment setting in the TextAlign extension

First off thank you to the development team and community for this great tool. So far it has been just what I needed and I am hopeful that it will be a long-term solution for me!

With the TextAlign extension enabled, I have a Tiptap editor element for which `element.style.textAlign == ""`. However, it inherits a `textAlign` value, which can be accessed through `window.getComputedStyle(element).textAlign`. On loading, the Tiptap editor uses the inherited `textAlign` value rather than the default value assigned through the extension using the `defaultAlignment` setting.

Then, the editor allows you to change the text alignment as expected, unless you try to change it to the value set in `defaultAlignment`. In that case, it changes the alignment to the inherited/computed value.

- Using Firefox 107.0 (latest)
- In addition to Tiptap, I am using the following tools for development:
    - Svelte 3.49.0
    - Vite 3.1.0
    - Typescript 4.6.4
- No special extensions, just StarterKit and TextAlign

The issue seems to also occur when I try using the TextAlign extension in a simple webapp made with React, so as far as I can tell it will occur as long as you install the TextAlign extension. Then, you set `defaultAlignment` using `TextAlign.configure()` so that it does not match whatever the editor element's inherited/computed `textAlign` value is.

https://codesandbox.io/p/sandbox/wild-morning-cbyyyh?file=%2Fsrc%2FApp.js

_(Not built with Svelte, but the issue persists)_

I expect that when the Tiptap editor first loads, each node's alignment would match the value configured for `defaultAlignment` (regardless of any properties inherited by the editor element). Then, each node's alignment could be changed using buttons, keyboard shortcuts, etc.

Please don't hesitate to let me know if anything I've said is unclear and I'll be happy to update the post to clarify.

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

JosephASullivan:
Proposed fix to resolve #3459

### DIFF
--- a/packages/extension-text-align/src/text-align.ts
+++ b/packages/extension-text-align/src/text-align.ts
@@ -41,10 +41,6 @@ export const TextAlign = Extension.create<TextAlignOptions>({
             default: this.options.defaultAlignment,
             parseHTML: element => element.style.textAlign || this.options.defaultAlignment,
             renderHTML: attributes => {
-              if (attributes.textAlign === this.options.defaultAlignment) {
-                return {}
-              }
-
               return { style: `text-align: ${attributes.textAlign}` }
             },
           },


JosephASullivan:
Apologies -- I am new to pull requests and a bit confused. If someone could explain the automatic messages here and let me know what further action is needed from me, I'd greatly appreciate it.

Mohrip (changes requested):
needs to modify

packages/extension-text-align/src/text-align.ts
```diff
@@ -41,10 +41,6 @@ export const TextAlign = Extension.create<TextAlignOptions>({
             default: this.options.defaultAlignment,
             parseHTML: element => element.style.textAlign || this.options.defaultAlignment,
             renderHTML: attributes => {
-              if (attributes.textAlign === this.options.defaultAlignment) {
```
bdbch, line 44: I think the idea here was to not add any style attributes to the element when the alignment is equal to the default alignment.

If we remove this, there's no way to get rid of alignment inline styles without removing a node first even though they are mostly not needed.

I think the only issue that could come up with this is, that the text alignment can diverge from the default because of a wrapping element that updates the text alignment and passes it down via inheritance but this is expected browser behaviour I'd say?

Open for discussion on this one.

bdbch:
Moved this to #5891

Issue #3209, opened 2022-09-18
Duplicate use of selection JSON ID gapcursor


justinbalaguer:
### What’s the bug you are facing?

installing `@tiptap-pro/extension-details` extension shows this error
`Uncaught RangeError: Duplicate use of selection JSON ID gapcursor`

### Which browser was this experienced in? Are any special extensions installed?

Microsoft Edge, installed `@tiptap-pro/extension-details` 

### How can we reproduce the bug on our side?

install `@tiptap-pro/extension-details` 

### What did you expect to happen?

no error

### Did you update your dependencies?

- [X] Yes, I’ve updated my dependencies to use the latest version of all packages.

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

zanemcca:
This issue is actually quite broad and could be very problematic.

Prosemirror packages will often have some local runtime cache or they will use `instanceOf` checks or they will generate some id field that is used later on that will break when there are multiple copies of the package installed.

Long story short there should only ever be a single copy of virtually any `prosemirror-` lib in any given project. The way prosemirror packages are listed as dependencies instead of peerDependencies in tiptap and hocuspocus will result in situations where npm cannot deduplicate packages so it will add two copies. Whenever this happens things will break. In my experience I have found that things usually break in far more insidious and less insightful ways than this particular error.

To resolve this all prosemirror packages at all levels in hocuspocus and tiptap should be peerDependencies not dependencies.

**Workaround**

To work around this issue I have been adding offending packages to the `overrides` section of my package.json.

This is not a good solution though as it can still result in issues. For instance my code uses `@hocuspocus/transformers` which uses `@tiptap/starter-kit` which uses `@tiptap/extension-gapcursor` which then uses `prosemirror-gapcursor`. I have fixed my `@hocuspocus/transformers` packages to a specific version and I have added `prosemirror-gapcursor` to a specific version as well with overrides in place. However since `@hocuspocus/transfomers` allows for any version of `@tiptap/starter-kit` above a certain level I ended up with a newer than desired `@tiptap/starter-kit` which had a hard requirement for a newer version of `prosemirror-gapcursor`. That was all too much for the fairly new `overrides` option to sort through so it installed multiple copies of my override instead of a single deduplicated copy.

[...]

PR #3487, opened 2022-12-04
Draft: Moves all prosemirror deps to peerDependencies & devDependencies

janthurau:
refs  #3209, https://github.com/ueberdosis/hocuspocus/issues/445

### DIFF
--- a/packages/extension-bubble-menu/package.json
+++ b/packages/extension-bubble-menu/package.json
@@ -29,12 +29,16 @@
     "dist"
   ],
   "peerDependencies": {
-    "@tiptap/core": "^2.0.0-beta.193"
+    "@tiptap/core": "^2.0.0-beta.193",
+    "prosemirror-state": "^1.4.1",
+    "prosemirror-view": "^1.28.2"
   },
   "dependencies": {
+    "tippy.js": "^6.3.7",
+    "lodash": "^4.17.21",
+    "@tiptap/core": "^2.0.0-beta.193",
     "prosemirror-state": "^1.4.1",
-    "prosemirror-view": "^1.28.2",
-    "tippy.js": "^6.3.7"
+    "prosemirror-view": "^1.28.2"
   },
   "repository": {
     "type": "git",
@@ -43,7 +47,6 @@
   },
   "sideEffects": false,
   "devDependencies": {
-    "@types/lodash": "^4.14.187",
-    "lodash": "^4.17.21"
+    "@types/lodash": "^4.14.187"
   }
 }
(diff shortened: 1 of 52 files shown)


janthurau:
@zanemcca Don't have too much experience with peerDependencies, but I believe this should do the job?

bdbch (approved):
LGTM, thanks!

packages/extension-bubble-menu/package.json
```diff
@@ -29,12 +29,16 @@
...
+    "prosemirror-view": "^1.28.2"
   },
   "dependencies": {
+    "tippy.js": "^6.3.7",
+    "lodash": "^4.17.21",
+    "@tiptap/core": "^2.0.0-beta.193",
     "prosemirror-state": "^1.4.1",
-    "prosemirror-view": "^1.28.2",
-    "tippy.js": "^6.3.7"
+    "prosemirror-view": "^1.28.2"
```
zanemcca, line 41: I anticipate that these three remaining tiptap & prosemirror `dependencies` will still cause trouble. I think if you moved them to `devDependencies` then things should work more smoothly.

zanemcca (commented):
Thanks so much for doing this @janthurau! Aside from that one case I noticed I think what you have hear is a fantastic baseline.

I think the next step will be an important one as well.

When you guys upgrade your prosemirror packages in `devDependencies` it would be ideal if the version ranges in the `peerDependencies` were extended to include the new package versions instead of moved forward to match the new `devDependencies`. If you were to move the `peerDepenencies` forward in lock-step with the `devDependencies` then it will force us to use that specific package version even though it may not really be necessary and it could cause trouble with our old editor which was built for an older version.

Obviously if you guys run into breaking changes at specific version of whichever package you can and should remove outdated package versions from the `peerDependencies` but that should be a fairly rare event. I will leave it to you guys to decide how best to mange when to drop support for older packages but the wider the range you can support the easier it is for downstream developers.

thet:
I'm not sure what that change should actually solve.
I'm directly depending on individual tiptap packages. Now I had to add all those prosemirror dependencies to my package dependencies.
Not a huge deal though, but also quite some work to get the build running again.

jmamakeesic:
@bdbch Please see [my comment](https://github.com/ueberdosis/tiptap/commit/30e799c67f2a9d5ef275f965998bcbbeac330226#r97974157) that references a related installation error.

bdbch:
Hey @jmamakeesic - if not already done could you open an issue for that so we can track it better?

jmamakeesic:
https://github.com/ueberdosis/tiptap/issues/3664
https://github.com/ueberdosis/tiptap/pull/3665

PR #3512, opened 2022-12-07
Image - Computable HTMLAttributes, file upload function for dragNdrop and paste

nkonev:
This PR adds:
* Computable HTMLAttributes
* Disabled by default support of upload function for **dragNdrop file** and **pasted image** from grphic editor

### DIFF
--- a/packages/extension-image/src/image.ts
+++ b/packages/extension-image/src/image.ts
@@ -1,13 +1,22 @@
 import {
+  isFunction,
   mergeAttributes,
   Node,
   nodeInputRule,
 } from '@tiptap/core'
+import {
+  Node as ProseMirrorNode,
+} from 'prosemirror-model'
+import { Plugin, PluginKey } from 'prosemirror-state'
 
 export interface ImageOptions {
   inline: boolean,
   allowBase64: boolean,
-  HTMLAttributes: Record<string, any>,
+  HTMLAttributes: Record<string, any> | ((input: ProseMirrorNode) => Record<string, any>),
+  allowPasteFromClipboard: boolean,
+  allowDragNDrop: boolean,
+  // accept File and returns src attribute
+  imageUpload?: ((image: any) => Promise<string>) | null
 }
 
 declare module '@tiptap/core' {
@@ -31,6 +40,9 @@ export const Image = Node.create<ImageOptions>({
       inline: false,
       allowBase64: false,
       HTMLAttributes: {},
+      allowPasteFromClipboard: false,
+      allowDragNDrop: false,
+      imageUpload: null,
     }
   },
 
@@ -68,8 +80,12 @@ export const Image = Node.create<ImageOptions>({
     ]
   },
 
-  renderHTML({ HTMLAttributes }) {
-    return ['img', mergeAttributes(this.options.HTMLAttributes, HTMLAttributes)]
+  renderHTML({ node, HTMLAttributes }) {
+    const attrs = isFunction(this.options.HTMLAttributes)
+      ? this.options.HTMLAttributes(node)
+      : this.options.HTMLAttributes
+
+    return ['img', mergeAttributes(attrs, HTMLAttributes)]
   },
 
   addCommands() {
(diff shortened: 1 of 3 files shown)


nkonev:
@bdbch PTAL

packages/extension-image/src/image.ts
```diff
@@ -96,4 +112,82 @@ export const Image = Node.create<ImageOptions>({
...
+
+            items.forEach(item => {
+              if (item.type.indexOf('image') === 0) {
+                const { schema } = view.state
+
+                const image = item.getAsFile()
+
+                this.options.imageUpload?.(image).then(src => {
+                  const node = schema.nodes.image.create({
+                    src,
```
kylealwyn, line 137: adding an option for custom attributes like a file id for downstream extraction/relationships would be money

kylealwyn (commented):
random drive by but :chef_kiss:

rfgamaral:
I'm not a maintainer, so this is just my opinion, but personally I don't think this should be part of the `Image` extension.

IMO, the `Image` extension should be simple and provide core functionality to display images only, and providing a partial upload mechanism (partial because it doesn't upload anything, cause that depends on your server) doesn't feel like "core functionality" for images.

But that's just me, I don't make decisions, just want to be clear about that.

nkonev:
@rfgamaral thank you for your opinion.
I can't agree with you, 
I'm convinced that handling drag-n-drop and image pasting 
A) is not very special thing in 2022
B) is common in most of messengers, for example in Slack

So, having this, I think it's ok to enhance the existing extension for it.

kylealwyn:
Echo @nkonev. For a core TipTap use case - collaboration - an image extension without upload functionality is half baked. I was bummed to see there was no support for it initially.

rfgamaral:
I personally don't think those are very strong arguments to include this partial solution in the `Image` extension, especially because, and please don't take this the wrong way, I feel that the implementation in this PR is not very flexible/generic/extensible. We also have upload functionality on our side, and the solution here wouldn't work for us.

For instance, we need to show upload progress while the image is being uploaded, and this implementation does not take that into account. The technical details of implementing upload progress may differ from consumer to consumer, and one solution may not fit everyone's server implementation, forcing some consumers to completely override everything, increasing the final bundle size with code that is never going to be used.

Also, the way this is implemented, it looks to me that the image would only be inserted in the editor when the upload is finished, and for large images (or slow connections), this is not a very good user experience. You could say, "no, you can invoke `imageUpload` on your side whenever you want, not only when the upload is finished", and I would reply "sure, but that's not how our server upload mechanism works, we only get the final image URL once the upload is finished, so we would only be able to call it once it's done".

I wouldn't mind having this functionality built-in in the core `Image` extension if it was **flexible and extensible enough that would cater everyone's needs and server implementations**, but I don't think this one is that. It feels built for your own use case, not everyone else's. That's not "core functionality" material in my book, and that's also why we haven't opened a PR ourselves with our own solution, because it's very specific to us, to our server, to our use case, which may or may not work for everyone else.

But again, I don't make decisions, I'm only voicing my concerns as a contributor and as a consumer of Tiptap.

kylealwyn:
> flexible and extensible enough that would cater everyone's needs and server implementations
Agree with that, which led to my comment above regarding adding node attributes. It would be ideal to handle generic dom events of drag/dropping & pasting, and insertion, and call out to the consumer for any other details.

nkonev:
> It would be ideal to handle generic dom events of drag/dropping & pasting

Sounds interesting, will see

nkonev:
I've found (my bad=)) ability to extend, closing because of this.

```
    const embedUploadFunction = (chatId, fileObj) => {
        const formData = new FormData();
        formData.append('embed_file_header', fileObj);
        return axios.post('/api/storage/'+chatId+'/embed', formData)
            .then((result) => {
                let url = result.data.relativeUrl; // Get url from response
                console.debug("got embed url", url);
                return url;
            })
    }


    const uploadFunction = (image) => embedUploadFunction(this.chatId, image)

    const imagePluginInstance = Image.extend({
        addProseMirrorPlugins() {
            return [
                new Plugin({
                    key: new PluginKey('imageHandler'),
                    props: {
                        handlePaste: (view, event) => {
                            const items = (event.clipboardData || event.originalEvent.clipboardData).items;
                            for (const item of items) {
                                if (item.type.indexOf("image") === 0) {
                                    event.preventDefault();
                                    const {schema} = view.state;

                                    const image = item.getAsFile();

                                    uploadFunction(image).then(src => {
                                        const node = schema.nodes.image.create({
                                            src: src,
                                        });
                                        const transaction = view.state.tr.replaceSelectionWith(node);
                                        view.dispatch(transaction)
                                    });

                                }
                            }
                        },

                        handleDOMEvents: {
                            drop: (view, event) => {
                                const hasFiles =
                                    event.dataTransfer &&
                                    event.dataTransfer.files &&
                                    event.dataTransfer.files.length;

                                if (!hasFiles) {
                                    return false;
                                }

                                const images = Array.from(
                                    event.dataTransfer?.files ?? []
                                ).filter((file) => /image/i.test(file.type));

                                if (images.length === 0) {
                                    return false;
                                }

                                event.preventDefault();

                                const { schema } = view.state;
                                const coordinates = view.posAtCoords({
                                    left: event.clientX,
                                    top: event.clientY,
                                });
                                if (!coordinates) return false;

                                images.forEach(async (image) => {

                                    const node = schema.nodes.image.create({
                                        src: await uploadFunction(image),
                                    });
                                    const transaction = view.state.tr.insert(coordinates.pos, node);
                                    view.dispatch(transaction);

                                });

                                return true;
                            },
                        },
                    }
                }),
            ];
        }
    }).configure({
      inline: true,
      HTMLAttributes: {
          class: 'image-custom-class',
      },
    });

```

Issue #3538, opened 2022-12-13
`textSerializers` in `generateText` don't override `renderText` set in schema


tovaschreier:
### What’s the bug you are facing?

Providing `textSerializers` to `generateText` does not override the `renderText` function defined in the schema

### Which browser was this experienced in? Are any special extensions installed?

Chrome, no special extensions

### How can we reproduce the bug on our side?

Provide a custom `renderText` in your schema:
```js
  const extensions = [
    StarterKit,
    Image.extend({
      // default renderText defined in the schema
      renderText() {
        return "[Image text from renderText in schema]";
      }
    })
  ];

  const editor = useEditor({
    extensions,
    content: `
      <p>
        this is a basic example of tiptap with an image below. 
      </p>
      <img src="https://faculty.washington.edu/ejslager/random-generator/duck.gif" alt="duck swimming">
    `
  });
```

Call `generateText` and pass in a custom `textSerializer` for the node:
```js
  const textContent = generateText(editor.getJSON(), extensions, {
        // this should override the default renderText method
        textSerializers: {
          image: () => "[Image text from textSerializer]"
        }
      });
```

Expected behavior: the text returned in the text serializer should be output when calling `generateText`

### Can you provide a CodeSandbox?

Yes! https://codesandbox.io/s/infallible-keldysh-tf1ldm?file=/src/App.js

### What did you expect to happen?

The `textSerializers` provided to `generateText` would override the `renderText` function defined in the schema.

In fact, this seems to be the expected behavior laid out in [the PR that introduced](https://github.com/ueberdosis/tiptap/pull/1875#issue-992671651) the `generateText` method - the PR description says:

### usage of `generateText()`

```js
// with some options
const text = generateText(json, [
  Document,
  Paragraph,
  Text,
  HardBreak,
  // other extensions …
], {
  // define a custom block separator if you want to
  blockSeparator: '\n',
  // overwrite `renderText` method of nodes
  textSerializers: {
    hardBreak: () => '\n',
  },
})
```

Note the "overwrite `renderText` method of nodes" comment right before `textSerializers`

### Anything to add? (optional)

I believe this is just a matter of switching the order of [these 2 lines](https://github.com/ueberdosis/tiptap/blob/main/packages/core/src/helpers/generateText.ts#L26-L27): 
```
  return getText(contentNode, {
    blockSeparator,
    textSerializers: {
      ...textSerializers,
      ...getTextSerializersFromSchema(schema),
    },
  })
```

[...]

PR #3546, opened 2022-12-14
Fix: custom text serializers should override text serializers defined in the schema

tovaschreier:
Fixes https://github.com/ueberdosis/tiptap/issues/3538


bdbch:
Looks good to me! Thank you!

bdbch (approved):

PR #3550, opened 2022-12-15
feat: support for custom pasted or dragged-in images

fEyebrow:
I updated the doc for images. you can see the changes I've made.

### DIFF
--- a/demos/src/Examples/Images/Vue/index.vue
+++ b/demos/src/Examples/Images/Vue/index.vue
@@ -40,11 +40,24 @@ export default {
         Document,
         Paragraph,
         Text,
-        Image,
+        Image.configure({
+          upload: async fileOrUrl => {
+            if (typeof fileOrUrl === 'string') {
+              return {
+                src: fileOrUrl,
+              }
+            }
+
+            return {
+              src: URL.createObjectURL(fileOrUrl),
+            }
+          },
+        }),
         Dropcursor,
       ],
       content: `
         <p>This is a basic example of implementing images. Drag to re-order.</p>
+
         <img src="https://source.unsplash.com/8xznAGy4HcY/800x400" />
         <img src="https://source.unsplash.com/K9QHL52rE2k/800x400" />
       `,
--- a/docs/api/nodes/image.md
+++ b/docs/api/nodes/image.md
@@ -25,7 +25,9 @@ Renders the image node inline, for example in a paragraph tag: `<p><img src="spa
 
 It totally depends on what kind of editing experience you’d like to have, but can be useful if you (for example) migrate from Quill to Tiptap.
 
-Default: `false`
+Default: `true`
+
+If you pasted images and text, and the image is included in the p, image must be inline. p can't contain both block and inline. 
 
 ```js
 Image.configure({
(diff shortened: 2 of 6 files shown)


rfgamaral:
I have the exact same reservations for this PR as I mentioned [here](https://github.com/ueberdosis/tiptap/pull/3512#issuecomment-1349646048) and [here](https://github.com/ueberdosis/tiptap/pull/3512#issuecomment-1350890121).

The problems I see with this solution:

- You are changing the default options for the `Image` extension to cater for the upload mechanism needs, and while I'm sure that's something a lot of people want/need, not everyone does, and this could potentially "break" people using Tiptap assuming certain defaults
- It assumes that everyone wants a placeholder image while the image is being uploaded, and this will not be true for everyone
    - On top of that, a default placeholder image is hardcoded, I don't think this is the right way to do it
- I'm not a fan of including `prosemirror-replaceattrs` the way this PR is doing it, why not pull the dependency and add it as a peer dep instead?
    - To be honest, I'm not even sure this is strictly required and just make things more complex than they need to be (maintaining this will be harder and more time-consuming)

---

IMO, and as I've also mentioned in the other PR, I think this should be an additional extension extending the original `Image`, and not built-in in the base `Image` one. I'm a firm believer that this should not be part of the core extension due to many little details that will not fit everyone's needs and server requirements.

fEyebrow:
@rfgamaral 
I change the code. If you don't set upload, the placeholder of the image will be replaced with the original image.

Because the prosemirror-replaceattrs package has not been updated for a long time, so I included this. Maybe you can extract this, and put it in the right place.

rfgamaral:
@fEyebrow Apologies if I sent the wrong message, but I'm not a Tiptap maintainer, and I wasn't "requesting changes" per se, I was only voicing my concerns as a Tiptap contributor and consumer.

> Because the prosemirror-replaceattrs package has not been updated for a long time, so I included this. Maybe you can extract this, and put it in the right place.

To be honest with you, I didn't dig deep into that part of the code, so I'm not really sure what are you doing with that, and if it's required/useful. It's just that we also have upload functionality on our side, and we didn't felt the need to use anything like that, so I'm wondering if using `prosemirror-replaceattrs` is really useful or not.

But again, don't make any changes on my account, as I don't make decisions around here 😅

fEyebrow:
@rfgamaral 
image

rfgamaral:
@fEyebrow Thanks for clarifying that for me. We don't use collaboration (I'm assuming you're talking about hocuspocus, correct?), so I wasn't aware of that.

I'm wondering if instead of including `prosemirror-replaceattrs` the way you did (I'm particularly not a fan of this: `Transaction.prototype.replaceAttrs = function replaceAttrs`), perhaps we could have some additional helper methods and/or built-in Tiptap commands to achieve the same thing (based on the code you already have, of course)? I'm thinking that that would be more inline with Tiptap architecture.

But just to be clear, don't go make these changes on my account, let the Tiptap maintainers comment, and follow their directions instead.

nperez0111:
I agree with all of @rfgamaral's points, will not be accepting this contribution. Though, I very much appreciate you contributing it & hope it is useful to someone else.

As a pro extension we offer the file upload extension which can handle more than just images and is a much better abstraction

PR #3555, opened 2022-12-16
Change Build Process to Lerna + tsup & prepare for prosemirror-meta package

bdbch:
This PR changes our build process away from a custom rollup solution with package iterations to a simple tsup build. This includes:

- tsup builds with one global tsconfig
- individual package build tasks that extend the base tsconfig
  - this allows packages to have separate build processes without any super crazy configuration hacks
  - this is required for our new "prosemirror" meta package we're currently preparing

### IMPORTANT

This PR is still WIP and needs to be heavily tested, specially the NPM publish and installation process with `npm`, `yarn` and `pnpm`.

### TODO

- [x] Test package resolution with new bundles
- [x] Check Netlify demo build


demos/package.json
```diff
@@ -4,8 +4,7 @@
   "private": true,
   "scripts": {
     "start": "vite --host",
-    "build": "npm run ts && vite build",
-    "preview": "vite preview",
```
svenadlung, line 8: Shouldn't we keep the preview script?

demos/package.json
```diff
@@ -4,8 +4,7 @@
...
+    "build:demo": "npm run ts && vite build",
```
svenadlung, line 7: Should we call it `build:demos` for the sake of consistency?

package.json
```diff
@@ -65,7 +64,8 @@
     "rollup-plugin-sourcemaps": "^0.6.3",
```
svenadlung, line 64: Do we still need all the rollup deps?

packages/extension-bubble-menu/package.json
```diff
@@ -15,15 +15,14 @@
   "type": "module",
   "exports": {
     ".": {
-      "types": "./dist/packages/extension-bubble-menu/src/index.d.ts",
-      "import": "./dist/tiptap-extension-bubble-menu.esm.js",
-      "require": "./dist/tiptap-extension-bubble-menu.cjs"
+      "types": "./index.d.ts",
```
svenadlung, line 18: Is it missing `dist/`?

packages/core/package.json
```diff
@@ -18,15 +18,14 @@
   "type": "module",
   "exports": {
     ".": {
-      "types": "./dist/packages/core/src/index.d.ts",
-      "import": "./dist/tiptap-core.esm.js",
-      "require": "./dist/tiptap-core.cjs"
+      "types": "./dist/index.cjs",
```
svenadlung, line 21: I think the file extensions were mixed up here.

packages/core/package.json
```diff
@@ -18,15 +18,14 @@
...
+      "import": "./dist/index.js",
+      "require": "./dist/index.d.ts"
```
svenadlung, line 23: I think the file extensions were mixed up here.

packages/vue-3/tsconfig.json
```diff
@@ -1,9 +1,10 @@
...
-    "../../demos",
-    "../vue-2"
-  ]
+  "extends": "../../tsconfig.tiptap.json",
+  "compilerOptions": {
+    "paths": {
+      "@tiptap/*": ["../*/src"]
+    }
+  },
+  "include": ["**/*.d.ts", "**/*.ts", "**/*.tsx", "**/*.vue"],
```
svenadlung, line 8: Do we need those two lines (include, exclude) in every package specific tsconfig.json? Doesn't it get extended by base tsconfig?

svenadlung (changes requested):

demos/package.json
bdbch: Done!

package.json
bdbch: Negative, cleaning this up. Good find.

packages/vue-3/tsconfig.json
bdbch: Unfortunately yeah - those paths are resolved when the tsconfig is loaded. I tried to put this out and move it into the extended tsconfig, but then the paths were picked from the root directory instead of the relative one.

svenadlung (dismissed):

svenadlung (approved):

PR #3558, opened 2022-12-17
Export `createNodeFromContent` and other missing helpers

jacksleight:
I'd like to use the `createNodeFromContent` helper in a project, but it's not currently exported. While looking at this I found there are few others that also aren't exported, so this PR adds in all the missing ones.

If there's a reason any of these aren't exported I can remove them, but `createNodeFromContent` would be really useful!


Doerge:
I need to use `createChainableState` for a custom Keymap extension, so exporting everything that is not some sort of private API would be great for me.

bdbch (approved):

Issue #3604, opened 2023-01-08
Wrong insertion position when pasting code from VSCode


aliasliao:
### What’s the bug you are facing?

When pasting from VSCode, the code is inserted **BEFORE** the newly created codeBlock.

### Which browser was this experienced in? Are any special extensions installed?

- Google Chrome Version 108.0.5359.124.
- No.

### How can we reproduce the bug on our side?

1. go to https://tiptap.dev/api/nodes/code-block#usage, focus in the live demo in this page, clear the editor
2. type in some plain text
3. open a code file with VSCode, select some code and **right click**, click **copy** menu item
4. navigate back to tiptap live demo, paste the copied code

https://user-images.githubusercontent.com/22636233/211207075-d3d306e5-ac61-48e1-ac4c-690fd932c451.mp4

### Can you provide a CodeSandbox?

https://tiptap.dev/api/nodes/code-block#usage

### What did you expect to happen?

The code should be inserted **inside** the newly created codeBlock.

### Did you update your dependencies?

- [X] Yes, I’ve updated my dependencies to use the latest version of all packages.

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

Issue #3321, opened 2022-10-18
Code blocks are not highlighted properly


gethari:
### What’s the bug you are facing?

When we have some text in the editor, and now if we copy paste a codeblock , eg: a `JSON` like this

```json
{
  "type": "initial",
  "maximumWarning": "2mb",
  "maximumError": "5mb"
}
```
It is not highlighted properly.

### Which browser was this experienced in? Are any special extensions installed?

Google Chrome
106.0.5249.119 (Official Build) (64-bit), no extensions installed

### How can we reproduce the bug on our side?

- Navigate to [CodeBlockLowLight](https://tiptap.dev/api/nodes/code-block-lowlight) page
- Clear all the existing code
- Now paste the `JSON`, it will work now.
- Now clear everything again.
- Type some text , & try pasting the same json again & it looks like this now

![image](https://user-images.githubusercontent.com/20477956/196336584-495937b5-6e94-493e-8130-3c010ca10500.png)

### What did you expect to happen?

Expect the code block to be highlighted properly.

### Anything to add? (optional)

I tried to debug it, & I am able to see that in [this line](https://github.com/ueberdosis/tiptap/blob/2b69f344c713befecd4ec606df5e9ba680aa2ce8/packages/extension-code-block-lowlight/src/lowlight-plugin.ts#L72) the `decorations` are returned as `[]` & hence the highlight is not working properly.

_No response_

### Did you update your dependencies?

- [X] Yes, I’ve updated my dependencies to use the latest version of all packages.

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

aliasliao:
Looks like it's the same reason as this one https://github.com/ueberdosis/tiptap/issues/3604.

PR #3606, opened 2023-01-08
fix(extension-code-block): #3604 paste code from vscode

aliasliao:
**fixes https://github.com/ueberdosis/tiptap/issues/3604 , https://github.com/ueberdosis/tiptap/issues/3321**

This line (L267) causes the wrong insertion position:
https://github.com/ueberdosis/tiptap/blob/main/packages/extension-code-block/src/code-block.ts#L267
```ts
tr.setSelection(TextSelection.near(tr.doc.resolve(Math.max(0, tr.selection.from - 2))))
```

Removing L267 can fix the bug.
But I think the better solution is to replace selection with a text node, rather than imperatively inserting the copied content  : )

cc @bdbch


aliasliao:
@bdbch Can you help to review this?

bdbch (approved):
LGTM

@svenadlung cc

aliasliao:
@svenadlung need your review, please

svenadlung:
@aliasliao Thank you. I'll have a look next week due to public holidays.

svenadlung:
@aliasliao Sorry for the delay. There is a small difference in some cases. In that case the cursor after pasting is not in the code node, but in the following node. Can you please check again?

aliasliao:
> @aliasliao Sorry for the delay. There is a small difference in some cases. In that case the cursor after pasting is not in the code node, but in the following node. Can you please check again?

@svenadlung Hi, I fixed this issue in the latest commit. Could you please review it?

There is one more case that I found hard to handle: if there is a CodeBlock after the cursor before pasting, the cursor remains inside that CodeBlock after pasting, instead of moving to the newly created code block.

nperez0111 (dismissed):
LGTM

nperez0111 (approved):

PR #3675, opened 2023-01-30
add optionalSlashSlash to protocol options

taras-turchenko-moc:
## Why?
I need to link `tel:12312312`. linkifyjs can find `mailto` out of box but not `tel` however it can be done via custom protocol with `optionalSlashSlash` enabled

## Updated linkifyjs from `v3.0.5` to `4.1.0`
Some breaking changes ([changelist](https://github.com/Hypercontext/linkifyjs/releases))
1. Dropped support of ie11 and safari10
2. Links that begin with `mailto:` now have type `url` instead of `email`

## Changed `Link` extension
1. Added new custom protocol options
2. Updated protocols documentation


packages/extension-link/package.json
```diff
@@ -38,7 +38,7 @@
     "prosemirror-state": "^1.4.1"
   },
   "dependencies": {
-    "linkifyjs": "^3.0.5"
+    "linkifyjs": "^4.1.0"
```
bdbch, line 41: I guess this major version change won't lead to issues?

packages/extension-link/src/link.ts
```diff
@@ -6,6 +6,11 @@ import { autolink } from './helpers/autolink'
 import { clickHandler } from './helpers/clickHandler'
 import { pasteHandler } from './helpers/pasteHandler'
 
+export interface LinkProtocolOptions {
+  scheme: string;
+  optionalSlashSlash?: boolean;
```
bdbch, line 11: I'm not sure about the name of this option. How about `optionalSlashes`?

bdbch (changes requested):
Hey! Just two minor things. Otherwise looks good!

packages/extension-link/src/link.ts
taras-turchenko-moc: I used the argument name from linkify but I'm OK with `optionalSlashes` too

packages/extension-link/package.json
taras-turchenko-moc: 1. Tiptap doesn't use any removed plugins or custom ones. No issues there
2. Links that start with `mailto:` now have `link` type instead of `email`. We don't have checks based on link type but may be a potential breaking change if someone imported linkify as a transitive dependency. It's a pretty bad pattern so we may ignore it or add some note to the changelog
3. Dropped ie11 and safari10 support. I hope no issues with it

bdbch (dismissed):

bdbch (dismissed):

bdbch:
Looks good @taras-turchenko-moc could you resolve the conflict so we can merge this?

taras-turchenko-moc:
@bdbch done

bdbch:
Could you resolve the conflict again? I'll merge afterwards

taras-turchenko-moc:
I wanna cry 🥲

taras-turchenko-moc:
@bdbch Merged again

taras-turchenko-moc:
@bdbch Hi! Can you take a look?

Issue #3518, opened 2022-12-08
Interactive node views can be dragged bypassing the drag handle (to no avail)


fenok:
### What’s the bug you are facing?

It's possible to drag an interactive node view without using the drag handle. The drag doesn't actually do anything, presumably because it only works when the drag handle is used.

### Which browser was this experienced in? Are any special extensions installed?

Latest Chrome, no extensions.

### How can we reproduce the bug on our side?

It can be seen in the drag example: https://tiptap.dev/guide/node-views/examples#drag-handles

https://user-images.githubusercontent.com/15672677/206466322-734b850d-b218-4e2e-ab91-81aba77385a2.mov

### Can you provide a CodeSandbox?

Doesn't seem to be necessary.

### What did you expect to happen?

Drag should only be possible by using the drag handle.

### Anything to add? (optional)

I was able to narrow it down to this line of code: https://github.com/ueberdosis/tiptap/blob/5380758365a2007aa7400edfad7044cbc3286fb7/packages/core/src/NodeView.ts#L127

It seems that a node view renderer, due to being contentEditable, is considered an input, and the function returns too early, without a chance to cancel the event here: https://github.com/ueberdosis/tiptap/blob/5380758365a2007aa7400edfad7044cbc3286fb7/packages/core/src/NodeView.ts#L148

I was able to resolve this by patching the library so that the check looks like this: `(isInput && !isDropEvent && !isDragEvent)`. However, I'm not sure what's going on here, and it might be better to always exclude the renderer here (I'm not sure how to do this).

### Did you update your dependencies?

- [X] Yes, I’ve updated my dependencies to use the latest version of all packages.

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

imownbey:
We are also running into this, it makes using a drag-handle on a contenteditable="false" basically useless and also disables selecting any text inside that component (because clicking anything will instead begin dragging)

PR #3677, opened 2023-01-30
Fix/drag not respecting handles

matthewmullin01:
Fixes https://github.com/ueberdosis/tiptap/issues/3518

> It's possible to drag an interactive node view without using the drag handle. The drag doesn't actually do anything, presumably because it only works when the drag handle is used.

With this fix nodes are only draggable via their drag handle.

| Before  | After |
| ------------- | ------------- |
| TODO  | TODO  |


bdbch (approved):

matthewmullin01:
Thanks for the review @bdbch! Just had some last min conflicts to solve, otherwise it's exactly as it was before.

bdbch (approved):

PR #3704, opened 2023-02-04
build(vue3-deps): Moves extension deps to peerDependencies & devDependencies

jambonn:
Do not affect when there is a new version

### DIFF
--- a/packages/extension-floating-menu/package.json
+++ b/packages/extension-floating-menu/package.json
@@ -28,8 +28,8 @@
     "dist"
   ],
   "peerDependencies": {
-    "@tiptap/core": "^2.0.0-beta.209",
-    "@tiptap/pm": "^2.0.0-beta.209"
+    "@tiptap/core": "^2.0.0-beta.212",
+    "@tiptap/pm": "^2.0.0-beta.212"
   },
   "devDependencies": {
     "@tiptap/core": "^2.0.0-beta.212",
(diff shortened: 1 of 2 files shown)


bdbch:
Hey @jambonn - while I think it's a good idea to not have any hard dependencies we had a few issues with yarn and pnpm not installing those automatically. Do you think we would make the installation process for react/vue project more complicated that way?

If so I think we should update the documentation of said library installation guides to let users now that those extensions need to be installed too.

jambonn:
@bdbch After the release of version `v2.0.0-beta.212`, I’m detecting a problem. If the project doesn’t have *lock, then after an installment, dependencies will automatically get the latest version (due to our library being in beta). Therefore I think it’s best to remove the dependencies. If you don’t see any problem here, I’ll update the installment docs accordingly.

bdbch:
> I’ll update the installment docs accordingly.

That would be great.
Yeah right now the beta versioning makes updates tedious. We hope to get to the stable version soon so we can start proper beta/alpha releases without breaking to many things when introducing big changes for a lot of people.

jambonn:
I've updated the documentation.

jambonn:
@bdbch Do you see a problem if I update both react?

jambonn:
Move new merge request [#3723](https://github.com/ueberdosis/tiptap/pull/3723)

Issue #2659, opened 2022-03-29
Placeholder not rendering


sellsjonathan:
### What’s the bug you are facing?

When the editor is empty, either initially, or after deletion, no placeholder is displayed to the user.

A paragraph node is added with the appropriate classes, and a `data-placeholder` as set by my configuration, however, there is no displayed placeholder for the user.  The Paragraph tag just contains an empty `<br/>`

![image](https://user-images.githubusercontent.com/8787299/160528186-a3337f09-4de2-4f2f-b512-3ed09db1f0ce.png)

### How can we reproduce the bug on our side?

Set up basic editor, use `@tiptap/extension-placeholder` as per docs

### What did you expect to happen?

The placeholder text gets displayed when there is no content.

### Did you update your dependencies?

- [X] Yes, I’ve updated my dependencies to use the latest version of all packages.

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

cadars:
Placeholder text is displayed via a pseudo element in CSS, have a look at the demo setup: https://embed.tiptap.dev/preview/Extensions/Placeholder

KarolMilewski:
add some CSS to your component
```css
/* Placeholder (at the top) */
.ProseMirror p.is-editor-empty:first-child::before {
  content: attr(data-placeholder);
  float: left;
  color: #adb5bd;
  pointer-events: none;
  height: 0;
}

/* Placeholder (on every new line) */
/*.ProseMirror p.is-empty::before {
  content: attr(data-placeholder);
  float: left;
  color: #adb5bd;
  pointer-events: none;
  height: 0;
}*/
```

sellsjonathan:
Oh wow, I was trying to hack it via this type of approach, didn't realize it's how they want us to actually render it.

Does this seem a little hacky though at first glance? What benefit does this have vs, a default display through just the `PlaceHolder.configure`.

At the very least I feel this should be made a little more clear in the docs.  I completely missed the styles tab in the example myself at 1am last night.

Appreciate the replies everyone.  Thanks.

drummerjolev:
Any guidance on how to style the placeholder with Tailwind?

I'm using this:
```javascript
Placeholder.configure({
  placeholder,
  emptyNodeClass:
    'first:before:text-gray-400 first:before:float-left first:before:content-[attr(data-placeholder)] first:before:pointer-events-none',
}),
```

The placeholder works but the cursor isn't placed at the beginning of the placeholder text:
[...]

PR #3711, opened 2023-02-06
Added CSS Required for Setup

koffietiger:
It was previously not clear that additional CSS was required to display placeholders, as discussed in #2659.


bdbch (approved):
Great idea. Thanks for your PR!

Issue #3717, opened 2023-02-07
The entrypoint for '@tiptap/extension-blockquote' could not successfully be resolved. The file '/turbo_modules/@tiptap/extension-blockquote@2.0.0-beta.212/dist/index.d.cjs' does not exist.


splincode:
### What’s the bug you are facing?

<img width="925" alt="image" src="https://user-images.githubusercontent.com/12021443/217267751-07a1ae95-cfc8-4122-996e-cf5e9aabcfae.png">

### Which browser was this experienced in? Are any special extensions installed?

Chrome 109, MacOS

### How can we reproduce the bug on our side?

<img width="1362" alt="image" src="https://user-images.githubusercontent.com/12021443/217267528-4bda1d76-5cec-4985-a6cf-0c0c8f4afb1e.png">

https://stackblitz.com/edit/angular-pj96to?file=src%2Fapp%2F%40stackblitz%2Fall-taiga-modules.ts&hideExplorer=1

### What did you expect to happen?

Not error

### Did you update your dependencies?

- [X] Yes, I’ve updated my dependencies to use the latest version of all packages.

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

splincode:
https://github.com/ueberdosis/tiptap/pull/3716

bdbch:
I just released `2.0.0-beta.214` which uses our old build process with a few minor changes and the new PM package. Would be nice to see if this fixes the issues introduced by the esbuild builds.

splincode:
https://github.com/Tinkoff/taiga-ui/pull/3626
Our test are failed
<img width="770" alt="image" src="https://user-images.githubusercontent.com/12021443/217517432-8dd4b5bc-d212-4ced-9c52-99378b33078c.png">

bdbch:
Could you try again with the most recent package?

splincode:
fixed

Issue #3708, opened 2023-02-06
Bug with build project after release v2.0.0-beta.212 (lodash/debounce issue)


Intiligent:
After release v2.0.0-beta.212 corrupt build of my project.
Update all tiptap packages to latest version not solve the problem.

<img width="1139" alt="image" src="https://user-images.githubusercontent.com/2857625/216983653-566683e1-bb7f-45cc-bb21-41b9479172e2.png">

ERROR in ./node_modules/@tiptap/extension-bubble-menu/dist/index.js 11:0-39
Module not found: Error: Can't resolve 'lodash/debounce' in '/Users/vitaliy/Documents/domains/infusemedia.loc/node_modules/@tiptap/extension-bubble-menu/dist'
Did you mean 'debounce.js'?
BREAKING CHANGE: The request 'lodash/debounce' failed to resolve only because it was resolved as fully specified
(probably because the origin is strict EcmaScript Module, e. g. a module with javascript mimetype, a '*.mjs' file, or a '*.js' file where the package.json contains '"type": "module"').
The extension in the request is mandatory for it to be fully specified.
Add the extension to the request.

### Which browser was this experienced in? Are any special extensions installed?

not depend on browser. error in webpack build or in tiptap dependency

### How can we reproduce the bug on our side?

```
rm -rf node_modules
npm cache clean --force
npm i
// build the project
// in my case
npm run dev

### Can you provide a CodeSandbox?

_No response_

### What did you expect to happen?

expect build the project without error

### Anything to add? (optional)

_No response_

### Did you update your dependencies?

- [X] Yes, I’ve updated my dependencies to use the latest version of all packages.

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

**Comment by nephew92** [comment]
same error here in production
👉 no package have change since last build
![image](https://user-images.githubusercontent.com/9345442/216995531-132301d8-9d89-475d-b0a9-9067475e40c1.png)

👉 specify old versions dindnt solve the problem

<img width="675" alt="image" src="https://user-images.githubusercontent.com/9345442/216995431-c1fc8d61-9377-4e97-a8e6-894279b961c3.png">

-- please, don't do big changes affecting all verisons, this is not useful --

**Comment by benbowler** [comment]
I'm also seeing this.

**Comment by Intiligent** [comment]
@nephew92 problem with resolve internal package solve as add package to package.json
```
"@tiptap/pm": "^2.0.0-beta.212",
```
or install manualy
```
[...]

Issue #3694, opened 2023-02-03
TypeError: Cannot set property Editor of #<Object> which has only a getter


imranbarbhuiya:
### What’s the bug you are facing?

I've just updated tiptap to `^2.0.0-beta.212` and my build is failing with error

```sh
🕙 09:17 ❯ yarn build
unocss v0.49.4                                                                                                             09:17:14
start UnoCSS for production...                                                                                             09:17:14
√ 510 utilities generated to src/styles/unocss.css                                                                         09:17:14

info  - Loaded env from D:\Discord\Sofi-Site\apps\website\.env.production
warn  - You have enabled experimental feature (appDir) in next.config.js.
warn  - Experimental features are not covered by semver, and may cause unexpected or broken application behavior. Use at your own risk.
info  - Thank you for testing `appDir` please leave your feedback at https://nextjs.link/app-feedback

info  - Creating an optimized production build
info  - Compiled successfully
info  - Skipping linting
info  - Checking validity of types
info  - Collecting page data ...TypeError: Cannot set property Editor of #<Object> which has only a getter
    at Object.<anonymous> (D:\Discord\Sofi-Site\node_modules\@tiptap\react\dist\index.cjs:437:49)
    at Module._compile (node:internal/modules/cjs/loader:1218:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1272:10)
    at Module.load (node:internal/modules/cjs/loader:1081:32)
    at Module._load (node:internal/modules/cjs/loader:922:12)
    at Module.require (node:internal/modules/cjs/loader:1105:19)
    at require (node:internal/modules/cjs/helpers:103:18)
    at Object.<anonymous> (D:\Discord\Sofi-Site\node_modules\@mantine\tiptap\cjs\Content\Content.js:7:13)
    at Module._compile (node:internal/modules/cjs/loader:1218:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1272:10)

> Build error occurred
Error: Failed to collect page data for /profile/[userId]
    at D:\Discord\Sofi-Site\node_modules\next\dist\build\utils.js:963:15
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5) {
  type: 'Error'
}
info  - Collecting page data .
```

### Which browser was this experienced in? Are any special extensions installed?

At build time

### How can we reproduce the bug on our side?

running `yarn build` in below link should show the error but instead, it's showing #3689 

### Can you provide a CodeSandbox?

[...]

Issue #3692, opened 2023-02-02
Bug in release process


benatshippabo:
### What’s the bug you are facing?

Hello, it is my suspicion that when a release is cut for tiptap it overwrites previous releases.

If we install tiptap version `2.0.0-beta.195`, we can see unmet dependencies on `@tiptap/pm`, but `@tiptap/pm` did not exist when `2.0.0-beta.195` was versioned:

https://github.com/ueberdosis/tiptap/tree/v2.0.0-beta.195/packages

### Which browser was this experienced in? Are any special extensions installed?

N/A

### How can we reproduce the bug on our side?

```sh
yarn add 
```

Notice warnings, about `@tiptap/pm`:

```
➤ YN0000: ┌ Resolution step
➤ YN0002: │ @tiptap/react@npm:2.0.0-beta.195 [f2cea] doesn't provide @tiptap/pm (p56fa6), requested by @tiptap/extension-bubble-menu
➤ YN0002: │ @tiptap/react@npm:2.0.0-beta.195 [f2cea] doesn't provide @tiptap/pm (pa23e8), requested by @tiptap/extension-floating-menu
```

This indicates that the older package `2.0.0-beta.195` was overwritten with a newer release.

### What did you expect to happen?

Versioned packages to not be overwritten with newer ones.

### Did you update your dependencies?

- [X] Yes, I’ve updated my dependencies to use the latest version of all packages.

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

bdbch:
It seems like `@tiptap/extension-bubble-menu` or/and `@tiptap/extension-floating-menu` are requested in a newer version (higher than beta.195) which leads to it requesting the new beta.211 version which then requires `@tiptap/pm`.

I'll see if we can fix this in retroperspective.

ShowBamboo:
The same problem, excuse me how to solve

jk-baseer:
Hi @bdbch 
Just asking, if you managed to solve this and have some example steps on how to achieve this? It's happening in my next JS project and this is the package version I have. 

`
"@tiptap/extension-color": "^2.0.0-beta.209",	
"@tiptap/extension-text-align": "^2.0.0-beta.31",
	
"@tiptap/extension-text-style": "^2.0.0-beta.209",	

"@tiptap/extension-underline": "^2.0.0-beta.25",	

"@tiptap/react": "^2.0.0-beta.114",	

[...]

PR #3720, opened 2023-02-07
Move back from tsup/esbuild to rollup

bdbch:
This PR should fix a few issues introduced in version beta.210 to beta.213 by going back to the old build process for all packages except the `@tiptap/pm` package.

Each package now has it's own build process to avoid build memory leaks and building now is happening via lerna running through each package individually.

Issues connected to this:
- https://github.com/ueberdosis/tiptap/issues/3717
- https://github.com/ueberdosis/tiptap/issues/3708
- https://github.com/ueberdosis/tiptap/issues/3694
- https://github.com/ueberdosis/tiptap/issues/3692
- https://github.com/ueberdosis/tiptap/issues/3689


packages/extension-blockquote/package.json
```diff
@@ -15,14 +15,15 @@
   "type": "module",
   "exports": {
     ".": {
-      "types": "./dist/index.d.ts",
+      "types": "./dist/packages/extension-blockquote/src/index.d.ts",
       "import": "./dist/index.js",
       "require": "./dist/index.cjs"
     }
   },
   "main": "dist/index.d.cjs",
```
splincode, line 23: is it correct? not `"main": "dist/index.cjs",`

bdbch: Thanks, good find!

svenadlung (approved):

PR #3750, opened 2023-02-15
Update regex to exclude channel URL unfurling

JustMaier:
Just ran into this issue on our site and thought I'd share the fix.


bdbch (approved):
Awesome, thanks for your PR!

PR #3759, opened 2023-02-17
Extend `nodePasteRule` `find` type to most generic `PasteRuleFinder`

jiegillet:
Hello team!
First off, thank you for building and maintaining `tiptap`, it is truly a pleasure to work with.

While working on our editor, we ran across `nodePasteRule` which we wanted to use with a non-regexp pattern.
The specs in `PasteRule.ts` doe also allow functions as argument for `find`, and functions actually work perfectly with `nodePasteRule` as well, but there will be a type error.

This PR changes the type of `nodePasteRule`'s `find` to the most general type it can accept.
I hope it's  OK to open a PR directly for this two-line change.

### DIFF
--- a/packages/core/src/pasteRules/nodePasteRule.ts
+++ b/packages/core/src/pasteRules/nodePasteRule.ts
@@ -1,6 +1,6 @@
 import { NodeType } from '@tiptap/pm/model'
 
-import { PasteRule } from '../PasteRule'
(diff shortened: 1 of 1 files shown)


bdbch (approved):
Thanks for your PR @jiegillet.
Looks good to me!

PR #3776, opened 2023-02-23
Adds attributes to toggleList

katerlouis:
When dealing with different variants of bullet lists, I wanted to adopt the same technique I used for different paragraph variants. Since `wrapInList` is capable of receiving attributes, just like `setNode` is, I don't see any reason why `toggleList` should not be capable of the same. 

Here's my bullet list extension in action that is in need of attributes support.

```js
export const CustomBulletList = BulletList.extend({
  content: 'listItem*',

  addAttributes() {
    return {
      variant: {
        default: DEFAULT_LIST,

        renderHTML: attributes => {
          return {
            class: `list-${attributes.variant}`,
          };
        },
      },
    };
  },

  addCommands() {
    return {
      toggleBulletList: attributes => (c) => {
        return c.commands.toggleList(this.name, this.options.itemTypeName, true, attributes);
      },
    };
  },
});
```

### DIFF
--- a/docs/api/commands/toggle-list.md
+++ b/docs/api/commands/toggle-list.md
@@ -10,6 +10,14 @@ The type of node that should be used for the wrapping list
 
 The type of node that should be used for the list items
 
+`keepMarks?: boolean`
+
+If marks should be kept as list items or not
+
+`attributes?: Record<string, any>
+
+The attributes that should be applied to the list. **This is optional.**
+
 ## Usage
 ```js
 // toggle a bullet list with list items
--- a/docs/api/commands/toggle-list.md
+++ b/docs/api/commands/toggle-list.md
@@ -14,7 +14,7 @@ The type of node that should be used for the list items
 
 If marks should be kept as list items or not
 
-`attributes?: Record<string, any>
+`attributes?: Record<string, any>`
 
 The attributes that should be applied to the list. **This is optional.**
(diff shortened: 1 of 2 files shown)


docs/api/commands/toggle-list.md
```diff
@@ -10,6 +10,14 @@ The type of node that should be used for the wrapping list
 
 The type of node that should be used for the list items
 
+`keepMarks?: boolean`
+
+If marks should be kept as list items or not
+
+`attributes?: Record<string, any>
```
bdbch, line 17: I think you're missing a trailing ` here.

bdbch (changes requested):
Nice, thank you. Just one minor thing.

@svenadlung you also want to review?

svenadlung (commented):
LGTM besides what @bdbch has already mentioned

bdbch (approved):

bdbch:
Could you check the linting issues again @katerlouis ?

bdbch:
Nevermind - seems to come from outside. I'll fix it!

gethari:
@bdbch I could'nt able to find this attribute code (`variant`) in latest default branch `develop` ? Is it missed or am I missing something ?

bdbch:
@gethari can you double check that rc.1 is installed?

PR #3777, opened 2023-02-23
Only allow left mouse button to open links

thecodrr:
Currently links inside the editor can be opened by pressing any of the mouse buttons. This changes fixes that by allowing only the left mouse button to open links (as is the behavior every where).


bdbch (approved):

bdbch (approved):
lgtm @svenadlung?

svenadlung (approved):

PR #3778, opened 2023-02-23
Optimize empty document detection in `documentClear` plugin

thecodrr:
Currently Tiptap checks if the document is empty on almost every transaction using `doc.textBetween` which is quite expensive. This is not required most of the time because Select all + delete is a rare case. This PR moves the `doc.textBetween` based check after the other checks to it only runs when it is 100% that user has actually selected all the text.


bdbch (approved):
Looks good to me. What about you @svenadlung?

svenadlung (approved):

PR #3533, opened 2022-12-13
fix: Queue flushSync call

kylealwyn:
Do the same thing as https://github.com/ueberdosis/tiptap/pull/3188 - confirmed this got rid of the warning locally

PR #3781, opened 2023-02-23
fix(react): reset initialized when editorcontent is unmounting

### LINKED ISSUE #3580: Error: flushSync was called from inside a lifecycle method in <PureEditorContent>

The component `import { EditorContent } from "@tiptap/react"` is crashing for some reason.

I'm getting the following error in console:

```
Uncaught Error: flushSync was called from inside a lifecycle method. It cannot be called when React is already rendering.
    at flushSync (react-dom.development.js:21917:1)
    at PureEditorContent.maybeFlushSync (EditorContent.tsx:82:1)
    at PureEditorContent.removeRenderer (EditorContent.tsx:100:1)
    at ReactRenderer.destroy (ReactRenderer.tsx:93:1)
    at ReactNodeView.destroy (ReactNodeViewRenderer.tsx:173:1)
...
```

More specifically here:
```
The above error occurred in the <PureEditorContent> component:
    in PureEditorContent
    in PureEditorContent (at TipTap.js:154)
    in div (at TipTap.js:145)
...
```

Reduced code:
```javascript
const TipTap = ({ editor, ...props }) => {
  return (
    <div className="tip-tap-editor-container text-editor">
      <EditorContent className="tip-tap-editor-content mt-5" editor={editor} />
    </div>
  )
}
```

If I remove the `<EditorContent>` the app doesn't crash.

Not sure if this bug is only related to our implementation.
Anyway, any help, clue or information is welcome in order to fix this issue which is urgent.

Google Chrome
Version 108.0.5359.124 (Official Build) (arm64)

macOs ventura 13.1

Not sure.

_No response_

Not crashing.

Tip-tap package versions:
```json
{
    "@tiptap/extension-color": "^2.0.0-beta.209",
    "@tiptap/extension-highlight": "^2.0.0-beta.209",
    "@tiptap/extension-image": "^2.0.0-beta.209",
    "@tiptap/extension-link": "^2.0.0-beta.209",
    "@tiptap/extension-placeholder": "^2.0.0-beta.209",
    "@tiptap/extension-text-style": "^2.0.0-beta.209",
    "@tiptap/extension-underline": "^2.0.0-beta.209",
    "@tiptap/react": "^2.0.0-beta.209",
    "@tiptap/starter-kit": "^2.0.0-beta.209",
}
```

React version: `16.13.1`
Node version: `v14.19.1`

You can find the entire log here: [tiptap.log](https://github.com/ueberdosis/tiptap/files/10315868/tiptap.log)

---
**Note:** my app depends of this node version only

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

### LINKED ISSUE #3764: flushSync gets called on render when the Editor gets updated with content that uses an extension with addNodeView

I noticed that `flushSync` gets called on render when editorContent gets updated with content that uses `addNodeView`. This results in the warning message:
```Warning: flushSync was called from inside a lifecycle method. React cannot flush when React is already rendering. Consider moving this call to a scheduler task or micro task.```

**Which browser was this experienced in?**
Google Chrome

**Are any special extensions installed?**
 I have copied the demo extension from "https://github.com/ueberdosis/tiptap/blob/main/demos/src/Examples/InteractivityComponentContent/React/Extension.js" to the "App.js" file in the CodeSandbox.io example to illustrated the bug.

** To recreate manually **
1) Create an extension that utilizes addNodeView as part of the render
2) Add content that renders the extension to the editorcontent.content after the first render. To do this, I used router.isReady to trigger the re-render by setting the content data after the router is ready

** Sample code **
```
import {
  EditorContent,
  useEditor,
  ReactNodeViewRenderer,
  NodeViewContent,
  NodeViewWrapper,
} from "@tiptap/react";
import { useRouter } from "next/router";
import { useEffect, useState } from "react";

import React from "react";

import { mergeAttributes, Node } from "@tiptap/core";

const Component = () => {
  return (
    <NodeViewWrapper className="react-component-with-content">
      <span className="label" contentEditable={false}>
        React Component
      </span>

      <NodeViewContent className="content" />
    </NodeViewWrapper>
  );
};

const Extension = Node.create({
  name: "reactComponent",

  group: "block",

  content: "inline*",

  parseHTML() {
    return [
      {
        tag: "react-component",
      },
    ];
  },

    return ["react-component", mergeAttributes(HTMLAttributes), 0];
  },

  addNodeView() {
    return ReactNodeViewRenderer(Component);
  },
});

const TiptapTest = ({ data }: { data: string }) => {
  const editor = useEditor(
    {
      extensions: [StarterKit, Extension],
      content: data,
    },
    [data]
  );

  return (
    <div>
    </div>
  );
};
export default function Index() {
  const router = useRouter();
  const [data, setData] = useState("");
  useEffect(() => {
    if (router.query) {
      setData(`<p>
      This is still the text editor you’re used to, but enriched with node views.
    </p>
    <react-component>
      <p>This is editable.</p>
    </react-component>
    <p>
      Did you see that? That’s a React component. We are really living in the future.
    </p>`);
    }
  }, [router.isReady]);

  return TiptapTest({ data });
}
```

https://codesandbox.io/s/modern-sun-h7r3kg

`flushSync` to not be called during render.

I suspect that the problem is caused when flushSync gets called from inside a lifecycle method when rendering content that has extensions that utilizes addNodeView when the editor unmounts and re-mounts.

In fix #3533, flushSync was queued onto the microtask in order to prevent flushSync from being called from a lifecycle method. This does not account for what happens when the editor un-mounts and re-mounts and after editorContent.initialized is set to "true".

We propose a modification to EditorContent.tsx that unsets the initialized state when the component is about to be unmounted:
```
componentWillUnmount() {
    this.initialized = false;
}
```

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

### PR DESCRIPTION
**Pull request by bdbch**
This PR should resolve the flushSync issues (see #3331, #3580, #3764) and pr #3533 

From my local tests I didn't receive any error messages nor did my cursor stay in the old position after inserting a new React NodeView with an editor content.

closes #3331
closes #3580
closes #3764

### DIFF
--- a/packages/react/src/EditorContent.tsx
+++ b/packages/react/src/EditorContent.tsx
@@ -79,9 +79,7 @@ export class PureEditorContent extends React.Component<EditorContentProps, Edito
     // lifecycle methods, and React doesn't allow calling flushSync from inside
     // a lifecycle method.
     if (this.initialized) {
-      queueMicrotask(() => {
-        flushSync(fn)        
-      })
+      flushSync(fn)
     } else {
       fn()
     }
@@ -115,6 +113,10 @@ export class PureEditorContent extends React.Component<EditorContentProps, Edito
 
     if (!editor) {
       return
+    }
+
+    if (this.initialized) {
+      this.initialized = false
     }
 
     if (!editor.isDestroyed) {

### CODE REVIEW

**Review by svenadlung — DISMISSED** [review]

File: packages/react/src/EditorContent.tsx [hunk 1]
```diff
@@ -117,6 +115,10 @@ export class PureEditorContent extends React.Component<EditorContentProps, Edito
       return
     }
 
+    if (this.initialized) {
```
jamesopti, line 118: I don't think this check is necessary.

jamesopti (approved):
I have tested this as well and it appears to work well.

KentoMoriwaki:
I think this PR doesn't completely resolve this issue https://github.com/ueberdosis/tiptap/pull/3331 which is reproduced by https://github.com/ueberdosis/tiptap/pull/3533 
Here's a [CodeSandbox](https://codesandbox.io/s/react-nodeview-flushsync-debug-alternate-forked-ut7oo0) that demonstrates the problem.

Could you please consider an alternative approach that doesn't use `queueMicrotask` during normal updates?

KentoMoriwaki:
I opened a PR about this. 
https://github.com/ueberdosis/tiptap/pull/3862

PR #3783, opened 2023-02-23
fix: update typings for node view decorations

### LINKED ISSUE #3772: Decoration Types doesnt match what gets passed for node views

`NodeViewProps` has `Decorations` which is an `Decoration[]`, when we iterate over the Decorations we actually see

```
{ to, from, type } 
````
instead of the actual typing described in `@tiptap/pm/view`, so we get type errors when we want to check if `has-focus` decoration is active per node

All browsers :)

add a new paragraph, check console and you should see this

https://codesandbox.io/s/nice-cdn-6ktib5?file=/src/App.vue

image

logged from the custom paragraph component
```
  mounted() {
    if (this.decorations) {
      console.log(this.decorations);
    }
  },
  ```

https://codesandbox.io/s/nice-cdn-6ktib5?file=/src/App.vue

the type should match the type in { to, from, type } instead of the actual typing described in `@tiptap/pm/view`

_No response_

### Are you sponsoring us?

- [X] Yes, I’m a sponsor. 💖

bdbch:
This PR should close #3772 by adding a new type called `DecorationWithType` which includes the default `Decoration` type coming from `@tiptap/pm/view` but includes the NodeType.

This way people can access the `decoration.type` property without experiencing typing issues.

### DIFF
--- a/packages/core/src/NodeView.ts
+++ b/packages/core/src/NodeView.ts
@@ -38,7 +38,7 @@ export class NodeView<
     } as Options
     this.extension = props.extension
     this.node = props.node
-    this.decorations = props.decorations
+    this.decorations = props.decorations as DecorationWithType[]
     this.getPos = props.getPos
     this.mount()
   }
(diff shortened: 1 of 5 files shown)


svenadlung (changes requested):
Please check tests:

> Error: Webpack Compilation Error
> [tsl] ERROR in /home/runner/work/tiptap/tiptap/packages/core/src/NodeView.ts(41,5)
>       TS2322: Type 'Decoration[]' is not assignable to type 'DecorationWithType[]'.
>   Type 'Decoration' is not assignable to type 'DecorationWithType'.
>     Property 'type' is missing in type 'Decoration' but required in type '{ type: NodeType; }'.

PR #3790, opened 2023-02-24
fix(core): allow insertContentAt and insertContent text node arrays

### LINKED ISSUE #2770: Problem when using insertContent with a text node list

I just found a regression. When I paste text in my editor, I call the server so it can parse it. I just saw that it doesn't work anymore I had a result showing : 
image

I thought the problem was coming from somewhere else but in fact when I execute this piece of code in a blank project I also have the problem. Before it was working. 

```
editor
              .chain()
              .insertContent([
                {
                  type: 'text',
                  text: 'Second paragraph',
                },
                {
                  type: 'text',
                  text: 'First paragraph',
                }
              ])
              .focus()
              .run();
```

Demo here : [https://codesandbox.io/s/tiptap-demo-tailwind-forked-ewe285?file=/src/index.js](https://codesandbox.io/s/tiptap-demo-tailwind-forked-ewe285?file=/src/index.js)

Chrome 100, Safari, Firefox Latest

Use this code :

```
editor
              .chain()
              .insertContent([
                {
                  type: 'text',
                  text: 'Second paragraph',
                },
                {
                  type: 'text',
                  text: 'First paragraph',
                }
              ])
              .focus()
              .run();
```

https://codesandbox.io/s/tiptap-demo-tailwind-forked-ewe285?file=/src/index.js

It used to insert the nodes without any problem.

_No response_

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

bdbch:
This PR fixes #2770 by checking the inserted text content if it's of type array. If so we can be sure that the content will be an array of text content.

### DIFF
--- a/demos/src/Commands/InsertContent/Vue/index.spec.js
+++ b/demos/src/Commands/InsertContent/Vue/index.spec.js
@@ -1,0 +1,11 @@
+context('/src/Commands/InsertContent/Vue/', () => {
+  before(() => {
+    cy.visit('/src/Commands/InsertContent/Vue/')
+  })
+
+  beforeEach(() => {
+    cy.get('.ProseMirror').then(([{ editor }]) => {
+      editor.commands.clearContent()
+    })
+  })
+})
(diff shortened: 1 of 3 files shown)


svenadlung (approved):

svenadlung:
Thx @bdbch 🎉

Issue #3128, opened 2022-08-26
Confusion and frustration when dealing with empty list items (aka "Bullet List Limbo")


domjost:
### What’s the bug you are facing?

Exiting lists in TipTap can be done by hitting `Enter`, which is standard behavior in all editors. But the way this editor deals with `Backspace` in an empt list item suggests that users can exit lists also using `Backspace`. The problem is that Backspace only visually removes the bullet, suggesting to the user that they are about to write a paragraph, when in fact they are still writing bullet items.

This leads to lots of unwanted formatting issues, confusion and frustrating editing experience.

### Which browser was this experienced in? Are any special extensions installed?

Safari, Version 15.6.1 (17613.3.9.1.16)
macOS, 12.5.1

### How can we reproduce the bug on our side?

I used https://tiptap.dev/examples/default and recorded the issue as well as my expected behavior here: https://www.loom.com/share/001d5b52095a4caa889db51c58a44932

This is what causes "Bullet List Limbo".
1. You need a list with at least two items, the last one being empty:

<img width="761" alt="CleanShot 2022-08-26 at 18 37 57@2x" src="https://user-images.githubusercontent.com/90393429/186952009-9ce2e98d-232a-41af-bdb8-a1c24b973cbd.png">

2. Place the cursor on the second item (the empty one), and hit backspace. What you get looks like this:

<img width="762" alt="CleanShot 2022-08-26 at 18 38 31@2x" src="https://user-images.githubusercontent.com/90393429/186952066-5e5a5b5c-d4fc-4a53-9aa0-15ad2f05f686.png">

3. At this point, lots of users believe they have exited the bulleted list and continue as if it where a paragraph:
<img width="764" alt="CleanShot 2022-08-26 at 18 39 50@2x" src="https://user-images.githubusercontent.com/90393429/186952262-a02cffed-344a-4111-bf5f-259b84fa3f92.png">

4. After finishing the paragraph, they hit enter and are surprised to see this:
<img width="755" alt="CleanShot 2022-08-26 at 18 40 25@2x" src="https://user-images.githubusercontent.com/90393429/186952350-49f2b82d-4958-4fdc-a92a-351e41f5e684.png">

At this stage the editing experience goes in all sorts of wrong directions, because they struggle to get the editor to do what they want it to do. In other words: the state of the editor is no more in sync with the mind of the user, causing immense frustration and believing the editor is broken.

### What did you expect to happen?

[...]

PR #3819, opened 2023-03-03
[WORK IN PROGRESS] - List Improvements

bdbch:
This PR will implement several list item improvements to make handling with lists easier.

## Todo

- [x] Correctly handle backspace when at start of list item
  - [x] And no previous list item on the same level exists (lift item)
  - [x] A previous list item exists (join paragraphs)
- [x] Correctly handle delete key when at end of list item
  - [x] And no following list item on the same level exists (merge next content in if possible?)
  - [x] A following list item exists (join paragraphs)
- [ ] Handle backspace when the cursor is behind a bulletList (right now the current paragraph gets turned into a list item)
- [x] Fix undoInputrule keymap handler with new backspace behavior


bdbch:
The current new behavior would be the following:

## if the cursor is at the start and there is no direct sibling list item on the previous line

in this case, the current list item will just be lifted up one level

```
From:
- A
  - A.1
    - A.1.a
  - A.2 (we move cursor before A.2 and press backspace)
- B

To:
- A
  - A.1
    - A.1.a
- A.2 (since we pressed backspace on this level, we lifted the item one level up
- B
```

## if the cursor is at the start of a list item and the previous list item is on the same level plus doesn't have a sublist

in this case, the list items plus their paragraphs are merged

```
from:
- People
  - John
  - Maria
  - Mark (cursor is before "Mark" and we press backspace)
  - Fred

to:
- People
  - John
  - MariaMark (the content from the "Mark" list item gets merged with the previous list item)
  - Fred
```

in case the merged list item had child items those are lifted up accordingly:

```
from:
- Recipes
  - Fruits (lets say we press backspace with a cursor in front of "Fruits")
    - Apple
    - Banana
  - Soups
    - Tomato Soup
    - Noodle Soup

to:
- RecipesFruits (recipes+fruits are joined, sub items are lifted)
  - Apple
  - Banana
  - Soups
    - Tomato Soup
    - Noodle Soup
```

bdbch:
Just a quick update:

This PR is **almost** ready to be merged. Since we have a strict deadline right now for `2.0.0`, we're going to release those changes with the first `2.1.0-rc.1` release candidate version.

Reason for that is, that those changes are quite dangerous to get merged fast without testing as a lot of editors could break if some bugs occur. When the `2.1.0-rc.1` version is out, I'd love to invite everyone to test the new list behavior with me, report bugs and issues etc. so we can get list behavior right.

bdbch:
Thanks for pointing this one out @bkempe - will fix that behavior and get those list changes into 2.1.0-rc.0 by the end of the week!

kalda341:
Any updates on when this is going to land?

Slapbox:
Further findings I thought should be considered regarding a (I think) previously unknown list copy/paste bug: https://github.com/ueberdosis/tiptap/issues/3128#issuecomment-1560221547

Slapbox:
Courtesy of AdventureBeard I tried the [extensionized version](https://gist.github.com/AdventureBeard/1ae977213f9343221e78e250cc9a67d8) of these changes with `2.1.0-rc.4`.

**Tested with this input**

```
<ul>
  <li>
    <p>Item one</p>
    <ul>
      <li>
        <p>Item two</p>
      </li>
      <li>
        <p>Item three</p>
      </li>
    </ul>
  </li>
</ul>
```

## Findings so far:

#### Resolved:
1. Primary backspace issue (list limbo) is resolved

#### Introduced problems:
1. New backspace issue where you cannot backspace into an existing list unless the last item in the list is only one item deep (otherwise it toggles list on/off indefinitely) 
2. New backspace issue where if you start at the end of the last listItem and Ctrl+Backspace until the list is gone, you cannot undo the changes - First undo brings back the bullet list and second undo errors with `Uncaught RangeError: Position 12 outside of fragment (<paragraph, bulletList(listItem(paragraph)), paragraph>)`
3. The newly reported copy/paste bug is unaffected. A single `listItem` can still contain multiple `paragraph` and `ul` tags

New problem \#3
image

#### Changes in behavior
1. One of our custom list commands had a minor change in behavior for top-level listItems only. I expect this to be easily resolved but I've not investigated further yet.

Slapbox:
Friendly bump.

Slapbox:
Regarding this item on the list:

>  - [ ] Handle backspace when the cursor is behind a bulletList (right now the current paragraph gets turned into a list item)

Is this not a more generic problem? It affects all sorts of lists as well as blockquotes. Is this something that can be fixed in a more generalizable way? Or does the (hypothetical) fix need to be applied to each `node` type individually? @bdbch

Slapbox:
@bdbch I implemented your changes as an Extension to try to fix the problems related to `Ctrl/Ctrl+Shift` in combination with `Backspace/Delete`.

https://gist.github.com/Slapbox/1a7ca430c5d50c7e656d96d4570b8d4f

I'd encourage everyone interested in this issue to try this extension and report back their findings.

I'm not sure why, but some of the issues I was previously seeing no longer seem to be appearing. Perhaps I'm testing improperly, or perhaps I previously tested improperly?

**The only issues I now see are:**

1. Deleting nodes with children leaves behind empty items - I believe this should instead lift the child items up so that the parents can be deleted. Additionally the resulting selection is incorrectly at the `from` position instead of the `to` position

https://github.com/ueberdosis/tiptap/assets/6835891/97916b43-b53a-46b9-93d4-76a786da83d6

2. Pasting content copied across `listItem`s can still insert extra levels into the list - This actually seems partially mitigated now but still could use an additional check to prevent inserting any more crud than necessary

https://github.com/ueberdosis/tiptap/assets/6835891/f3067131-3b39-493d-aa80-897faa2c8f18

kalda341:
@Slapbox The extension works great!

One minor issue I've found is that pressing backspace while the cursor is at the first character of the second line of a paragraph in a list item containing multiple paragraphs deletes the previous list item.

IMO a list item containing multiple paragraphs is an invalid state anyway (with this extension the only way I can get get to it is by Ctrl+Shift+V pasting multiple paragraphs into a list item), but it should probably be considered if this state is to be allowed.

Slapbox:
Great find @kalda341. I added a (poorly named) check for that but haven't tested the new version:

```js
  // avoid unindenting at start of paragraph if paragraph is not the first in the listItem
  if (atStartOfSecondOrLaterParagraph(state)) {
    return commands.joinBackward();
  }
```

This fix will _not_ help users who use the default `content` rule on `ListItem` if the second/third/etc node is not a paragraph. I advise people use the following instead of the default:

```
ListItem.extend({
        content: 'paragraph (bulletList|orderedList|paragraph)*',
})
```

Slapbox:
I discovered there are two remaining problems:

1. Backspace on Android unindents and lifts with a single press of <kbd>Backspace</kbd> but it should take two presses to achieve that. And when that problem occurs, the caret is also off-by-one (@bdbch I could really use some advice on possible causes for this)
2. <kbd>Delete</kbd> at the end of a node fails to pull the following `listItem` up and join it to the focused `listItem` if the following item has a child

image

kalda341:
@Slapbox Yes, I suggest everyone makes the content change - it seems like a much more sensible default.

Those changes have improved things enough for me that I'm going to trial them in production. Wish me luck!

Slapbox:
@bdbch I noticed you've done some more work on this; thank you very much for it!

I know I've commented quite extensively on many related issues here to the point it would be overwhelming to respond to everything, [but I hope you could find time to reply to this comment](https://github.com/ueberdosis/tiptap/pull/3819#issuecomment-1585827680), which I feel is the summation of dozens of hours of investigation into this subject.

Most especially to point \#1 and the idea to release these changes as an extension for guaranteed backwards compatibility for all users.

Slapbox:
Thanks for your reply and your hard work @bdbch!

Unfortunately, as released these changes break our existing work and lock us into being stuck at `2.1.0-rc.8`. I really think this would be best released as a separate extension, especially if it's being included in a release candidate. I wouldn't expect going from `2.1.0-rc.8` to `2.1.0-rc.9` to introduce any substantial changes, and especially not potentially breaking ones.

I'm not sure right now what would be involved for me to fix our project to work with `2.1.0-rc.9`.

As much as I want these changes integrated, I really think that the way they're released right now should be rolled back. I'd totally support a release-candidate release of them via an extension though, because that would be guaranteed not to break anything. Otherwise I'd suggest they wait for `2.2.0` to give people who are already using the `2.1.0` version lineage the opportunity to make it to the stable release and not be locked into a release candidate indefinitely.

kalda341:
@bdbch I really appreciate you coming back to this - it will be great to have an official solution to these issues.
However, there are a few issues that I've noticed:
- As @Slapbox has said, these changes don't work with `Mod-Backspace` and `Mod-Delete` + the backspace away of a `listItem` with a child.
- Backspacing doesn't work correctly with a `listItem` with multiple paragraph children, which has also been fixed in the extension.
- Backspace works incorrectly at the start of the paragraph after a list. I don't believe anyone has a fix for this yet.

For this reason, I'm going to stick with @Slapbox's extension for now instead of upgrading, though I hope these issues can be resolved officially.

Given this is a potentially breaking change for some users this late in an RC, I too support including the fixes as an extension. Including the fixes in 2.2 would be much more palatable, especially if they are a bit more complete and robust.
It's a pity this didn't make 2.0 as that would have been the perfect time to introduce these breaking changes!

Thanks again @bdbch, I really hope we can get to a great solution that works for everyone.

Slapbox:
I found that the problems I thought were interference from `2.1.0-rc.9` were actually a subtle bug I introduced the same day into our project. It does seem that disabling the default commands/keyboardShortcuts for `ListItem` allows us to use `2.1.0-rc.9`. Hooray!

But I still feel that this change should be reverted. It may fix some issues but it generally makes lists even more unpredictable to work with since the partial fixes introduce new edge cases users won't have seen before. I still really think it should be rolled back deferred until `2.2.0`.

bdbch:
Good point. I already thought about reverting this for now.

In the meantime: I created a PR (see #4140) which should make handling positions (specially nested) easier in the future. As soon as I'm done with this, I'll continue work on the lists. In the end I want lists to not be unpredictable anymore and don't break / lock users into unavoidable traps in the content while only navigating with the keyboard.

bdbch:
> Backspace works incorrectly at the start of the paragraph after a list. I don't believe anyone has a fix for this yet.

Just made a fix for this in a new branch.

> As @Slapbox has said, these changes don't work with Mod-Backspace and Mod-Delete + the backspace away of a listItem with a child.

Also made a fix for this in said branch

> Backspacing doesn't work correctly with a listItem with multiple paragraph children, which has also been fixed in the extension.

I tried this with a listItem with 2 paragraphs. On backspace, it seems to work fine - just delete is kind of funky. need to look into this. in general this blocks a stable release right now but I also don't really want to revert anything or move it into an optional package.

List behavior on vanilla Prosemirror is **really** weird and having this as an optional extension would really not fulfill what I'd expect Tiptap to do (I install Tiptap + List extensions and I expect a well working list editing behavior already).

I think I'd rather think about how we can share all that list logic between the list and taskList extensions.

> Unfortunately, as released these changes break our existing work and lock us into being stuck at 2.1.0-rc.8. I really think this would be best released as a separate extension, especially if it's being included in a release candidate. I wouldn't expect going from 2.1.0-rc.8 to 2.1.0-rc.9 to introduce any substantial changes, and especially not potentially breaking ones.

Can you give me more information what exactly is breaking right now or even create a new ticket out of this?

cc @Slapbox @kalda341

bdbch:
Maybe @rfgamaral can also drop his 5 cents in here since he pointed me towards this issue. I'd like lists to work reliable but also work as expected from the enduser.

rfgamaral:
@bdbch The discussion is a bit long already, and I'm not exactly sure what you are asking me. Can you please rephrase?

bdbch:
I think the main discussion right now is to if the list changes I made should be a separate extension so it doesn't break existing projects or it should be included in the core / list extension itself.

The bugs described in the thread here are already handled as far as I can see.

Slapbox:
@bdbch I think releasing it integrated into the existing lists extension would be fine since it turns out my issues were due to my own mistake - but I don't see how it would work with taskLists in that case which was another reason for my suggestion of a separate extension, or otherwise it seems like there's no way to avoid code duplication.

I'd not release it as part of the existing lists extension until `2.2.0` since it technically would constitute a breaking change, although I think the partially merged changes are still in `2.1.0-rc.x`.

kalda341:
@bdbch 2.2.0 would be great.

rfgamaral:
I think fixes like this belong in the core extensions. It doesn't make sense to me that I need to install multiple extensions to get lists behaviour working as they should by default.

What problem does the a separate extension solve? Code duplication for task lists? If that's the only problem, I think we should have code duplication for lists and task lists in interim releases (so that we can get these fixes out to everyone ASAP), and iterate on that, perhaps finding a better solution that doesn't rely on an extra extension?

bdbch:
> Code duplication for task lists?

I solved this by moving those helper functions handling the list positioning into the core where it belongs to (as it is used by multiple extension and could be used by other list extensions in the future)

I think we'll stick to it being in core and bump the version higher up so people know about those breaking changes regarding lists and can react accordingly.

Issue #3813, opened 2023-03-03
`Code` not `exitable` in `StarterKit`


nholden:
### What’s the bug you are facing?

https://github.com/ueberdosis/tiptap/pull/2925 allowed users to exit inline code blocks with the right arrow key. This behavior works when `Code` is included as a standalone extension, but it does not work when `Code` is included as part of `StarterKit`.

### Which browser was this experienced in? Are any special extensions installed?

Chrome Version 110.0.5481.177, no special extensions

### How can we reproduce the bug on our side?

Confirm that `Code` is not exitable as part of `StarterKit`:
1. Go to the [default code editor](https://tiptap.dev/examples/default) in the docs.
2. At the bottom of the editor, type an inline code block in the editor by placing text between backticks.
3. Click inside of the inline code block.
4. Press the right arrow key many times, you won't be able to exit the inline code block.

Confirm that `Code` is exitable as a standalone extension:
1. Go to [the usage section of the `Code` docs](https://tiptap.dev/api/marks/code#usage).
2. Click into the "This is code." inline code block in the editor.
3. Press the right arrow key until you have exited the inline code block.

### What did you expect to happen?

I expected `Code` to have the same `exitable` behavior as part of the `StarterKit` as it does as a standalone extension.

### Anything to add? (optional)

My suspicion is that something within the [`extend` method](https://github.com/ueberdosis/tiptap/blob/bdc49803438415b7c3ea69108d3af600a111c15d/packages/core/src/Extension.ts#L357-L388), which is [called in `StarterKit` via `configure`](https://github.com/ueberdosis/tiptap/blob/bdc49803438415b7c3ea69108d3af600a111c15d/packages/starter-kit/src/starter-kit.ts#L61), is breaking this behavior. These CodeSandboxes seem to confirm it:

- [Including `Code` in `extensions` supports `exitable`](https://codesandbox.io/s/charming-aj-1emdcu)
- [Including `Code.extend()` in extensions does not support `exitable`](https://codesandbox.io/s/red-sound-64b71w)

### Did you update your dependencies?

- [X] Yes, I’ve updated my dependencies to use the latest version of all packages.

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

PR #3822, opened 2023-03-04
Retain existing config when calling `configure` on Marks and Extensions

nholden:
Currently, when you call `.configure()` on a `Mark` or `Extension`, the `Mark` or `Extension` that's returned doesn't retain the config from the original.

This is a problem for StarterKit. When `Code` is added to an editor added as a standalone extension, [an inline code block is `exitable`](https://github.com/ueberdosis/tiptap/blob/43970fd496ffa22eae73ce95a11ed71d6efa4c03/packages/extension-code/src/code.ts#L47). When added as part of StarterKit, [an inline code block is not `exitable`](https://github.com/ueberdosis/tiptap/issues/3813) because [StarterKit calls `Code.configure()`](https://github.com/ueberdosis/tiptap/blob/43970fd496ffa22eae73ce95a11ed71d6efa4c03/packages/starter-kit/src/starter-kit.ts#L61), blowing away the default config.

This PR ensures that when you call `.configure()` on a `Mark` or `Extension`, the `Mark` or `Extension` that's returned retains the config from the original.

closes https://github.com/ueberdosis/tiptap/issues/3813

### DIFF
--- a/packages/core/src/Extension.ts
+++ b/packages/core/src/Extension.ts
@@ -340,7 +340,7 @@ export class Extension<Options = any, Storage = any> {
   configure(options: Partial<Options> = {}) {
     // return a new instance so we can use the same extension
     // with different calls of `configure`
-    const extension = this.extend()
+    const extension = this.extend(this.config)
 
     extension.options = mergeDeep(this.options as Record<string, any>, options) as Options
 
--- a/packages/core/src/Mark.ts
+++ b/packages/core/src/Mark.ts
@@ -464,7 +464,7 @@ export class Mark<Options = any, Storage = any> {
   configure(options: Partial<Options> = {}) {
     // return a new instance so we can use the same extension
     // with different calls of `configure`
-    const extension = this.extend()
+    const extension = this.extend(this.config)
 
     extension.options = mergeDeep(this.options as Record<string, any>, options) as Options
 
--- a/tests/cypress/integration/core/configureExtensions.spec.ts
+++ b/tests/cypress/integration/core/configureExtensions.spec.ts
@@ -1,0 +1,12 @@
+/// <reference types="cypress" />
+
+import { Extension } from '@tiptap/core'
+
+describe('configure extensions', () => {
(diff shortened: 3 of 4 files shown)


bdbch (approved):
Good find. Thanks, LGTM! @svenadlung ?

fleon:
@bdbch Can we get this merged? Looks pretty straightforward. 😄

svenadlung:
@bdbch Looks good to me. Feel free to merge.

huangmb:
I think the main problem is `ExtensionManager`，`extension.type === 'mark' && extension.config.exitable` not call callOrReturn(getExtensionField(extension,'exitable'))

huangmb:
my temporary solution is 
```
addProseMirrorPlugins() {
        const extensions = this.options.extensions || [];
        return extensions.filter(extension => extension.type === 'mark')
            .map(extension => {
                const editor = this.editor
                const context = {
                    name: extension.name,
                    options: extension.options,
                    storage: extension.storage,
                    editor,
                }
                if (extension.config.exitable) {
                    // ExtensionManager已经正常处理了,跳过
                    return null
                }
                const exitable = callOrReturn(getExtensionField(extension, 'exitable', context))
                if (exitable) {
                    return keymap({
                        ArrowRight: () => Mark.handleExit({editor, mark: extension})
                    })
                }
                return null
            })
            .filter(plugin=>plugin)
    }
```
if tiptap core fix this bug, i will remove this extension

bdbch:
Sorry but I had to revert this PR as it was breaking a lot of the core functionality. It's included in `v2.1.0-rc.2` if you want to test it yourself.

Basically `this.config` will be `{ name: 'mark', defaultOptions: {} }` by default which means type names won't be correct anymore and the defaultOptions warning will be appear by default.

Could you check into this again @nholden ?

nholden:
Good catch, @bdbch! I took another stab at this in https://github.com/ueberdosis/tiptap/pull/3998.

PR #3830, opened 2023-03-07
Fix: add repetitiveSeparator to getTextBetween

ivolkoff:
I made a fix so that there were lines when generating text.

I needed line breaks in empty paragraphs, so I added an option for this behavior, it doesn't break the current logic, and makes it possible to configure it for myself.

I use html

```html
<p>row1</p>
<p></p>
<p></p>
<p>row4</p>
```

And I get the text

```
row1
row4
```

after setting the option `repetitiveSeparator: true`, I get the text

```
row1


row4
```


packages/core/src/Editor.ts
```diff
@@ -427,14 +427,15 @@ export class Editor extends EventEmitter<EditorEvents> {
     blockSeparator?: string
     textSerializers?: Record<string, TextSerializer>
   }): string {
-    const { blockSeparator = '\n\n', textSerializers = {} } = options || {}
+    const { blockSeparator = '\n\n', textSerializers = {}, repetitiveSeparator = false } = options || {}
```
bdbch, line 430: I'm not sure about the name of this option. For me it wouldn't be really clear what is meant by this.

bdbch (changes requested):
Just a small note to the option name. Also the build is currently failing because of a TS error. Could you check this?

packages/core/src/Editor.ts
ivolkoff: Thanks for the comment. Please suggest a name that would be convenient?

bdbch: Not sure exactly. From what I see this is adding the blockSeperator even when blocks are not separated?

bdbch: maybe `forceBlockSeperator`? I'd just like to explain to the users what this option exactly does. Maybe keep the name and add a jsDoc explaining the option?

I myself couldn't exactly tell what the option does just from the name.

bdbch: I'd go for `allowRepetitiveSeparators` to indicate that this allows repetitive separators to be parsed

bdbch (changes requested):

bdbch:
Closed and moved here: #5904

PR #3833, opened 2023-03-08
fix: Ordered list start support broke in #3541

### LINKED ISSUE #3831: Ordered list `start` support broke in #3541 / v2.0.0-beta.220

Typing a number > 1 (e.g. 2) followed by `. ` no longer creates an `<ol start="…">`. 

For example, typing…
```
2. two
3. three
```
…no longer outputs…
```
<ol start="2">
  <li>two</li>
  <li>three</li>
</ol>
```
This can be seen here in GitHub with the same input quoted below:
> 2. two
> 3. three

All browsers

Use the text shortcut to create an ordered list with a number > 1 in an editor running v2.0.0-beta.220.

<details><summary>👀</summary>

2023-03-07 at 09 50 03

</details>

This worked in v2.0.0-beta.219 prior to #3541 being merged.

_No response_

The resulting ordered list is numbered as typed. Typing `2. two` should generate `<ol start="2"><li>two</li></ol>`.

I’m fairly certain [this change](https://github.com/ueberdosis/tiptap/pull/3541/files#diff-eadcb068d8e57ac1cb3bfea983f0cec87f826391e1be6d95a59a39e509c0ab30L85-R105) is the cause:

2023-03-07 at 09 55 35

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

gethari:
Fixes #3831

### DIFF
--- a/packages/extension-ordered-list/src/ordered-list.ts
+++ b/packages/extension-ordered-list/src/ordered-list.ts
@@ -99,7 +99,8 @@ export const OrderedList = Node.create<OrderedListOptions>({
         type: this.type,
         keepMarks: this.options.keepMarks,
         keepAttributes: this.options.keepAttributes,
-        getAttributes: () => { return this.editor.getAttributes(TextStyle.name) },
+        getAttributes: match => ({ start: +match[1], ...this.editor.getAttributes(TextStyle.name) }),
+        joinPredicate: (match, node) => node.childCount + node.attrs.start === +match[1],
         editor: this.editor,
       })
     }
--- a/packages/extension-ordered-list/src/ordered-list.ts
+++ b/packages/extension-ordered-list/src/ordered-list.ts
@@ -91,6 +91,8 @@ export const OrderedList = Node.create<OrderedListOptions>({
     let inputRule = wrappingInputRule({
       find: inputRegex,
       type: this.type,
+      getAttributes: match => ({ start: +match[1] }),
+      joinPredicate: (match, node) => node.childCount + node.attrs.start === +match[1],
     })
 
     if (this.options.keepMarks || this.options.keepAttributes) {


bdbch (approved):

bdbch:
Looks good to me. @svenadlung feel free

gethari:
> This doesn’t, out of the box, restore support for _inputting_ an ordered list with a `start` offset due to the relevant code being inside a condition…
> 
> https://github.com/ueberdosis/tiptap/blob/87c9dbd96bcb9efd918300cd0ea81372cd7578d0/packages/extension-ordered-list/src/ordered-list.ts#L96
> 
> …that’s bypassed by default (both options default to `false`).
> To reproduce type `2. two`.

@javan great catch once again !, I've resolved it.

gethari:
@javan I think the code is not deployed properly, because in this same branch I can see that it is working fine in the scenario that you mentioned above. Attached a video from localhost


https://user-images.githubusercontent.com/20477956/224339424-d5065770-56d2-4eed-9acd-4820d334e323.mp4

@bdbch do you see any problems here ? Am I missing something

javan:
> @javan I think the code is not deployed properly, because in this same branch I can see that it is working fine in the scenario that you mentioned above. Attached a video from localhost

That looks perfect! I’m not familiar with how PRs get built + deployed and am just working off [the comment here from Netlify](https://github.com/ueberdosis/tiptap/pull/3833#issuecomment-1459260606). Thanks for taking a closer look and fixing!

javan (approved):
💖

gethari:
@bdbch would it be possible to merge this along with next release ?

bdbch:
sure. @svenadlung you want to take another look?

svenadlung:
Just rebuilt the preview. LGTM 👍

gethari:
@bdbch would be really helpful if we can get a new release with this as early as possible

Issue #3809, opened 2023-03-02
Horizontal rule by input rule adds extra paragraph node above


thefinnomenon:
### What’s the bug you are facing?

When adding a horizontal rule using an input rule (--- or ___ ), it adds an empty paragraph node before the divider node.

### Which browser was this experienced in? Are any special extensions installed?

Doesn't apply.

### How can we reproduce the bug on our side?

You can recreate it in the horizontal rule playground https://tiptap.dev/api/nodes/horizontal-rule
enter ---

### What did you expect to happen?

I expect it to only add the divider

### Did you update your dependencies?

- [X] Yes, I’ve updated my dependencies to use the latest version of all packages.

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

rfgamaral:
Yes, I use horizontal rules a lot, and this one is very annoying. Hope it gets some attentiont soon, it shouldn't be too hard to fix 😅

rfgamaral:
Still valid.

steven-tey:
@rfgamaral @thefinnomenon here's a patch for this that worked for me: https://github.com/ueberdosis/tiptap/pull/3859#issuecomment-1536799740

thefinnomenon:
Thanks, yeah I have been using that patch and it works perfectly!

rfgamaral:
Thank you, @steven-tey.

Issue #1964, opened 2021-10-01
HorizontalRule extension: inconsistent behavior


cadars:
### What’s the bug you are facing?

Inserting an HorizontalRule produces different results depending on:
- How is it created, shortcut or command
- Where in the document it is created; first, last,  or in between nodes
- The presence of the StarterKit extensions (maybe because of GapCursor?)

### How can we reproduce the bug on our side?

With just the HorizontalRule extension: https://embed.tiptap.dev/preview/Nodes/HorizontalRule
- Type `---` or `___` or `***`, an empty `<p>` is created before the node.

With the StarterKit extensions: https://embed.tiptap.dev/preview/Examples/Default
- Type `---` or `___` or `***`, an empty `<p>` is created before the node.
- Empty the editor, click <kbd>**horizontal rule**</kbd>, this markup gets created:

  ```html
  <hr contenteditable="false">
  <p><br></p>
  <p><br><br></p>
  ```

( I am using the tiptap.dev examples but I also can reproduce these issues locally, with all dependencies updated)

### What did you expect to happen?

- Adding an horizontal rule with markdown shortcuts should not add any empty paragraphs **before** it.
- Adding an horizontal rule as the first node with the `setHorizontalRule` command should not add more than one empty paragraph **after** it.

### Anything to add? (optional)

As always, thanks for your time, and this amazing editor!

### Did you update your dependencies?

- [X] Yes, I’ve updated my dependencies to use the latest version of all packages.

### Are you sponsoring us?

- [X] Yes, I’m a sponsor. 💖

rfgamaral:
I came looking for this issue, and I just want to add a few things:

> Adding an horizontal rule with markdown shortcuts should not add any empty paragraphs before it.

This still happens, even after @philippkuehn commit above, and I agree that it shouldn't.

> What did you expect to happen?

IMO, an empty paragraph should be added **after** adding the horizontal rule with the Markdown shortcut, if it's the last node being inserted in the document, so that the cursor is placed after the horizontal rule, and the user can continue typing.

---

@cadars I'm wondering if you ended up working on a fixed version of this extension on your side? If so, do you think you could share that?

autumninflab:
@bdbch 
Hi, :)
Is this issue will fix soon?
I have similar issue.

1. typing `---`, a horizontal rule created.
2. typing any character (without line break)
3. horizontal rule is disappeared.

I guess it's because cursor still on the horizontal rule.
so the cursor should place after horizontal rule.

[...]

Issue #1665, opened 2021-07-30
Horizontal Rule adds empty line


wmonecke:
### **Description**

When adding horizontal rule, you get a blank new line on top of the `<hr/>` tag.
I even used the following to make sure no `<p/>` would be rendered above with no avail.

`````
const CustomHorizontalRule = HorizontalRule.extend({
  draggable: true,
  renderHTML({ HTMLAttributes }) {
    return ["hr", HTMLAttributes, 0]
  },
})
`````
https://user-images.githubusercontent.com/24256469/127713605-31254229-2af6-47e4-96ed-2378baf1cd72.mov

### **Question**
In other words: Is it possible to delete certain nodes programmatically?
In my case finding the last empty `<p/>` and deleting it after inserting a `<hr/>`

### **Solution**

I dont know if this is the right way to do it, but I am just listening every transaction and dipatching a prosemirror action to delete the node.

`````

onTransaction(event) {
      const { transaction, editor } = event
      if (transaction.steps.length > 0) {
        const stepOfInterest = transaction.steps[0]
        const arrayOfNodes = stepOfInterest.slice.content.content
        const foundHrNode = arrayOfNodes.find(
          (el) => el.type.name === "horizontalRule"
        )

        if (foundHrNode) {
          const tr = editor.view.state.tr
          tr.delete(stepOfInterest.from - 1, stepOfInterest.to - 1)
          editor.view.dispatch(tr)
        }
      }
    },
`````

HeebeLee:
I think this is a personal demand, not the general public. Deleting the <p> is not an ideal way if someone wants to keep one more line blank on the horizontal line.

And why delete a <p>? Instead of all the blank lines?

wmonecke:
@HeebeLee To me it feels more intuitive and in line with tip tap (unopinionated).

cadars:
@philippkuehn Looks like this issue is still here…
To reproduce, go to https://tiptap.dev/examples/default and create an `<hr>`.

PR #3859, opened 2023-03-17
fix: replace the whole node in nodeInputRule

ngryman:
Hi,

This PR fixes some issues with the horizontal rule:

* https://github.com/ueberdosis/tiptap/issues/3809
* https://github.com/ueberdosis/tiptap/issues/1964
* https://github.com/ueberdosis/tiptap/issues/1665

The new behavior creates a new paragraph below the horizontal rule, instead of above.

## Implementation

I changed the implementation of `nodeInputRule` for the case when there's no group matched in the input rule regexp.

As far as I understand, the previous implementation was trying to replace the content of the existing node, not the node itself. We need to include the node boundaries as well to do so: `start - 1` and `end + 1`.

The new implementation first inserts the new node at the existing node position (`start - 1`) and then delete the remaining content. Let me know if that makes sense or if there's a better approach.

Grepping the codebase quickly, it seems that other usages of `nodeInputRule` always use a matching group, so this change should only affect the `HorizontalRule` extension.

### DIFF
--- a/packages/core/src/inputRules/nodeInputRule.ts
+++ b/packages/core/src/inputRules/nodeInputRule.ts
@@ -43,7 +43,10 @@ export function nodeInputRule(config: {
         // insert node from input rule
         tr.replaceWith(matchStart, end, config.type.create(attributes))
       } else if (match[0]) {
-        tr.replaceWith(start, end, config.type.create(attributes))
+        tr.insert(start - 1, config.type.create(attributes)).delete(
+          tr.mapping.map(start),
+          tr.mapping.map(end),
+        )
       }
     },
   })


chrisdeeming:
Anyone wishing to shim this to fix the `HorizontalRule` issue can use this extension:

```javascript
const ExtHorizontalRule = HorizontalRule.extend({
	addInputRules() {
		return [
			new InputRule({
				find: /^(?:---|—-|___\s|\*\*\*\s)$/,
				handler: ({ state, range, match }) => {
					const attributes = {};

					const { tr } = state
					const start = range.from
					let end = range.to

					tr.insert(start - 1, this.type.create(attributes)).delete(
						tr.mapping.map(start),
						tr.mapping.map(end),
					);
				}
			})
		]
	}
});
```

steven-tey:
@chrisdeeming you're a lifesaver – that worked! Thank you!

rfgamaral:
FYI, Tiptap v2.1 was meant to fix this issue (there was some work done in that regard [here](https://github.com/ueberdosis/tiptap/pull/4319)), but I'm finding that the change on this PR works much better, more details in [this comment](https://github.com/ueberdosis/tiptap/pull/4319#issuecomment-1681956601).

bdbch:
I merged this fix with develop on this new PR here where your commit is included @ngryman:
https://github.com/ueberdosis/tiptap/pull/4341

I'll close this PR in favor for the new one

PR #3862, opened 2023-03-20
 Fixing reoccurring issue #3331 and improving related PR #3533

KentoMoriwaki:
In this pull request, I addressed the reoccurring issue from https://github.com/ueberdosis/tiptap/pull/3331 which resurfaced due to PR #3533. I found that the solution in the comment https://github.com/ueberdosis/tiptap/pull/3533#pullrequestreview-1225920630 was appropriate, and it was implemented in https://github.com/ueberdosis/tiptap/pull/3781, but it needed some improvements.

I removed the remaining `queueMicrotask` for a more complete fix. To easily reproduce the issue, I added an example using the Paragraph node view, which helped me confirm that the problem is resolved before and after this commit.

I observed the flushSync warning by removing the line https://github.com/ueberdosis/tiptap/pull/3781/files#diff-c1d83c0fa32f3b18f2d8e5b3eb81f9d951c0d6de903191d495d169d248188ceeR116, , further verifying that the original fix was correct.

### DIFF
--- a/demos/src/Examples/CustomParagraph/React/index.spec.js
+++ b/demos/src/Examples/CustomParagraph/React/index.spec.js
@@ -1,0 +1,23 @@
+context('/src/Examples/CustomParagraph/React/', () => {
+  beforeEach(() => {
+    cy.visit('/src/Examples/CustomParagraph/React/')
+  })
+
+  it('should have a working tiptap instance', () => {
+    cy.get('.ProseMirror').then(([{ editor }]) => {
+      // eslint-disable-next-line
+      expect(editor).to.not.be.null
+    })
+  })
+
+  it('should have a paragraph and text length', () => {
+    cy.get('.ProseMirror p').should('exist').should('have.text', 'Each line shows the number of characters in the paragraph.')
+    cy.get('.ProseMirror .label').should('exist').should('have.text', '58')
+  })
+
+  it('should have new paragraph', () => {
+    cy.get('.ProseMirror').type('{selectall}{moveToEnd}{enter}')
+    cy.get('.ProseMirror p').eq(1).should('exist').should('have.text', '')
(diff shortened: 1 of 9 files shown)


demos/src/Examples/CustomParagraph/React/index.spec.js
```diff
@@ -0,0 +1,23 @@
...
+  it('should have a paragraph and text length', () => {
+    cy.get('.ProseMirror p').should('exist').should('have.text', 'Each line shows the number of characters in the paragraph.')
+    cy.get('.ProseMirror .label').should('exist').should('have.text', '58')
+  })
+
+  it('should have new paragraph', () => {
+    cy.get('.ProseMirror').type('{selectall}{moveToEnd}{enter}')
+    cy.get('.ProseMirror p').eq(1).should('exist').should('have.text', '')
+    cy.get('.ProseMirror .label').eq(1).should('exist').should('have.text', '0')
+  })
```
KentoMoriwaki, line 22: I tried to test the behavior addressed in this PR, but I could not reproduce in Cypress environment. 
But I think this example is still helpful to find this problem easily.

packages/react/src/EditorContent.tsx
```diff
@@ -75,9 +75,7 @@ export class PureEditorContent extends React.Component<EditorContentProps, Edito
     // lifecycle methods, and React doesn't allow calling flushSync from inside
     // a lifecycle method.
     if (this.initialized) {
-      queueMicrotask(() => {
```
bdbch, line 78: Does this still prevent the initial issue we had before all those changes regarding the cursor position / rendering?

Our Netlify build didn't work for this so I can't test it right now. I'll check it out on my machine in the meantime to test your changes.

bdbch (approved):
Just checked it out locally. Looks fine to me. @svenadlung ?

Issue #2827, opened 2022-05-25
`CollaborationCursor` history not working together with `BubbleMenu` component


ArminasTelesoftas:
### What’s the bug you are facing?

`CollaborationCursor` extension comes in with its own history and its working as expected until I've added `<BubbleMenu />` component. 

I tried dropping `<BubbleMenu />` component and used `BubbleMenu` extension from `@tiptap/extension-bubble-menu` instead and history started working as expected again. But I would prefer to use `<BubbleMenu />` component.

Worth mentioning that I've disabled `history` extension from `StarterKit`, otherwise it produces following warning:
```
index.js:27 [tiptap warn]: "@tiptap/extension-collaboration" comes with its own history support and is not compatible with "@tiptap/extension-history".
```

### Which browser was this experienced in? Are any special extensions installed?

Chrome, Firefox. No special extensions.

### How can we reproduce the bug on our side?

Try using history options `editor.chain().focus().undo().run()` or `editor.chain().focus().redo().run()` from `CollaborationCursor` together with  `<BubbleMenu />` component. History should not work.

Comment out `<BubbleMenu />` and history should work again.

### Can you provide a CodeSandbox?

https://codesandbox.io/s/sad-chaum-ti1296?file=/src/_app.js

### What did you expect to happen?

I have expected that history functions should work as described.

### Did you update your dependencies?

- [X] Yes, I’ve updated my dependencies to use the latest version of all packages.

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

ArminasTelesoftas:
Just found [related issue ](https://github.com/ueberdosis/tiptap/issues/2761)which is more detailed

Issue #2761, opened 2022-05-06
registerPlugin() will break UndoManger in Collaboration Extension


Alecyrus:
### What’s the bug you are facing?

`Collaboration` Extension cannot not work well with `BubbleMenu`, which is registered by using `registerPlugin()` after editori s initialized.

Specifically, once you call `registerPlugin()` after editor is initialized, `UndoManger` does not work.
```javascript

import { Editor, EditorContent } from '@tiptap/vue-3';
import Collaboration from '@tiptap/extension-collaboration';
import StarterKit from '@tiptap/starter-kit';
import CollaborationCursor from '@tiptap/extension-collaboration-cursor';
import * as Y from 'yjs';
import { WebrtcProvider } from 'y-webrtc';
import Placeholder from '@tiptap/extension-placeholder';
import { Plugin } from 'prosemirror-state';

const editor = shallowRef<Editor>();

const ydoc = new Y.Doc();

const provider = new WebrtcProvider(
  'tiptap-collaboration-cursor-extension',
  ydoc,
);

editor.value = new Editor({
  extensions: [
    StarterKit.configure({
      // The Collaboration extension comes with its own history handling
      history: false,
    }),
    Collaboration.configure({
      document: ydoc
    }),
    CollaborationCursor.configure({
      provider: provider,
      user: {
        name: 'Cyndi Lauper',
        color: '#f783ac',
      },
    }),
    Placeholder.configure({
      placeholder:
        'Write something … It’ll be shared with everyone else looking at this example.',
    }),
  ],
});
// comment this line to make undoManager work
editor.value.registerPlugin(new Plugin({}));
</script>
```

If you comment last line, `UndoManger` is back to work.

### Can you provide a CodeSandbox?

[Reproduction Link](https://codesandbox.io/s/kind-dew-urdc6t?file=/src/App.vue)

### Did you update your dependencies?

- [X] Yes, I’ve updated my dependencies to use the latest version of all packages.

### Are you sponsoring us?

- [X] Yes, I’m a sponsor. 💖

Alecyrus:
@hanspagel  @philippkuehn  🙏🙏🙏

Alecyrus:
https://github.com/yjs/y-prosemirror/issues/114

Alecyrus:
There is a workaround (make sure yUndoPlugin() is the last plugin registered ):

```javascript
// collaboration.ts
addProseMirrorPlugins() {
    const fragment = this.options.fragment
[...]

PR #3901, opened 2023-03-28
Collaboration: Fix history after late-registering plugins

svenadlung:
UndoManger (collaboration extension, `y-prosemirror`) fails if a ProseMirror plugin is registered after initialization (using `registerPlugin()`), as in any menu (FloatingMenu, BubbleMenu).

See 
- https://github.com/ueberdosis/tiptap/issues/2761
- https://github.com/ueberdosis/tiptap/issues/2827


SK-Runner:
Which of the following versions can be used to view the effect when tiptap is introduced?

PR #3934, opened 2023-04-03
Hotfix/fix event binding react

### LINKED ISSUE #3932: [Bug]: The callback registered using `editor.on("update", callback)` is never getting called when `onUpdate()` is defined

core, react

2.0.1

If both `onUpdate: callback1()` and `editor.on("update", callback2)` are specified, only the first gets invoked always whereas the latter never gets called.

Both used to work at the same time in v2.0.0-beta.220, and now it has stopped working as intended in the latest v2.0.1.

Chrome

### Code example

https://codesandbox.io/s/silly-lalande-82577w?file=/src/App.js

In console, it should print both `in onUpdate()` and `in on("update")` when changing the content in the editor.

It prints only `in onUpdate()` right now. If you comment on `onUpdate()` callback, it should start printing `in on("update")` to the console.

[This recent change](https://github.com/ueberdosis/tiptap/pull/3811/files#diff-13acb5e551812e195c1140c10ba5ac298a0005996d07756cfd77c2d4194cb350R70-R73) might have introduced this regression.

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

bdbch:

This PR fixes the issue described in #3932 by unbinding specific event handlers instead of **all** event handlers in a hook inside `useEditor`.

1. Save the event handlers into refs to compare them afterwards for unbinding
2. Whenever a changed event handler is registered, unbind the old one, register the new one
3. Save the new one was the old one for next check

I created a local demo to test this behavior. It didn't create to many callbacks & correctly unregistered old ones. Also events registered via `Editor.on` were working correctly.

Create a local React demo and register a callback via `editor.on`.

## Remarks

Nothing to remark

## Checklist

- [x] Followed the guidelines
- [x] Fixed linting issues

## Related issues

- closes #3932


bdbch:
I worked directly from `develop`, dummy. :) I'll reopen one with the origin coming from `main`.

PR #3966, opened 2023-04-17
VueNodeViewRenderer should return `null` for `contentDOM` for a non-leaf node with no `NodeViewContent`

### LINKED ISSUE #3937: [Bug]: `VueNodeViewRenderer` with a NodeView that doesn't have a static `NodeViewContent` will result in overriding the entire DOM with the children

vue-3

2.0.2

This line:
https://github.com/ueberdosis/tiptap/blob/ae10f0d58ee1edb16a8ebe34e44d6a33e8361374/packages/vue-3/src/VueNodeViewRenderer.ts#L134

Results in a `Node` that has `NodeView` that has children but no statically defined `NodeViewContent` component will result in the entire node being thrashed with it's children node, often resulting in messed up rendering.

It also doesn't react if the `NodeViewContent` is rendered dynamically and changes as the component changes.

prosemirror might also know how to handle when `contentDOM` is `null` and there are children. Allowing the `NodeView` to render the children manually. Which can't be done due to this line.

Chrome

### Code example

_No response_

`contentDOM` to be `null` if there is no `NodeViewContent`. For it to react when the `NodeViewContent` node is dynamic. e.g. Async components, `v-if`, etc.

_No response_

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

segevfiner:

For a non-leaf node with no `NodeViewContent` return `null` instead of `this.dom`, this allows the node view itself to handle rendering the child content, instead of overriding the entire node with its child contents, not allowing to render it at all.

I changed a line of code.

In my own project.

Create a node view that has child `content`, but no `NodeViewContent`.

## Remarks

## Checklist

- [x] Followed the guidelines
- [x] Fixed linting issues

## Related issues

Fixes #3937

### DIFF
--- a/.changeset/shy-clouds-smoke.md
+++ b/.changeset/shy-clouds-smoke.md
@@ -1,0 +1,5 @@
+---
+"@tiptap/vue-3": patch
+---
+
+VueNodeViewRenderer should return `null` for `contentDOM` for a non-leaf node with no `NodeViewContent`
--- a/packages/vue-3/src/VueNodeViewRenderer.ts
+++ b/packages/vue-3/src/VueNodeViewRenderer.ts
@@ -136,9 +136,7 @@ class VueNodeView extends NodeView<Component, Editor, VueNodeViewRendererOptions
       return null
     }
 
-    const contentElement = this.dom.querySelector('[data-node-view-content]')
-
-    return (contentElement || this.dom) as HTMLElement | null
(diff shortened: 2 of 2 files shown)


RealAlphabet:
This is also necessary in my use case. I found out about this PR by stopping at the same line in the code that was modified by the OP. If this PR were merged, how could a node view modify its own content? Do you know if there's a Tiptap or ProseMirror API for this?

segevfiner:
Why was this closed?

PR #3975, opened 2023-04-19
fix(extension-link): fix link not being kept when pasting url with link

bdbch:

This PR fixes an issue where for example a link gets copied from another richtext (for example Google Docs). When the copied text is a URL (for example `https://github.com/`) but this URL text is already a link (for example `https://github.com/ == LINKS TO ==> https://github.com/my/repo`) the link gets dismissed and the text gets pasted as a URL.

I removed the default pasteHandler as the paste handler will always trigger before any custom logic. The paster handler itself also doesn't know about any marks of the pasted content but only the content text so it wasn't possible to handle the check if a pasted content is already a link in there.

I moved the logic into it's own plugin that will handle said behavior.

I created a Google Docs file with two lines. The document looked like this:

- [tiptap.dev](https://tiptap.dev/hello)
- https://tiptap.dev (this one is not auto linked in Google Docs)

When copying the first line, the pasted content was:

```
tiptap.dev (linking correctly to https://tiptap.dev/hello as defined in the Google Doc)
```

When copying the second line, the pasted content was:

```
https://tiptap.dev (linking to https://tiptap.dev - this was done via the autolink paste handler)
```

Clone and checkout this branch, then create a demo (or use the existing Marks/Link demo) and paste in content as I described above.

## Remarks

Nothing to remark here

## Checklist

- [x] Followed the guidelines
- [x] Fixed linting issues

## Related issues

- No relating issues

### DIFF
--- a/packages/extension-link/src/helpers/linkifyPaste.ts
+++ b/packages/extension-link/src/helpers/linkifyPaste.ts
@@ -1,0 +1,49 @@
+import { Editor } from '@tiptap/core'
+import { MarkType } from '@tiptap/pm/model'
+import { Plugin, PluginKey } from '@tiptap/pm/state'
+import { find } from 'linkifyjs'
+
+export type LinkifyPastePluginOptions = {
+  editor: Editor
+  type: MarkType
+}
+
+export const LinkifyPastePlugin = ({ editor, type }: LinkifyPastePluginOptions) => {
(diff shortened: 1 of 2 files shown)


svenadlung (approved):
Was testing some cases. Worked out well for me. Good work 🙌

Issue #2892, opened 2022-06-17
Extending a Table with a NodeViewContent inserts a "div" tag where the "tbody" should be


githubbob42:
### What’s the bug you are facing?

I'm trying to make a draggable table by wrapping the `table` with a drag handle  (https://tiptap.dev/guide/node-views/examples#drag-handles and https://tiptap.dev/guide/node-views/react#adding-a-content-editable).  I'm probably doing this all wrong but I can't seem to find the proper way to do this.  Or even to wrap a `Table` with another element.

It appears that by default,  `NodeViewContent` creates two `div` tags that are nested but by using `as='table'` it creates the outer as a `table`.  Unfortunately, according to the source code, it creates a `contentDOMElement` node inside that is only ever a `span` or a `div`.  It correctly creates the `tableRow` nodes inside of that though.  

Granted a `table` is created and it does render on the page and it is editable but a number of the features that we need, and that make the `Table` so powerful, are gone: E.g. column resizing, cell selecting, cell merging/splitting, etc.

This is what get's rendered:
```html
<div class="react-renderer node-table">
  <div class="draggable-table" intent="layoutTable" data-node-view-wrapper="" style="white-space: normal;">
    <table class="content" as="table" data-node-view-content="" style="white-space: pre-wrap; min-width: 314px;">
      <div style="white-space: inherit;">
        <tr>
          <th colspan="1" rowspan="1" colwidth="264">
            <p><br class="ProseMirror-trailingBreak"></p>
          </th>
          <th colspan="1" rowspan="1">
            <p><br class="ProseMirror-trailingBreak"></p>
          </th>
          <th colspan="1" rowspan="1">
            <p><br class="ProseMirror-trailingBreak"></p>
          </th>
        </tr>
        <tr>
          <td colspan="1" rowspan="1" colwidth="264">
            <p><br class="ProseMirror-trailingBreak"></p>
          </td>
          <td colspan="1" rowspan="1">
            <p><br class="ProseMirror-trailingBreak"></p>
          </td>
          <td colspan="1" rowspan="1">
            <p><br class="ProseMirror-trailingBreak"></p>
          </td>
        </tr>
        <tr>
          <td colspan="1" rowspan="1" colwidth="264">
            <p><br class="ProseMirror-trailingBreak"></p>
          </td>
          <td colspan="1" rowspan="1">
            <p><br class="ProseMirror-trailingBreak"></p>
          </td>
          <td colspan="1" rowspan="1">
[...]

Issue #2675, opened 2022-04-03
NodeViewContent generates an unnecessary div


anton-liubushkin:
### What’s the bug you are facing?

I'm using React and working on a custom extension for tables and encountered that the NodeViewContent component always creates a nested div element.

![Screen Shot 2022-04-04 at 12 54 24 AM](https://user-images.githubusercontent.com/5649032/161450336-4e078246-245f-44e2-932c-98a6a5294c5c.png)

↓

![Screen Shot 2022-04-04 at 12 55 04 AM](https://user-images.githubusercontent.com/5649032/161450352-dea65c6b-8c3d-4ada-9296-0aaf9e60a171.png)

### How can we reproduce the bug on our side?

1. [Create a node view with React](https://tiptap.dev/guide/node-views/react) or add the Node view to a table extension
3. Use `NodeViewContent` to render child content

```
export const TableComponent: FC<NodeViewProps> = (props) => {
 /* Some sort of code */
  return (
    <table>
      <NodeViewContent as="tbody" />
    </table>
  )
}
```

### What did you expect to happen?

I would like to avoid unnecessary elements in the DOM.

### Did you update your dependencies?

- [X] Yes, I’ve updated my dependencies to use the latest version of all packages.

### Are you sponsoring us?

- [X] Yes, I’m a sponsor. 💖

philippkuehn:
Duplicate of https://github.com/ueberdosis/tiptap/issues/1527

anton-liubushkin:
@philippkuehn, if I understand correctly, there is no way to get rid of the internal div element?

`<NodeViewContent as="table">`
converts into
```javascript
<table as="table" data-node-view-content="" style="white-space: pre-wrap;">
  <div style="white-space: inherit;"> // <-- This ugly wrapper
    /* content */
  </div>
</table>
```

henrykuzmick:
@philippkuehn I don't think this should be closed? #2213 doesn't seem to fix the unnecessary div problem.

This causes a number of issues.

Let's say I want to add some controls around a table. I could extend the Table plugin and add `addNodeView`.

Inside of the component used I'd render some control handles and `<NodeViewContent as="table" />` in the middle.

This in turn breaks the structure of the table and most of the commands from the Table extension.

sradu:
I'm running into too. Any chance this will get merged soon?

levi-putna:
Beyond tables, this also creates issues when you want to use a CSS grid layout on the children.

[...]

PR #3984, opened 2023-04-25
feature: allow to customize the tag of the contentDOMElement

nikgraf:

Added the option `contentDOMElementTag` to the `NodeViewRendererOptions` to allow users to provide as the desired element tag they would like to render.

This allows developers to solve the issue with injected divs inside a `table` when extending the table extension:
- https://github.com/ueberdosis/tiptap/issues/2675
- https://github.com/ueberdosis/tiptap/issues/2892

Added the option `contentDOMElementTag`

In `demos/src/GuideNodeViews/ReactComponentContent/React/Extension.js` we tested it using:

```js
return ReactNodeViewRenderer(Component, { contentDOMElementTag: 'main' })
```

You can see it in action here where we applied the patch via patch-packages: https://github.com/serenity-kit/Serenity/blob/68d7bb29979abf9833116c81f2db27156724a93d/packages/editor/extensions/tableExtension/tableExtension.ts#L15

Reproduce the test?

## Checklist

- [x] Followed the guidelines
- [x] Fixed linting issues

## Related issues

- https://github.com/ueberdosis/tiptap/issues/2675
- https://github.com/ueberdosis/tiptap/issues/2892

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -170,6 +170,7 @@ export interface NodeViewRendererOptions {
   ignoreMutation:
     | ((props: { mutation: MutationRecord | { type: 'selection'; target: Element } }) => boolean)
     | null
+  contentDOMElementTag: string
 }
 
 export type NodeViewRendererProps = {
--- a/packages/react/src/ReactNodeViewRenderer.tsx
+++ b/packages/react/src/ReactNodeViewRenderer.tsx
@@ -80,9 +80,13 @@ class ReactNodeView extends NodeView<
 
     ReactNodeViewProvider.displayName = 'ReactNodeView'
 
-    this.contentDOMElement = this.node.isLeaf
-      ? null
-      : document.createElement(this.node.isInline ? 'span' : 'div')
+    if (this.node.isLeaf) {
+      this.contentDOMElement = null
+    } else if (this.options.contentDOMElementTag) {
+      this.contentDOMElement = document.createElement(this.options.contentDOMElementTag)
+    } else {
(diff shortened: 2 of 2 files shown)


sradu:
Running into this issue too.

nikgraf:
@sradu sorry, forgot to fix the description. I now did.

History: Initially we wanted to have `preventContentDOMElement` but that led to more and more changes until we stopped following this path and instead replaced it with allowing to define `contentDOMElementTag`. Minimal change, but fixed our issue.

sradu:
Circling back here. The issue is that if the parent element is a grid, and the child nodes are columns, this still breaks.

nikgraf:
@sradu do you have an example? or even better a solution how to solve both cases?

sradu:
@nikgraf my very broken fix is:

```
.node-TYPE [data-node-view-content] > div {
  @apply contents;
}
```

What is the purpose of that extra div though? Why is it needed?

nikgraf:
@sradu interesting trick, need to give it a try.

Initially we tried to remove the div, but realized this is a larger change since plenty of other places reference the div. So this was the lowest impact change to achieve the goal. I hoped this would make it easier to get it merged.

bdbch (approved):
LGTM. I'll add docs for this as I'm at it.

bdbch:
@svenadlung you also want to take another look?

svenadlung (approved):

nikgraf:
awesome thanks 🙌

bdbch:
I'll include this for 2.2.0-rc.1

PR #3998, opened 2023-04-30
Retain existing config when calling configure on Marks and Extensions (attempt 2)

nholden:
*This is a second attempt at addressing https://github.com/ueberdosis/tiptap/issues/3813. My first attempt, https://github.com/ueberdosis/tiptap/pull/3822, had to be reverted because of issues described in [this comment](https://github.com/ueberdosis/tiptap/pull/3822#issuecomment-1523489076). That comment described a defaultOptions warning that I was able to reproduce in my original approach, and I confirmed it's no longer present in this approach. cc @bdbch*

Currently, when you call `.configure()` on a `Mark` or `Extension`, the `Mark` or `Extension` that's returned doesn't retain the config from the original.

This is a problem for StarterKit. When `Code` is added to an editor added as a standalone extension, [an inline code block is `exitable`](https://github.com/ueberdosis/tiptap/blob/43970fd496ffa22eae73ce95a11ed71d6efa4c03/packages/extension-code/src/code.ts#L47). When added as part of StarterKit, [an inline code block is not `exitable`](https://github.com/ueberdosis/tiptap/issues/3813) because [StarterKit calls `Code.configure()`](https://github.com/ueberdosis/tiptap/blob/43970fd496ffa22eae73ce95a11ed71d6efa4c03/packages/starter-kit/src/starter-kit.ts#L61), blowing away the default config.

This PR ensures that when you call `.configure()` on a `Mark` or `Extension`, the `Mark` or `Extension` that's returned retains the config from the original.

closes https://github.com/ueberdosis/tiptap/issues/3813


bdbch:
Hey @nholden - sorry for not reviewing this PR earlier. This looks promising – at least I didn't encounter the issues we experienced with the first draft. Do you mind we merge this and bring it into `2.2.0-rc.1` first so we can test it out with more people?

svenadlung (approved):

bdbch (approved):

nholden:
Sounds great. Thanks, @bdbch!

nperez0111:
This has already been resolved by https://github.com/ueberdosis/tiptap/pull/4191

PR #4001, opened 2023-04-30
Require file extensions for imports and exports

### LINKED ISSUE #3488: Type imports broken in TypeScript's NodeNext mode (ESM support)

This used to work in `beta.199`:

```ts
import type { JSONContent } from "@tiptap/core";
```

Now (`beta.204`) it shows the `Module has no exported member` error.

This could be due to this [TS issue](https://github.com/microsoft/TypeScript/issues/50482), meaning the `index.d.ts` exported by tiptap is using expressions like `export * from './CommandManager';`, while it should be `export * from './CommandManager.js';`.

I'm using:
-  ESM, i.e. `"type": "module"` in my `package.json`.
- TS version `4.9.3`.
- TS `NodeNext` mode, as supposed to be used for ESM projects.

This happens in TS.

1. Create an ESM TS project with `"type": "module"` in your `package.json` and `"module": "NodeNext"` in your `tsconfig.json`.
2. Write `import type { JSONContent } from "@tiptap/core";`
3. Run `tsc`
4. See the error.

No, that's in TS.

TS should resolve the type.

_No response_

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

benasher44:

Updates all imports and exports to include file extensions.

Careful find/replace, eslint, and iterations of building and fixing issues.

Built the project and demos, tested the changes with our main project, and tried a few examples in the deployed demos.

I'm not really sure how you all test something like this, but it sounded like from a comment in #3488 there was some method in mind. Let me know how I can help here.

## Remarks

- Updated an eslint rule to ensure these changes will stick going forward
- Used no-restricted-imports to specifically target '..' and '.' imports, which weren't covered by `import/extensions` for some reason.
- Generally speaking, I think this should work fine with CJS and improve compatibility for TS folks migrating to ESM. I'm less familiar with other supported packaging/module models.
- In cases where Vite/Rollup didn't like .js extensions for .ts files when building demos, I used .ts endings.

## Checklist

- [x] Followed the guidelines
- [x] Fixed linting issues

## Related issues

Closes #3488


benasher44:
In my latest update, I did a more comprehensive test build again our internal app and found a final issue that wasn't caught by the updated eslint rule. I filed it as a bug in import-js/eslint-plugin-import#2777 and added a no-restricted-imports rule to cover the imports from ".." and "." case

benasher44:
Hi @bdbch! Any thoughts here? prosemirror also recently published nodenext-compatible updates as well, so with this, we should be all set in that area.

benasher44:
Hi @bdbch. Is this solution okay, or were you looking for something different? Would love to get this merged. This is the only library we need to fix in order to migrate our node backend to ESM.

bdbch:
Hey @benasher44, sorry for the late update. I'll give your PR a try later this day and will merge if it works! Wanted to do this too earlier to bring in support for NodeNext.

Thanks already for contributing and keeping me in the loop

benasher44:
Amazing, thank you! I have updated develop. I'll watch CI to make sure it passes.

benasher44:
It looks like 1 workflow still has to be approved to run

benasher44:
Okay I added an extensionAlias to the webpack ts-loader config that cypress is using, so it should properly understand/resolve the .js endings now. I also updated with develop again. Mind approving the workflow run once more? 🤞

bdbch:
Thanks @benasher44! Let's see what the CI is doing - afterwards the PR looks good for me. I'll let @svenadlung also take a look.

.eslintrc.js
```diff
@@ -43,7 +43,21 @@ module.exports = {
         'no-console': ['warn', { allow: ['warn', 'error'] }],
         semi: ['error', 'never'],
         'import/order': 'off',
-        'import/extensions': 'off',
+        'import/extensions': ['error', 'ignorePackages'],
```
bdbch, line 46: Do you know if this is autofixeable?

bdbch (approved):

.eslintrc.js
benasher44: I don't think it is. The only way to get something like that is to convert all of the packages to use moduleResolution: NodeNext, and then TypeScript will suggest imports with .js endings from the start.

I can try that, if you all are open to it. It's a bigger change / commitment.

benasher44: I briefly tried locally, and it creates a bunch of TS type resolution issues that need to be resolved. The project builds, but there are lots of type issues that appear in the editor.

benasher44: If you attempt this after merging this, see ProseMirror/prosemirror#1305, where we discussed issues here in depth. In order to build with NodeNext, here's what I did to start:

1. Update main tsconfig.json moduleResolution to "NodeNext"
2. Override moduleResolution in tests/tsconfig.json to be "node", so cypress tests still build/run
3. Update all prosemirror packages to get the compatible dependencies

benasher44: @bdbch is auto-fixability a blocker?

bdbch: Not really. I just need another review from @svenadlung since we share PR responsibilities.

svenadlung (approved):

svenadlung:
Sorry for the delay on my part. Let's give it a try. Thank you for the contribution @benasher44  🙌

benasher44:
Amazing, thank you!

benasher44:
I did a quick pass on a handful of open PRs in the last week and left comments, but expect to maybe have to quickly follow up with fixes on develop, if people aren't up-to-date with develop

.eslintrc.js
benasher44: Heads up that there is now the `typescript.preferences.importModuleSpecifierEnding` for VSCode, which will make .js endings the default when auto-importing!

Issue #1451, opened 2021-06-11
TypeError: Cannot read property 'matchesNode' of null (TipTap for React)


chasinhues:
## Description
Same issue as in https://github.com/ueberdosis/tiptap/issues/438. TypeError that crashes my app.

## Steps to reproduce the bug
Happens on occasion when navigating to a page with a TipTap editor on it. 

## CodeSandbox
In the CSB below, run the tests, and observe the console spits out a bunch of these type errors for the browser. I don't know why it does this...

https://codesandbox.io/s/tiptap-prose-matchesnode-error-yhvdc

![image](https://user-images.githubusercontent.com/13565648/121638434-51d63a80-cace-11eb-87a2-698f0bfe4a3e.png)

## Expected behavior
No error.

## Environment?
 - operating system: OSX
 - browser: N/A
 - mobile/desktop: N/A
 - tiptap version: `@tiptap/core@^2.0.0-beta.65`

## Additional context
I tried to keep the CSB example above as close to my current code as possible, minus other third-party libraries that would unnecessarily bloat the example. 

I am able to replicate the error consistently in my tests:

```
    Error: Uncaught [TypeError: Cannot read property 'matchesNode' of null]
        at reportException (/Users/jasonhughes/sites/innovator-portal/innovator/client/node_modules/jsdom/lib/jsdom/living/helpers/runtime-script-errors.js:62:24)
        at Timeout.task [as _onTimeout] (/Users/jasonhughes/sites/innovator-portal/innovator/client/node_modules/jsdom/lib/jsdom/browser/Window.js:521:9)
        at listOnTimeout (internal/timers.js:549:17)
        at processTimers (internal/timers.js:492:7) TypeError: Cannot read property 'matchesNode' of null
        at EditorView.updateStateInner (/Users/jasonhughes/sites/innovator-portal/innovator/client/node_modules/prosemirror-view/src/index.js:141:45)
        at EditorView.updateState (/Users/jasonhughes/sites/innovator-portal/innovator/client/node_modules/prosemirror-view/src/index.js:114:10)
        at Editor.dispatchTransaction (/Users/jasonhughes/sites/innovator-portal/innovator/client/node_modules/@tiptap/core/src/Editor.ts:299:15)
        at EditorView.dispatch (/Users/jasonhughes/sites/innovator-portal/innovator/client/node_modules/prosemirror-view/src/index.js:374:50)
[...]

PR #4029, opened 2023-05-08
fix: Don't init when editor is destroyed

tomi-bigpi:

Adds a check to see if the `EditorContent` editor is already destroyed when `init` is called. This can happen due to using `React.StrictMode` where the editor is re-rendered back-to-back automatically by React.

Extended the existing check to include verifying that the editor hasn't already been destroyed.

Manual testing.

Unfortunately, I don't have a clean reproduction available at this time.

## Checklist

- [x] Followed the guidelines
- [x] Fixed linting issues

## Related issues

This issue is similar to the various problems described in https://github.com/ueberdosis/tiptap/issues/1451 (and other linked issues in that thread) and manifests itself with the same `TypeError: Cannot read property 'matchesNode' of null`.

### DIFF
--- a/packages/react/src/EditorContent.tsx
+++ b/packages/react/src/EditorContent.tsx
@@ -49,7 +49,7 @@ export class PureEditorContent extends React.Component<EditorContentProps, Edito
   init() {
     const { editor } = this.props
 
-    if (editor && editor.options.element) {
+    if (editor && !editor.isDestroyed && editor.options.element) {
       if (editor.contentComponent) {
         return
       }


alizand1992 (approved):

gl3nn:
We are currently seeing an error that might be fixed by this. It only occurs with strict mode on; and it is clear that the editor is destroyed in the stack trace. Did you find a temporary work around?

tomi-bigpi:
> We are currently seeing an error that might be fixed by this. It only occurs with strict mode on; and it is clear that the editor is destroyed in the stack trace. Did you find a temporary work around?

We're still patching this locally...

nperez0111 (approved):

PR #4068, opened 2023-05-23
Feature/wrapper class

patrickbaber:

We can be proud of what we have built. So let's show this to all devs in the world! I pretend the class name `tiptap` to the editor element 🎉

I added a new method in the Editor class to pretend `tiptap` in the list of classes. I have updated the docs and demos.

I ran the demos and looked for the class name in the wrapper element. The tests are also working as expected.

You can check the demos and search for the new class name.

## Checklist

- [  ] Added tests where possible
- [X] Followed the guidelines
- [X] Fixed linting issues

### DIFF
--- a/demos/setup/helper.ts
+++ b/demos/setup/helper.ts
@@ -29,7 +29,7 @@ export function debug() {
   // @ts-ignore
   sendData('source', window.source)
 
-  waitUntilElementExists('.ProseMirror', element => {
+  waitUntilElementExists('.tiptap', element => {
     // @ts-ignore
     const editor = element.editor
 
--- a/demos/setup/style.scss
+++ b/demos/setup/style.scss
@@ -73,7 +73,7 @@ select {
   }
 }
 
-.ProseMirror:focus {
+.tiptap:focus {
   outline: none;
 }
 
--- a/demos/src/Commands/InsertContent/Vue/index.vue
+++ b/demos/src/Commands/InsertContent/Vue/index.vue
@@ -94,7 +94,7 @@ export default {
 
 <style lang="scss">
 /* Basic editor styles */
-.ProseMirror {
+.tiptap {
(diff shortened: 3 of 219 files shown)


rfgamaral:
Should this be considered a breaking change? Not sure if it's documented anywhere but I'm thinking so, because `.ProseMirror` class could've been used to stylize certain elements that weren't possible any other way. Not to mention that other ProseMirror internals and/or plugins might introduce `.ProseMirror-*` classes, for instance `.ProseMirror-selectednode` and `.ProseMirror-trailingBreak`. We use this across our products to make sure our editors behave exactly the way we intend them too, and a change like this would be breaking for us. Food for thought.

svenadlung (approved):

rfgamaral:
@janthurau Thanks for clarifying.

patrickbaber:
For things like styling Tiptap it's a bit inconsistent to use `.ProseMirror` instead of `.tiptap`, but as @janthurau mentioned, the old class names of ProseMirror are still existing.

The more specific classes like `.ProseMirror-selectednode` are completly untouched.

rfgamaral:
> For things like styling Tiptap it's a bit inconsistent to use `.ProseMirror` instead of `.tiptap`

IMO, styling should be flexible enough where you can use whatever tooling your project requires without explicitly targetting global classes such as `.ProseMirror` or `.tiptap`. However, that's how we initially implemented things (no good reason) and we are now in the process of reviwing that and using our own custom mechanism, but we still have some legacy `.ProseMirror` usages here and there.

patrickbaber:
@rfgamaral You are right! I mean simpler setups like the ones mentioned here: https://github.com/ueberdosis/tiptap/blob/v2.0.3/docs/guide/styling.md#option-1-style-the-plain-html

PR #4088, opened 2023-05-29
issue: add test for mergeDeep fail

YousefED:
This test demonstrates there's a bug in the implementation of `mergeDeep`, which can cause other issues down the line


nperez0111 (approved):
Handled the issue with some simpler logic, thanks for pointing it out

Issue #4103, opened 2023-06-03
[Documentation]: Explain how to add image width & height attribute customly 


ashrafchowdury:
### What’s the URL to the page you’re sending feedback for?

https://tiptap.dev/api/nodes/image

### What part of the documentation needs improvement?

The documentation for the Image element does not explain how to **add custom width and height attributes**. This makes it difficult to control the size of images in the editor.

### What is helpful about that part?

The helpful part of the documentation is that it provides a clear example of how to add an image to the editor. However, it would be more helpful if it also **explained how to add custom width and height attributes**.

### What is hard to understand, missing or misleading?

The hard-to-understand, missing, or misleading part of the documentation is the lack of explanation for how to add custom width and height attributes. This makes it difficult to control the size of images in the editor.

ashrafchowdury:
I can fix this issue by adding how to add custom width and height on the documentation, would please assign me to fix this issue?

rezaffm:
this should do it (just copied and pasted from my extension):

```js
    addAttributes() {
        return {
            src: {
                default: null
            },
            width: {
              default: null
            },
            height: {
              default: null
            },
            alt: {
                default: null
            },
            title: {
                default: null
            },
            class: {
              default: null
            },
            srcset: {
              default: null
            },
            sizes: {
              default: null
            },
            loading: {
              default: null
            }
        };
    },
```

ashrafchowdury:
@rezaffm I thinking to add this to the documentation Image section so that people can understand easily how to add external attributes.

rezaffm:
Go for it if it helped you I am happy

bernardoenock:
This has not been added.
I hope to help, in case anyone is already looking for something like this.

I made a "workaround" to work in Typescript:

https://tiptap.dev/docs/editor/guide/custom-extensions#attributes
```typescript
type ImageAttributes = {
  src: string
  alt?: string
  title?: string
  width?: string | undefined
  height?: string | undefined
  style?: string
}

export const ExtendedImage = Image.extend({
  addAttributes() {
    return {
      src: {
        default: '',
      },
      alt: {
        default: undefined,
      },
      title: {
        default: undefined,
      },
      width: {
        default: undefined,
      },
      height: {
        default: undefined,
      },
      style: {
        default: undefined,
      },
    }
  }
})

  const addImage = () => {
    const url = window.prompt('URL')
    const width = window.prompt('Width')
    const height = window.prompt('Height')
[...]

PR #4113, opened 2023-06-07
documentation - Updated the image extension doc and added a section on img tag attributes.

ashrafchowdury:

#4103 I have updated the image extension doc and added a new section to show how to use other attributes on img tag, like width and height.

I accomplished these changes by first adding the new section to the doc. I then went through the doc and added the code for new attributes.

No

You can verify my changes by running the documentation image section.

## Checklist

- [x] Followed the guidelines
- [ ] Fixed linting issues

## Related issues

No

### DIFF
--- a/docs/api/nodes/image.md
+++ b/docs/api/nodes/image.md
@@ -55,6 +55,45 @@ Image.configure({
 })
 ```
 
+### Custom Attributes
+Custom HTML attributes can be configured to use on img tag.
+
+```js
+Image.configure({
+  addAttributes() {
+      return {
+        src: {
+            default: null
+        },
+        width: {
+          default: null
+        },
+        height: {
+          default: null
+        },
+        alt: {
+            default: null
+        },
+        title: {
+            default: null
+        },
+        class: {
+          default: null
+        },
+        srcset: {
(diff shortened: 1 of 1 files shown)


bdbch:
Hey @ashrafchowdury – first thanks for your PR!

I don't think it makes sense to just add custom attribute documentation to the Image node but not anything else – since we don't do the same for the other documentation sites.

Since the extend/config functionality is more of a "core functionality" of Tiptap I'd rather have a link to https://tiptap.dev/guide/custom-extensions/#extend-existing-attributes instead of explaining each attribute on it's own here.

What do you think?

cesar-ibr2:
I see value on adding this in the documentation about Images, for me it was hard to find something like this because I expected to find how to set `width` and `height` (very common attributes in images) in the section dedicated to Images.

bdbch:
@cesar-ibr2 I'm pretty sure we have this documented here: https://tiptap.dev/guide/custom-extensions/#attributes

Wouldn't it make sense to just link there instead of having this kind of duplicated on the image documentation page? Otherwise we'd had to add this to any extensions that *could* also have custom attributes added to it.

PR #2610, opened 2022-03-11
feat: Add a generic type for suggestion items

rfgamaral:
Provides a generic type for suggestion items, defaulting to `any` as before (no breaking change).

PR #4136, opened 2023-06-16
Fix TypeScript types for Suggestion `command`, allowing for generic override

sjdemartini:

As of https://github.com/ueberdosis/tiptap/pull/2610 (https://github.com/ueberdosis/tiptap/commit/7cae9673f0086973b4d31e1343375ed5ad04ee0a), new generics were added for Suggestion options and props. However, there is a subtle bug in the current typing: the object selected with the suggestion `command` need not have the same types as the `items` in the suggestion options. For instance, in Tiptap's official demo https://tiptap.dev/api/nodes/mention, the suggestion `items` are each `string`s, but the selected Mention is of type `{id: string;}` (for the attributes of the Mention node):

```ts
  const selectItem = index => {
    const item = props.items[index]

    if (item) {
      props.command({ id: item })
    }
  }
```

i.e., there should be no restriction that when you select something with the suggestion `command`, it must use the identical structure as the suggested items, as that results in a type error in the above perfectly functional code. When using the suggestion plugin within the Mention extension, for instance, the value passed to the SuggestionProps `props.command()` function must be a `Record<string, any>`, as it's directly/exclusively used to set the `attrs` of a `Node` via `insertContentAt` (and you need not use that identical shape for suggestion options themselves):

https://github.com/ueberdosis/tiptap/blob/44996d60bebd80f3dcc897909f59d83a0eff6337/packages/extension-mention/src/mention.ts#L42
    https://github.com/ueberdosis/tiptap/blob/f8695073968c5c6865ad8faf05351020abb2a3cc/packages/core/src/types.ts#L79

This fixes the type definitions so that suggestions can correctly refer separately to their own items (of `any` type), while ensuring the `command`ed item can be defined with its own type definition. This allows use with the Mention extension, where you can still require the correct object structure needed for mention node attributes. In addition, the `Mention` extension now has more specific typing so that it requires a configured `suggestion` to pass in objects of the expected structure (`id` and `label`).

Tested these types with my local version of suggestion/mention extensions, included an extension that `extend`s `Mention` and further overrides its attributes, and it resolved type errors that were otherwise present.

Use TypeScript with the original mention example from Tiptap.

## Remarks

There are two commits: the first is the simplest fix, which just sets the `command` argument to `any` (most basic way to resolve the type error). The second commit adds a generic so that the type can be specified and narrowed by callers, which the Mention extension now does by default as well.

## Checklist

- [x] Followed the guidelines
- [x] Fixed linting issues

## Related issues

n/a

### DIFF
--- a/packages/extension-mention/src/mention.ts
+++ b/packages/extension-mention/src/mention.ts
@@ -3,10 +3,24 @@ import { Node as ProseMirrorNode } from '@tiptap/pm/model'
 import { PluginKey } from '@tiptap/pm/state'
 import Suggestion, { SuggestionOptions } from '@tiptap/suggestion'
 
-export type MentionOptions = {
+// See `addAttributes` below
+export interface MentionNodeAttrs {
+  /**
+   * The identifier for the selected item that was mentioned, stored as a `data-id`
+   * attribute.
+   */
+  id: string | null;
+  /**
+   * The label to be rendered by the editor as the displayed text for this mentioned
+   * item, if provided. Stored as a `data-label` attribute. See `renderLabel`.
+   */
+  label?: string | null;
+}
+
+export type MentionOptions<SuggestionItem = any, Attrs extends Record<string, any> = MentionNodeAttrs> = {
   HTMLAttributes: Record<string, any>
-  renderLabel: (props: { options: MentionOptions; node: ProseMirrorNode }) => string
-  suggestion: Omit<SuggestionOptions, 'editor'>
+  renderLabel: (props: { options: MentionOptions<SuggestionItem, Attrs>; node: ProseMirrorNode }) => string
+  suggestion: Omit<SuggestionOptions<SuggestionItem, Attrs>, 'editor'>
 }
 
 export const MentionPluginKey = new PluginKey('mention')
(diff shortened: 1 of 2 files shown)


bdbch (approved):
LGTM - @svenadlung you want to take another look?

sjdemartini:
Thanks for the initial review! I have a complete example [here](https://github.com/sjdemartini/mui-tiptap/pull/127/files#diff-ade29e0573c5eb74da16256dc1ebb06a5ce334b8fe55753ac6a63566877f85d8R42-R61) which I think is helpful to illustrate the problem and how this PR solves it. (Notice the `@ts-expect-error` that's necessary to silence the TS error for `SuggestionProps["command"]`.)

devth:
Would love to see this merged!

DhenPadilla:
Would also love this in!

nperez0111 (dismissed):
Looks good to me too

bdbch (approved):
LGTM

bdbch:
@sjdemartini we have a few issues with the branches. Could you
1. Rebase your changes to the current main branch
2. Change the target branch to main?

sjdemartini:
@bdbch and @nperez0111 Thanks for following up on this! I've changed the target branch to `main`, and rebased and handled some conflicts. I tested this branch's updated version locally against https://github.com/sjdemartini/mui-tiptap (using https://github.com/wclr/yalc) and all worked well for me, and I was able to remove the `@ts-expect-error` in that project related to `props.command()` thanks to the updated types here. 

Let me know if there's anything else needed to merge.

nperez0111 (approved):
Happy to merge

PR #4138, opened 2023-06-19
fixed: check props.clientRect before creating ReactRenderer

### LINKED ISSUE #4137: [Documentation]: Commands
### What’s the URL to the page you’re sending feedback for?

https://tiptap.dev/experiments/commands

### What part of the documentation needs improvement?

This example has a problem with using the Chinese input method. When I use the Chinese input method, onStart in the suggestion.js file is executed, but there is no props.clientRect property.

 Although this case is already filtered using “if(props.clientRect)”, it should be filtered on the first line of the function, otherwise it will cause rendering-related errors。

The problem I encountered was that I used the element-plus infinite-scroll component in the CommandsList.vue file. When I entered the full active string in Chinese, no menu appeared. When I moved the cursor one bit to the left, an error occurred.

**Cannot destructure property 'container' of 'el[SCOPE]' as it is undefined.**

please move this part to the first line of onStart function！！！
`if (!props.clientRect) {
          return
}`

### What is helpful about that part?

Reduce bugs risk for Chinese users

### What is hard to understand, missing or misleading?

 
`if (!props.clientRect) {
          return
        }`
should run at the first line of onStart Function！！！

_No response_

HariAcharya80:

Moved validation to top of the `onStart` function.

Moved Validation:
```javascript
 if (!props.clientRect) {
          return
        }
```
to the first line of  `onStart`, see commits.

## Remarks

Reduce bugs risk for Chinese users

## Checklist

- [x] Followed the guidelines
- [x] Fixed linting issues

## Related issues

closes #4137


bdbch:
Thanks - good find.

bdbch:
Could you fix those linting issues? @HariAcharya80

HariAcharya80:
> Could you fix those issues we have with the PR? @HariAcharya80

Ohh sorry for not seeing this comment, Ok let me get that just few mins..

HariAcharya80:
@bdbch **Done!**

PR #4140, opened 2023-06-20
feat(core): NodePositions

bdbch:

This PR adds a new class to the core library called `NodePosition`. This is a wrapper around ProseMirror's `ResolvedPos` aiming to make handling content & content positions easier.

With this class, you could do stuff like this:

```js
const { $anchor } = editor.state.selection
const nodePos = new NodePosition($anchor)

// Delete the node before the currently anchored node
editor
  .chain()
  .deleteRange({ from: nodePos.before.from, to: nodePos.before.to })
  .focus()
  .run()
```

or

```js
const { $anchor } = editor.state.selection
const nodePos = new NodePosition($anchor)

// Add some text behind the current list
const bulletListParent = nodePos.getParentByType('bulletList')
if (!bulletListParent) {
  return false
}

editor
  .chain()
  .insertContentAt(bulletListParent.to + 1, 'Hello World!')
  .focus()
  .run()
```

The `NodePosition` class has the following values & getters:

- `NodePosition.node` - The node found at the current ResolvedPos
- `NodePosition.$pos` - The current resolved position
- `NodePosition.doc` - The parent document node
- `NodePosition.depth` - The depth of the current NodePosition
- `NodePosition.name` - The node typename found at the NodePosition
- `NodePosition.from` - The start position of the NodePosition
- `NodePosition.to` - The end position of the NodePosition
- `NodePosition.before` - The NodePosition before this NodePosition
- `NodePosition.after` - The NodePosition after this NodePosition
- `NodePosition.parent` - The parent NodePosition for this NodePosition
  - This could be used for handling nested elements
- `NodePosition.children` - Returns an array of direct children if there are any
- `NodePosition.firstChild` - Returns the first child if there is one
- `NodePosition.lastChild` - Returns the last child if there is one
- `NodePosition.range` - The range of this NodePosition
- `NodePosition.createNodeRange` - Creates a new NodeRange for this NodePosition
- `getParentByType(typeOrName)` - Finds a specific Node type as parent and returns it's NodePosition
- `hasParentByType(typeOrName)` - Checks if a specific Node type as parent exists
- `NodePosition.getChildrenOnDepth(targetDepth: number)` - Returns all children on the given depth of this NodePosition
- `NodePosition.getDeepChildren()` - Returns ALL children inside this NodePosition (including deep childs like `paragraph` and `text` nodes)

It also has two static functions called the same as the dynamic functions for parents.
I'm currently investigating how I can make the handling even better (for example also going through children, finding specific child NodePositions, etc.)

In the end something like this is now possible:

```js
// Get the position of the parent of this parents predecessor
const targetPosition = myNodePos.parent.before.parent.from
```

I'm working on this to make my own work on some of our issues like List behavior #3819 or other nested nodes.

Take a look at the class:
https://github.com/ueberdosis/tiptap/compare/feature/node-positions?expand=1#diff-2d825cb76dab95991ece6347a3abec26ba8185c10374d30329656622976126aa

I created a local demo page showing and tracking all NodePositions important to me with Decorations & I console logged them on update and selectionChange to make sure the values come out correctly.

1. Create a local demo from this branch
2. Create an editor
3. Inside the onUpdate and onSelectionChange hooks add this:

```js
const { $anchor } = editor.state.selection
console.log(new NodePosition($anchor))
```

4. Check out the console output and try different setups

## Remarks

This is still early – I'll also need to see how well it performs but since most of the more complex things only happen when the getters are called I think it's already pretty good.

## Checklist

- [x] Add documentation where needed
- [x] Followed the guidelines
- [x] Fixed linting issues


bdbch:
I also added a demo for this:
https://deploy-preview-4140--tiptap-embed.netlify.app/src/experiments/nodepositions/react/

Here you can see it in action (prev, before, current and parent nodes are highlighted - below you get an output of all the NodePositions).

Slapbox:
This looks very nifty, even if the case I'm about to ask about isn't supported.

I assume the following case would not be possible, right?

```
const targetPosition = myNodePos.parent.before.parent.parent.parent // Accessing parents of parents
```

If the above code would work, that would be a truly monumental quality of life improvement and make code dramatically more readable.

bdbch:
Theoretically that could work @Slapbox as those NodePositions should be able to resolve from ResolvedPos to ResolvedPos.

I could test it out later to see if there are some errors coming up. Would it be okay if in the worst case you end up with a `null` if it's not resolveable (for example in this chain a parent is not existing?)

bdbch:
Just tested it out and it seems to work. Just one bug where it's throwing an error when the position becomes NaN because of the depth being messed up there but I'll look into this.

Would you rather always receive `null` or the `doc` if the parent is not resolvable?

I tested it with a deeply nested list (around 8-9 levels deep) and I was able to find the correct parent with the method you asked for (Just need to keep in mind that you have to do 2x parent in a list since you always need to step up one `listItem` and one `bulletList`.

bdbch:
I also added the following getters and functions:

### Getters

- `NodePosition.children` - Returns an array of direct children if there are any
- `NodePosition.firstChild` - Returns the first child if there is one
- `NodePosition.lastChild` - Returns the last child if there is one

### Functions

- `NodePosition.getChildrenOnDepth(targetDepth: number)` - Returns all children on the given depth of this NodePosition
- `NodePosition.getDeepChildren()` - Returns ALL children inside this NodePosition (including deep childs like `paragraph` and `text` nodes)

bdbch:
More things to add:

- [ ] `NodePosition.isFirstNodePosition` - true if it's the first NodePosition in it's parent
- [ ] `NodePosition.isLastNodePosition` - true if it's the last NodePosition in it's parent
- [ ] `NodePosition.isEmpty` - true if no content is inside the NodePosition

Slapbox:
Thanks again for your work on TipTap @bdbch!

> Would it be okay if in the worst case you end up with a null if it's not resolveable (for example in this chain a parent is not existing?)

I think that would be a reasonable approach since that would most closely match ProseMirror's underlying behavior.

docs/api/node-position.md
```diff
@@ -0,0 +1,238 @@
...
+
+```js
+// either look up a single type
+nodePosition.getParentByType('bulletList')
+
+// or look up multiple types
+nodePosition.getParentByType(['bulletList', 'orderedList'])
+```
+
+#### `NodePosition.hasParentByType``
```
svenadlung, line 194: Remove last `

docs/api/node-position.md
```diff
@@ -0,0 +1,238 @@
...
+
+```js
+// make sure to import the NodePosition class
+import { NodePosition } from '@tiptap/core'
+
+// first lets get the current selection position
+// as a ResolvedPos
+const { $anchor } = editor.state.selection
+
+// Now lets create a NodePosition out of it
```
svenadlung, line 226: keep the lower case, as in the comments before?

docs/api/node-position.md
```diff
@@ -0,0 +1,238 @@
...
+const nodePosition = new NodePosition(nodePosition)
+
+// Lets find the closest parent list
```
svenadlung, line 229: keep the lower case, as in the comments before?

docs/api/node-position.md
```diff
@@ -0,0 +1,238 @@
...
+const parentList = nodePosition.getParentByType(['bulletList', 'orderedList'])
+
+if (!parentList) {
+  return null
+}
+
+// Lets find the next node after the parent list
```
svenadlung, line 236: keep the lower case, as in the comments before?

packages/core/src/helpers/index.ts
```diff
@@ -24,6 +24,8 @@ export * from './getNodeAtPosition'
 export * from './getNodeAttributes'
 export * from './getNodeType'
 export * from './getRenderedAttributes'
+export * from './getResolvedPosAfter'
+export * from './getResolvedPosBefore'
```
benasher44, line 28: heads up that these and others will need .js endings now

bdbch:
Closed in favor of #4701

PR #4141, opened 2023-06-20
New Cut Command & Position Mapping for insertContentAt

bdbch:

This PR adds two new functionalities:

1. A new `cut` command to move content from one place to another
2. Position mapping for `insertContentAt` to take positional changes on the transaction and maps it to the given position

See changed files

I used a local demo to test my changes

Create a local demo with a button that cuts content from one position to another and test it.

## Remarks

This may introduce a **breaking change** due to some people maybe doing the position mapping manually already. This is something for 2.1.0.

## Checklist

- [x] Followed the guidelines
- [x] Fixed linting issues


packages/core/src/commands/cut.ts
```diff
@@ -0,0 +1,25 @@
+import { RawCommands } from '../types'
```
benasher44, line 1: heads up that this needs a .js ending now
```suggestion
import { RawCommands } from '../types.js'
```

PR #4151, opened 2023-06-24
fix(core) Allow text style to be clearable on new lines

### LINKED ISSUE #3702: TextStyle cannot be cleared when cursor is on new line

When you have your cursor on a new line and **some text style is applied** (either from the Font Family or Color extensions), you cannot clear out the text styles when the cursor is on a new line.

Chrome Version 109.0.5414.75 (Official Build) (64-bit)

Extensions:
- Text Style,
- Font Family,
- Color

For either the font-family or color extensions, scroll down to the text editor demos:
https://tiptap.dev/api/extensions/font-family
https://tiptap.dev/api/extensions/color

1. Enter content in the editor, being sure to have either a font type selected or color active
2. Hit "Enter" to move the cursor to a new paragraph
3. Toggle off or unset the color or font type
4. Continue typing
5. **Bug:** the text styles are still applied

_No response_

The text style should be removed from your stored marks even if the cursor is at the beginning of a new block

gif97

I don't know why this issue only applies when the cursor is at the beginning of a new paragraph/block(?). It seems to be caused by weird interactions between `.setMark("textStyle", { myAttr: null })` and `.removeEmptyTextStyle()`.

Combining the two commands into a single custom transactional command seems to resolve this issue:
```typescript
  chain()
      .command(({ tr: subCmdTr, state, commands }) => {
          const oldAttributes = getMarkAttributes(state, type);

          subCmdTr.addStoredMark(
              type.create({
                  ...oldAttributes,
                  ...{ fontSize: null },
              })
          );

          const newAttributes = getMarkAttributes(
              // Note the state.apply here. This may be part of the issue
              state.apply(subCmdTr),
              type
          );
          const hasStyles = Object.entries(newAttributes).some(
              ([, value]) => !!value
          );

          if (hasStyles) {
              return true;
          }

          return commands.unsetMark("textStyle");
      })
      .run();

```

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

C-Hess:

When you have your cursor on a new line and some text style is applied (either from the Font Family or Color extensions), you cannot clear out the text styles when the cursor is on a new line.

It turns out there are two main factors that cause this behavior

### Factor 1
We begin with the unset command of an extension that extends text-style. For example: font-family

https://github.com/ueberdosis/tiptap/blob/ccf05b04e3a2e2ee36c5bdac4d3b1a88bfe7a71d/packages/extension-font-family/src/font-family.ts#LL64C1-L67C17

Note that it first calls setMark then calls removeEmptyTextStyle. It doesn't matter what the two commands do. Really all that is causing this issue is the fact that we are calling another command after `setMark` within a `chain()`

### Factor 2
Chained commands fail to persist stored marks. Look at this code snippet; it looks like in between chained commands, we are resetting the transaction's stored marks based on what is in state:

https://github.com/ueberdosis/tiptap/blob/ccf05b04e3a2e2ee36c5bdac4d3b1a88bfe7a71d/packages/core/src/CommandManager.ts#LL123C1-L125C6

As a result, the storedMarks fields cannot be updated if you run more than one chained command.

Removing the logic that resets the transaction stored marks for chained commands resolves the issue. I really would need to know what prompted the change to begin with. All I have to go off of is this commit that I'm struggling to find an associated PR/issue for:
https://github.com/ueberdosis/tiptap/commit/532eaa4e869180d639a696f2feb26c7ca6d334d5

Added automated tests around this behavior.

Follow issue reproduction's steps in the color demo

## Checklist

- [X] Added tests where possible
- [X] Followed the guidelines
- [X] Fixed linting issues

## Related issues

fixes #3702


C-Hess:
Hold: looks like I just figured out what that line does. Looks like without it, stored marks are lost when creating a list...

Looking at writing a test for that and trying to find a solution

C-Hess:
Closing for now. Will need to think of a different approach

C-Hess:
Okay false alarm, sorry for the mess here :)

>  Hold: looks like I just figured out what that line does. Looks like without it, stored marks are lost when creating a list...
Looking at writing a test for that and trying to find a solution

I was wrong about this, this behavior never existed. I can't find anything different before and after my changes

packages/core/src/CommandManager.ts
```diff
@@ -120,10 +120,6 @@ export class CommandManager {
     const { rawCommands, editor, state } = this
     const { view } = editor
 
-    if (state.storedMarks) {
```
bdbch, line 123: https://prosemirror.net/docs/ref/#state.EditorState.storedMarks

From the context of the file it seems like this is used when chains are created / can commands are created.

From the Prosemirror doc it seems like this is used to save marks between transactions and reapply them after a transaction.

I could guess there are some edge cases where this could happen. Do you think your commit would break this behavior? Maybe it's not even that bad since this is only relevant for chain commands anyway?

bdbch (commented):
I'm not sure if I can approve this. @svenadlung you want to take a look too? I could imagine this being at least a breaking change for some users who do work with styles being kept on new lines

bdbch (approved):

PR #4162, opened 2023-06-26
Update markInputRule.ts

mcshaz:

trivial change - the `markEnd` variable was assigned outside the `if` block and then reassigned within the block without ever using the initial value. Using const assignment just above the first use of the variable improves readability.

search through the file for any other occurences of `markEnd` - there are none

N/A trivial change to help people trying to read and understand what the code and variables are achieving

look at the tiny changes to the file, verify there are no other occurences of markEnd with a simple text search of the file

## Remarks

This will make no difference to the function, but improves readability & most style guides do not like useless assignment or assigning outside the block a variable is used and prefer const over let when never reassigning etc.

## Checklist

- [ x] The changes are not breaking the editor
- [ ] Followed the guidelines
- [ ] Fixed linting issues

### DIFF
--- a/packages/core/src/inputRules/markInputRule.ts
+++ b/packages/core/src/inputRules/markInputRule.ts
@@ -30,7 +30,6 @@ export function markInputRule(config: {
       const { tr } = state
       const captureGroup = match[match.length - 1]
       const fullMatch = match[0]
-      let markEnd = range.to
 
       if (captureGroup) {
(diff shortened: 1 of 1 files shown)


bdbch (approved):
Thanks for the PR. Looks good to me. @svenadlung what about you?

bdbch:
@svenadlung I'll merge this for now as this should not break anything.

PR #4192, opened 2023-07-08
React context implementation for Tiptap

bdbch:

This PR adds few new features to the React package - mainly:

- `<EditorProvider />`: This now creates a react context and the editor instance and passes it through to child components
- `useCurrentEditor()`: Is used to get the current wrapping editor

so instead of doing

```jsx
const MyEditor = () => {
  const editor = useEditor({
    extensions: [...],
    content: `...`,
  })

  if (!editor) {
    return null
  }

  return (
    <>
      <MyFloatingMenu editor={editor} />
      <MyBubbleMenu editor={editor} />
    </>
  )
}
```

you can now do

```jsx
const MyEditor = () => {
  return (
    <EditorProvider extensions={[...]} content={`...`}>
      <MyFloatingMenu />
      <MyBubbleMenu />
    </EditorProvider>
  )
}
```

1. Created a Context, hooks and a wrapping provider component
2. Added some functionality to the FloatingMenu and BubbleMenu to consume the provider if no editor is provided via props

I tried the default demo and a demo with custom node views and both worked fine.

Check out the deployment preview - if you can't find a difference then everything should be fine.

## Checklist

- [x] Followed the guidelines
- [x] Fixed linting issues


docs/installation/react.md
```diff
@@ -83,9 +118,41 @@ const App = () => {
...
+**Important**: This won't work if you use the `useEditor` hook to setup your editor.
+
 You should now see a pretty barebones example of Tiptap in your browser.
 
-#### 5. The complete setup (optional)
+#### 6. Add before or after slots
+Since the EditorContent component is rendered by the `EditorProvider` component, we now can't directly define where to render before or after content of our editor. For that we can use the `slotBefore` & `slotAfter` props on the `EditorProvider` component.
+
+```jsx
+<EditorPrivider
```
svenadlung, line 147: EditorProvider

romellem:
What purpose is the `<EditorConsumer>` serving in the `<EditorProvider>`? That component already has the `editor` instance - it came from the `useEditor()` call, and in fact we are already passing it to the `EditorContext.Provider`.

That seems unnecessary?

That is, can we go from this

```typescript
export function EditorProvider({
  children, slotAfter, slotBefore, ...editorOptions
}: EditorProviderProps) {
  const editor = useEditor(editorOptions)

  if (!editor) {
    return null
  }

  return (
    <EditorContext.Provider value={{ editor }}>
      {slotBefore}
      <EditorConsumer>
        {({ editor: currentEditor }) => (
          <EditorContent editor={currentEditor} />
        )}
      </EditorConsumer>
      {children}
      {slotAfter}
    </EditorContext.Provider>
  )
}
```

To this?

```typescript
export function EditorProvider({
  children, slotAfter, slotBefore, ...editorOptions
}: EditorProviderProps) {
  const editor = useEditor(editorOptions)

  if (!editor) {
    return null
  }

  return (
    <EditorContext.Provider value={{ editor }}>
      {slotBefore}
      {/* Instead use `editor` directly */}
      {children}
      {slotAfter}
    </EditorContext.Provider>
  )
}
```

Unless there is something I'm missing.

Thanks!

bdbch:
You can pass in the editor instance from `useEditor` like this, yes. The consumer is only used to avoid prop drilling with the editor prop deeper in the component tree. (Similar how you can get the editor down the component tree via `const { editor } = useCurrentEditor`.

PR #4210, opened 2023-07-13
Fix List issues & add support for Mod keys

### LINKED ISSUE #2747: Backspacing through a nested list doesn't work properly

When backspacing through a list and not starting at the last item, it clears the current item and the text of it's parent... but that is about it.

Video provided

[https://www.loom.com/share/0dc205342dae40ecba2b73e715abf030](https://www.loom.com/share/0dc205342dae40ecba2b73e715abf030)

Version 99.0.4844.51 (Official Build) (x86_64)

My video above demonstrates the bug.

This is done within your sandbox on tiptap.dev

_No response_

I expect to naturally backspace through the list.

_No response_

### Are you sponsoring us?

- [X] Yes, I’m a sponsor. 💖

### LINKED ISSUE #3128: Confusion and frustration when dealing with empty list items (aka "Bullet List Limbo")

bdbch:

This PR fixes various issues with lists (for example when the cursor is placed right behind a list and backspace is pressed).

It also moves all list helpers into core so they can be used by taskLists too.

## Checklist

- [x] Followed the guidelines
- [x] Fixed linting issues

## Related Issues

closes #3128
closes #2747


terlan4:
When will this version be released? I would love to see my backspace work asap 🙂❤

bdbch:
@svenadlung could you take a look?

svenadlung (approved):

Slapbox:
I was looking through these changes and I'm curious what the changes to the `cut` command accomplished? I'm curious because we're reliant on that command in our own code. Was there some `mapping`-related issue with the old implementation?

bdbch:
If I remember correctly the issue occured when something was cut and moved after the cut content. Since the mapping wasn't respected beforehand, the insert position was invalid.

I could check again if I find some time what it was about exactly.

Slapbox:
I also noticed the logic in these Backspace handler lines lacks corresponding logic in the Delete handler: https://github.com/ueberdosis/tiptap/blob/2a11d39fe90f1e5a1e251414f262a148a33b92ef/packages/core/src/helpers/listHelpers/handleBackspace.ts#L20-L42

Slapbox:
> If I remember correctly the issue occured when something was cut and moved after the cut content. Since the mapping wasn't respected beforehand, the insert position was invalid.

Thanks for the reply @bdbch! I was curious looking at it because it seems related to the weird mapping misbehavior in #4283

Issue #4182, opened 2023-07-05
[Bug]: splitListItem should support an `attrs` argument identical to prosemirror-schema-list@1.3.0


Nantris:
### Which packages did you experience the bug in?

core

### What Tiptap version are you using?

2.1.0-rc.9

### What’s the bug you are facing?

We need to be able to split `listItem`s, setting the desired attributes on the new `listItem`, but it appears the TipTap implementation of `splitListItem` doesn't actually mirror the ProseMirror implementation, nor does it support the second argument for `attrs` as I expected.

https://prosemirror.net/docs/ref/#schema-list.splitListItem

### What browser are you using?

Chrome

### What did you expect to happen?

Calling `editor.commands.splitListItem('listItem', { foo: true })` should result in the new node having attribute `foo="true"` specified (so long as that's been appropriately configured inside of `addAttributes`, which it has for us) 

### Anything to add? (optional)

Not being able to support this use case is majorly limiting for working with `listItems`. This limitation is important to address, but is only needed towards the end of work on `listItem`s in our project where we're addressing an edge case, so **it may not be immediately apparent just how many people actually will need this. I suspect it's a lot; especially considering it was added to ProseMirror's API too and marijnh is very cautious about increasing the surface area of the API.**

### Did you update your dependencies?

- [X] Yes, I’ve updated my dependencies to use the latest version of all packages.

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

Slapbox:
Friendly bump. Can anyone advise on this? I've tried to work around this limitation but it doesn't seem to be possible. Support for `attrs` is required to achieve the aim here of setting attributes on the newly created item that results from the split.

Slapbox:
Any update on this/any interest in supporting it? A week later I'm hitting this issue again and the workaround is crazy to have to use once, but much worse the second time.

bdbch:
Hey @Slapbox - time wise i'm a bit capped right now but will try to get back to a lot of issues still open. In the meantime is it possible to maybe take a look at the splitList command yourself to see if you could provide a PR for that? Getting it merged / released would not be a big of a deal - otherwise I'd have to ask for some more time.

Slapbox:
Thanks for your reply @bdbch; totally fair and super relatable! I've submitted a skeleton of a PR to do this in #4253, but it needs some testing/linting work which I probably won't find the time to complete unless I hit this a third time.

nperez0111:
[...]

PR #4253, opened 2023-07-26
Update splitListItem.ts to support overrideAttrs

Nantris:

This change addresses #4182

Use object spread to allow `overrideAttrs` to override the default determined attributes

_Very minimally!_ I only tested these changes via `patch-package` and then free-handed them in using the online Github editor

I have not added any relevant tests.

## Remarks

I have not updated the docs

## Checklist

- [x] Followed the guidelines
- [x] Fixed linting issues

## Related issues

#4182


Slapbox:
This PR needs additional work to be merge-worthy. I hope someone will pickup and complete the remaining tasks as I will not have the time.

packages/core/src/commands/splitListItem.ts
```diff
@@ -68,11 +68,11 @@ export const splitListItem: RawCommands['splitListItem'] = typeOrName => ({
         const depthAfter = $from.indexAfter(-1) < $from.node(-2).childCount ? 1 : $from.indexAfter(-2) < $from.node(-3).childCount ? 2 : 3
 
       // Add a second list item with an empty default start node
-      const newNextTypeAttributes = getSplittedAttributes(
+      const newNextTypeAttributes = {...getSplittedAttributes(
```
bdbch, line 71: We have some linting issues in here @Slapbox, could we get those resolved?

packages/core/src/commands/splitListItem.ts
```diff
@@ -19,7 +19,7 @@ declare module '@tiptap/core' {
   }
 }
 
-export const splitListItem: RawCommands['splitListItem'] = typeOrName => ({
+export const splitListItem: RawCommands['splitListItem'] = (typeOrName, overrideAttrs = {}) => ({
```
bdbch, line 22: Maybe I'm a bit blind but won't Typescript define the type for `overrideAttrs` to an empty object? Don't you need to add a type for `overrideAttrs` here?

https://github.com/ueberdosis/tiptap/blob/fc62eb3c4d6d784b5f9b43a58e0bb0a1073cb612/packages/core/src/commands/splitListItem.ts#L17C23-L17C23

I'd suggest just to set it to `[key: string]: any`?

bdbch (changes requested):
Linting and Typing issues I guess? Otherwise this looks fine, thank you.

packages/core/src/commands/splitListItem.ts
Slapbox: Good catch!

Slapbox:
I made these edits in a proper editor with ESLint support this time, so I think they should pass the tests.

Slapbox:
Although this commit still needs the associated docs update.

bdbch (approved):

bdbch:
LGTM for now – feel free to add docs to the PR

nperez0111 (approved):

PR #4286, opened 2023-08-03
made starterkit options in packages/starter-kit/src/starter-kit.ts a …

Brentably:
…partial

Refactored the type for starter kit options.

## Checklist

- [x] Followed the guidelines
- [x] Fixed linting issues


Brentably:
Ok, the way I needed to fix this was by strongly typing my options instead.

something like 

```
const starterKitOptions:Partial<StarterKitOptions> = {<starter kit options here>}
```
and then 
```
StarterKit.configure(starterKitOptions)
```
made the compiler happy.

PR #4288, opened 2023-08-05
fix(strikethrough): update strikethrough shortcut

### LINKED ISSUE #2540: Keyboard shortcut for strikethrough overwrites browser shortcut for switching to right to left language layout

When using a tiptap editor with strikethrough nodes pressing `ctrl+shift+x` results in switching to strikethrough. However this shortcut is the default for switching to a right to left language layout and back in firefox. Using it in the github editor for example changes the text input layout to provide for better input experience for hebrew and arabic and other rtl languages.

Press `ctrl+shift+x` in the editor on tiptap.dev
Open a comment here with firefox and press `ctrl+shift+x` to see the expected behavior.

tiptap.dev ;)

Open a comment here with firefox and press `ctrl+shift+x` to see the expected behavior.

I'm not sure how relevant this is. Chrome only seems to offer the switch in a browser extension. If the language settings of the browser are a rtl language this will probably also be the default for the text entry. So it's only necessary if you want to write non-rtl text in an rtl language browser or the other way round. However this can be a frequent in a multi language environment.

I'm also not aware of any workarounds to switch the text entry direction without the keyboard shortcut - but i am also not writing rtl languages.

Just wanted to flag this as a possible issue. Feel free to close if picking a different default for the keyboard shortcut is problematic.

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

bdbch:

This PR changes the shortcut for strikethrough to a more standardized version similar to other apps (Notion, Dropbox Paper).

Now the default shortcut is `Mod+Shift+s` for MacOS and `Ctrl+Shift+s` for other operation systems.

See https://defkey.com/search?irq=strikethrough

Changed the addKeyboardShortcuts function

Tested them locally on MacOS and Windows.

Check the deployed demo

## Remarks

nothing

## Checklist

- [x] Followed the guidelines
- [x] Fixed linting issues

## Related issues

closes #2540

### DIFF
--- a/demos/src/Marks/Strike/React/index.spec.js
+++ b/demos/src/Marks/Strike/React/index.spec.js
@@ -58,15 +58,15 @@ context('/src/Marks/Strike/React/', () => {
 
   it('should strike the selected text when the keyboard shortcut is pressed', () => {
     cy.get('.tiptap')
-      .trigger('keydown', { modKey: true, shiftKey: true, key: 'x' })
+      .trigger('keydown', { ctrlKey: true, shiftKey: true, key: 's' })
       .find('s')
       .should('contain', 'Example Text')
   })
 
   it('should toggle the selected text striked when the keyboard shortcut is pressed', () => {
     cy.get('.tiptap')
-      .trigger('keydown', { modKey: true, shiftKey: true, key: 'x' })
-      .trigger('keydown', { modKey: true, shiftKey: true, key: 'x' })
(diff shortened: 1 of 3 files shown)


svenadlung (approved):

Slapbox:
This PR needs to be amended to be case-insensitive like the others (eg [bold](https://github.com/ueberdosis/tiptap/blob/ec0b4a3771e31bcb9443c960a0fb946563edc6d6/packages/extension-bold/src/bold.ts#L79-L84))

PR #4292, opened 2023-08-05
fix(link): Fix autolinking and pasting

### LINKED ISSUE #4256: [Bug]: linkOnPaste: false doesn't work at all - pasted links are still linkified

extension-link

2.1.0-rc.12

When `linkOnPaste` is set to false, I'd expect the `pasteHandler` code not to run, but instead nearly all of it runs.

Chrome

### Code example

_No response_

`pasteHandler` code should be fully or almost fully skipped over during execution.

_No response_

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

### LINKED ISSUE #4258: [Bug]: Links are partially or entirely unlinked when modified

extension-link

2.1.0-rc.12

The way this problem manifests in my editor is different than it manifests on the demos, even when I reduce my code to only use `StarterKit` + `Link` with no configuration options set. **Edit: (I suspect this is due to `2.0.4` vs `2.1.0-rc.12`)**

In my environment, deleting the colon, slash, or any of the characters in `http` (but not the `s`) causes the last character to become un-marked.

https://github.com/ueberdosis/tiptap/assets/6835891/0d402fca-bbff-48d2-b507-896258812608

In the examples, deleting those characters un-marks the entire mark, which is more reasonable but still not really desirable. I can't trace where this behavior is originating from. Can anyone advise?

Chrome

### Code example

_No response_

Links should not be removed or mangled once set

_No response_

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

bdbch:

This PR will fix multiple issues with the current auto linking.

- First fix is that pasteOnLink option was not working anymore
- I'll also fix an apparent issue that content after a pasted link is also autolinked? (wasn't able to reproduce for now)

## Checklist

- [x] Followed the guidelines
- [x] Fixed linting issues

## Related issues

fixes #4256 
fixes #4258

### DIFF
--- a/packages/extension-link/src/helpers/autolink.ts
+++ b/packages/extension-link/src/helpers/autolink.ts
@@ -7,7 +7,7 @@ import {
 } from '@tiptap/core'
 import { MarkType } from '@tiptap/pm/model'
 import { Plugin, PluginKey } from '@tiptap/pm/state'
-import { find, test } from 'linkifyjs'
+import { find } from 'linkifyjs'
 
 type AutolinkOptions = {
   type: MarkType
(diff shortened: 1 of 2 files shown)


rfgamaral:
@bdbch The [docs for the `linkOnPaste` option](https://tiptap.dev/api/marks/link#link-on-paste) say:

> Adds a link to the current selection if the pasted content only contains an url.

But I don't think this is quite true...

Take the [link editor example](https://tiptap.dev/api/marks/link#usage) and paste something like `[tiptap](https://tiptap.dev)`. You can see that `https://tiptap.dev` will be converted into a link, but the pasted content **does not** only contain a URL, it contains more than that.

firefox_paI0aAJ33T

This became a problem for us because we have Markdown support on our editor, and we have our own paste rules to handle the Markdown syntax. However, this new `linkOnPaste` option (default: `true`) caused us an issue by auto-linking when it shouldn't (based on my interpretation of the documentation).

As a workaround, we are setting `linkOnPaste` to `false`, and this seems to fix everything for us, but I'm wondering if this shouldn't be fixed in Tiptap itself? Although Tiptap does not support Markdown, I don't think the link found in `[tiptap](https://tiptap.dev)` should be auto-linked.

Here's another example of a similar situation where auto-linking shouldn't have worked:

firefox_oa8cLw6arn

What do you think?

PR #4298, opened 2023-08-07
Update horizontal-rule.ts

courthead:
Allow HTML attributes to be applied to horizontal rules, which makes it easy to apply custom classes for example, and different styles for hr elements.


packages/extension-horizontal-rule/src/horizontal-rule.ts
```diff
@@ -38,10 +38,10 @@ export const HorizontalRule = Node.create<HorizontalRuleOptions>({
   addCommands() {
     return {
       setHorizontalRule:
-        () => ({ chain }) => {
+        options => ({ chain }) => {
```
bdbch, line 41: I think this `options` is not typed in the Commands typings as far as I can see? Could you please add Typescript typings for this?

bdbch (changes requested):

PR #4313, opened 2023-08-11
fix demo for cut command

bdbch:

Fixes the demo for the `cut` command.


Slapbox:
@bdbch I can't actually find this command on the site. Is it included in the docs?

Issue #3133, opened 2022-08-28
How delete content when drop to other editor use dropcursor, in multiple editors


songispm:
### What problem are you facing?

I have multiple editors, when I use dropcursor to drop content to other editor, the content still there in source editor.

![image](https://user-images.githubusercontent.com/38745323/187059484-077e5f61-48f5-4380-bcf8-ff6106f8da4d.png)

![image](https://user-images.githubusercontent.com/38745323/187059231-1ed818c4-83d0-4373-9fcd-4011532fbf27.png)

![image](https://user-images.githubusercontent.com/38745323/187059245-0a556903-d705-4ace-9526-8a93ee8dd5ba.png)

### What’s the solution you would like to see?

Tell me how delete it

### What alternatives did you consider?

Have not

### Anything to add? (optional)

example:

[https://codesandbox.io/s/dropcursor-multiple-editors-m0iti5](https://codesandbox.io/s/dropcursor-multiple-editors-m0iti5)

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

bdbch:
Thanks for the issue, this seems to be a bug we should look into at some point.

LuchoCateura:
Hello, I would like to help resolve this issue @bdbch

LuchoCateura:
https://github.com/ueberdosis/tiptap/pull/3279

bdbch:
I'm looking into your PR! Thanks for your submission ✌️

bdbch:
When I'm done with the PR for sure!

bdbch:
Fixed by https://github.com/ueberdosis/tiptap/pull/5893

PR #3429, opened 2022-11-18
Feature/delete content when drop on another editor

LuchoCateura:
Now it works for Firefox too

PR #4328, opened 2023-08-13
feat: drag and drop across multi editors

songispm:

Delete content when drop on another editor, 

At the same time, it does not affect the normal drop event effect of the editor.

Work in nested nodes too.

Core change of code:

https://github.com/ueberdosis/tiptap/pull/4328/files#diff-6eddae5591c39b131ab5597e8ac574df71fafda96023c4573c45fd73ed066b9f

## Related issues

[https://github.com/ueberdosis/tiptap/issues/3133](https://github.com/ueberdosis/tiptap/issues/3133)

## Demo

https://deploy-preview-4328--tiptap-embed.netlify.app/preview/Experiments/MultipleEditors

or checkout this pull request

1. npm run dev
2. Visit http://localhost:3000/preview/Experiments/MultipleEditors

https://github.com/ueberdosis/tiptap/assets/38745323/d9e6c03a-edb4-4543-b723-b7511c3a844c


songispm:
Another related pull request: https://github.com/ueberdosis/tiptap/pull/3429

packages/core/src/PasteRule.ts
```diff
@@ -164,13 +164,25 @@ export function pasteRulesPlugin(props: { editor: Editor; rules: PasteRule[] }):
           dragSourceElement = view.dom.parentElement?.contains(event.target as Element)
             ? view.dom.parentElement
             : null
+
+          if (dragSourceElement) {
+            (window as any).tiptapDragFromOtherEditor = editor
```
bdbch, line 169: Do we need to save into the global window instance here? Is there no other way to keep track of the editor instance?

bdbch (changes requested):

packages/core/src/PasteRule.ts
songispm: > Do we need to save into the global window instance here? Is there no other way to keep track of the editor instance?

Because when dragging across editors, there are multiple editor instances. If you delete the selection content of another editor instance in the drop method of one editor instance, you must get a reference to another editor instance.

I can only simply store it on the window, unless tipap provides a similar global store or variable to store this kind of thing. @bdbch

songispm: If you have a better way, we can modify it again.

songispm: pushed again

https://github.com/ueberdosis/tiptap/pull/4328/commits/5b906e7456c3dc03ded851588a424dedaca58cf4

colinramsay:
@songispm is this ready to merge given the change which was requested was made?

songispm:
yes, I think it's worked.

https://github.com/ueberdosis/tiptap/pull/4328/files

colinramsay:
Great work! Is it straightforward to do this React? I see there’s currently only a Vue example. @bdbch if you are happy do you think this could get in to a release?

crystalball808:
Would like to use this feature!

bdbch:
Moved to #5893

Issue #3747, opened 2023-02-14
Mixing up the words 'that' and 'this'


raarts:
### What’s the URL to the page you’re sending feedback for?

https://tiptap.dev/guide/styling

### What part of the documentation needs improvement?

The sentence: 

`You can even pass classes to the element which contains the editor like that:`

Needs to change into:

`You can even pass classes to the element which contains the editor like this:`

This error happens on multiple pages.

### What is helpful about that part?

I'm only pointing out a spelling error.

### What is hard to understand, missing or misleading?

The word 'that' is just used in the wrong way. No biggie, it's just wrong.

PR #4334, opened 2023-08-16
Improve introducing examples in the documentation

lode:

This improves copy to introduce example code. Previously it added 'like that:'. In most cases this was just a stop word and can be removed. In other cases the sentence needs to be rewritten for that, or, what I did, replace it with 'like this:' because it refers to an upcoming example, instead of one that was shown before.

n/a

n/a

I double checked my logic here: https://forum.wordreference.com/threads/like-this-like-that.2851037/

## Remarks

n/a

## Checklist

- [x] Followed the guidelines
- [x] Fixed linting issues

## Related issues

This fixes https://github.com/ueberdosis/tiptap/issues/3747.

### DIFF
--- a/docs/api/commands.md
+++ b/docs/api/commands.md
@@ -140,7 +140,7 @@ editor.first(({ commands }) => [
 ])
 ```
 
-Inside of commands you can do the same thing like that:
+Inside of commands you can do the same thing:
 
 ```js
 export default () => ({ commands }) => {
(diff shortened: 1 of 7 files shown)


bdbch (approved):
Thanks. I'd be fine with those changes, what do you think @svenadlung

svenadlung (approved):

Issue #4343, opened 2023-08-18
[Bug]: autolink does not work and throw an exception


steven-qi:
### Which packages did you experience the bug in?

core, extension-link, react

### What Tiptap version are you using?

2.1.2

### What’s the bug you are facing?

With `autolink` enabled for link extension, whenever I inputted a URL manually and then type a space (expecting that it will turn into a link automatically), the editor removed the latest space and the link remain unlinked.
There is an error in the console.

I first encountered this issue in version `2.1.0`, and then upgraded to `2.1.1` and `2.1.2` with no luck.

![261482485-dd9cb92c-fc8f-48fe-b9ac-b9153415d6c2](https://github.com/ueberdosis/tiptap/assets/6123222/38950be2-89da-4ddb-8f82-957f0c6e78c8)

and here is a stack trace I captured at the exact time it is going to throw (`type` is `undefined`, but it is trying to invoke `type.isInSet`).

![261483943-d493eee0-d96e-4840-b8d6-8db34b9d52d2](https://github.com/ueberdosis/tiptap/assets/6123222/80279a4b-6fb4-4f53-9625-cc49e6dcc7dc)

I dig into it a bit, it seems that the cause is that, `newState.schema.marks` does not have the property `code`.

https://github.com/ueberdosis/tiptap/blob/develop/packages/extension-link/src/helpers/autolink.ts#L93

![image](https://github.com/ueberdosis/tiptap/assets/6123222/689d56e6-8407-4b70-a1f5-2389d770a105)

### What browser are you using?

Chrome

### What did you expect to happen?

the inputted text should turn into a link node without any error

### Anything to add? (optional)

paste link does work ok

### Did you update your dependencies?

- [X] Yes, I’ve updated my dependencies to use the latest version of all packages.

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

svenadlung:
Seems this got added [here](https://github.com/ueberdosis/tiptap/commit/b24df3aa4c2f3fdb2ed6122d2d32fb7c4e07f2af#diff-b6af6a239569036b1e9e01542e5ef9c2a41ae95134d7c75090c0e8a67b08e4d7R126).

I can imagine that you have not enabled the code extension and that is why the test ends with an error. 

CC @rfgamaral

bdbch:
Merged your PR @steven-qi – fix should be included in 2.1.3. I'll close this issue for now – feel free to reopen if the problem still persists in the new version

[...]

PR #4344, opened 2023-08-18
fix autolink when code is not enabled for editor

steven-qi:
fix autolink issue when code extension is not enabled for the editor. See issue https://github.com/ueberdosis/tiptap/issues/4343 for more details

There is a filter in the auto link handler to avoid auto link in a code block. However, when code is not enabled for a particular editor instance, `schema.marks.code` will be `undefined`. The handler passed it to `rangHasMarks` without any validation, and cause an exception.

To fix the issue, we can just return early from the handler when `schema.marks.code` is a falsy value.

Yes. As the code change is very small, I verified it by applying the same change to my local `node_modules` folder.

To reproduce the issue, just setup an editor instance with `extension-link` included, but without `extension-code` included, typo a URL in the editor, then type a space, you should see an exception in the console, and link is not created.

Note, `extension-code` is part of `StarterKit`, that's why it does not happen for demos.

After applying this change, auto-link should work without any console error, with or without `extension-code`

## Remarks

## Checklist

- [x] Followed the guidelines
- [x] Fixed linting issues

## Related issues

https://github.com/ueberdosis/tiptap/issues/4343

### DIFF
--- a/packages/extension-link/src/helpers/autolink.ts
+++ b/packages/extension-link/src/helpers/autolink.ts
@@ -87,6 +87,7 @@ export function autolink(options: AutolinkOptions): Plugin {
             }))
             // ignore link inside code mark
             .filter(link => {
+              if (!newState.schema.marks.code) return true; // maybe code is not enabled for this editor
               return !newState.doc.rangeHasMark(
                 link.from,
                 link.to,


bdbch (approved):
Good find. I moved the base branch to `main` so we can include this as a hotfix instead of going the whole rc => release round.

@svenadlung I'd merge this asap so we don't have broken editors

PR #4346, opened 2023-08-18
prevent opening link when clicking on selection containing a link

aldinezi:

This code will check whether a clicked node name is 'A'. This will fix the issue when `openOnClick` is enabled, that causes clicking on a selection to open a link that's somewhere within the selection, even though user clicked some other element.

I have tried these changes in our repo that uses Tiptap for our pages editor.

Our QA engineer tested whether this fix solves the issue.

To reproduce the current issue:
 1. `openOnClick` must be enabled in link-extension config
 2. Add some link somewhere within the paragraph and some other text
 3. Select the text and the link
 4. Click somewhere, not directly on the link
 5. It will open a link

These changes will make click open link only when user clicks on the link itself.
 
## Remarks

No remarks.

## Checklist

- [x] Followed the guidelines
- [x] Fixed linting issues

### DIFF
--- a/packages/extension-link/src/helpers/clickHandler.ts
+++ b/packages/extension-link/src/helpers/clickHandler.ts
@@ -15,8 +15,11 @@ export function clickHandler(options: ClickHandlerOptions): Plugin {
           return false
         }
 
-        const eventTarget = event.target as HTMLElement;
-        if (eventTarget.nodeName !== 'A') return false;
+        const eventTarget = event.target as HTMLElement
+
+        if (eventTarget.nodeName !== 'A') {
+          return false
+        }
 
         const attrs = getAttributes(view.state, options.type.name)
         const link = (event.target as HTMLLinkElement)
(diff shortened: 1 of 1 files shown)


bdbch:
Moved target branch to `main` so we can include this for the next hotfix

packages/extension-link/src/helpers/clickHandler.ts
```diff
@@ -15,6 +15,12 @@ export function clickHandler(options: ClickHandlerOptions): Plugin {
           return false
         }
 
+        const eventTarget = event.target as HTMLElement
+
+        if (eventTarget.nodeName !== 'A') {
```
bdbch, line 20: This would still open the wrong link when I have my selection in Link A and click on Link B right?

aldinezi: Good point. It seemed so to me also, but after checking out, when you have two links within the selection it will not open the wrong one. It always opens the one you clicked. 

In another case, if you have selected link A, and you click on the other link outside selection, this handler will not run.

In conclusion, it works in both cases.

ioma8:
Can this be merged? I have the issue with link opening

PR #4354, opened 2023-08-19
feature(core): pass through paste event to pasteHandler `getAttributes`

### LINKED ISSUE #4347: [Bug]: YouTube video pasting broken with Link extension since 2.1.0-rc.2

extension-link, extension-youtube

2.1.X

The paste handler configured for the YouTube extension does not work when Link extension is used as well. The paste rule is defined here:
https://github.com/ueberdosis/tiptap/blob/af017d18ade9fc741c37c325b187ae344daff040/packages/extension-youtube/src/youtube.ts#L127-L133

Instead of the YouTube video being embedded automatically on paste, a link appears pointing to the YouTube URL.

This works fine with 2.0.4 but in any version 2.1.X it is broken (broken since stable release 2.1.0 and from RC release 2.1.0-rc.2).

Chrome

### Code example

https://codesandbox.io/s/objective-edison-zxxc94

I would expect the YouTube extension's paste rule to work alongside the Link extension.

Here is a working example using version 2.0.4 (the same example using 2.0.4):
https://codesandbox.io/s/reverent-dust-j9cmm5

To drill down even deeper, it seems to also work with `2.1.0-rc.1` but it is broken in `2.1.0-rc.2`.

Potentially breaking change between these releases #3975 based on the `autolink.js` change history within the `extension-link` package.

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

bdbch:

This PR passes through the `ClipboardEvent` coming from the actual paste event down to the `getAttributes` function inside the pasteEvent handler for `MarkPasteRule` and `NodePasteRule`s.

This allows us to do more checks on the paste without breaking out of the Tiptap priority system.

This way I also fixed an issue with the Link extension being to agressive as it was ignoring all priorities on extensions when doing the pasteHandling.

This will fix #4347 as it was the issue bringing this up to me.

- Inside the PasteHandlerPlugin I added variables to store the paste/drop events in
- Those are passed down to the handler from where they are passed to the `getAttributes` function
- After that, the paste and drop event are set back to empty events

I tested my changes locally to see if the YouTube extension works fine again while the linking from the link extension still works on paste.

1. Checkout this PR
2. Add the Link extension to the YouTube demo
3. Try pasting a normal link (should work as usual)
4. Try pasting a youtube video (this now should embed the video)

## Remarks

- This should be backwards compatible and non breaking as the new parameter is optional to use.

## Checklist

- [x] Followed the guidelines
- [x] Fixed linting issues

## Related issues

- fixes #4347


svenadlung (approved):

chhuang:
@bdbch any chance of getting this merged?

PR #4380, opened 2023-08-25
Fix suggestion collaboration bug

fsvergara:

Fix a bug where the suggestion re-renders when someone is typing above the cursor while collaborating.
Here's a screencast explaining the bug and the code:
https://www.loom.com/share/bbb7e73feae44a8c8ebfcbae77ad3eb5?sid=8fd24818-0b3b-4cea-9a54-b7bcb017d682

Change when onStart and onExit are called in the suggestion extension, so that they won't be called if we are just moving the suggestion but it will remain active. 

Used the changed version of the suggestion extension in our local environment, here's how it looks after the changes:
https://www.loom.com/share/facfc0b58a7a496199f2be07a8306969

1. Create an editor with the Suggestion and Collaboration extensions
2. Open two editors, with one of them initiate the suggestion and with the other one type above the cursor where the suggestion is
3. The suggestion should not re-render

## Remarks

## Checklist

- [x] Followed the guidelines
- [x] Fixed linting issues


nperez0111:
Thank you for your contribution @fsvergara 

Sorry it took so long to get a review on this, I'm going to add to the beta

nperez0111 (approved):

rfgamaral:
@nperez0111 We found out that this fix introduced a bug that **didn't exist before** when using the suggestion utility with `allowSpaces: true`, any chance this can be reverted until a proper a fix that doesn't introduce new bugs is implemented?

The newly introduced bug by this can be reproduced like this:

- Add `allowSpaces: true` to `demos/src/Nodes/Mention/React/suggestion.js`
- Open the `Nodes/Mention/React` demo (i.e. `npm run start`)
- Type the following in this order: <kbd>@lea</kbd>, <kbd>Esc</kbd>, <kbd>Space</kbd>, and <kbd>@</kbd>
    - Observe that the suggestion dropdown **does not appear** (this didn't happen before)

nperez0111:
@rfgamaral I'd prefer to roll forward, will look into it tmrw

There are too many changes queued up to be doing a patch release right now

nperez0111:
@fsvergara I wonder if this change (to latest tiptap version on develop) works for what you were running into?

```diff
diff --git a/demos/src/Nodes/Mention/React/suggestion.js b/demos/src/Nodes/Mention/React/suggestion.js
index 554d9ee11..2695199d2 100644
--- a/demos/src/Nodes/Mention/React/suggestion.js
+++ b/demos/src/Nodes/Mention/React/suggestion.js
@@ -4,6 +4,7 @@ import tippy from 'tippy.js'
 import MentionList from './MentionList.jsx'
 
 export default {
+  allowSpaces: true,
   items: ({ query }) => {
     return [
       'Lea Thompson',
diff --git a/packages/suggestion/src/suggestion.ts b/packages/suggestion/src/suggestion.ts
index 80744aaac..c0dd97808 100644
--- a/packages/suggestion/src/suggestion.ts
+++ b/packages/suggestion/src/suggestion.ts
@@ -194,10 +194,9 @@ export function Suggestion<I = any, TSelected = any>({
           const started = !prev.active && next.active
           const stopped = prev.active && !next.active
           const changed = !started && !stopped && prev.query !== next.query
-
-          const handleStart = started
+          const handleStart = started || moved
           const handleChange = changed || moved
-          const handleExit = stopped
+          const handleExit = stopped || moved
 
           // Cancel when suggestion isn't active
           if (!handleStart && !handleChange && !handleExit) {
```

rfgamaral:
@nperez0111 Yes, that will fix it. Actually, just changing `handleStart` to `started || moved` is enough to fix it. But won't that reintroduce the issue at hand? 🤔

nperez0111:
My understanding is that it would be okay to trigger more updates than needed (which would re-render a slash command), rather than forcing it to re-render and recreate from scratch by flagging the start and stop.
@rfgamaral I'm fairly confident in this, but I will let it soak over night as it is late over here

rfgamaral:
@nperez0111 That's okay, rest up, and come back tomorrow energized 💪

Hopefully, you can figure it out tomorrow morning, and if you could cut a patch release, I'd really appreciate it. 🙏 This is blocking us from updating Tiptap to the latest version, which brings a couple of important fixes for us.

nperez0111:
@rfgamaral consider it done 😄

rfgamaral:
@nperez0111 This is much appreciated, thank you!

rfgamaral:
@nperez0111 Sorry to bring this up again, but while doing test on our side with the new version, we realized is still not working as it was before this PR.

The remaining issue can be reproduced like this:

- Add `allowSpaces: true` to `demos/src/Nodes/Mention/React/suggestion.js`
- Open the `Nodes/Mention/React` demo (i.e. `npm run start`)
- Type the following in this order: <kbd>@</kbd>, <kbd>Space</kbd>, <kbd>@</kbd>, and <kbd>Esc</kbd>
    - Observe that the suggestion dropdown **does not close** (this didn't happen before this PR)

And while I was defining this test plan, I found yet another issue caused by this PR:

https://github.com/user-attachments/assets/8b842d98-5859-4956-9dfa-a869fb8e08d0

When I selected all the text, and deleted the selection, the dropdown **does not close**.

And yet another bug (they are all related to the same change):

https://github.com/user-attachments/assets/4821d159-7257-446d-bae9-a490d75a4c66

I always found the suggestion utility code a bit confusing, so I'm unsure if the best solution is to revert the code to what it was before this PR (and try to find a better solution for that issue), or try to keep rolling forward, and find a solution for all these issues.

nperez0111:
@rfgamaral I have too many things to do today so I won't be able to get to it until next week.

My suggestion would be to attempt to add the same condition that I put at the `handleStart` event onto the `handleExit` event & see if that resolves it. I tried the minimal change because I couldn't come up with a scenario where that would happen, but it seems like you found them.

Would be willing to take a PR for a fix sometime within the day

rfgamaral:
> My suggestion would be to attempt to add the same condition that [I put at the handleStart event](https://github.com/ueberdosis/tiptap/commit/3c3d7148c2faff56085340868051bc4ef6746e96) onto the handleExit event & see if that resolves it.

@nperez0111 Yes, that solves all those issues I posted, and keeps the issue at hand also fixed 👍

Will open a quick PR.

nperez0111:
This is released with v2.7.0 https://github.com/ueberdosis/tiptap/releases/tag/v2.7.0

PR #4388, opened 2023-08-28
fix orderedList truncated after the codeBlock is deleted in the order…

wang1xiang:

fix orderedList truncated after the codeBlock is deleted in the orderedList.

Currently using clearNodes to delete the code block when the code block is empty. However, when the code block is in an ordered list, deleting this code block causes the ordered list to be truncated.

In order to solve this problem, we can additionally process the ordered list in the delete event, instead of deleting the code block through clearNodes, delete it through deleteNode.

Yes. As the code change is very small, I verified it by applying the same change to my local node_modules folder.

To reproduce the problem, just insert a block code in the ordered list, then type a Backspace and you should see the exception, the ordered list is truncated.

After applying this change, the issue was fixed.

## Remarks

## Checklist

- [x] Followed the guidelines
- [x] Fixed linting issues

## Related issues

### DIFF
--- a/packages/extension-code-block/src/code-block.ts
+++ b/packages/extension-code-block/src/code-block.ts
@@ -138,6 +138,12 @@ export const CodeBlock = Node.create<CodeBlockOptions>({
         }
 
         if (isAtStart || !$anchor.parent.textContent.length) {
+          if (this.editor.isActive('orderedList'))
+            return this.editor
+              .chain()
+              .insertContent('<p></p>')
+              .deleteNode(this.name)
+              .run()
           return this.editor.commands.clearNodes()
         }
 


packages/extension-code-block/src/code-block.ts
```diff
@@ -138,6 +138,12 @@ export const CodeBlock = Node.create<CodeBlockOptions>({
         }
 
         if (isAtStart || !$anchor.parent.textContent.length) {
+          if (this.editor.isActive('orderedList'))
```
bdbch, line 141: I think using defined extensions here to solve this is the wrong way because this issue will appear in other types of nodes too I guess.

wang1xiang: Oh, you are right. It's because I haven't thought about it comprehensively enough.

I'm currently modifying it like this in the project.

bdbch: I'll think about this for some time. The solution should work for all kind of wrapping nodes if possible.

arnaugomez:
The issue seems to be that `clearNodes()` conflates two separate behaviors:

1. **Reset the current textblock in place**

   * Empty `codeBlock` → parent-compatible default textblock, usually `paragraph`.
   * Preserve wrappers like list items, blockquotes, table cells, details content, or custom containers.
   * Do **not** lift or unwrap ancestors.

2. **Clear/lift surrounding structure**

   * The broader `clearNodes()` behavior, useful when intentionally normalizing a selected range.

For this bug, Backspace in an empty `codeBlock` should use behavior **1**.

Example solution:

```ts
if (isAtStart || !$anchor.parent.textContent.length) {
  return editor.commands.resetCurrentTextblock()
}
```

`resetCurrentTextblock()` could be a generic core command/helper that:

* Requires a collapsed selection inside a textblock.
* Looks at the current node’s parent and index.
* Finds the parent-compatible default textblock.
* Replaces only the current textblock.
* Never calls `lift()`.

Simplified example implementation of `resetCurrentTextblock`:

```ts
const parent = $from.node($from.depth - 1)
const index = $from.index($from.depth - 1)
const defaultType = parent.contentMatchAt(index).defaultType

if (!defaultType?.isTextblock) {
  return false
}

tr.setNodeMarkup($from.before(), defaultType)
```

This would handle ordered lists, bullet lists, blockquotes, table cells, details content, and custom wrappers consistently, without special-casing `codeBlock`.

bdbch:
@coderabbitai full review

bdbch:
I'll close this PR for now as it's old and not maintained anymore.

Issue #4327, opened 2023-08-13
[Bug]: `renderText` does not propagate to `NodeSpec.leafText` and `extension-floating-menu` uses `Node.textContent` which uses `NodeSpec.leafText`


segevfiner:
### Which packages did you experience the bug in?

core, extension-floating-menu

### What Tiptap version are you using?

2.0.4

### What’s the bug you are facing?

1.  Open a new empty paragraph, see the "FLOATING MENU".
2. Click "Insert foo".
3. The "FLOATING MENU" still shows! Even though the paragraph is not empty!

### What browser are you using?

Chrome

### Code example

https://codesandbox.io/p/sandbox/tiptap-leaftext-issue-skv4vm?file=%2Fsrc%2Fcomponents%2FEditor.vue%3A4%2C1

### What did you expect to happen?

The "FLOATING MENU" should not show on a non empty paragraph.

### Anything to add? (optional)

This is caused by:
https://github.com/ueberdosis/tiptap/blob/f387ad3dd4c2b30eaea33fb0ba0b42e0cd39263b/packages/extension-floating-menu/src/floating-menu-plugin.ts#L38-L55

It uses `Node.textContent` which uses `NodeSpec.leafText` to fill it. But Tiptap does not fill `NodeSpec.leafText` using Tiptap's `Node.renderText`, rather Tiptap has its own separate mechanism to generate text content from Nodes that doesn't integrate with ProseMirror's one (Maybe ProseMirror's one didn't exist at the time?). 

It does hook that implementation into ProseMirror text copy paste using [packages/core/src/extensions/clipboardTextSerializer.ts](https://github.com/ueberdosis/tiptap/blob/243bdb8c3f8f3497fa3e70cc49040894c44870f0/packages/core/src/extensions/clipboardTextSerializer.ts#L33-L34), but not into the Tiptap's own floating menu. Though maybe a simpler check can be done in the floating menu anyhow to check if the paragraph is empty or not.

### Did you update your dependencies?

- [X] Yes, I’ve updated my dependencies to use the latest version of all packages.

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

Issue #4310, opened 2023-08-10
[Bug]: textBetween cannot get text inside custom extensions


deathg0d:
### Which packages did you experience the bug in?

core

### What Tiptap version are you using?

2.0.4

### What’s the bug you are facing?

I created a simple extension that displays an attribute value. 

```
import { Node } from '@tiptap/core'

export default Node.create({
  name: 'inlineText',
  group: 'inline',
  inline: true,
  content: 'block*', // adding text* throws error in collaborative editors if another user selects the text
  // atom: true,
  draggable: true,
  selectable: true,
  addAttributes() {
    return {
      value: {
        default: '...',
      },
    }
  },
  parseHTML() {
    return [
      {
        tag: 'inline-text',
      },
    ]
  },

  renderHTML({ HTMLAttributes }) {
    return ['inline-text', HTMLAttributes, 0]
  },

  renderText({ node }) {
    return node.attrs.value
  },

  addNodeView() {
    return ({ editor, node, getPos }) => {
      const { view } = editor
      const dom = document.createElement('span')
      dom.innerHTML = node.attrs.value;
      return {
        dom,
      }
    }
  },
})
```

Even when renderText is set, tiptap assumes that the content of this span is empty. Everything displays properly. However if I try to get the content over this extension it shows up empty. I selected text including this extension and triggered the following code using the bubble menu.

```
const { view, state } = editor;
const { from, to } = view.state.selection;
const text = state.doc.textBetween(from, to);
console.log("text-between", from, to, text);
```
All other texts show up fine except the text inside the extension that shows up empty.

Weird behavior: If I Ctrl+c the selected text and paste it in a notepad, it shows full text, including the text inside the extension.

### What browser are you using?

Chrome

### What did you expect to happen?

When using textBetween over the extension, it should capture the text value of the extension as well.

### Did you update your dependencies?

- [X] Yes, I’ve updated my dependencies to use the latest version of all packages.

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

tonyhallett:
For your issue "textBetween cannot get text inside custom extensions" the problem is Node.nodesBetween is unaware of renderText - ProseMirror has leafText instead.  TipTap has a helper getTextBetween but you need to provide text serializers which you can get from getTextSerializersFromSchema.  This will use your renderText.

PR #4444, opened 2023-09-13
fix: #4327

tonyhallett:

Fix issue 4327

> It uses Node.textContent which uses NodeSpec.leafText to fill it. But Tiptap does not fill NodeSpec.leafText using Tiptap's Node.renderText, rather Tiptap has its own separate mechanism to generate text content from Nodes

Used Tiptap method of getting text, which use renderText, instead of textContent which does not.

Added a new example with associated spec.

Run the test or view the example and see that there is no menu shown editor has focus.

## Remarks

Note that the spec is currently not with the ui example due to [cypress issue on windows](https://github.com/cypress-io/cypress/issues/8599). (Perhaps demos folder could go in tests ? )
If pr is approved then I will move the spec in to the demos folder.
Also the ui is in Examples - do you want in Experiments / a new Issues folder ?
Should I expose a helper - getText(node,schema,blockseparator) ?

As this change uses TipTap renderText and not ProseMirror leafText the result will differ if Node extension is provided that has added leafText with extendNodeSchema or changed the schema in onBeforeCreate.
I don't think that any public repos are doing this.
https://github.com/search?q=tiptap+leafText&type=code
Regardless, The getTextSerializersFromSchema method could be changed to invoke the spec leafText if it had been provided.  This would also ensure that editor.getText, generateText and the clipboardTextSerializer can also work with leafText.

Similar issues to this one where renderText is not used ( and  spec.leafText is )
[Node.textContent](https://github.com/search?q=repo%3Aueberdosis%2Ftiptap+.textcontent&type=code) is used in other places in Tiptap.  Potentially issues there too.
[Node.textBetween](https://github.com/ProseMirror/prosemirror-model/blob/387576a887335b5dccba4c940b7935e1da13ef73/src/node.ts#L100) => [Fragment.textBetween](https://github.com/ProseMirror/prosemirror-model/blob/387576a887335b5dccba4c940b7935e1da13ef73/src/fragment.ts#L64) => invokes leafText and not renderText.
Should [Tiptap usages of Node.textBetween](https://github.com/search?q=repo%3Aueberdosis%2Ftiptap+.textBetween&type=code) be changed to use [Tiptap getTextBetween ( and schema text serializers)](https://github.com/ueberdosis/tiptap/blob/7832b96afbfc58574785043259230801e179310f/packages/core/src/helpers/getTextBetween.ts#L5) ?

renderText and how to ensure it is used is not well documented in the docs.  This is the only mention
https://tiptap.dev/api/schema

> Here is the example from the [Mention](https://tiptap.dev/api/nodes/mention) extension:
```
renderText({ node }) {
  return `@${node.attrs.id}`
},
```
(This is not the case anymore for mention extension )
```
  renderText({ node }) {
    return this.options.renderLabel({
      options: this.options,
      node,
    })
  },
```
As such users expect renderText to be used when using the ProseMirror api - https://github.com/ueberdosis/tiptap/issues/4310

It is possible to add leafText to the spec so that it calls renderText although the signatures are different.  ( Could use onBeforeCreate to get the editor and the doc for parent and index but impossible to get the pos argument.  ) 

## Checklist

- [x] Followed the guidelines
- [x] Fixed linting issues

## Related issues

https://github.com/ueberdosis/tiptap/issues/4327

### DIFF
--- a/demos/src/Examples/Issue4327/React/foo.ts
+++ b/demos/src/Examples/Issue4327/React/foo.ts
@@ -1,0 +1,26 @@
+import { mergeAttributes, Node } from '@tiptap/core'
+
+export default Node.create({
+  name: 'foo',
+
+  group: 'inline',
+
+  inline: true,
+
+  parseHTML() {
+    return [
+      {
+        tag: 'span',
+        getAttrs: node => (node as HTMLElement).hasAttribute('data-foo') && null,
+      },
+    ]
+  },
+
+  renderHTML({ HTMLAttributes }) {
+    return ['span', mergeAttributes({ 'data-foo': '', HTMLAttributes }), 'foo']
+  },
+
+  renderText() {
+    return 'foo'
+  },
+})
--- a/demos/src/Examples/Issue4327/React/index.tsx
+++ b/demos/src/Examples/Issue4327/React/index.tsx
@@ -1,0 +1,27 @@
+import './styles.scss'
+
+import { EditorContent, FloatingMenu, useEditor } from '@tiptap/react'
+import StarterKit from '@tiptap/starter-kit'
+import React from 'react'
+
+import Foo from './foo.js'
+
+export default () => {
+  const editor = useEditor({
+    extensions: [
+      StarterKit, Foo,
+    ],
+    content: `
+      <p><span data-foo=''>foo</span></p>
+    `,
+  })
+
+  return (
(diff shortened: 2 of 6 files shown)


bdbch:
@tonyhallett I moved your commit over to #5838 and added a little bit on top. Closing this PR for that.

PR #4453, opened 2023-09-15
Fix React Dependency Array creating new editor instances instead of reusing existing one

### LINKED ISSUE #2403: onUpdate callback does not update after re-render

I have a wrapper around tiptap, whenever the text changes I trigger a request to my back-end, this works fine the first time  tiptap is mounted but the parent component (where tiptap is mounted) can change its internal variables, and therefore the closure should capture a new context, the problem is that it doesn't after the parent component changes the state, the closure/lambda passed on the `onUpdate` function remains the same and therefore tiptap tries to update the wrong component.

Here is some of the code, my high level component on the parent, notice the `id` param, which is the param that changes at some point:

```ts
<Tiptap
  onFocus={({ editor }) => editor.commands.blur()}
  initialContent={project.notes ? JSON.parse(project.notes) : null}
  placeholder="You can add a default checklist in the settings."
  className="md:max-w-2xl lg:max-w-none"
  onChange={async (e) => {
    console.warn("URL PARAM ID", id) // ALWAYS REMAINS THE SAME, THEREFORE CANNOT UPDATE THE PROJECT CORRECTLY

    await updateProjectMutation({
      id,
      notes: JSON.stringify(e),
    })
    refetch()
  }}
  ref={tiptapRef}
/>
```

My internal TIptap implementation, notice the `onUpdate` function that I'm passing to the `useEditor` hook:

```ts
import Link from "@tiptap/extension-link"
import Placeholder from "@tiptap/extension-placeholder"
import TaskItem from "@tiptap/extension-task-item"
import TaskList from "@tiptap/extension-task-list"
import { BubbleMenu, EditorContent, Extension, useEditor } from "@tiptap/react"
import StarterKit from "@tiptap/starter-kit"
import React, { forwardRef, useImperativeHandle, useState } from "react"
import { useBoolean } from "../hooks/useBoolean"
import { Button } from "./Button"

interface IProps {
  editable?: boolean
  onClick?: (this: unknown, view: any, pos: number, event: MouseEvent) => boolean
  initialContent?: any
  // content?: any
  onChange?: (content: any) => void
  autofocus?: boolean | null | "end" | "start"
  onFocus?: (params: { editor: any }) => void
  placeholder?: string
  className?: string
}

export const Tiptap = forwardRef<any, IProps>(
  (
    {
      editable = true,
      onClick,
      initialContent,
      onChange,
      autofocus,
      onFocus,
      placeholder,
      className,
      // content,
    },
    ref
  ) => {
    const [isAddingLink, addLinkOn, addLinkOff] = useBoolean()
    const [link, setLink] = useState("")
    const editor = useEditor({
      autofocus,
      onFocus: onFocus ? onFocus : () => {},
      editorProps: {
        attributes: {
          class: "prose focus:outline-none dark:prose-dark dark:text-gray-300 text-base",
        },
        editable: () => editable,
        handleClick: onClick,
      },
      content: initialContent,
      onUpdate: ({ editor }) => {
        onChange?.(editor.getJSON())
      },
      extensions: [
        StarterKit,
        Placeholder.configure({
          showOnlyWhenEditable: false,
          placeholder,
        }),
        TaskList.configure({
          HTMLAttributes: {
            class: "pl-0",
          },
        }),
        TaskItem.configure({
          HTMLAttributes: {
            class: "before:hidden pl-0 flex items-center dumb-prose-remove",
          },
        }),
        Extension.create({
          // Do not insert line break when pressing CMD+Enter
          // Most of the time handled by upper components
          addKeyboardShortcuts() {
            return {
              "Cmd-Enter"() {
                return true
              },
              "Ctrl-Enter"() {
                return true
              },
            }
          },
        }),
        Link,
      ],
    })

    useImperativeHandle(ref, () => ({
      getEditorInstance() {
        return editor
      },
    }))


    return (
        <EditorContent editor={editor} className={className} />
    )
  }
)
```

In any case, it seems the useEditor hook saves only the first passed `onUpdate` function and does not update it in sub-sequent renders

Attached the code above, but if necessary I can try to reproduce the issue in a code sandbox

_No response_

The passed callback `onUpdate` should be updated when a new value is passed to it, instead of constantly re-using the first memoized value

I tried to update tiptap to the latest version but then I faced this other crash: https://github.com/ueberdosis/tiptap/issues/577 so I reverted to my old/current versions

```
"@tiptap/extension-bubble-menu": "2.0.0-beta.51",
    "@tiptap/extension-link": "2.0.0-beta.33",
    "@tiptap/extension-placeholder": "2.0.0-beta.45",
    "@tiptap/extension-task-item": "2.0.0-beta.30",
    "@tiptap/extension-task-list": "2.0.0-beta.24",
    "@tiptap/react": "2.0.0-beta.98",
    "@tiptap/starter-kit": "2.0.0-beta.154",
```

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

bdbch:

Before this PR everytime dependencies change the whole editor gets replaced with a new instance. This would remove entered content, cause flashes and gets rid of the editor state (which sucks for example because of the history living in the state for example)

This PR implements a check that will update the editor options when the deps change instead of creating a new instance with said options.

Added a check in the react useEditor hook to update the editor instead of replacing it.

I tested my changes on the demo page I created (Link: https://deploy-preview-4453--tiptap-embed.netlify.app/src/issues/2403/react/)

See above

1. Write something - in the console you should see `0`
2. Click on Inc a few times
3. Write something - now the onUpdate function inside the editor should be updated
4. Check the editor class (it should be `my-editor`
5. Click on the Update class button
6. The editor class should now correctly be `my-editor updated-editor`

## Remarks

On my tests everything worked fine for now but I think we'd need to verify that this change is not breaking some special cases.

## Checklist

- [x] Followed the guidelines
- [x] Fixed linting issues

## Related issues

- fixes #2403


bdbch:
@svenadlung could you take a look?

svenadlung (approved):

packages/react/src/useEditor.ts
```diff
@@ -98,24 +98,31 @@ export const useEditor = (options: Partial<EditorOptions> = {}, deps: Dependency
...
-    })
+    } else {
+      editor.setOptions(options)
+    }
 
     return () => {
       isMounted = false
     }
-  }, deps)
+  }, [deps])
```
Slapbox, line 125: This line wrongly double-wraps the dependencies in an additional array.

Slapbox:
I noticed an issue that I commented on.

In theory this seems like exactly what we need to avoid the insanity of passing a zustand store to our Extensions just to give them up-to-date functions to read from, or otherwise passing down all of their relevant arguments. In practice it doesn't seem to work that way though.

But even after implementing this with the line I commented on fixed, it seems like the option passed to an Extension are not updated. Eg we are trying to get data from an external store but the function is still using the outdated version of the store. I'd expect that `editor.setOptions(options)` would make the extensions use the latest versions of their `options`. Is that not intended to be the case?

Specifically I'd expect `addProseMirrorPlugins` to run anew, but I guess that may not be the intention since I imagine that would break plugin states. @bdbch can you clarify how this should work?

Also, just for reference, this will break any code that relied on the editor re-rendering on every update of the editor, eg for a `ControlledBubbleMenu` or the like, (although that should be trivial for users to fix.)

andrictham:
@Slapbox 

> exactly what we need to avoid the insanity of passing a zustand store to our Extensions just to give them up-to-date functions to read from, or otherwise passing down all of their relevant arguments

This is close to the approach I’m taking right now for the Mention suggestions list to be reactive, except I used Legend State.

I’m also unsure if this PR being merged will mean I no longer have to do that. 

I’m not familiar with TipTap’s internals and how all their extensions are authored. But in my specific case, I was interested in using the Mention extension and wanted the suggestions list to be reactive. 

Reading their code for the Mention and Suggestion extensions, it seems that, potentially, the issue is caused by suggestions being passed into Prosemirror as a prop. From what I’ve gathered, unlike React, Prosemirror views don’t update when props change, only when state changes. So, if you pass in something that you expect to change, you lose reactivity once you pass it down. 

I don’t know if this is exactly the issue, but here’s [a related discussion thread](https://discuss.prosemirror.net/t/props-to-plugin-views/4672/7) on the Prosemirror forums.

Palanikannan1437:
@bdbch @svenadlung would love to see this get merged 🙌

packages/react/src/useEditor.ts
bdbch: Thanks - fixed

bdbch:
> But even after implementing this with the line I commented on fixed, it seems like the option passed to an Extension are not updated. Eg we are trying to get data from an external store but the function is still using the outdated version of the store. I'd expect that editor.setOptions(options) would make the extensions use the latest versions of their options. Is that not intended to be the case?

Right now `setOptions` only updates the editors state & editor props but does not run any extension setup again, see here:

https://github.com/ueberdosis/tiptap/blob/6f6e49ecdf78cf6fd63ad1eb7006fb5afd4064a1/packages/core/src/Editor.ts#L159-L164

Before this PR the editor instance actually gets completely replaced which would mean that the extension manager will also run through another time.

Could you maybe createa Codesandbox / Stackblitz showcasing what you want to do with Prosemirror plugins accessing external data, I could try to use that to work off from.

> Specifically I'd expect addProseMirrorPlugins to run anew, but I guess that may not be the intention since I imagine that would break plugin states. @bdbch can you clarify how this should work?

I think that would be a case yes - there's no easy way to persist a plugin state.

> Also, just for reference, this will break any code that relied on the editor re-rendering on every update of the editor, eg for a ControlledBubbleMenu or the like, (although that should be trivial for users to fix.)

That's true - @svenadlung what your thought on that? That would mean we need to push the version up to 3.0.0 for the next release with this PR merged.

---

> I’m not familiar with TipTap’s internals and how all their extensions are authored. But in my specific case, I was interested in using the Mention extension and wanted the suggestions list to be reactive.

This PR won't fix that as the mention plugin won't rerun (see above). But when it comes to updating Prosemirror props & state the `setOptions` function does that:

https://github.com/ueberdosis/tiptap/blob/6f6e49ecdf78cf6fd63ad1eb7006fb5afd4064a1/packages/core/src/Editor.ts#L159-L164

---

In general I think there are a few misconceptions right now when it comes to reactivity and the `useEditor` hook - while the hook will run whenever you update the dependencies - the editor won't just take all reactive props and update them down to all plugins and extensions registered (as for now).

Feel free to send in Sandboxes which we can use to work from as references or maybe even open a PR for that issue if you find a good fix for this - I'd be down to play around with those reactivity problems.

Back to the suggestion issue: What I did was completely doing data fetching / reactivity inside my React SuggestionList component rendered inside my suggestion popup which was reacting to query changes. That means I didn't use the items array at all. (Not the cleanest solution but worked).

nperez0111:

Issue #4467, opened 2023-09-21
[Bug]: Inline code formatting not applied consistently with backticks


marwan37:
### Which packages did you experience the bug in?

extension-code

### What Tiptap version are you using?

2.2.0-rc.3

### What’s the bug you are facing?

This issue was originally identified in the BlockNote project. For additional context and discussions, see [Issue #338 in BlockNote](https://github.com/TypeCellOS/BlockNote/issues/338).

When using backticks to format inline code, the expected behavior is that any text enclosed between two backticks should be formatted as inline code, regardless of the surrounding characters. However, the current implementation in TipTap does not consistently apply this formatting, especially when the backticks are surrounded by characters like parentheses, brackets, or other punctuation marks.

Issue can be replicated from the TipTap playground for the `Code` extension:
<img width="777" alt="TipTapPlayground-Code" src="https://github.com/ueberdosis/tiptap/assets/99911676/51d84749-54d4-49b7-be28-80b00cc50ec1">

### What browser are you using?

Chrome

### What did you expect to happen?

I expected that any text enclosed between two backticks, irrespective of surrounding characters, would be consistently formatted as inline code. This behavior should be uniform across different contexts and scenarios.

### Did you update your dependencies?

- [X] Yes, I’ve updated my dependencies to use the latest version of all packages.

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

nperez0111:
v2.10.3 resolves this

Issue #338, opened 2019-06-03
Image: Support `inputRules`


michalsnik:
**Is your feature request related to a problem? Please describe.**
Currently the only ways to insert images are:
- dragging image into the editor
- using command declared by `Image` to insert `<img>` tag with proper attribute

It's not possible however to use Markdown syntax to automatically insert images by using typing the following:
```
![Example alt text](image.jpg "Optional title")
```

**Describe the solution you'd like**
The proposed solution includes creating inputRule that would transform above syntax into Image node automatically.

**Describe alternatives you've considered**
This is a fairly simple feature to implement and I my current solution is creating new Image class that extends one provided by this lib, but as some of the other Nodes already have this feature and support Markdown syntax out of the box I though it might be a good addition to the core implementation too.

philippkuehn:
It's released.

PR #4468, opened 2023-09-21
Fix inline code formatting for text enclosed in backticks

marwan37:

In the current implementation of the [@tiptap/extension-code](https://github.com/ueberdosis/tiptap/blob/39cf6979c49e953118bbbe4b894a1dc296128932/packages/extension-code/src/code.ts#L4), the regex pattern does not correctly format text enclosed in backticks when they are immediately preceded or followed by any character other than whitespace.

Updated the `inputRegex` and `pasteRegex` patterns to the following:

```ts
export const inputRegex = /(?<!`)`([^`]+)`(?!`)/;
export const pasteRegex = /(?<!`)`([^`]+)`(?!`)/g;
```

The regex uses negative lookbehind and lookahead assertions to ensure that only single backticks are considered for formatting to avoid conflicts with potential code block syntax or other uses of multiple backticks.

Tested in various scenarios to ensure that it correctly formats text as inline code blocks, and that it doesn't conflict with other extensions.

TipTap-LocalTest

1. Clone the repository and switch to the branch containing these changes.
2. Start the app via `npm run start` and navigate to any playground
3. Type text enclosed in backticks, with various surrounding characters.
4. You should observe the text within the backticks being correctly formatted as inline code.
5. Additionally, test with multiple backticks to ensure they don't interfere with the formatting.

## Remarks

This change enhances the flexibility of the inline code formatting.

## Checklist

- [x] Followed the guidelines
- [x] Fixed linting issues

## Related issues

#4467 
[Issue #338 in BlockNote](https://github.com/TypeCellOS/BlockNote/issues/338) - Original issue where the bug was identified and discussed.


packages/extension-code/src/code.ts
```diff
@@ -34,14 +34,21 @@ declare module '@tiptap/core' {
...
+ * Regular expressions to match inline code blocks enclosed in backticks.
+ *  It matches:
+ *     - An opening backtick, followed by
+ *     - Any text that doesn't include a backtick (captured for marking), followed by
+ *     - A closing backtick.
+ *  This ensures that any text between backticks is formatted as code,
+ *  regardless of the surrounding characters (exception being another backtick).
  */
-export const inputRegex = /(?:^|\s)(`(?!\s+`)((?:[^`]+))`(?!\s+`))$/
+ */
```
nperez0111, line 45:
```suggestion

```

nperez0111 (approved):
Thanks for your contribution @marwan37, sorry for the long wait!

rfgamaral:
This change introduced the following issue:

firefox_gjEVM9odIA

I spent some time trying to tweak the RegExp to keep this fix and avoid the bug that introduced, but I have a feeling that tweaking the RegExp will not be enough, I think the `markInputRule` and `markPasteRule` functions play a part in this.

@marwan37 Think you can have a look since you were the original author of this change?

_(cc @nperez0111 for awareness; not sure if a fix for this is that straightforward, but if it's not, maybe a revert could be in order?)_

rfgamaral:
> @rfgamaral what version are you using, because we've changed the regex since this was closed to remove the backtracking in: #5916

I tried with both the version that introduced this PR, and the latest, the issue exists on both.

That PR changed the implementation of the regex to not include lookbehind assertions (unsupported by Safari below 16.4), but it kept the RegExp change that introduces the new behaviour described in this PR (which also introduces the bug I just described).

PR #4514, opened 2023-10-09
Update image.ts to include style and id

dsti-arobb:

We have incorporated the inclusion of ids and styles into the image extensions, enhancing the capability to control image style and dimensions. The implementation of this code has enabled us to efficiently assign and retrieve these IDs and styles as required within our system.

Added id and style attributes to the index.ts file.

Updating and changing the styles and ids on pictures that are added to the editor. 

(For Exmaple: this.editor.chain().focus().setImage({ src: imgSrc,  style: 'width: 100%', id: 'full-width-image'}).run() )

Try to add images to the editor with id or style attributes (example above).

## Remarks

Let me know your thoughts. 

Thanks!

## Related issues

https://github.com/ueberdosis/tiptap/issues/4103

### DIFF
--- a/packages/extension-image/src/image.ts
+++ b/packages/extension-image/src/image.ts
@@ -55,6 +55,12 @@ export const Image = Node.create<ImageOptions>({
       title: {
         default: null,
       },
+      style: {
+        default: null,
+      },
+      id: {
+        default: null,
+      },
     }
   },
 


bdbch:
Thanks for your PR @dsti-arobb

I don't think those attributes to the image node should be added by default since we don't do the same on other nodes. If we would like to have those added to all nodes, we should add them via `addGlobalAttributes` somewhere else, but I think we should register less if not needed.

bdbch:
I'm open for discussion though - if we can find a good solution on why those should be added by default I'm open to reopen this PR.

PR #4545, opened 2023-10-18
Added CSS.escape to font-family.ts

SanderLeenders:

Added CSS.escape to renderHTML. Prevents invalid css when using fonts with numbers in their names, like https://fonts.google.com/specimen/Exo+2

Added CSS.escape() to renderHTML

Tested local 

Before Screenshot 2023-10-18 at 15 07 03
After
image

## Remarks
-

## Checklist

- [x] Followed the guidelines
- [x] Fixed linting issues

### DIFF
--- a/packages/extension-font-family/src/font-family.ts
+++ b/packages/extension-font-family/src/font-family.ts
@@ -44,7 +44,7 @@ export const FontFamily = Extension.create<FontFamilyOptions>({
               }
 
               return {
-                style: `font-family: ${attributes.fontFamily}`,
+                style: `font-family: ${CSS.escape(attributes.fontFamily)}`,
               }
             },
           },
--- a/packages/extension-font-family/src/font-family.ts
+++ b/packages/extension-font-family/src/font-family.ts
@@ -44,7 +44,7 @@ export const FontFamily = Extension.create<FontFamilyOptions>({
               }
 
               return {
-                style: `font-family: ${CSS.escape(attributes.fontFamily)}`,
+                style: `font-family: ${attributes.fontFamily.split(',').map((fontFamily: string) => CSS.escape(fontFamily.trim())).join(', ')}`,
               }
             },
           },


janthurau:
@SanderLeenders I think this will also escape `,`, if you specify multiple fonts, right? So we'd probably need to escape only the real font names. See on the demo:

image

SanderLeenders:
@janthurau  you are right! I've added a fix for this.

janthurau (approved):

ricardasnav:
Hi
This change broke rendering on the server side where CSS is not available as a global variable. I use Cloudflare workers to render the output and now it throws.

nperez0111:
I'm not sure that this was the right move, I think I'm going to make a PR to revert this

jonahallibone:
> I'm not sure that this was the right move, I think I'm going to make a PR to revert this

This SO answer provides an alternate solution:

https://stackoverflow.com/questions/7638775/do-i-need-to-wrap-quotes-around-font-family-names-in-css

nperez0111:
> > I'm not sure that this was the right move, I think I'm going to make a PR to revert this
> 
> This SO answer provides an alternate solution:
> 
> [stackoverflow.com/questions/7638775/do-i-need-to-wrap-quotes-around-font-family-names-in-css](https://stackoverflow.com/questions/7638775/do-i-need-to-wrap-quotes-around-font-family-names-in-css)

Sorry there are a few recommendations here are you saying that we should wrap with `&quot;`? That feels hackish.

Or maybe we can do what we did before and just wrap in `"` if it contains a number or space

SanderLeenders:
Hi @nperez0111, have you seen my previous comment? I’d appreciate your input on this. 

I agree that my solution was causing issues (e.g., CSS variables), but I also believe that TipTap's output should be a valid input for TipTap. If you have any suggestions, I’d love to hear them.

nperez0111:
Sorry for not getting back to you @SanderLeenders, I think the easiest thing to do here is to just have a basic polyfill for CSS.escape like this:

```ts
function cssEscape(value: string): string {
  const length = value.length;
  let result = '';

  for (let i = 0; i < length; i++) {
    const char = value.charAt(i);
    const code = char.charCodeAt(0);

    // If the character is a control character, or a non-printable character, or a space
    if (code < 0x20 || code === 0x7F || (code >= 0x80 && code <= 0x9F) || char === ' ') {
      result += '\\' + code.toString(16) + ' ';
    } else {
      // If the character is a special character in CSS
      switch (char) {
        case '\0':
          result += '\uFFFD';
          break;
        case '"':
        case '#':
        case '$':
        case '%':
        case '&':
        case '\'':
        case '(':
        case ')':
        case '*':
        case '+':
        case ',':
        case '.':
        case '/':
        case ':':
        case ';':
        case '<':
        case '=':
        case '>':
        case '?':
        case '@':
        case '[':
        case '\\':
        case ']':
        case '^':
        case '`':
        case '{':
        case '|':
        case '}':
        case '~':
          result += '\\' + char;
          break;
        default:
          result += char;
      }
    }
  }

  return result;
}

// Example usage
const escaped = cssEscape('Hello, world!');
console.log(escaped); // Output: Hello\,\ world\!
```

nperez0111:
@SanderLeenders some comments on this here: https://github.com/ueberdosis/tiptap/pull/5164

nperez0111:
@SanderLeenders on further review, I think you are right that we should not be doing this quote removal on parsing the HTML. Can I leave this to you to open a new PR for this?

As long as we have our existing tests pass & any new cases that were not being handled I'd be happy to accept a PR here

PR #4553, opened 2023-10-22
collab / cloud: adds documentation about the PATCH call

janthurau:
This adds documentation about the new PATCH document endpoint of the collab API.
I've also added docs about the new GET /documents endpoint, which returns a list of all documents available on the 
server.
Also adds docs about axios pitfall when trying to get the document in yjs format, and adds code example to duplicate a document


svenadlung (dismissed):

janthurau:
Moved to https://github.com/ueberdosis/tiptap/pull/4553

PR #4561, opened 2023-10-24
feat(suggestion): add option to allow triggering chars to also be in suggestions

### LINKED ISSUE #4447: [Documentation]: Mentions - allowSpaces also allows email search
### What’s the URL to the page you’re sending feedback for?

https://tiptap.dev/api/utilities/suggestion

### What part of the documentation needs improvement?

allowSpaces
Allows or disallows spaces in suggested items.

Default: false

### What is helpful about that part?

It is helpful to know that we can allow spaces without closing the mentioning functionality

### What is hard to understand, missing or misleading?

It is missing to explain that it also allows you to add a second "@" char in the string without closing the mention. Without it, it was not possible to search by email, as typing "@sara@" would close mentions on that last char input.

I dont know if that was intentional or unintentional, but it was an issue that I encountered with it so really glad that it can get fixed that way!

tcitworld:

It's currently not possible to mention emails or federated usernames for instance, because the triggering char (often `@`) cannot be included in the mention suggestion itself. This option allows to have some, so that the mention will only end with a space. That makes it incompatible with the `allowSpaces` option.

Added the option in question and modified the regex matching the suggestion text to use it.

Tested that the new regex will also include the part after a secondary triggering character.

Add `allowToIncludeChar` to the suggestion settings, then type an email as suggestion.

## Remarks

## Checklist

- [x] Followed the guidelines
- [x] Fixed linting issues

## Related issues

Closes #4447

### DIFF
--- a/docs/api/utilities/suggestion.md
+++ b/docs/api/utilities/suggestion.md
@@ -21,6 +21,11 @@ Allows or disallows spaces in suggested items.
 
 Default: `false`
 
+### allowCharInSuggestedItem
+Allows or disallows the triggering char in suggested items. This option is incompatible with the `allowSpaces` option.
+
+Default: `false`
+
 ### allowedPrefixes
 The prefix characters that are allowed to trigger a suggestion. Set to `null` to allow any prefix character.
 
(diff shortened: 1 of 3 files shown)


bdbch:
Moved this into #5895

tcitworld:
Thanks !

PR #4579, opened 2023-10-29
Remove intermediate render cycle from useEditor

### LINKED ISSUE #3345: Editor is being rendered with setTimeout (async)

The Editor is being rendered asynchronously. Here on this line, the initialization is set after timeout (which makes it asynchronous). 

https://github.com/ueberdosis/tiptap/blob/4851fc5e9b6daccc15a1839e471db489401eca0c/packages/core/src/Editor.ts#L98

This makes it hard to use TipTap with other libraries that work with `clientHeight` (e.g. [react-window](https://github.com/bvaughn/react-window)).

All browsers. 

Yes, please see attached codesandbox.

Here's a codesandbox demo demonstrating the problem - there is height calculated immediately after rendering and one after 1ms timeout. 

https://codesandbox.io/s/cranky-feather-9tf083?file=/src/App.js

I would expect that TipTap is initialized and rendered synchronously.

_No response_

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

C-Hess:
~~May or may not be a breaking change depending on if any people _actually_ required this behavior. That would surprise me though.~~

Was a breaking change because it broke SSR, though that's probably not something TipTap should support long term anyways. Instead, opted for an alternative hook to be used until a major version change.

Possibly fixes #3345 (issue can probably be re-opened)

We've ran issues getting TipTap operating smoothly with virtual scrolling libraries as well because they ideally expect the height of a component be defined after the first render cycle. While we can't guarantee that Prosemirror doesn't introduce their own delays in rendering, I think the least we can do is remove the unnecessary delay in the React TipTap wrapper.

The change really is just moving initialization to the first render cycle. As a nice side effect, this also means you also no longer have to check if the editor is null before using it ❤

~~I ran through most of the tests and they seem to be passing. I spot checked a few test failures and they appear unrelated to the changes made.~~

No longer relevant. New approach does not touch existing hook.

Verification might be a little involved. Didn't really want to get react virtual scrolling all setup in an example at this time, plus like I said, there is no guarantee's it would eliminate the problem entirely. At a minimum though, it removes an unnecessary render cycle.

## Checklist

- [X] Added tests where possible
    - It doesn't really seam testable with the current fixures
- [X] Followed the guidelines
- [X] Fixed linting issues

## Related issues

#3345


C-Hess:
@janthurau , Sorry to put pressure, but any chance that we could get this merged? It would be awesome for us because it could potentially help resolve issues with virtual scrolling libraries.

janthurau:
hey @C-Hess, absolutely. We haven't had a look here yet as the tests are still failing:

```
7:37:54 AM: ../packages/react/src/Context.tsx(36,26): error TS2322: Type "Editor | null" is not assignable to type "Editor".
7:37:54 AM:   Type "null" is not assignable to type "Editor".
```

Do you think you can fix them?

C-Hess:
Oops 😳, at one point I thought the build was failing for unrelated reasons. I'll resolve the issue, sorry

ondrejsevcik:
@C-Hess thanks for taking initiative to fix this issue 🙏🏼 highly appreciated.

ondrejsevcik:
It would be good to hear from the owners of the TipTap to understand their perspective on this topic.

In my opinion it would be better to split it up into two separate use cases. SSR for TipTap editor does not make much sense. The WYSIWYG is user facing component that will always struggle in server side environment. Trying to bend it will bring unecessary complexity. I would rather encourage to use [HTML helpers](https://tiptap.dev/docs/editor/api/utilities/html) for SSR instead.

C-Hess:
I changed the approach to be non-breaking. In a future major version of TipTap, I'm hoping that useEditorForImmediateRender can replace the logic within useEffect with the caveat that TipTap will no longer be usable through server side rendering (which from my understanding, it shouldn't have been usable via server-side rendering anyways)

packages/react/src/useEditor.ts
```diff
@@ -113,14 +113,121 @@ export const useEditor = (options: Partial<EditorOptions> = {}, deps: Dependency
 
     return () => {
       isMounted = false
+      editorRef.current?.destroy()
```
C-Hess, line 116: This also should fix the memory leak concerns from my understanding following the change to ref here

C-Hess:
Going to rely on CI for testing as I'm not home yet to verify my updates. I updated the React editor context and main demo file to optionally use the newer hook.

I'll test this later tonight

C-Hess:
Should be good to review @bdbch and @svenadlung. Let me know if you want to tweak the approach I went with here. Thanks!

corbinkems:
I'm also running into this issue and it's causing a lot of headaches. @janthurau

packages/react/src/useEditor.ts
```diff
@@ -113,14 +114,130 @@ export const useEditor = (options: Partial<EditorOptions> = {}, deps: Dependency
...
-  return editorRef.current
+  const [, forceUpdate] = useState({})
+  const editor = useMemo(() => {
+    const instance = new Editor(options)
+
+    instance.on('transaction', () => {
+      requestAnimationFrame(() => {
+        requestAnimationFrame(() => {
+          if (isMounted.current) {
+            forceUpdate({})
```
ondrejsevcik, line 167: Not required in this PR, but it would be nice to add comment here why is this section necessary (to call the forceUpdate wrapped in 2 requestAnimationFrame calls on every editor transaction). This is a bit black magic to me.

C-Hess:
@bdbch , @svenadlung, @janthurau sorry to bother you all. Is there any chance this could be looked at? I see two other PRs introduced by you both that both have different ways of doing this on top of this one. Each approach has pros and cons, but it would be awesome for us to be able to resolve this issue. This is a high priority for my organization to help hopefully resolve virtual scrolling issues when using TipTap, so I'd be more than happy to make any suggested changes as soon as possible if requested.

Thank you all for your hard work on this Repo!

packages/react/src/useEditor.ts
C-Hess: Same here. I just copied this over from the original hook, but it looks to me like it's logic to get around timing issues between Prosemirror transactions and React render cycles

hjoelh:
Ran into this issue today, until this or another fix is merged - my current workaround is like this

```
  {/* 
      due to https://github.com/ueberdosis/tiptap/pull/4579
      editor is null on initial render 😢, to prevent a flicker we render 
      an empty box with the same styles as the edtior 
      */}
      {editor ? (
      ) : (
        <Box className={styles["editor"]} />
      )}
```

packages/react/src/useEditor.ts
szymonchudy: This seems to solve the flickering issue mentioned here: https://github.com/ueberdosis/tiptap/issues/2040

packages/react/src/Context.tsx
```diff
@@ -19,18 +19,38 @@ export const useCurrentEditor = () => useContext(EditorContext)
 
 export type EditorProviderProps = {
   children: ReactNode;
+  /**
+   * This option will create and immediately return a defined editor instance. The editor returned in the context consumer will never be null if
+   * this is enabled. In future major versions, this property will be removed and this behavior will be the defualt.
```
Nantris, line 24: Typo defualt to default

Nantris:
@C-Hess would this potentially address #5166? I think probably not based on quickly looking over it, but figured I'd ask to be sure.

packages/react/src/useEditor.ts
Nantris: Unfortunately there is a habit in this repo of closing non-fixed issues.

@szymonchudy is that issue separate from #5166? It's hard to tell if they're related or truly the same.

nperez0111:
I took a stab at this, taking a slightly different approach for it: https://github.com/ueberdosis/tiptap/pull/5161

nperez0111:

nperez0111:
@C-Hess Very much appreciate your contribution, was definitely a source of inspiration for my change

Issue #3908, opened 2023-03-29
style attributes are getting duplicated


gethari:
### Which packages did you experience the bug in?

core

### What Tiptap version are you using?

2.0.0.rc-1

### What’s the bug you are facing?

We are trying to support `style` tags which can be manipulated by the end users who consume the editor, due to which we are doing following steps to achieve it. But the problem is on the editor's content DOM, the style attributes values are getting duplicated multiple times. We are unsure if this is a bug in the framework, or we are doing something wrong here. 

## Minimal explanation of what we are trying to do

```ts
 addAttributes() {
    return {
      ...this.parent?.(),
      style: {
        default: null,
      },
    };
  },

parseHTML() {
    return [
      {
        tag: "span",
        getAttrs: (dom) => {
          const element = dom;

          return {
            style: element.getAttribute("style"),
          };
        },
      },
    ];
  },
```

Attached an example [Stackblitz](https://stackblitz.com/edit/react-dbfmbq?file=src%2Fstyle.css,src%2FTiptap.js,src%2FMenuBar.js,src%2Findex.js), in which the behavior can be seen by just hitting `Enter` or changing the color of the text inside the Editor.

### What browser are you using?

Chrome

### Code example

https://stackblitz.com/edit/react-dbfmbq?file=src%2Fstyle.css,src%2FTiptap.js,src%2FMenuBar.js,src%2Findex.js

### What did you expect to happen?

Style attributes should not get duplicated.

### Anything to add? (optional)

- We are extending the default's editor extensions such as `Paragraph` / `TextStyle` / & when the color/ background color attribute is getting applied on top of the paragraph, the values are being duplicated
- We could see the same problem happening across all the extensions when we follow the same pattern
- We are unsure if this is a internal bug in tiptap or even maybe from ProseMirror itself,

### Did you update your dependencies?

- [X] Yes, I’ve updated my dependencies to use the latest version of all packages.

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

bdbch:
Perfect issue, thanks. I'll see that we prioritize it soon!

tonyhallett:
Looks like user error to me.  If it is not then this is why you are getting your result.

The TextStyle, in original form, is a Mark that matches a span with a style attribute ( yours on the other hand matches any span).
```
  parseHTML() {
    return [
      {
        tag: 'span',
        getAttrs: element => {
[...]

PR #4610, opened 2023-11-08
Core: Fix styles being duplicated on mergeAttributes function

sifat009:

Before this change, the properties of style attribute on HTMLAttributes would be duplicated inside the mergeAttributes function, because there was no check if a style property already exists on the HTMLAttributes style properties.

This PR fixes this.

Added a check if the style property is already included inside the style attribute.

Tested locally in the demo with the following code
```
export default () => {
  const CustomTextStyle = TextStyle.extend({
    name: 'textStyle',

    addOptions() {
      return {
        HTMLAttributes: {},
      }
    },

    addAttributes() {
      return {
        ...this.parent?.(),
        style: {
          default: null,
        },
      }
    },

    parseHTML() {
      return [
        {
          tag: 'span',
          getAttrs: dom => {
            const element = dom

            return {
              style: element.getAttribute('style'),
            }
          },
        },
      ]
    },

      return [
        'span',
        mergeAttributes(this.options.HTMLAttributes, HTMLAttributes),
        0,
      ]
    },
  })

  const editor = useEditor({
    extensions: [Document, Paragraph, Text, CustomTextStyle, Color],
    content: `
        <span style="color: red; background: orange;  ">This has a &lt;span&gt; tag without a style attribute, so it’s thrown away.</span>
      `,
  })

  if (!editor) {
    return null
  }

  return <EditorContent editor={editor} />
}


```

You can just use the above code in any of the demo.

## Checklist

- [x] Followed the guidelines
- [x] Fixed linting issues

## Related issues

fixes https://github.com/ueberdosis/tiptap/issues/3908


packages/core/src/utilities/mergeAttributes.ts
```diff
@@ -23,7 +23,29 @@ export function mergeAttributes(...objects: Record<string, any>[]): Record<strin
...
+            updatedExistingStyleProperty => updatedExistingStyleProperty.join(':'),
+          ).join('; ')
+
+          const newStyleAttributes = newStyleProperties.map(
+            valueStyleProperty => valueStyleProperty.join(':'),
+          ).join('; ')
+
+          const mergedStyleAttributes = [existingStyleAttributes, newStyleAttributes].filter(attribute => !!attribute)
+
+          mergedAttributes[key] = mergedStyleAttributes.join('; ')
```
nperez0111, line 48: There has to be an easier way to do this, couldn't you just roll this up into a Map or an object (acting as `Record<AttributeName, AttributeValue>`)

Also would be great to have tests to understand the desired behaviors

nperez0111:
And sorry for the wait, really appreciate the contribution

nperez0111 (approved):

Issue #2690, opened 2022-04-07
Cannot select inline node at the end of a line or document


Billiam:
### What’s the bug you are facing?

Tiptap does not support selecting inline nodes with the mouse when dragging from right to left, if the node occurs at the end of a line (or is the only element on the line).

Selecting from left-to-right works, as does right-to-left _if_ the node is not at the end of the line.

Oddly, if the cursor is very slightly inside the node on mouse down, selection works as expected (see gif), but only for non-draggable nodes.

![tiptap-selection](https://user-images.githubusercontent.com/242008/162326170-76818e92-8ff5-42c9-9fd0-e76e6e5e465a.gif)

### How can we reproduce the bug on our side?

1. Visit https://tiptap.dev/api/nodes/emoji
2. Insert an emoji at the end of a line
3. Try to select the emoji with the mouse, beginning at the right side of the emoji and dragging to the left.

### Can you provide a CodeSandbox?

https://tiptap.dev/api/nodes/emoji

### What did you expect to happen?

Element should be selected, the same way it is if element is not at the end of the line.

### Anything to add? (optional)

This issue _does not_ occur in Prosemirror:
 
https://prosemirror.net/examples/dino/

![prosemirror-selection](https://user-images.githubusercontent.com/242008/162326534-63108b58-c6a4-4bc9-81d4-e2030c00bd98.gif)

### Did you update your dependencies?

- [X] Yes, I’ve updated my dependencies to use the latest version of all packages.

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

scottsidwell:
Having a look into this, it appears this behaviour is caused by the `ProseMirror-separator` node that is inserted by ProseMirror and custom styles Tiptap has applied to it. 

Specifically, setting `width: 1px` on `ProseMirror-separator` appears to trigger this behaviour in Chrome's content-editable (I couldn't reproduce the original bug on Safari or Firefox). This was introduced in https://github.com/ueberdosis/tiptap/commit/0e94afe42a5c15a47698152b3bc88e6bc4f8c01f. Demo of this being the culprit below:

https://user-images.githubusercontent.com/38820281/162338901-5170b9b2-68e1-450b-9305-ed88cb80b2da.mov

Perhaps @philippkuehn can weigh in with what styling the above commit was trying to solve; to work out if it's safe to revert - or to hunt around for CSS that won't cause this conflict :)

Billiam:
[...]

PR #4646, opened 2023-11-21
Make `img.ProseMirror-separator` 0px

tjenkinson:

Make `img.ProseMirror-separator` 0px because otherwise if there is an atom as the last thing on a line in chrome you can't click and drag to select back from it.

Tested on the demo in Chrome `119.0.6045.159` on https://tiptap.dev/api/nodes/mention with a few tweaks:

1. Make the mention nodes `display: inline-block` instead of `inline`. Without this it's still broken even with this fix, and I think `inline-block` still makes sense. Maybe the plugin should be updated to use `inline-block`? You can see with `inline` when you click at the end of the line the cursor actually ends up inside the mention instead of outside, which seems a bit strange.
2. Update the css for the img hack to have `width` and `height` as `0`.
3. Update the text to be a string and then a tag followed by nothing (which means you need to remove the empty space the plugin adds)
4. Try and click and drag to select from the end. It should work. With the `1px` image, it doesn't.

https://github.com/ueberdosis/tiptap/assets/3259993/5dc36c4e-f831-42ff-97a1-1b2f9460111e

Also tested these changes in our real project and it seems to work.

And tested in the deploy preview here.

See the steps above.

## Checklist

- [x] Followed the guidelines
- [x] Fixed linting issues

## Related issues

https://github.com/ProseMirror/prosemirror-view/pull/162


Billiam:
Related: https://github.com/ueberdosis/tiptap/issues/2690

Pruxis:
What's blocking this PR? This fixes several bugs for text selection

nperez0111 (approved):
Thank you for your contribution, I have tested this and believe it resolves several realted bugs

PR #4661, opened 2023-11-30
fix(react): fix performance regression because of select/deselect

bdbch:

This PR should fix a performance regression that caused to many React state updates because the new select/deselection fix introduced in #4148

I added checks to make sure those updates are only made when a node was actually affected by the change

## Checklist

- [x] Followed the guidelines
- [x] Fixed linting issues


rajatkapoor:
Thanks for getting this fixed!!

PR #4706, opened 2023-12-17
fix: make all link marks non-inclusive fixes #4702

### LINKED ISSUE #4702: [Bug]: Strange behavior after deleting link/autolink text

extension-link

2.2.0-rc.7 (latest code from develop branch)

### Link deletion bug

When you delete the full text of a link and then keep typing without moving your mark, the entered text will still be a link.

Steps to reproduce:
1. Open the Marks/Link demo when running `npm run start` from the project folder
2. Type any link (e.g., `google.com` then creating an autolink by hitting space)
3. Backspace until you've deleted the entire link text (but no farther).
4. Type more text (e.g. `this should not be linked text`)

Now, `this should not be linked text` will be a link to google.com.

Chrome

### Code example

_No response_

After you delete `google.com` and keep typing, the new text is not be a link.
(This is the default behavior in most rich text editors, for example, Notion)

### Related autolink creation/deletion feature request

Additionally, it would be great if there were a feature/default behavior were changed where autolinks were unset immediately after the text is no longer a link, e.g.
1. `google.c`| <- mark, text is not a link
2. type 'o',
`google.co`| <- mark, text becomes a link
3. type <backspace>
`google.c`| <- mark, text is not a link

This behavior is fairly common and something that users would likely want or expect. For example, it is default behavior in the Twitter/X post editor, as well as in Remirror.

Currently, the autolink will not be unset even if the text is no longer a link. Autolinks are also not created until whitespace is inserted after the link text.

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

dcourv:

The link mark was inclusive when autolinks were turned on, causing unwanted behavior when deleting links fully.

Please see #4702 for full problem description.

Changed link mark to always have `inclusive: false`, even when autolink is enabled.

I repeated the steps described in #4702. Links now behave as desired.

Follow the "steps to reproduce" in #4702 and see that `this should not be linked text` is no longer linked text.

## Checklist

- [x] Followed the guidelines
- [x] Fixed linting issues

## Related issues

#4702

### DIFF
--- a/packages/extension-link/src/link.ts
+++ b/packages/extension-link/src/link.ts
@@ -80,9 +80,7 @@ export const Link = Mark.create<LinkOptions>({
     reset()
   },
 
-  inclusive() {
-    return this.options.autolink
-  },
+  inclusive: false,
 
   addOptions() {
     return {


bdbch:
LGTM but this also changes the behavior of links. If you set inclusive to false you won't be able to add new characters to a link when your cursor is at the end of a link.

Did you consider this? Maybe a better solution would be to implement some kind of cleanup when a transaction comes in where you check if you are currently inside a link that is empty?

bdbch (commented):
@dcourv 

LGTM but this also changes the behavior of links. If you set inclusive to false you won't be able to add new characters to a link when your cursor is at the end of a link.

Did you consider this? Maybe a better solution would be to implement some kind of cleanup when a transaction comes in where you check if you are currently inside a link that is empty?

dcourv:
Hi @bdbch,
I see your point. However,  there are more problems with having `inclusive: true`, especially with autolinks. For example, do the following:
1. type "google.com"
2. type \<space\>, google.com becomes a link
3. type \<backspace\>
4. type anything, and this text does not add to the link

Now instead, do the following
1. type "google.com"
2. type \<space\>, google.com becomes a link
3. press left arrow, so the cursor is at the end of google.com
4. type anything, and _**text is added to the link**_

This is inconsistent behavior for what should happen when your cursor is at the end of the link.

The default behavior for links in Google Docs and Pages is to have `inclusive: false`-like behavior all the time (you cannot continue link text just by putting the cursor at the end). I think users would expect this in Tiptap, and it looks like it would fix any bugs like the one mentioned in the PR and the one in this comment.

bdbch:
@dcourv could you rebase this to the current develop branch as we released V3, deprecating the next branch & would need to update most of our PR's anyway? Thanks a lot! I'll take a look again afterwards, sorry for the delay - lots of stuff to do.

PR #4721, opened 2023-12-21
code-block: on arrow down move to next node

hollykurt:

when nodeAfter exists. move selection to the next node

I used prosemirror state's Selection.near to find the correct next position, and set the current selection to it

Manually tested with
* code block -> paragraph
* code block -> code block

* create a document with a code block and a text paragraph after it
* move the cursor into the code block
* arrow down to navigate out of the code block and on to the paragraph that follows

## Remarks

The issue this solves is that when code block is not the last leaf in the doc, ArrowDown had no effect.
Leaving the user stuck within the code block, without the possibility to navigate out (except by shift+enter to add a new paragraph).

Contributed on behalf of [Swimm](https://swimm.io/)

## Checklist

- [x] Followed the guidelines
- [x] Fixed linting issues


hollykurt:
@nperez0111 did you get a chance to review this PR?

nperez0111:
Hi @hollykurt could you show me a screen recording of the original issue & what this fixes. I'm testing on the default editor https://tiptap.dev/docs/examples/basics/default-text-editor and it is behaving like I would expect so I'm confused to what issue your PR is resolving

nperez0111 (approved):
NVM, I understand it. This is a good behavior to have, thank you for your contribution!

PR #4749, opened 2024-01-06
extract dependencies to menus

### LINKED ISSUE #3170: Popper & Tippy should not be in a bundled app by default

Event if I don't use the bubble & floating menus, popper.js and tippy.js are included in the final bundle of my app (vue3 vite app).
While the (undocumented) **enableCoreExtensions: false** option on the editor is a good addition, this architecture does not allow to exclude popper and tippy with tree shaking.

It would be great to have the possibility to use a "bare" editor and decide what core extension to import.

I don't see any alternative, am i missing something ?

_No response_

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

YousefED:
closes #3170 

This implements the approach suggested by @bdbch in https://github.com/ueberdosis/tiptap/issues/3170, to extract the react menu components. This way, @tiptap/react won't pull in tippy / popper dependencies anymore

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5171,6 +5171,14 @@
       "resolved": "packages/react",
       "link": true
     },
+    "node_modules/@tiptap/react-bubble-menu": {
+      "resolved": "packages/react-bubble-menu",
+      "link": true
+    },
+    "node_modules/@tiptap/react-floating-menu": {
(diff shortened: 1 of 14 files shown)


nperez0111:
Closing in favor of a separate export

Issue #2720, opened 2022-04-19
The insertContent command adds unnecessary blank paragraphs with HTML as value


rfgamaral:
### What’s the bug you are facing?

The `insertCommand` seems to be adding unnecessary blank paragraphs to the editor when newlines are found between HTML tags.

### Which browser was this experienced in? Are any special extensions installed?

n/a

### How can we reproduce the bug on our side?

Take the following HTML as an example:

```js
const htmlInput= `<h1><a href="https://tiptap.dev/">Tiptap</a></h1>
<p><strong>Hello World</strong></p>`
```

And then call `editor.commands.insertCommand(htmlInput)`. This is what gets rendered by Tiptap:

```html
<div
    translate="no"
    role="textbox"
    aria-readonly="false"
    aria-multiline="true"
    class="ProseMirror"
    tabindex="0"
    contenteditable="true"
>
    <p><br class="ProseMirror-trailingBreak" /></p>
    <h1>
        <a target="_blank" rel="noopener noreferrer nofollow" href="https://tiptap.dev">Tiptap</a>
    </h1>
    <p></p>
    <p><strong>Hello World</strong></p>
</div>
```

As you can see, there's one paragraph before and after the header that shouldn't be there.

### What did you expect to happen?

Based on the example above, Tiptpa should've rendered this:

```html
<div
    translate="no"
    role="textbox"
    aria-readonly="false"
    aria-multiline="true"
    class="ProseMirror"
    tabindex="0"
    contenteditable="true"
>
    <h1>
        <a target="_blank" rel="noopener noreferrer nofollow" href="https://tiptap.dev">Tiptap</a>
    </h1>
    <p><strong>Hello World</strong></p>
</div>
```

### Did you update your dependencies?

- [X] Yes, I’ve updated my dependencies to use the latest version of all packages.

### Are you sponsoring us?

- [X] Yes, I’m a sponsor. 💖 _(on behalf of Doist)_

rfgamaral:
@bdbch This issue still exists, do you think this could be reopened, and possibly some attention?

bdbch:
bump

evanshe:
The same with u, instead , i call insertcontent, value is the json
```
inserContent({
  type: 'xxx',
content: []
})
```

ondrejsevcik:
@determinationlove I don't think that comment like this is appropriate.

> Errr... Can't believe this bug has been reported for more than a year without being fixed?

[...]

PR #4767, opened 2024-01-10
fix(core): fix new lines being added via elementFromString

bdbch:

This PR fixes an issue that adds new lines on the parser level of the browser. It traverses through the actual dom instead of using string replacements removing all newlines added additionally by the parser.

See above

I added a new test to check if newlines are kept in pre tags.

Check out the local demo of Issue #2720

## Checklist

- [x] Followed the guidelines
- [x] Fixed linting issues

### DIFF
--- a/packages/core/src/utilities/elementFromString.ts
+++ b/packages/core/src/utilities/elementFromString.ts
@@ -1,6 +1,26 @@
+const removeWhitespaces = (node: HTMLElement) => {
+  const children = node.childNodes
+
+  for (let i = children.length - 1; i >= 0; i -= 1) {
+    const child = children[i]
+
+    if (child.nodeType === 3 && child.nodeValue && !/\S/.test(child.nodeValue)) {
+      node.removeChild(child)
+    } else if (child.nodeType === 1) {
+      removeWhitespaces(child as HTMLElement)
+    }
+  }
+
+  return node
+}
+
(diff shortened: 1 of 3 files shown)


packages/core/src/utilities/elementFromString.ts
```diff
@@ -1,6 +1,26 @@
...
 export function elementFromString(value: string): HTMLElement {
   // add a wrapper to preserve leading and trailing whitespace
   const wrappedValue = `<body>${value}</body>`
 
-  return new window.DOMParser().parseFromString(wrappedValue, 'text/html').body
+  const html = new window.DOMParser().parseFromString(wrappedValue, 'text/html').body
+
+  removeWhitespaces(html)
+
+  return removeWhitespaces(html)
```
yurtsiv, line 25: Sorry, but does it have to be two calls to `removeWhitespaces`?

Marvelousy20:
Updated @tiptap/core to v2.2.1 and this issue is still there.

okejminja:
Any updates regarding this issue? 
When trying to call `.insertContentAt()` with `<p>content</p>` I get an unnecessary paragraph inserted above my content

```javascript
editor.commands.insertContentAt(
  insertAfter,
  {
    type: "paragraph",
    content: [{ type: "text", text: modifiedContent }],
  },
  { parseOptions: { preserveWhitespace: false } },
)
```

Issue #4770, opened 2024-01-10
[Bug]: typescript build error in suggestion package


remyma:
### Which packages did you experience the bug in?

suggestion

### What Tiptap version are you using?

2.1.15

### What’s the bug you are facing?

Hi
I am using tiptap suggestion in my code and I'm facing this issue during typescript build.

my typescript version: `"typescript": "^5.3.3",`

Here is the error:

```
ERROR in ./node_modules/@tiptap/suggestion/src/suggestion.ts:185:32
TS6133: 'oldState' is declared but its value is never read.
    183 |
    184 |       // Apply changes to the plugin state from a view transaction.
  > 185 |       apply(transaction, prev, oldState, state) {
        |                                ^^^^^^^^
    186 |         const { isEditable } = editor
    187 |         const { composing } = editor.view
    188 |         const { selection } = transaction
  ```

I can disable the `noUnusedLocals` rule in my `tsconfig.json` but that would apply to my whole project.

As `oldState` is unused in the context of suggestion plugin, can we replace it by `_oldState` so that it's ignored by typescript?

Thanks

### What browser are you using?

Chrome

### What did you expect to happen?

Typescript buids without error when `noUnusedLocals` rule is set to true.

### Did you update your dependencies?

- [X] Yes, I’ve updated my dependencies to use the latest version of all packages.

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

remyma:
oups, I just realized that it was caused due to bad resolved import on my project :sweat_smile: 

So no issue anymore on my side, but yes we can remove this parameter as unused.

I think I can close the issue then.

PR #4772, opened 2024-01-11
fix(suggestion): typescript build error in suggestion package

Mishra-Pravin:
## you can simply remove the oldState parameter from the function declaration if it’s not being used.

The error you’re seeing is because the oldState parameter in the apply function is declared but not used anywhere in the function.

## This should resolve the TypeScript error you’re seeing. If oldState is needed in the future, you can add it back. Remember, the noUnusedLocals rule is there to help keep your code clean and maintainable. It’s generally a good practice to remove unused variables. However, if you find this rule too restrictive for your project, you can turn it off in your tsconfig.json, but as you mentioned, it would apply to your whole project.

## Checklist

- [ ] Followed the guidelines
- [ ] Fixed linting issues

## Related issues

[https://github.com/ueberdosis/tiptap/issues/4770]

### DIFF
--- a/packages/suggestion/src/suggestion.ts
+++ b/packages/suggestion/src/suggestion.ts
@@ -184,7 +184,7 @@ export function Suggestion<I = any>({
       },
 
       // Apply changes to the plugin state from a view transaction.
-      apply(transaction, prev, oldState, state) {
+      apply(transaction, prev, state) {
         const { isEditable } = editor
         const { composing } = editor.view
         const { selection } = transaction


packages/suggestion/src/suggestion.ts
```diff
@@ -184,7 +184,7 @@ export function Suggestion<I = any>({
       },
 
       // Apply changes to the plugin state from a view transaction.
-      apply(transaction, prev, oldState, state) {
```
nperez0111, line 187: state is used below, it should not be referencing oldState

PR #4830, opened 2024-01-25
Prepare stable release 2.2.0

svenadlung:
Prepare stable release and fix demos and some obvious issues.

### DIFF
--- a/packages/core/src/utilities/elementFromString.ts
+++ b/packages/core/src/utilities/elementFromString.ts
@@ -20,7 +20,5 @@ export function elementFromString(value: string): HTMLElement {
 
   const html = new window.DOMParser().parseFromString(wrappedValue, 'text/html').body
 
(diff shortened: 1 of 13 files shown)


bdbch (approved):
LGTM

packages/core/src/utilities/elementFromString.ts
```diff
@@ -20,7 +20,5 @@ export function elementFromString(value: string): HTMLElement {
 
   const html = new window.DOMParser().parseFromString(wrappedValue, 'text/html').body
 
-  removeWhitespaces(html)
```
bdbch, line 23: I'll check this issue again on another branch afterwards.

svenadlung: Yeah, only removed the duplicate function call here. Shouldn't have changed anything with the other issue.

PR #4835, opened 2024-01-26
fix(core): check if pos is valid before resolving in backspace core keymap handler

solastley:
So I'm working on a TipTap editor where I don't want to allow block elements in the document. Not exactly a common use case I'm guessing, but it makes sense for what I'm doing.

I discovered that when I override the `content` of the `Document` extension to only allow text like so...

```
Document.extend({
    content: 'text*',
})
```

...it causes an error to be thrown whenever I press `Backspace` and either:

1. The editor is empty OR
2. All of the text is selected with the anchor at the beginning of the text

The error comes from the backspace keyboard shortcut handler which is added in TipTap core. This handler attempts to resolve a position in the document at `pos - 1` where `pos` is equal to `editor.state.selection.$anchor.pos`. I think this works fine when there are block elements in the editor since even an empty editor will still have at least an empty block element. Whenever only text is allowed, though, this `pos` is equal to zero whenever the editor is empty. This results in the handler attempting to resolve the position in the document at a `pos` of `-1`, which throws an error.

Added a check that `pos > 0` before attempting to resolve `pos - 1` in the document.

Tested both cases where I found the error manually, and with this check, the error no longer occurs.

I also tested backspace in a more typical editor configuration with the default `Document` extension. Everything works as expected, including backspace when there is a lot of text, backspace when there is no text, and backspace when all text is selected.

To reproduce bug:
1. Create an editor and extend `Document` as I've outlined above.
2. Press `Backspace` when the editor is empty
3. See the error

To verify fix, repeat the process with my changes applied and see the error is gone.

## Remarks
Loving TipTap so far :)

## Checklist

- [x] Followed the guidelines
- [x] Fixed linting issues

## Related issues
N/A


alexmalev:
> ## Please describe your changes
> So I'm working on a TipTap editor where I don't want to allow block elements in the document. Not exactly a common use case I'm guessing, but it makes sense for what I'm doing.
> 
> I discovered that when I override the `content` of the `Document` extension to only allow text like so...
> 
> ```
> Document.extend({
>     content: 'text*',
> })
> ```
> 
> ...it causes an error to be thrown whenever I press `Backspace` and either:
> 
> 1. The editor is empty OR
> 2. All of the text is selected with the anchor at the beginning of the text
> 
> The error comes from the backspace keyboard shortcut handler which is added in TipTap core. This handler attempts to resolve a position in the document at `pos - 1` where `pos` is equal to `editor.state.selection.$anchor.pos`. I think this works fine when there are block elements in the editor since even an empty editor will still have at least an empty block element. Whenever only text is allowed, though, this `pos` is equal to zero whenever the editor is empty. This results in the handler attempting to resolve the position in the document at a `pos` of `-1`, which throws an error.
> 
> ## How did you accomplish your changes
> Added a check that `pos > 0` before attempting to resolve `pos - 1` in the document.
> 
> ## How have you tested your changes
> Tested both cases where I found the error manually, and with this check, the error no longer occurs.
> 
> I also tested backspace in a more typical editor configuration with the default `Document` extension. Everything works as expected, including backspace when there is a lot of text, backspace when there is no text, and backspace when all text is selected.
> 
> ## How can we verify your changes
> To reproduce bug:
> 
> 1. Create an editor and extend `Document` as I've outlined above.
> 2. Press `Backspace` when the editor is empty
> 3. See the error
> 
> To verify fix, repeat the process with my changes applied and see the error is gone.
> 
> ## Remarks
> Loving TipTap so far :)
> 
> ## Checklist
> * [x]  The changes are not breaking the editor
> * [ ]  Added tests where possible
> * [x]  Followed the guidelines
> * [x]  Fixed linting issues
> 
> ## Related issues
> N/A

very much waiting for this. same exact problem. thanks!

solastley:
@svenadlung gentle bump

bdbch (approved):

Issue #4785, opened 2024-01-15
[Bug]: Link parse error when inserted


stevobm:
### Which packages did you experience the bug in?

core

### What Tiptap version are you using?

2.1.15

### What’s the bug you are facing?

When a link is inserted into the document, I only get the text instead of the link, and an error below:
![image](https://github.com/ueberdosis/tiptap/assets/5198256/06fe3533-30e2-4b84-b170-63392ea47bc8)
This problem occurred when I upgraded tiptap to 2.1.15, and it also exists in 2.1.16, wish it to be solved soon.

### What browser are you using?

Chrome

### What did you expect to happen?

Link be parsed correctly.

### Did you update your dependencies?

- [X] Yes, I’ve updated my dependencies to use the latest version of all packages.

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

albingroen:
I'm running into the same issue here. Using v2.1.16

jakebarshickf1nx:
2.1.10 is the last version that I don't experience any pasting issues FYI. Seems like 2.1.11 is where things start breaking for me maybe #4354 is the initial cause?

stevobm:
Fixed in https://github.com/ueberdosis/tiptap/pull/4836.

PR #4836, opened 2024-01-27
fix: fix bug #4785

stevobm:
https://github.com/ueberdosis/tiptap/issues/4785


felipe-sme:
Hi there! Just wanted to report that after upgrading to v2.2.0, a chunk containing two links pasted to the editor will still silently log the error to the console "Uncaught TypeError: Cannot read properties of undefined (reading 'nodeSize')". It will however work as expected, besides the console error. Happy to create an issue if it helps. Thanks!

PR #4841, opened 2024-01-30
Refactor NodePos class to support block nodes

### LINKED ISSUE #4840: [Bug]: `NodePos` methods does not work with non-text nodes

core

2.2.0

Newly implemented `NodePos` methods such as `$node(...)` does not work with non-text block nodes like `horizontalRule` or `blockquote`.

Chrome

### Code example

_No response_

Given document state,
```
{
  "type": "doc",
  "content": [
    {
      "type": "paragraph"
      "content": [
        {
          "type: "text",
          "text": "hello"
        }
    },
    {
      "type": "horizontalRule"
    }
  ]
}
```

`editor.$node('horizontalRule')` should return corresponding `NodePos`, but instead it incorrectly returns `null`.

_No response_

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

bdbch:

This PR adds support for block nodes in the NodePos implementation. This is still a bit janky but at least block nodes are now found. Reason for this jank is that [ResolvedPos in Prosemirror won't find block nodes at their positions but the `doc` node and the depth of 0](https://discuss.prosemirror.net/t/resolvepos-works-unexpected-for-atom-block-node/5087/3).

This is bad because I used ResolvedPos for all the position mapping inside the NodePos class. In the future we should see if we can improve on this. This means that currently `.before`, `.after`, etc. won't be very exact and could lead to wrong positions as I'm trying to find those positions by checking if a nodepos is a block nodepos.

1. Added information about `isBlock` into a nodePos together with it's actual `depth` that I try to get from the context of the `children` getter.
2. Updated the children getter to pass down those informations to the newly found NodePos instances

I used a customized Examples/Default demo locally.

You can do the same above - bind the editor created in the demo to window.editor and then try to do things like `editor.$node('bulletList')`

## Remarks

As I said above this still is a bit jank as the ResolvedPos implementation of Prosemirror doesn't give me correct information about a nodes depth. We'll probably need to improve this class over time as we run into more and more issues.

## Checklist

- [x] Followed the guidelines
- [x] Fixed linting issues

## Related issues

closes #4840

### DIFF
--- a/packages/core/src/NodePos.ts
+++ b/packages/core/src/NodePos.ts
@@ -8,23 +8,35 @@ import { Content, Range } from './types.js'
 export class NodePos {
   private resolvedPos: ResolvedPos
 
+  private isBlock: boolean
+
   private editor: Editor
 
-  constructor(pos: ResolvedPos, editor: Editor) {
-    this.resolvedPos = pos
-    this.editor = editor
+  private get name(): string {
+    return this.node.type.name
   }
 
+  constructor(pos: ResolvedPos, editor: Editor, isBlock = false, node: Node | null = null) {
+    this.isBlock = isBlock
+    this.resolvedPos = pos
+    this.editor = editor
+    this.currentNode = node
+  }
+
+  private currentNode: Node | null = null
+
   get node(): Node {
-    return this.resolvedPos.node()
+    return this.currentNode || this.resolvedPos.node()
   }
 
   get element(): HTMLElement {
     return this.editor.view.domAtPos(this.pos).node as HTMLElement
   }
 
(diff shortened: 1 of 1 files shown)


svenadlung (approved):

shaileshiyer:
This refactor to support block nodes has introduced a bug where none of the items other than block nodes are found. #4997 #5003. It essentially breaks a lot of the examples mentioned in the documentation website for core and leads to being unable to find any items other than at the current depth. Essentially only searching for the nodes in the current children but never searching the grandchildren if the isaBlock is set to true.

PR #4843, opened 2024-01-31
feature(core): add onPaste and onDrop events to editor

bdbch:

[This message](https://discord.com/channels/818568566479257641/818569721934774272/1202162639317508158) on our Discord server made me realize that there's apparently no easy way to just add a simple paste or drag event handler to the actual editor instance. This user tried to add it to the editors DOM element which works sometimes (doesn't work properly if there's no content inside the contenteditable yet).

I created two very simple plugins to handle paste and drop events passing up their relative events to the option passed through.

Those can be configured on the editor constructor.

I tried to add `onPaste` and `onDrop` on a local demo.

See above

## Remarks

No remarks from my side

## Checklist

- [x] Followed the guidelines
- [x] Fixed linting issues

## Related issues

No related issues


packages/core/src/plugins/DropPlugin.ts
```diff
@@ -0,0 +1,13 @@
+import { Plugin, PluginKey } from '@tiptap/pm/state'
+
+export const DropPlugin = (onPaste: (e: DragEvent) => void) => {
+  return new Plugin({
+    key: new PluginKey('tiptap__drop'),
```
svenadlung, line 5: Maybe keep this in lowerCamelCase (`tiptapDrop`), like we do in other plugins?

packages/core/src/plugins/PastePlugin.ts
```diff
@@ -0,0 +1,13 @@
+import { Plugin, PluginKey } from '@tiptap/pm/state'
+
+export const PastePlugin = (onPaste: (e: ClipboardEvent) => void) => {
+  return new Plugin({
+    key: new PluginKey('tiptap__paste'),
```
svenadlung, line 5: Same here.

packages/core/src/plugins/DropPlugin.ts
svenadlung: Or maybe call it `dropEvent`, analog to `focusEvents`?

packages/core/src/index.ts
```diff
@@ -11,6 +11,8 @@ export * from './NodePos.js'
 export * from './NodeView.js'
 export * from './PasteRule.js'
 export * from './pasteRules/index.js'
+export * from './plugins/DropPlugin.js'
```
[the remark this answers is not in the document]
svenadlung: Oh, I just realized focusEvents is a wrapped plugin (as a Tiptap extension). But maybe we keep it the same style?

svenadlung (commented):
Please check my comments.

packages/core/src/plugins/DropPlugin.ts
bdbch: how about `tiptap__dropEvent` - I'm using the `tiptap__` prefix to make sure to never accidentially conflict with other pluginKeys

packages/core/src/plugins/PastePlugin.ts
bdbch: how about `tiptap__pasteEvent` - I'm using the `tiptap__` prefix to make sure to never accidentially conflict with other pluginKeys

packages/core/src/plugins/DropPlugin.ts
nperez0111: PluginKeys are symbols by default no? So this should only ever exist once.

packages/core/src/plugins/DropPlugin.ts
```diff
@@ -0,0 +1,13 @@
...
+
+    props: {
+      handleDrop: (_, e) => {
+        onPaste(e)
```
nperez0111, line 9:
```suggestion
        onDrop(e)
```

nperez0111 (approved):

packages/core/src/types.ts
```diff
@@ -121,6 +121,8 @@ export interface EditorOptions {
   onFocus: (props: EditorEvents['focus']) => void;
   onBlur: (props: EditorEvents['blur']) => void;
   onDestroy: (props: EditorEvents['destroy']) => void;
+  onPaste?: (e: ClipboardEvent) => void
+  onDrop?: (e: DragEvent) => void
```
nperez0111, line 125: I think these types are intentionally not optional, there is a Partial<EditorOptions> wherever we use it

packages/core/src/plugins/PastePlugin.ts
```diff
@@ -0,0 +1,13 @@
+import { Plugin, PluginKey } from '@tiptap/pm/state'
+
+export const PastePlugin = (onPaste: (e: ClipboardEvent) => void) => {
+  return new Plugin({
+    key: new PluginKey('tiptapPaste'),
+
+    props: {
+      handlePaste: (_, e) => {
+        onPaste(e)
+      },
```
nperez0111, line 10: On looking at this again, would we want to also pass through any of these other values?

I see that `handlePaste` is defined as:
```
    /**
    Can be used to override the behavior of pasting. `slice` is the
    pasted content parsed by the editor, but you can directly access
    the event to get at the raw content.
    */
    handlePaste?: (this: P, view: EditorView, event: ClipboardEvent, slice: Slice) => boolean | void;
```
I think at least `slice` might be useful to passthrough?

packages/core/src/index.ts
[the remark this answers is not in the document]
nperez0111: I think that is fair, I'm okay with this organization

packages/core/src/plugins/PastePlugin.ts
bdbch: Good idea - done.
Added `slice` for onPaste and `slice` and `moved` for onDrop.

YousefED:
It seems a bit weird that these methods are initialized to `() => null` and a bit later in the code they're checked with `if(this.options.onPaste) {`.

imo it would make more sense to initialize them as undefined

nperez0111:
> It seems a bit weird that these methods are initialized to `() => null` and a bit later in the code they're checked with `if(this.options.onPaste) {`.
> 
> imo it would make more sense to initialize them as undefined

Good point, resolved with: https://github.com/ueberdosis/tiptap/pull/5684

PR #4857, opened 2024-02-05
Change to unnecessarily resolved positions

aguydan:

I think in the `hasValidSelection` boolean expression inside of the Bubble Menu plugin there is no obvious reason for why `$from.pos` and `$to.pos` are used instead of the properties of `selection` object with the same name.

I swapped resolved positions with unresolved ones in `hasValidSelection`.

I compared the values returned by both versions of the expression while tinkering with the menu. Also, as far as I'm aware, the ProseMirror documentation states that they're literally the same.

## Checklist

- [x] Followed the guidelines
- [ ] Fixed linting issues


svenadlung (approved):

aguydan:
Well, performance is probably not an issue here, so I suppose it just makes the code a little bit more consistent.

PR #4858, opened 2024-02-06
feat(react): add useEditorForImmediateRender hook

svenadlung:
To be added …

## Please describe your changes

[add a description of your changes here]

## How did you accomplish your changes

[add a detailed description of how you accomplished your changes here]

## How have you tested your changes

[add a detailed description of how you tested your changes here]

## How can we verify your changes

[add a detailed description of how we can verify your changes here]

## Remarks

[add any additional remarks here]

## Checklist

- [ ] The changes are not breaking the editor
- [ ] Added tests where possible
- [ ] Followed the guidelines
- [ ] Fixed linting issues

## Related issues

[add a link to the related issues here]

PR #4864, opened 2024-02-07
fix: initially set editor object only on non-ssr environments

bdbch:

This PR gets rid of the initial null render on non-ssr environments.

We're doing a check on the window object to see if the current env is a browser environment.

## Checklist

- [x] Followed the guidelines
- [x] Fixed linting issues

### DIFF
--- a/packages/react/src/useEditor.ts
+++ b/packages/react/src/useEditor.ts
@@ -9,7 +9,9 @@ import {
 import { Editor } from './Editor.js'
 
 export const useEditor = (options: Partial<EditorOptions> = {}, deps: DependencyList = []) => {
-  const editorRef = useRef<Editor | null>(null)
+  // if window is undefined (SSR), return the editor instance
+  // otherwise we can start with a null value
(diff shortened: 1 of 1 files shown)


packages/react/src/useEditor.ts
```diff
@@ -99,7 +101,9 @@ export const useEditor = (options: Partial<EditorOptions> = {}, deps: Dependency
   useEffect(() => {
     let isMounted = true
 
-    editorRef.current = new Editor(options)
+    if (!editorRef.current) {
```
C-Hess, line 104: Take a look at https://github.com/ueberdosis/tiptap/pull/4579 if you get the chance. That pull request uses memo instead of a ref and also eliminates the double initialization that also causes issues. Best not to create the editor instance in an effect, but a memo instead; that way it gets created first render and after any deps array changes.

Also heads up, this is _also_ similar to this draft PR by https://github.com/ueberdosis/tiptap/pull/4858/files, which has the same issue of double editor initialization

hjoelh: pls 🙏 @bdbch  @svenadlung

nperez0111:

PR #4865, opened 2024-02-07
fix(extension-link): Avoid auto-linking partial text for invalid TLDs

rfgamaral:

This PR implements better autolink validation to ensure that only complete and valid text is hyperlinked, preventing cases where a valid top-level domain (TLD) is immediately followed by an invalid character, like a number.

For instance, with the `find` method from [Linkify](https://linkify.js.org/), typing the text `example.com1` would result in `example.com` being linked and the trailing `1` left as plain text:

firefox_ybHTgnid0x

This behaviour is a bit unexpected, and this PR implements a more comprehensive validation on the input text.

With the `tokenize` method instead of `find`, which allows us to perform a more comprehensive validation on the input text.

> [!NOTE]
> The `find` method uses `tokenize` internally, but it does a much simpler validation.

Used the `Link` extension demo to test multiple link formats to make sure that autolinking still worked as expected.

Use the `Link` extension demo as I did.

### DIFF
--- a/packages/extension-link/src/helpers/autolink.ts
+++ b/packages/extension-link/src/helpers/autolink.ts
@@ -7,7 +7,29 @@ import {
 } from '@tiptap/core'
 import { MarkType } from '@tiptap/pm/model'
 import { Plugin, PluginKey } from '@tiptap/pm/state'
-import { find } from 'linkifyjs'
+import { MultiToken, tokenize } from 'linkifyjs'
+
+/**
+ * Check if the provided tokens form a valid link structure, which can either be a single link token
+ * or a link token surrounded by parentheses or square brackets.
+ *
+ * This ensures that only complete and valid text is hyperlinked, preventing cases where a valid
+ * top-level domain (TLD) is immediately followed by an invalid character, like a number. For
+ * example, with the `find` method from Linkify, entering `example.com1` would result in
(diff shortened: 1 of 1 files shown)


rfgamaral:
There's one test failing, `allows images to be added via URL`, but it looks a flaky one, since I had this passing before in this PR, and fail in a different PR. Both PRs have changes unrelated to the failing test, as far as I can tell.

bdbch (approved):
LGTM! Just a quick question regarding the `tokenize` function. It says it's an external function only - but I still assume it's safe to use?

Also what does this do?

```
return ['()', '[]'].includes(tokens[0].value + tokens[2].value)
```

bit hard to understand why we check that `()` and `[]` array.

rfgamaral:
> LGTM! Just a quick question regarding the `tokenize` function. It says it's an external function only - but I still assume it's safe to use?

Yes, it should be safe to use it. I mean, it's used by the `test` function itself.

> 
> Also what does this do?
> 
> ```
> return ['()', '[]'].includes(tokens[0].value + tokens[2].value)
> ```

It allows one to **type** a URL surrounded by parentheses or square brackets. So, if you were to type `[example.com]<space>`, pressing `<space>` would auto-link `example.com` inside the square brackets.

PR #4866, opened 2024-02-07
fix: Disallow only whitespace between markdown shortcuts delimiters

rfgamaral:

This PR disallows Markdown shortcuts (through typing or pasting) for all marks (`Bold`, `Code`, `Highlight`, `Italic`, and `Strike`) when the **content between the delimiters is only whitespace**.

Changed the input and paste Regex for all the marks mentioned above to keep supporting as much as possible from the current syntax, while only disallowing "empty strings" between the delimiters for the Markdown shortcuts of each mark.

> [!NOTE]
> The regular expression for all marks follows the exact same pattern, the only difference being the delimiter characters for each mark.

I've used both [regex101.com](https://regex101.com/):

firefox_wzo9ORdntq

And the marks demos to test multiple patterns, here's an example for the `Highlight` extension:

NVIDIA_Share_RoSfI1pELP

The same way I did above.

### DIFF
--- a/packages/extension-bold/src/bold.ts
+++ b/packages/extension-bold/src/bold.ts
@@ -28,10 +28,10 @@ declare module '@tiptap/core' {
   }
 }
 
-export const starInputRegex = /(?:^|\s)((?:\*\*)((?:[^*]+))(?:\*\*))$/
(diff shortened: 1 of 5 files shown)


bdbch (approved):
LGTM

Issue #4893, opened 2024-02-16
[Bug]: setContent breaks mentions when provided as html


stefanholzapfel:
### Which packages did you experience the bug in?

core, extension-mention

### What Tiptap version are you using?

2.2.3

### What’s the bug you are facing?

I remember in earlier versions it was working perfectly fine to re-add a mention tag that was previously extracted from the editor as plain html like e.g.:

```
editor.commands.setContent('<p><span class="mention" data-id="4221" data-mention="" data-label="Stefan Holzapfel" contenteditable="false">@Stefan Holzapfel</span></p>')
```

In version 2.2 that behavior doesn't work anymore, something parses the mention and transforms it to that:

```
<p><span data-decoration-id="id_3447837711" class="suggestion">@Stefan Hol</span>zapfel</p>
```

Anyone has an explanation what happens here? Especially the cropping and re-attaching of a part of the string after the closing </span> tag is weird.

I have defined the ``renderHtml`` in the ``Mention.configuration`` like this:
```
return [
            "span",
            {
                class: 'mention',
                'data-id': id,
                'data-mention': '',
                'data-label': label
            },
            `@${label}`
        ]
```

This is the complete initialization of the editor:

```
        editor = new Editor({
            element: this.shadowRoot.querySelector('#container'),
            parseOptions: {
                preserveWhitespace: true
            },
            onBlur: () => {
                this.closeMentionList();
            },
            autofocus: false,
            extensions: [
              StarterKit.configure({
                  blockquote: false,
                  bulletList: false,
                  codeBlock: false,
                  hardBreak: false,
                  heading: false,
                  horizontalRule: false,
                  listItem: false,
                  orderedList: false,
                  code: false
              }),
              Highlight,
              Typography,
              Placeholder.configure({
                  placeholder: this.placeholder
              }),
              Mention.configure({
                renderHTML: ({ options, node }) => {
                    return [
                      "span",
                      {
                          class: 'mention',
                          'data-id': id,
                          'data-mention': '',
                          'data-label': label
                      },
                      `@${label}`
                  ]
                },
                suggestion: {
                    allowSpaces: true,
                    items: query => {
                        return this.mentionData;
                    },
                    render: () => {
                        return {
                            onStart: props => {
                                // ...
                            },
                            onKeyDown: props => {
                                // ...
                            },
                            onUpdate: props => {
                                // ...
                            },
                            onExit: () => {
                                this.closeMentionList();
                            },
                        };
                    }
                }
[...]

Issue #4862, opened 2024-02-07
[Bug]: 2.1.15 and up breaks dozens of our tests


Nantris:
### Which packages did you experience the bug in?

core

### What Tiptap version are you using?

2.1.15

### What’s the bug you are facing?

This version and 2.1.16 break a dozen of our tests which seemingly have nothing to do with the only documented change:

> fix(core): fix insertContentAt keeping new lines in html content by @bdbch in https://github.com/ueberdosis/tiptap/pull/4465

The tests that are broken are as diverse as special modes of splitting list items to handling pasted links to inserting lines above the current line.

Even tests which don't seem to use `insertContentAt` in any way are broken.

### What browser are you using?

Chrome

### What did you expect to happen?

Patch versions should never break tests.

### Anything to add? (optional)

 I really don't want to be needing to rethink complex logic all over our app because of this change caused by an issue that originated almost 2 years ago (https://github.com/ueberdosis/tiptap/issues/2720)

I tried updating our `insertContentAt` calls to use `preserveWhitespace` with `'full'`, `true` and `false` - all result in still broken tests.

Do we have any remedy here besides just not upgrading until we find spare weeks for this? We're never going to have time to fix all of the problems that change caused.

**Example of broken output:**

```
    Expected: "<p>https://example.org\"&gt;https://example.org|&lt;/a&gt;&lt;/p&gt;&lt;/li&gt;&lt;/ul&gt;&lt;p&gt;&lt;/p&gt;|</p>"
    Received: "<p>|</p>"
```

@svenadlung @bdbch

### Did you update your dependencies?

- [X] Yes, I’ve updated my dependencies to use the latest version of all packages.

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

Slapbox:
Possible reasons the breaking effects are so wide ranging:

- `insertContent` relies on it under the hood: https://github.com/ueberdosis/tiptap/blob/56a5737ed102ee75ec72f9cc2847e3c977f431bd/packages/core/src/commands/insertContent.ts#L23

bdbch:
Could you give me an example on what kind of content you're setting that leads to the wrong output and what the output should actually look like so I can do local tests and fix them?

Slapbox:
@bdbch the trouble is I'm not really sure what information I can share to help you reproduce this given that I can't understand the relation between the documented change and the broken tests.

[...]

Issue #4034, opened 2023-05-10
[Bug]: Difference in behavior of setContent and insertContent commands


ashu12chi:
### Which packages did you experience the bug in?

core

### What Tiptap version are you using?

latest

### What’s the bug you are facing?

In our application, we are using `insertContent` and `setContent` command. The issue we are facing is the final output for the same content is different for the two commands.

Example to this:
```javascript
const content = `<h1>Welcome to your fresh Tiptap Code Sandbox</h1>
  <p>You can create a demo for your issue inside of this sandbox and share it with us.</p>`
```
For above content, output of `setContent` is:
![image](https://github.com/ueberdosis/tiptap/assets/45908393/3ce4743e-1059-4dbd-8d26-54d9d266b8c1)
But, for `insertContent` it is coming with two blank lines in-between the content
![image](https://github.com/ueberdosis/tiptap/assets/45908393/676fe983-2dbe-4855-ac2c-1c058b3e8b28)
if I pass `parseOptions` in `insertContent` command as
```javascript
editor.commands.insertContent(content, {
    parseOptions: { preserveWhitespace: false }
})
```
It will still come with one blank line like this:
![image](https://github.com/ueberdosis/tiptap/assets/45908393/cb68fc00-cc24-41d3-9ca8-939beb6d611c)

Ideally, behavior should be same for both the cases.

I have some observations regarding this:
1. If I modify my content to have no new line characters like this:
```javascript
const content = `<h1>Welcome to your fresh Tiptap Code Sandbox</h1><p>You can create a demo for your issue inside of this sandbox and share it with us.</p>`
```
Behavior will be same for both `insertContent` and `setContent`.

2. On taking a cursory at the internal implementation of the two commands, `setContent` is using [parseSlice](https://prosemirror.net/docs/ref/#model.DOMParser.parseSlice) and `insertContent` is using [parse](https://prosemirror.net/docs/ref/#model.DOMParser.parseSlice) function from Prosmirror.

Since, external content can possibly have new lines in them, we want behavior of both the commands to be same (Ideally the behavior of `setContent`)

### What browser are you using?

Chrome

### Code example

https://codesandbox.io/s/green-monad-surcnd?file=/src/App.js

### What did you expect to happen?

`setContent` and `insertContent` should produce same output for given content.

### Anything to add? (optional)

NA

### Did you update your dependencies?

- [X] Yes, I’ve updated my dependencies to use the latest version of all packages.

### Are you sponsoring us?

[...]

Issue #4704, opened 2023-12-16
[Bug]: 2.2.0-rc.7 upgrade - "Error: Adding different instances of a keyed plugin (plugin$)" -- worked in 2.1.x


Nantris:
### Which packages did you experience the bug in?

core

### What Tiptap version are you using?

2.2.0-rc.7

### What’s the bug you are facing?

This error appears for two plugins after upgrading, but we only call them once and there's no discernible difference between the problematic two and all the rest. Adding keys to the plugins doesn't help - it only changes the key in the error.

### What browser are you using?

Chrome

### What did you expect to happen?

Upgraded editor doesn't throw mystery errors.

### Did you update your dependencies?

- [X] Yes, I’ve updated my dependencies to use the latest version of all packages.

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

svenadlung:
I feel like this could be related to this: https://github.com/ueberdosis/tiptap/commit/55377ebd3e3e19429447aaa4467ce4a2eaa6b1fe

Could you please send your output of 
`npm list y-prosemirror` ?

svenadlung:
Should be fixed with newest version (RC right now).

K3TH3R:
@svenadlung I'm seeing this error now when I upgrade our repo from `2.1.16` to either `2.2.0` or `2.2.1`.  Even the most basic setup is still throwing this error for me:

```js
const editor = useEditor({
	content: '',
	extensions: [Document, Paragraph, Text],
})
```

```
"@tiptap/core": "2.2.1",
"@tiptap/extension-bold": "2.2.1",
"@tiptap/extension-document": "2.2.1",
"@tiptap/extension-italic": "2.2.1",
"@tiptap/extension-link": "2.2.1",
"@tiptap/extension-paragraph": "2.2.1",
"@tiptap/extension-text": "2.2.1",
"@tiptap/extension-underline": "2.2.1",
"@tiptap/pm": "2.2.1",
"@tiptap/vue-3": "2.2.1",
```

svenadlung:
@K3TH3R can you please provide a sandbox? 

It sounds like something has been installed with different versions. If possible, delete the lock file and reinstall the packages.

K3TH3R:
@svenadlung I've been able to confirm now that deleting the `yarn.lock` file and re-installing dependencies @ v2.2.1 is now working without a problem, thank you.

Slapbox:
[...]

PR #4895, opened 2024-02-16
add insertContent logic to setContent

bdbch:

This PR removes some logical differences between `setContent` and `insertContent` or `insertContentAt`. `setContent` was using `createDocument` which is using a completely different way of parsing the content and creating an **initial document**.

Since `setContent` is never used on a non-existing document as the doc already exists as it is initialized while setting up the prosemirror editor inside the Editor class, we can just simply use `insertContentAt` to replace the whole document content instead of initializing a completely new doc to set.

This should help to make sure that in the future we **only have one way of setting content** into the editor removing the confusion regarding setContent and insertContent.

I removed the old code from `setContent` and added a simple `insertContent` call to make sure all content commands are using the same way of content insertion logic.

I created several tests for `setContent` and made sure they run in positive.

Run the tests locally on your machine via `npm run test:open` and open the `setContent` spec. Otherwise you can also just run `setContent` in any editor.

## Remarks

While I don't think this is a **breaking change** I think we should bump this change into a new minor version. In theory it does the same as the `setContent` logic (taking a document, setting the specified content as child content) - the only difference now is that escape characters `\n` or `\t` are now included while setting content and are not stripped out.

I'm open for this discussion on this one though.

## Checklist

- [x] Followed the guidelines
- [x] Fixed linting issues

## Related issues

- related to #4893 as I added a test for mentions via `setContent`
- related to #4862 as I added a test for the case specified in this issue which now runs positive
- related to #4034 as this issue describes exactly what this fix is created for


Slapbox:
Thanks again for your work on this!

This all seems sound. One concern though.

> I think we should bump this change into a new minor version.

If that's the case, I think that the `2.1.x` branch should have `2.1.15` reverted since people affected by #4862 can't upgrade past `2.1.15` without this patch being applied to the `2.1.x` version. If waiting for the next minor version then this won't land until `2.3.x`. I'm guessing `2.3.x` is a long ways off, and currently we're blocked from even `2.2.0` by other issues.

That means we'd be stuck at `2.1.14` for potentially a very long time.

I don't know how long the team is planning to support the `2.1.x` branch, but it seems quite possible that some fix will land on that branch that we need in the meanwhile.

bdbch:
@Slapbox could you point me to the issues which are blocking you from upgrading to 2.2.x right now? Not sure if we should start releasing or reverting patch versions of an older minor.

Slapbox:
Sure @bdbch - the only known blocker for us is #4704, but there's always a chance there's more blockers we haven't been able to get to yet because #4704 caused crashes.

Nantris:
Thanks again for your work on this @bdbch. I'm wondering if you have any updates about how/when this change will be rolled out, or if `2.1.15` could be reverted? I think reverting `2.1.15` is the most appropriate approach, but if it's also incorporated into `2.2.x` then I know that's a bit more complicated.

Nantris:
Friendly bump.

Is there any way to provide some clarity on plans for this fix?

bdbch:
I'll take a look at the failing tests today to see if they are coming from the changes introduced by this PR - otherwise I'll try to get another review.

Nantris:
Ah yes I hadn't even realized there were failing tests. I expect the failures are unrelated based on it being complaining about editable vs non-editable, but maybe you'll find otherwise.

Nantris:
`2.3.0` seems to handle inserting `/t` properly, but not `/n`. If I insert `/r/n` it inserts a `/n` but it's not at all clear why this is necessary and it might have unintended consequences.

Inserting `/n/n` inserts two `/n` characters, leaving me all the more confused about why `/n` inserts zero of them.

Do you know why this might be? This is now the last known blocker for us to get off of the `2.1.x` branch (so long as #5061 is also merged)

Nantris:
@svenadlung any update on when this might be reviewed? I'm increasingly concerned we have no upgrade path and that it will only become more difficult to find one as this issue is baked into more and more versions of TipTap.

Patch releases should not break editors and when they do that should really be addressed. The lack of attention to fixing this unexpected breakage has been discouraging.

packages/core/src/commands/insertContentAt.ts
```diff
@@ -64,11 +64,6 @@ export const insertContentAt: RawCommands['insertContentAt'] = (position, value,
       },
     })
 
-    // don’t dispatch an empty fragment because this can lead to strange errors
-    if (content.toString() === '<>') {
-      return true
-    }
-
```
nperez0111, line 71: This was removed because `commands.clearContent` relies on `setContent` to be able to reset the document to the initial schema. This was early exiting and causing a no-op making `clearContent` not actually clear anything.

This is very old code that wasn't very well documented but after some testing, I'm confident in this not "leading to strange errors". `insertContentAt` is much more robust than when this was written.

I tested inserting empty content into a node and a block and it essentially resulted in a no-op.

Nantris:
Thanks for reviewing @nperez0111. Will this be merged soon? We're still stuck on 2.1.x as 2.4.0 is released due to the underlying issue I hope this PR will resolve.

nperez0111:
> Thanks for reviewing @nperez0111. Will this be merged soon? We're still stuck on 2.1.x as 2.4.0 is released due to the underlying issue I hope this PR will resolve.

I understand your urgency. We are currently in discussion of whether this is actually a breaking change or not. We have discussed the possibility of releasing an `rc` to get feedback on whether it breaks user's workflows or not.

We will get back to you soon on this. We appreciate your patience.

Nantris:
I noticed `2.5.0-beta.0` dropped today. I was wondering if this is planned for `2.5.x`? This issue remains as the only one preventing us from upgrading past `2.1.13`.

Could this be addressed by adding a flag to `setContent` and `insertContent` to enable either behavior as the end-developer requires? Otherwise I fear this is never going anywhere because it seems bound to be a breaking change for some people no matter what the final behavior. Many people have been developing with the new default behavior since November, whereas everyone on versions `2.1.13` and below is going to expect the old behavior.

Has any decision been reached? Thanks for reading!

Nantris:
I had the idea for a workaround trying to remove one of the inserted `\n` characters in a chained command so we could unblock ourselves by inserting `\n\n` instead of `\n` in the meantime, but unfortunately this produces a mismatched transaction, so this indeed remains a blocker.

nperez0111:
So, I've taken a slightly different approach with this one. I made it only ever apply the new "correct" logic when `parseOptions.preserveWhitespace === 'full' || parseOptions.preserveWhitespace === true` I did this because there is a weird mix between prosemirror parsing behaviors and tiptap parsing behaviors. Given that we already expose that property, we may as well respect what the developer intended.

`preserveWhitespace` is not entirely being respected in certain cases since tiptap does some whitespace removal of it's own on top. But with this additional it is at least a bit more in-line with prosemirror by keeping whitespace when the developer explicitly said that it should be kept entirely.

This therefore keeps backwards compatibility, while also exposing the correct behavior behind an existing option that now is respected.

Eventually this should be removed and we rely on insertContentAt's behaviors for everything, but I left only a comment to deal with that in the next major.

Thoughts @svenadlung  & @bdbch ?

Nantris:
Any update here? This has been broken for over half a year now.

nperez0111:
@Nantris it is now part of v2.5.0-beta.5

Thanks for pushing this forward

Nantris:
Many thanks to you all! Looks good!

PR #4896, opened 2024-02-16
feat(extension-link): add openOnMetaClick option to handle meta/ctrl on click

sudo97:

Add new option to link extension. `openOnMetaClick`. When it is true clicks on the link will work only when CMD/Ctrl are pressed.

## Remarks

## Checklist

- [x] Followed the guidelines
- [x] Fixed linting issues

## Related issues

### DIFF
--- a/packages/extension-link/CHANGELOG.md
+++ b/packages/extension-link/CHANGELOG.md
@@ -5,6 +5,11 @@ See [Conventional Commits](https://conventionalcommits.org) for commit guideline
 
 ## [2.2.3](https://github.com/ueberdosis/tiptap/compare/v2.2.2...v2.2.3) (2024-02-15)
 
+Add openOnMetaClick option to link extension
+**Note:** Version bump only for package @tiptap/extension-link
+
+## [2.2.3](https://github.com/ueberdosis/tiptap/compare/v2.2.2...v2.2.3) (2024-02-15)
+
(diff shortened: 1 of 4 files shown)


nperez0111:
What is the use case for this? Doesn't this break the default behavior of CMD+Click being to open in a new tab? I'm not sure that we would want to use this in everyone's editor.

Perhaps this is a good extensions for you that you can override in your editor use case?

sudo97:
Does nothing unless the option is explicitly used, so no it shouldn't break any defaults, it just adds a popular (from my point of view) option. But maybe you're right, I actually don't have resource to finish it the right way, so maybe it's better to close it

nperez0111:
I agree on closing it too. I'm not sure that we would want this behavior as an option. Though we do appreciate you trying to contribute and would encourage you to contribute again.

Extensions are very flexible so we want their behavior to be easily overridable by the consumer for such situations. We prefer extensions to configuration of every possibility.

nperez0111:
No worries! 

I'd also recommend that instead of copy-pasting the entire extension whole-sale that you can just `.extend` the existing extension and override what you need out of it. [See this example](https://tiptap.dev/docs/editor/guide/custom-extensions#extend-existing-extensions) of using a custom keyboard combo for bullet lists as an example. For your situation, you'd override the `addProseMirrorPlugins` to use your custom handler.

Happy tip tapping!

Issue #4887, opened 2024-02-14
[Bug]: setHorizontalRule inserts in a wrong position


yan-yanishevsky:
### Which packages did you experience the bug in?

extension-horizontal-rule

### What Tiptap version are you using?

2.2.2

### What’s the bug you are facing?

Placing cursor after an image block and calling `setHorizontalRule` will insert horizontal line before that image.

### What browser are you using?

Chrome

### What did you expect to happen?

`setHorizontalRule` should insert horizontal line under the cursor

### Anything to add? (optional)

Probably because of this line of code in `setHorizontalRule` function

```js
currentChain.insertContentAt(Math.max($originTo.pos - 2, 0), { type: this.name })
```

### Did you update your dependencies?

- [X] Yes, I’ve updated my dependencies to use the latest version of all packages.

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

bdbch:
Could you check if this issue still exists in the most recent versions (specially after #4898)?

I will close this issue in the meanwhile, feel free to reopen.

PR #4898, opened 2024-02-17
fix(horizontal-rule): fix insertion behavior

aguydan:

This PR contains fixes to the `HorizontalRule` extension, which behaves a bit unintuitively in certain cases.

- When a horizontal rule is placed in the position under an empty leaf node (such as an image), it is no longer inserted before the node.

| **Before** | **After** |
|:----------:|:---------:|
| no1 |      yes1     |

- If a selection ends at the beginning of a text node, it does not prevent the rule from replacing the entire selection.

| **Before** | **After** |
|:----------:|:---------:|
|       no2     |     yes2      |

- If only one node is selected, the rule is placed underneath it instead of replacing it.

| **Before** | **After** |
|:----------:|:---------:|
| no3 | yes3 |

This clause now checks if there's indeed just a cursor at this position (ensuring `from` and `to` match). The value decremented from `$originTo.pos` is now `1` instead of `2`, which prevents the rule from appearing before the node.

    if (isCursor && $originTo.parentOffset === 0) {
        currentChain.insertContentAt(Math.max($originTo.pos - 1, 0), {
            type: this.name,
        })
    }

This clause is responsible for preventing replacement of nodes by the horizontal rule.

    } else if (isNodeSelection(selection)) {
        currentChain.insertContentAt($originTo.pos, {
            type: this.name,
        })
    }

By adding an image node to the horizontal rule demo and then setting the rule.

## Remarks

The behavior of the rule replacing image nodes when selected might be something that only I find undesirable, making it more of a feature proposal than a fix, I guess.

## Checklist

- [X] Followed the guidelines
- [x] Fixed linting issues

## Related issues

- #4887


bdbch:
Looks good. Could you add a test case for this in cypress?

aguydan:
@bdbch 

Sure! I'll add it shortly when I'm back home.

aguydan:
@bdbch 

I've added test cases. I wasn't sure exactly what I should do in order to test the insertion related to image nodes. That is, whether I should modify the horizontal rule demo to include another block leaf node or add a test case to the integration tests folder. I thought it'd be best to do the latter.

Also I refactored the code a bit to get rid of trailing breaks left after selection replacement.

nperez0111 (approved):
Thank you for your contribution @aguydan

PR #4916, opened 2024-02-23
Initialize editor before setting event handlers

ondrejsevcik:

Currently, the editor is initialized after the callback bindings are hooked up. The two `useEffect` call order should be swapped. The one that hooks up callbacks is run first, then the one where the instance is created is run second. This means there is an attempt to set bindings when the editor does not exists. And once it is created, it needs to re-render and hook up the callbacks again.

Moving `useEffect` that creates editor instance to be the first one to run.

I have tested it with my basic editor. I haven't run extensive testing.

It should still work the same way it does now from API and user perspective. There will be just one less re-render on initial mount. 

## Checklist

- [x] Followed the guidelines
- [x] Fixed linting issues


nperez0111:

PR #4955, opened 2024-03-07
Fixes resizeable columns when the first row has a colspan

jaapvanhoek:
The width of columns get lost when the first row has a colspan that spans the sized columns because the colwidth (which has a comma separated list of column-widths) gets cast to an int and thus removing the other column-widths

The comma separated column-widths for the spanned columns are now split before that are cast to an int.

By calling .split on the colwidth variable each width can be parsed separately. 

I added the colwidth attribute to the existing example and noticed that it did not behave as expected. Only the first of the spanned columns got the right width. After I changed the code the result was like expected and all columns have the right size, either the given pixel value or it fills the remaining space.

1. Create a table
2. Merge the columns/cells of the first row
3. Resize multiple columns that are spanned by the first row
4. Use the output HTML as the input for a new table when tiptap loads
5. The size of each column must be equal to the value that it was given.

## Remarks

The issue was with both TableCell and TableHeader

## Checklist

- [v] The changes are not breaking the editor
- [v] Followed the guidelines
- [ ] Fixed linting issues

### DIFF
--- a/demos/src/Examples/Tables/React/index.jsx
+++ b/demos/src/Examples/Tables/React/index.jsx
@@ -155,8 +155,8 @@ export default () => {
       <table>
         <tbody>
           <tr>
-            <th>Name</th>
-            <th colspan="3">Description</th>
+            <th colwidth="200">Name</th>
+            <th colspan="3" colwidth="150,100">Description</th>
           </tr>
           <tr>
             <td>Cyndi Lauper</td>
--- a/packages/extension-table-cell/src/table-cell.ts
+++ b/packages/extension-table-cell/src/table-cell.ts
@@ -28,7 +28,7 @@ export const TableCell = Node.create<TableCellOptions>({
         parseHTML: element => {
           const colwidth = element.getAttribute('colwidth')
           const value = colwidth
-            ? [parseInt(colwidth, 10)]
+            ? colwidth.split(',').map(width => parseInt(width, 10))
             : null
 
           return value
--- a/packages/extension-table-header/src/table-header.ts
+++ b/packages/extension-table-header/src/table-header.ts
@@ -27,7 +27,7 @@ export const TableHeader = Node.create<TableHeaderOptions>({
         parseHTML: element => {
           const colwidth = element.getAttribute('colwidth')
           const value = colwidth
-            ? [parseInt(colwidth, 10)]
+            ? colwidth.split(',').map(width => parseInt(width, 10))
             : null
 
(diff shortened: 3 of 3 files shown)


bdbch (approved):
lgtm

jaapvanhoek:
Are there any things I could or should do to ensure that this pull request can be merged?

nperez0111:
@jaapvanhoek I've looked at this before. The thing that worries me is why colwidth would have a comma separated list of values. I haven't looked into whether that is how it works right now or not so maybe you can shed some light on it?

Also seeing more tests would give me higher confidence that it actually achieves the intended behavior without regressions

jaapvanhoek:
> The thing that worries me is why colwidth would have a comma separated list of values. I haven't looked into whether that is how it works right now or not so maybe you can shed some light on it?
@nperez0111 This is how it currently works when the steps to reproduce are followed, and it has nothing to do with my changes. The only thing I did was change the way the value is handled and make it possible to read the comma-separated values.

Later today I will try add some tests for the table-header extension.

nperez0111:
@jaapvanhoek thanks for letting me know, I haven't really touched the table extension or read through it to understand it yet. So you can understand why I'm cautious of it. Tests would definitely give me more confidence, so I appreciate it

jaapvanhoek:
After spending some time trying to run the tests, they keep failing at the Cypress step with the error: "Cypress failed to verify that your server is running". Based on what I've found online, this issue might be related to the setup I'm using (MacOS Sonoma 14.6.1), but none of the suggested solutions have worked for me. As a result, I'm currently unable to add tests to the project. Hopefully, someone else can shed some light on why I can't run the tests at all.

nperez0111:
Hi @jaapvanhoek , so to run the tests you need to have both the development server and the test server running. the development server runs on localhost:3000

So in one terminal you'll need `npm run dev` and in another, `npm run test:open`

it will look like this:
image

This opens the cypress UI that allows you to run tests individually in a nide interface rather than only in the terminal which is useful for writing tests.

I hope this is helpful for you

jaapvanhoek:
@nperez0111 I have added the tests you requested, any suggestions?

SanderLeenders:
@nperez0111 @bdbch can we get an update on this issue?

nperez0111 (approved):
Sorry for the long wait, @jaapvanhoek really appreciate it, I checked on my side & it is working correctly. Thanks for your contribution!

PR #4973, opened 2024-03-12
React: fix editor cleanup

kfirmanty:

Hi! I noticed that editor.destroy() is not being called for each reference that is being created, this should fix the problem.

Cleaning up editor was moved to the return of useEffect that creates a new editor, this way for each editor instantiated there will be a respective cleanup fn.

## Checklist

- [x] Followed the guidelines
- [x] Fixed linting issues


bdbch (approved):
Thanks, lgtm!

PR #4976, opened 2024-03-13
Feature/y sync options: allow passing ySyncOptions to extension-collaboration

janthurau:
This change makes it possible to pass ySyncOptions to extension-collaboration, which enabled advanced features such as `permanentUserData` or `colors` and `colorMapping`.

### DIFF
--- a/packages/extension-collaboration/src/collaboration.ts
+++ b/packages/extension-collaboration/src/collaboration.ts
@@ -7,8 +7,9 @@ import {
   yUndoPlugin,
   yUndoPluginKey,
 } from 'y-prosemirror'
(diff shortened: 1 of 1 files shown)


packages/extension-collaboration/src/collaboration.ts
```diff
@@ -7,6 +7,7 @@ import {
   yUndoPlugin,
   yUndoPluginKey,
 } from 'y-prosemirror'
+import type { YSyncOpts } from 'y-prosemirror/dist/src/plugins/sync-plugin'
```
janthurau, line 10: ..

bdbch (approved):

bdbch:
@janthurau we should merge this into main right?

Issue #1547, opened 2021-07-06
Bug: keyboard shortcuts fire in reverse order based on priority


thatsjonsense:
**Description**
If I have two extensions that both register the same shortcut, the one with the _lower_ priority takes precedence. The docs say it should be higher.

The issue is with this line: https://github.com/ueberdosis/tiptap/blob/28737b1631719315cebc354f7cb5a1d924b0b543/packages/core/src/ExtensionManager.ts#L209

I don't understand why it's reversing. The extensions are sorted from highest priority to low, but this reverses them for adding plugins.

According to [ProseMirror keymap docs](https://prosemirror.net/docs/ref/#keymap):
> The order in which they appear determines their precedence (the ones early in the array get to dispatch first).

Because we're reversing, the highest priority extensions are dispatching last.

HeebeLee:
I was also misled by this at that time. Later, when I looked at other components, I found that the smaller the more advanced it was.

However, at present, if it is to be executed in reverse order, I believe it should not be necessary, but the documents should be updated in time.

Issue #4006, opened 2023-05-02
[Bug]: Cmd+Shift+B should not trigger bold - and same for the other marks


Nantris:
### Which packages did you experience the bug in?

bold

### What Tiptap version are you using?

2.0.0

### What’s the bug you are facing?

The hotkeys for triggering marks should be limited to _exactly_ the keys specified and not trigger in response to additional keys like Shift.

Same for italics, underline, and perhaps more.

### What browser are you using?

Chrome on Linux (so specifically using ctrl, not cmd)

### Code example

https://tiptap.dev/api/marks/bold

### What did you expect to happen?

Cmd+Shift+B or Ctrl+Shift+B should do nothing - but instead they add bold.

### Anything to add? (optional)

This also interferes with the blockquote extension's default hotkeys of Cmd+Shift+B: https://tiptap.dev/api/nodes/blockquote

### Did you update your dependencies?

- [X] Yes, I’ve updated my dependencies to use the latest version of all packages.

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

ashu12chi:
I also faced this issue, the workaround I used here is to explicitly make `Ctrl+Shift+B` no operation.

apaar97:
+1

Slapbox:
It appears to also be the cause behind this issue: #4005 - and I assume others.

Slapbox:
There seems to be no way to work around this besides to tap into ProseMirror which seems like _extreme_ overkill.

@bdbch or @svenadlung could we get this added to the roadmap for TipTap 2.2.0?

Slapbox:
I've created #4111 to track the related hotkey failures that affect TipTap by default.

@fbolcic pinging you specifically because you're the only person who reacted but is not subscribed.

PR #4995, opened 2024-03-20
fix: make blockquote shortcut work in starter-kit

### LINKED ISSUE #4994: [Bug]: Blockquote shortcut does not work when using starter-kit

starter-kit

2.2.4

When using the starter-kit, using keyboard shortcut for Blockquote (`Control Shift B` / `Cmd Shift B`) instead triggers the Bold command (`Control B` / `Cmd B`). This is reproducible in the official [TipTap Default text editor example](https://tiptap.dev/docs/editor/examples/default).

Chrome

### Code example

https://tiptap.dev/docs/editor/examples/default

The Blockquote shortcut (`Control Shift B` / `Cmd Shift B`)  should fire the `toggleBlockquote` command, and not the `toggleBold` command, when using the starter-kit.

_No response_

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

Mathias-S:
Because blockquote uses `Mod-Shift-b`, and bold uses `Mod-b`, the bold shortcut will override the blockquote shortcut if added to the extensions later.

Fixes #4994

Changed the order in which the `bold`/`blockquote` extensions are added in the starter-kit. This is because plugins added later will override plugins added earlier, because of f8efdf797a10a01235b75091729b15aca076e47a (#1547).

Changed the order in which the `bold`/`blockquote` extensions are added in the starter-kit. I realise that this breaks the alphabetical ordering, but it's more important that the shortcuts work correctly.

Tested locally with the `Examples/Default` editor, which uses `starter-kit`. 

Load the Examples/Default example and check that `Mod-shift-b` (`Control Shift B` / `Cmd Shift B`) toggles Blockquote, and that `Mod-b` (`Control B` / `Cmd B`) toggles Bold, as expected.

## Remarks

It would be better if `Mod-b` didn't override `Mod-shift-b` regardless of plugin order, but it seems like this is handled by Prosemirror.

## Checklist

- [x] Followed the guidelines
- [x] Fixed linting issues

## Related issues

* Fixes #4994
* Possibly caused by f8efdf797a10a01235b75091729b15aca076e47a (#1547)

### DIFF
--- a/.changeset/five-tigers-kick.md
+++ b/.changeset/five-tigers-kick.md
@@ -1,0 +1,5 @@
+---
+"@tiptap/starter-kit": patch
+---
+
+fix(starter-kit): make blockquote shortcut work in starter-kit
--- a/packages/starter-kit/src/starter-kit.ts
+++ b/packages/starter-kit/src/starter-kit.ts
@@ -139,12 +139,12 @@ export const StarterKit = Extension.create<StarterKitOptions>({
   addExtensions() {
     const extensions = []
 
-    if (this.options.blockquote !== false) {
-      extensions.push(Blockquote.configure(this.options?.blockquote))
-    }
-
     if (this.options.bold !== false) {
       extensions.push(Bold.configure(this.options?.bold))
+    }
+
+    if (this.options.blockquote !== false) {
+      extensions.push(Blockquote.configure(this.options?.blockquote))
     }
 
     if (this.options.bulletList !== false) {


rfgamaral:
> [!IMPORTANT]
> I'm not a Tiptap maintainer, just someone that cares about it. Opinions are my own, and not the views of Tiptap maintainers.

Relying on code order execution is generally not a good idea to fix an issue like this. There's nothing in the `StarterKit` code that would make you think that order is important for one to worry about when modifying that file.

IMO, a better solution is to increase the `Blockquote` extension priority to be higher than the `Bold` one. This is something we do ourselves in Typist (see [here](https://github.com/Doist/typist/blob/main/src/extensions/rich-text/rich-text-kit.ts#L204)).

Mathias-S:
I agree that I'm not a huge fan of simply changing the order, and it should at least have a comment in the code explaining it.

I definitely agree with you that in your own project, it makes a lot of sense to explicitly set `priority`, but I'm not sure it's the best way to solve it in `starter-kit`. However, priority _is_ explicitly modified based on execution order (see f8efdf797a10a01235b75091729b15aca076e47a), and if we were to increase Blockquote's `priority` directly in `starter-kit`, this might cause other users' custom extensions to no longer override blockquote from the starter-kit.

rfgamaral:
> and if we were to increase Blockquote's `priority` directly in `starter-kit`, this might cause other users' custom extensions to no longer override blockquote from the starter-kit.

I personally don't see that as a problem, as long as such a change is marked as breaking change.

> Priority _is_ explicitly modified based on execution order (see [f8efdf7](https://github.com/ueberdosis/tiptap/commit/f8efdf797a10a01235b75091729b15aca076e47a))

I could be wrong, but I don't think that's related to **extensions priorities**, that's the ProseMirror plugin order, which is what is run when an extension makes use of `addProseMirrorPlugins`.

Mathias-S:
> > and if we were to increase Blockquote's `priority` directly in `starter-kit`, this might cause other users' custom extensions to no longer override blockquote from the starter-kit.
> 
> I personally don't see that as a problem, as long as such a change is marked as breaking change.

I agree, but I doubt TipTap is going to make a breaking release/major version bump to fix such a minor bug.

> > Priority _is_ explicitly modified based on execution order (see [f8efdf7](https://github.com/ueberdosis/tiptap/commit/f8efdf797a10a01235b75091729b15aca076e47a))
> 
> I could be wrong, but I don't think that's related to **extensions priorities**, that's the ProseMirror plugin order, which is what is run when an extension makes use of `addProseMirrorPlugins`.

The commit was made in response to #1547, which was about shortcut priority, so it's quite related. In fact, I think this bug is _caused_ by that commit.

rfgamaral:
> I agree, but I doubt TipTap is going to make a breaking release/major version bump to fix such a minor bug.

I don't see why not. If it's a breaking change, no matter how small, it deserves its own major version. It's just a number.

But again, I'm not a maintainer, I don't make decisions, just voicing my own opinion. I tend to prefer explicit code where comments explaining things are avoided, if possible. But that's just me.

bdbch (approved):

Nantris:
I haven't tested or reviewed these changes, but wanted to mention that if these changes work it would likely make #4006 able to be closed.

Nantris:
> yet it's something to be discussed as should all extension by themselves have the same priority as others or should they all come with their own "guessed" priorioty based on what could potentially overwrite them?

The team has already made `Link` extension a higher priority to address some issues. I think making one other exception is not unreasonable.

Mathias-S:
Anything that should be changed in order to get this merged?

I agree with other commenters that finding a more reliable way of fixing this issue would be a good idea, but this specific change will simply fix the bug in starter-kit without really introducing any other possible issues, so it should be the lowest-risk way of fixing the bug and will help everyone just using the starter-kit.

nperez0111:
Yea, I think that this is fine to merge in as is. I'll get it in

PR #5019, opened 2024-03-29
Add `blockSeparator` option to `clipboardTextSerializer` core extension

hivokas:

This PR adds `blockSeparator` option to `clipboardTextSerializer` core extension. 

It will allow to override block separator (without copy pasting the entire extension from core) and add it to my editor.

```js
import { extensions } from '@tiptap/core';

const editor = new Editor({
  extensions: [
    extensions.ClipboardTextSerializer.configure({
      blockSeparator: '\n',
    }),
  ],
});
```

By making it possible to provide `blockSeparator` option to `clipboardTextSerializer` core extension.

I've copied the implementation from this PR to my project and tested it with and without `blockSeparator` parameter.

By doing this:

```js
import { extensions } from '@tiptap/core';

const editor = new Editor({
  extensions: [
    extensions.ClipboardTextSerializer.configure({
      blockSeparator: '\n',
    }),
  ],
});
```

## Remarks

If there is a better way to implement it, I'm open to it. 

## Checklist

- [x] Followed the guidelines
- [x] Fixed linting issues

## Related issues

N/A

### DIFF
--- a/packages/core/src/Editor.ts
+++ b/packages/core/src/Editor.ts
@@ -9,7 +9,9 @@ import { EditorView } from '@tiptap/pm/view'
 import { CommandManager } from './CommandManager.js'
 import { EventEmitter } from './EventEmitter.js'
 import { ExtensionManager } from './ExtensionManager.js'
-import * as extensions from './extensions/index.js'
+import {
+  ClipboardTextSerializer, Commands, Editable, FocusEvents, Keymap, Tabindex,
+} from './extensions/index.js'
 import { createDocument } from './helpers/createDocument.js'
 import { getAttributes } from './helpers/getAttributes.js'
 import { getHTMLFromFragment } from './helpers/getHTMLFromFragment.js'
@@ -32,7 +34,7 @@ import {
 import { createStyleTag } from './utilities/createStyleTag.js'
 import { isFunction } from './utilities/isFunction.js'
 
-export { extensions }
+export * as extensions from './extensions/index.js'
 
 export interface HTMLElement {
   editor?: Editor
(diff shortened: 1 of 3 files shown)


bdbch:
Nice idea!

```js
const editor = new Editor({
  extensions: [
    extensions.ClipboardTextSerializer.configure({
      blockSeparator: '\n',
    }),
  ],
});
```

The `extensions.ClipboardTextSerializer.configure` approach looks weird though - since this is a core extension, maybe we should add the option to the editors `EditorOptions` object?

https://github.com/ueberdosis/tiptap/blob/6f6e49ecdf78cf6fd63ad1eb7006fb5afd4064a1/packages/core/src/Editor.ts#L56-L78

That way users would not need to do some weird extension-merging-shenanigans to get the option down to the core extensions.

`clipboardBlockSeparator: '\n'` would be the option to set the option for the extension then?

hivokas:
> `clipboardBlockSeparator: '\n'` would be the option to set the option for the extension then?

Yes, this approach looks good to me. Will you implement this change?

packages/core/src/Editor.ts
```diff
@@ -235,7 +238,17 @@ export class Editor extends EventEmitter<EditorEvents> {
    * Creates an extension manager.
    */
   private createExtensionManager(): void {
-    const coreExtensions = this.options.enableCoreExtensions ? Object.values(extensions) : []
+
+    const coreExtensions = this.options.enableCoreExtensions ? [
```
bdbch, line 242: Since we only have less than 10 core extensions I'd think it's okay to just load them manually here – that way we can pass configurations down to them.

bdbch:
@hivokas I added a commit that adds coreExtension options to the editor options props.

bdbch:
@svenadlung since I also made changes here it would be best if you could also just review again :)

svenadlung (approved):

hivokas:
Thanks @bdbch!

PR #5022, opened 2024-03-31
feat: add support for default link protocol

henryStelle:

Added the ability to specify a default protocol for the auto link features in the Link extension.

Added a configurable `defaultProtocol` setting (the setting defaults to the previous default so there is no breaking change) that is then passed to linkifyjs to do its magic.

Added 4 tests to the Link mark demo testing the positive and negative cases. I also confirmed these new tests fail without the added setting.

## Remarks

Open up the Link demo and set the `defaultProtocol` to something strange like `ftp` and then type example.com. The href of the generated link should be ftp://example.com. Then, type in http://example.com and confirm the href is still http://example.com.

## Checklist

- [x] Followed the guidelines
- [x] Fixed linting issues

## Related issues

https://github.com/ueberdosis/tiptap/discussions/5007


bdbch (dismissed):

packages/extension-link/src/link.ts
```diff
@@ -94,6 +99,7 @@ export const Link = Mark.create<LinkOptions>({
       linkOnPaste: true,
       autolink: true,
       protocols: [],
+      defaultProtocol: 'http',
```
nperez0111, line 102: A better default for the modern web should be `https`

Dragonite24:
Is there any information about when the feature is planned to be released? Thank you

nperez0111 (approved):
Looks good

Issue #5018, opened 2024-03-29
[Bug]: React DND broken when used within a NodeViewWrapper


johnsmithecon:
### Which packages did you experience the bug in?

@tiptap/react

### What Tiptap version are you using?

2.2.4

### What’s the bug you are facing?

`<NodeViewWrapper
      className="border p-4 shadow-lg rounded-lg bg-white"
      style={{ height: "500px" }}
    >
    <MyComponent />
</NodeViewWrapper>`

### What browser are you using?

Chrome

### What did you expect to happen?

When placing my custom react component (which includes react-dnd functionality) within tiptap, specifically a NodeViewWrapper, I expect the items to be draggable using react-dnd.

### Did you update your dependencies?

- [X] Yes, I’ve updated my dependencies to use the latest version of all packages.

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

Pruxis:
We had the same, I used https://dndkit.com. However it's having terrible side effects sometimes, I'd recommend to use https://github.com/atlassian/pragmatic-drag-and-drop

nperez0111:
This is released with version: `2.11.0`

PR #5023, opened 2024-04-01
Fix : React DND broken when used within a NodeViewWrapper

tej1562:
Drag event doesn't work on components wrapped inside a NodeViewWrapper.

Currently, drag events are explicitly prevented on NodeView. This is also applied to NodeView children due to event propagation, thus preventing drag events on the components inside NodeView. This PR adds a check to ensure drag events are only prevented on the NodeView parent.

Added event target check to ensure drag prevention on NodeView doesn't propagate to child components.

```
    if (!isDraggable && isSelectable && isDragEvent && event.target === this.dom) {
      event.preventDefault()
    }

    if (isDraggable && isDragEvent && !isDragging && event.target === this.dom) {
      event.preventDefault()
      return false
    }
```

Created a NodeView with a react dnd component using NodeViewWrapper. Drag works as expected on the react dnd component.

Create a NodeView with draggable component and check if component can be dragged.

## Remarks

## Checklist

- [x] Followed the guidelines
- [x] Fixed linting issues

## Related issues

#5018


bdbch:
Could be that it comes from these checks:

https://github.com/ueberdosis/tiptap/blob/05ed5560215bf81726d993925e9ed39930136d3d/packages/core/src/NodeView.ts#L135-L145

not sure though - would need some test case in our demo for that.

bdbch:
Since you check for `this.dom` - would that break something if I drag a child node of my prosemirror node that is not supposed to be draggable?

bdbch:
Sorry, misclicked there!

Oops

tej1562:
The drag on child node won't affect the drag behaviour of parent node. Here, `this.dom` refers to dom element representing the parent NodeView and it's drag behaviour only depends on the `draggable` property. If `draggable` is set to `false` on the NodeView, then it remains undraggable but child nodes within NodeView can still be dragged.

bdbch (approved):
LGTM

Issue #5031, opened 2024-04-04
[Bug]: Performance issue with large documents migrating from vue 2 to vue 3


Rirax:
### Which packages did you experience the bug in?

@tiptap/vue-3

### What Tiptap version are you using?

2.2.4

### What’s the bug you are facing?

Hello everyone,

In my company, we're fervent Tiptap users, and Tiptap pro subscribers. Today we're experiencing a blocking performance issue with `tiptap/vue-3`.

Since migrating from `tiptap/vue-2` to `tiptap/vue-3`, we've noticed a **slowdown on documents with a large number of nodes that need to be rendered in Vue**.

A look at the [thread on tiptap-vue3 implementation](https://github.com/ueberdosis/tiptap/issues/1166#issuecomment-777792156)](https://github.com/ueberdosis/tiptap/issues/1166#issuecomment-777792156) mentions the breaking changes between vue-2 and vue-3, and the use of Teleport to add elements to the DOM, which may explain this drop in performance.

To demonstrate the bug, you need content with at least 1500 nodes.

The loading time increases quadratically with the number of nodes.

The repository below reproduces the bug 👇

### What browser are you using?

Chrome

### Code example

https://codesandbox.io/p/devbox/tiptap-vue-3-vuenodeviewrenderer-mg4dp8?file=%2Fsrc%2FApp.vue%3A16%2C1

### What did you expect to happen?

We are a knowledge management software company, and our applications are particularly dependent on tiptap for the user experience.

We are in the process of migrating our stack from Vue 2 to Vue 3 and as a result are looking to adopt tiptap/vue-3.
The current performance issue is blocking our migration to Vue3.

### Anything to add? (optional)

Let us know if we can be of any help to reproduce this bug or contribute a fix to it.

### Did you update your dependencies?

- [X] Yes, I’ve updated my dependencies to use the latest version of all packages.

### Are you sponsoring us?

- [X] Yes, I’m a sponsor. 💖

bdbch:
Thanks for reporting! I tried to get a quick grasp via performance debugging in Chrome Dev Tools but seems like it even crashes my chrome instance on that.

I just saw that the NodeViewRenderer takes a lot of time for each component - I don't have a Vue 2 setup now that I could easily test. Do you still have your code around that you could put into a sandbox real quick so I can do my tests on that to compare where and what increases the rendering time that much?

bdbch:
[...]

PR #5041, opened 2024-04-08
Draft: (vue-3) render node only once

Rirax:

In this PR we change the way nodes are rendered into Vue components and held in locale state during content loading.

For each node in `EditorContent.ts`, the current node and all its predecessors are rendered for each new node, which implies a performance issue.

To avoid all nodes being rendered each time, we now keep the rendered nodes as a reference

Since we cannot mutate the `Editor` props in `render` method of `EditorContent`, we move the logic to render a node in `VueRenderer.ts`.
Instead of keeping a Map of VueRenderer instance for each node, we keep the rendered VNodes in Map.

For each new node, we render it and add it to the existing Map.
The `render` method of `EditorContent` now retrieves the VNodes and renders them.

## Remarks

## Checklist

- [x] Followed the guidelines
- [x] Fixed linting issues

## Related issues

[Link to the related issues here](https://github.com/ueberdosis/tiptap/issues/5031)

### DIFF
--- a/packages/vue-3/src/Editor.ts
+++ b/packages/vue-3/src/Editor.ts
@@ -7,9 +7,10 @@ import {
   markRaw,
   reactive,
   Ref,
+  RendererElement,
+  RendererNode,
+  VNode,
 } from 'vue'
-
-import { VueRenderer } from './VueRenderer.js'
 
 function useDebouncedRef<T>(value: T) {
   return customRef<T>((track, trigger) => {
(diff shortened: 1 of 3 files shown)


bdbch (approved):

bdbch:
@Rirax LGTM but it seems there is a test failing in a Vue demo (regarding a Vue demo for the Drawing example).
Could you verify and maybe take a look at that?

Rirax:
Hi @bdbch thanks for the review. I've revised my approach here
https://github.com/ueberdosis/tiptap/pull/5206

nperez0111:
Closed in favor of https://github.com/ueberdosis/tiptap/pull/5206

PR #5046, opened 2024-04-09
Optionally apply input and paste rules when using insertContent methods

svenadlung:

First call to enable the option of applying input and paste rules when using `insertContent`/`insertContentAt`. This is interesting, for example, if streamed content contains markdown characters and characters are added character by character within the editor.

Added `applyInputRules` and `applyPasteRules` to `insertContent`/`insertContentAt` options. Within the InputRule and PasteRule plugin I created some logic to trigger the rules.

I wrote tests and checked a few scenarios by hand.

Check the "InsertContentApplyingRules" demo.
https://deploy-preview-5046--tiptap-embed.netlify.app/src/commands/insertcontentapplyingrules/react/

## Checklist

- [x] Followed the guidelines
- [x] Fixed linting issues

### DIFF
--- a/packages/core/src/PasteRule.ts
+++ b/packages/core/src/PasteRule.ts
@@ -154,6 +154,16 @@ function run(config: {
   return success
 }
 
+const createClipboardPasteEvent = (text: string) => {
+  const event = new ClipboardEvent('paste', {
+    clipboardData: new DataTransfer(),
+  })
+
+  event.clipboardData?.setData('text/html', text)
+
+  return event
+}
+
 /**
  * Create an paste rules plugin. When enabled, it will cause pasted
  * text that matches any of the given rules to trigger the rule’s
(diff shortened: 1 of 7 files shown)


mkriegeskorte (approved):
From my understanding the changes are reasonable.
And as you haven’t added any dots in our test suite I’d say LGTM (looking at you xz ;))

packages/core/src/PasteRule.ts
```diff
@@ -167,6 +177,45 @@ export function pasteRulesPlugin(props: { editor: Editor; rules: PasteRule[] }):
...
+  const processEvent = ({
+    state,
+    from,
+    to,
+    rule,
+    pasteEvt,
+  }: {
+    state: EditorState
+    from: number
+    to: { b: number }
```
bdbch, line 189: Any reason why to is not just the number too?

packages/core/src/PasteRule.ts
```diff
@@ -167,6 +177,45 @@ export function pasteRulesPlugin(props: { editor: Editor; rules: PasteRule[] }):
...
+      transaction: tr,
+    })
+
+    const handler = run({
+      editor,
+      state: chainableState,
+      from: Math.max(from - 1, 0),
+      to: to.b - 1,
+      rule,
+      pasteEvent: pasteEvt,
```
bdbch, line 205: Can't we take `pasteEvent` as a variable name and also use a shorthand here?

bdbch (approved):
LGTM - just had a few comments. :)

packages/core/src/PasteRule.ts
svenadlung: There is already a global variable here with that name: https://github.com/ueberdosis/tiptap/pull/5046/files#diff-6eddae5591c39b131ab5597e8ac574df71fafda96023c4573c45fd73ed066b9fR177

packages/core/src/PasteRule.ts
svenadlung: > Any reason why to is not just the number too?

Yeah, this is a lil dirty. This is what `findDiffX()` returns [here](https://github.com/ueberdosis/tiptap/pull/5046/files#diff-6eddae5591c39b131ab5597e8ac574df71fafda96023c4573c45fd73ed066b9fR289). `{ a: number; b: number }`

packages/core/src/PasteRule.ts
```diff
@@ -212,45 +261,45 @@ export function pasteRulesPlugin(props: { editor: Editor; rules: PasteRule[] }):
...
-          state,
-          transaction: tr,
-        })
-
-        const handler = run({
-          editor,
-          state: chainableState,
-          from: Math.max(from - 1, 0),
-          to: to.b - 1,
+        return processEvent({
```
nperez0111, line 296: I see why this `processEvent` function was extracted, but all that it is doing is switching the arguments that it is operating on.
I think it'd be clearer to inline it again, and have the `if (isSimulatedPaste) {` switch the values it operates on.

svenadlung: Haha, tbh that's what I had done before and I thought it might be easier to understand that way 😂

PR #5058, opened 2024-04-11
tsx version

nicitaacom:

[add a description of your changes here]

## Remarks

## Checklist

- [ ] Followed the guidelines
- [ ] Fixed linting issues

## Related issues

### DIFF
--- a/demos/src/Examples/Community/React/MentionList.tsx
+++ b/demos/src/Examples/Community/React/MentionList.tsx
@@ -1,0 +1,73 @@
+import React, { forwardRef, useEffect, useImperativeHandle, useState, ReactNode } from 'react';
+
+interface MentionListProps {
+  items: string[];
+  command: (item: { id: string }) => void;
+}
+
+const MentionList = forwardRef((props: MentionListProps, ref) => {
+  const [selectedIndex, setSelectedIndex] = useState(0);
+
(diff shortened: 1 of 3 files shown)


bdbch:
Could you please explain what you are trying to solve with this PR or what kind of demo this is?

nicitaacom:
> Could you please explain what you are trying to solve with this PR or what kind of demo this is?

https://tiptap.dev/docs/editor/examples/suggestions
I see js here
I'm using TypeScript
So it would be nice if I would see TypeScript version there

bdbch:
I see. For our demos though we would like to stick to Javascript as it is easier to understand for more people. Typescript developers should be fine understanding Vanilla Javascript & Javascript devs should be able to also understand our examples without learning Typescript first. Thanks for your PR though!

nicitaacom:
> Typescript developers should be fine understanding Vanilla Javascript

That's right

I don't understand which type editor has
Also this is some uncommon type for clientRect

```ts
interface SuggestionProps {
  query: string;
  editor: any;
  clientRect?: (() => DOMRect | null) | null | undefined;
}
```

bdbch:
editor here can either be the type `Editor` of the core package or from one of the library packages. Could improve typings there.

The clientRect type is that way because it's the callback function that is required for the tippyjs props.

PR #5070, opened 2024-04-15
fix(docs): correct capitalization in signing header

EricHasegawa:
Lowercases the signing header for webhook requests.

The request header for signing webhook requests is displayed with capitalizations in the docs. 

HTTP headers are case insensitive, but if you're in (for example) a NextJS app and look for a `X-Hocuspocus-Signature-256` header in a webhook request, it'll return undefined. However, looking for a `x-hocuspocus-signature-256` header will work.

### DIFF
--- a/docs/collaboration/webhook.md
+++ b/docs/collaboration/webhook.md
@@ -22,7 +22,7 @@ A sample payload of the webhook request looks like this:
 
 ### Signing
 
-All requests to your webhook URL will contain a header called `X-Hocuspocus-Signature-256` that signs the entire message with your secret. You can find it in the [settings](https://collab.tiptap.dev/apps/settings) of your Tiptap Collab app.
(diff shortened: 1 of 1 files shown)


docs/collaboration/webhook.md
```diff
@@ -22,7 +22,7 @@ A sample payload of the webhook request looks like this:
 
 ### Signing
 
-All requests to your webhook URL will contain a header called `X-Hocuspocus-Signature-256` that signs the entire message with your secret. You can find it in the [settings](https://collab.tiptap.dev/apps/settings) of your Tiptap Collab app.
```
nperez0111, line 25: HTTP headers are, out of convention, capitalized see MDN's docs: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers

Your concern is around a Next.js implementation detail. The docs are more correct here

PR #5094, opened 2024-04-25
fix: NodePos querySelectorAll function

Dalcvi:
Current implementation of `querySelectorAll` has a few flaws:

- Doesn't send given attributes and `firstItemOnly` params when traversing deeper in to the tree using recursion:
`nodes = nodes.concat(childPos.querySelectorAll(selector)`
- If `firstItemOnly` is given, it only skips rest of the code for one cycle when a node with correct selector and attributes is found. It doesn't skip other cycles, so it keeps searching for more nodes!
For example, if we have this DOC structure:
```
[
  {
    type: 'COLLAPSIBLE_LIST',
    children: [
      { type: 'COLLAPSIBLE_LIST_ITEM', attrs: { color: 'red', id: '1' } },
      { type: 'COLLAPSIBLE_LIST_ITEM', attrs: { color: 'blue', id: '2' } },
    ],
  },
  {
    type: 'COLLAPSIBLE_LIST',
    children: [
      { type: 'COLLAPSIBLE_LIST_ITEM', attrs: { color: 'blue', id: '3' } },
      { type: 'COLLAPSIBLE_LIST_ITEM', attrs: { color: 'red', id: '4' } },
    ],
  },
];
```
And we are searching for `COLLAPSIBLE_LIST_ITEM` with `color: 'red'` as an attribute and `firstItemOnly` set to true, the returned result will be:
```
[
  { type: 'COLLAPSIBLE_LIST_ITEM', attrs: { color: 'red', id: '1' } },
  { type: 'COLLAPSIBLE_LIST_ITEM', attrs: { color: 'red', id: '4' } }
]
```
Even if the `{ type: 'COLLAPSIBLE_LIST_ITEM', attrs: { color: 'red', id: '4' } }` had more children with `COLLAPSIBLE_LIST_ITEM` that have `color: 'red'` it will take those too!

- Cleaned up the code a bit, as the indentation was getting too big (personal preference, hopefully you won't disagree with this change)


bdbch (approved):
lgtm

Dalcvi:
Should've fixed all issues in tests @bdbch

bdbch (approved):

PR #5097, opened 2024-04-25
Text Align Extension: issue with DOM update when multiple types exist and not every exist at the DOM of the editor.

echatzief:
…

I have changed the `setTextAlign` and `unsetTextAlign` because if you used multiple types lets say ['heading', 'paragraph'] and at the DOM of the editor we only have a paragraph only no heading. Then when we call `setTextAlign` then it will go the the every loop and at the first try to updateAttributes for heading it will fail and it will stop there an the paragraph won't update because that how the every loop works. It stops on the first falsy value. Same happens for the `unsetTextAlign` function

I changed the setTextAlign and unsetTextAlign to use a map to go through every update and then validate the updates. That how we avoid the bug described above

Yes. I have tested locally with a Vue3 project

Try the scenario that we described.

## Checklist

- [X] Followed the guidelines
- [X] Fixed linting issues


nperez0111 (approved):
This looks right to me

bdbch:
Nice change. LGTM!

bdbch (approved):

echatzief:
If it looks good then merge it 😀

bdbch:
@echatzief sometimes we wait for another maintainer to also take a look

PR #5106, opened 2024-04-30
fix issue with code pasting from VS Code when at the last line of code

bdbch:

This PR fixes an minor issue where code copied from VSCode wouldn't be correctly pasted at the end of a document because `replaceSelectionWith` would cause a buggy insertion when the end position of the selection is in a node at the end of the document.

I'm checking if the cursor is at the end - if so I insert the code block before the current position - otherwise I use the replaceSelectionWith function

Tested it locally in the default demo

See above

## Checklist

- [x] Followed the guidelines
- [x] Fixed linting issues

### DIFF
--- a/packages/extension-code-block/src/code-block.ts
+++ b/packages/extension-code-block/src/code-block.ts
@@ -260,8 +260,15 @@ export const CodeBlock = Node.create<CodeBlockOptions>({
 
             const { tr } = view.state
 
-            // create an empty code block
-            tr.replaceSelectionWith(this.type.create({ language }))
+            // create an empty code block´
+            // if the cursor is at the absolute end of the document, insert the code block before the cursor instead
+            // of replacing the selection as the replaceSelectionWith function will cause the insertion to
+            // happen at the previous node
+            if (view.state.selection.from === view.state.doc.nodeSize - (1 + (view.state.selection.$to.depth * 2))) {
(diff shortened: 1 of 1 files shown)


packages/extension-code-block/src/code-block.ts
```diff
@@ -260,8 +260,15 @@ export const CodeBlock = Node.create<CodeBlockOptions>({
 
             const { tr } = view.state
 
-            // create an empty code block
-            tr.replaceSelectionWith(this.type.create({ language }))
+            // create an empty code block´
+            // if the cursor is at the absolute end of the document, insert the code block before the cursor instead
+            // of replacing the selection as the replaceSelectionWith function will cause the insertion to
+            // happen at the previous node
+            if (view.state.selection.from === view.state.doc.nodeSize - (1 + (view.state.selection.$to.depth * 2))) {
```
nperez0111, line 267: I'm not going to pretend that I completely understand this. What is tripping me up is the `* 2`.

But otherwise this looks right

nperez0111 (approved):

packages/extension-code-block/src/code-block.ts
bdbch: The `2` is the offset of a prosemirror node to the next one. If you for example have a nested node - the wrapping node will always add an offset of 1 at the beginning and the end which you have to account for if you want to find the correct position after said node.

PR #5145, opened 2024-05-14
fix: tab shortcut always returns true in listitem

halil-pan:
## Changes Overview
When I click tab on listItem or taskItem, the default `sinkListItem` command is executed and the result is returned. When `sinkListItem` returns `false`, the browser default behavior is automatically triggered, such as focusing on input on the page.

## Implementation Approach
In the shortcut key handling function of listItem, always returns true for tab-related events, preventing subsequent behavior of the event

## Verification Steps
1. Open preview page: [local](http://localhost:3000/preview/Experiments/All) or [online](https://deploy-preview-5145--tiptap-embed.netlify.app/preview/Experiments/All)
2. Select the first line of any list
3. Click tab, before, the page will automatically jump to the taskItem position, after this change, nothing will happen.

## Checklist

### DIFF
--- a/packages/extension-list-item/src/list-item.ts
+++ b/packages/extension-list-item/src/list-item.ts
@@ -57,8 +57,14 @@ export const ListItem = Node.create<ListItemOptions>({
   addKeyboardShortcuts() {
     return {
       Enter: () => this.editor.commands.splitListItem(this.name),
-      Tab: () => this.editor.commands.sinkListItem(this.name),
-      'Shift-Tab': () => this.editor.commands.liftListItem(this.name),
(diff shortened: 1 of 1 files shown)


nperez0111:
We appreciate your contribution.

Unfortunately, this would break browser default behavior that may be depended on in other editor implementations.

And, it not returning true is by design so that other events may be triggered by the same keystroke.

So for that reason, I'm closing this PR. If your editor needs this behavior we encourage you [to extend the extension](https://tiptap.dev/docs/editor/guide/custom-extensions#extend-existing-extensions) & add that behavior yourself

Issue #4981, opened 2024-03-15
[Bug]:  Unexpected behaviour of configured extensions


yurtsiv:
### Which packages did you experience the bug in?

core

### What Tiptap version are you using?

2.2.4

### What’s the bug you are facing?

It applies to all methods in the extension config, but I'll give an example I stumbled upon with `addProseMirrorPlugins`

```typescript
const Base = Extension.create(...)

const pluginKey = new PluginKey("someKey");

const Example = Base.extend({
  addProseMirrorPlugins() {
    return [
      ...(this.parent?.() ?? []),
      new Plugin({ key: pluginKey }),
    ]
  },
})

new Editor({
   extensions: [
      Example.configure()
   ]
})
```

Crashes with ```RangeError: Adding different instances of a keyed plugin (someKey$)```.

It happens because [.configure() extends](https://github.com/ueberdosis/tiptap/commit/fda6310e7462a86724566d6630812c3ddb46a2d6#diff-5646f4773d47ddc5a0ff04bd8aa67213e7457fbac3b135e3237ffda72737b323L343), so we end up with the hierarchy `ExampleBase -> Example -> Example` instead of `ExampleBase -> Example` which results in duplicate plugins being registered.

The same problem occurs with other methods, but those that return lists are most vulnerable (e.g. `addInputRule`).

I guess  `.configure()` should copy the extension instead of extending it.

### What browser are you using?

Chrome

### What did you expect to happen?

No duplicate plugins are registered.

### Did you update your dependencies?

- [X] Yes, I’ve updated my dependencies to use the latest version of all packages.

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

yurtsiv:
Partially a duplicate of https://github.com/ueberdosis/tiptap/issues/4704

yurtsiv:
This commit https://github.com/ueberdosis/tiptap/commit/fda6310e7462a86724566d6630812c3ddb46a2d6 seems to be the reason why it broke. Before it `.extend()`  wasn't copying the config so the last extension in the hierarchy `ExampleBase -> Example -> Example` (described above) didn't have `addProseMirrorPlugins` defined

nperez0111:
This should be [fixed](https://github.com/ueberdosis/tiptap/pull/5147) with version 2.4.0, feel free to try it out!

Issue #4257, opened 2023-07-27
[Bug]: addProseMirrorPlugins fires twice at editor initialization


Nantris:
### Which packages did you experience the bug in?

extension-link

### What Tiptap version are you using?

2.1.0-rc.12

### What’s the bug you are facing?

I am reporting this specifically for `extension-link` but it may to be more wide ranging than that.

Note the plugin keys have `1` appended to them on the second run. I can't determine if this really means the plugin is loaded twice or if it matters in practice.

# `debugger` hit 1
![image](https://github.com/ueberdosis/tiptap/assets/6835891/0d27096c-9c82-40b9-9ddb-ee50c04a592d)

# `debugger` hit 2
![image](https://github.com/ueberdosis/tiptap/assets/6835891/47c733ff-5979-43e7-acb1-08b57cbae1cd)

### What browser are you using?

Chrome

### What did you expect to happen?

`addProseMirrorPlugins` should only run one time.

### Did you update your dependencies?

- [X] Yes, I’ve updated my dependencies to use the latest version of all packages.

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

Slapbox:
This seems to affect all nodes which you use the `.extend()` functionality on. When it runs a second time the keys change from like `codeBlockVSCodeHandler$` to `codeBlockVSCodeHandler$1`. I assume that means these are being loaded twice?

634750802:
Maybe related to https://github.com/ueberdosis/tiptap/discussions/5136

Nantris:
@634750802 that does seem related, but it should probably be filed as an issue, not a discussion. I guess the maintainers can probably change that though if they see fit.

nperez0111:
This should be [fixed](https://github.com/ueberdosis/tiptap/pull/5147) with version 2.4.0, feel free to try it out!

PR #5147, opened 2024-05-14
fix: configure should use the parent of the current instance, to avoid duplication

nperez0111:
## Changes Overview

Based on the approach taken by https://github.com/ueberdosis/tiptap/discussions/5136 and validated for appropriate behavior with a barrage of tests.

The gist of what is happening here is that `extend()` will always set the new child's `parent` property to be `this` (establishing a child -> parent relationship).

When doing `.configure()` it is [doing a `.extend()`](https://github.com/ueberdosis/tiptap/blob/8cbd8e237779e7b04c0af297159e1886e26da9f0/packages/core/src/Extension.ts#L458). Doing this directly on an extension (i.e. one that is not already `.extend`'d), as it will create a child -> parent relationship with pretty much the exact same config so it ends up overwriting itself with it's same options.
But what we really want is not a child -> parent relationship, but to just overwrite some options while maintaining an effective "copy" of the child (i.e. preserving it's child -> parent relationship).

This is pretty complicated, so I'm going to add a diagram explaining the current behavior and desired behavior here:

what it was

should be this

There was a lot of digging for this one:

It should resolve:
 - https://github.com/ueberdosis/tiptap/issues/4981
 - https://github.com/ueberdosis/tiptap/issues/4257

It may resolve:
 - https://github.com/ueberdosis/tiptap/issues/4704

## Implementation Approach
Someone had implemented this already: https://github.com/ueberdosis/tiptap/discussions/5136
I just validated that it worked and made sense with a barrage of tests. The root behavior I was trying to fix was https://github.com/ueberdosis/tiptap/issues/4704 

## Testing Done
Most of the code was adding tests to make sure that I understood what the problem was and if it fixed it.

## Verification Steps
Run the tests with the `Extension.ts` file as it was before and note the tests that it fails to pass now without it

## Checklist

### DIFF
--- a/packages/core/src/Extension.ts
+++ b/packages/core/src/Extension.ts
@@ -457,6 +457,7 @@ export class Extension<Options = any, Storage = any> {
     // with different calls of `configure`
     const extension = this.extend()
 
+    extension.parent = this.parent
     extension.options = mergeDeep(this.options as Record<string, any>, options) as Options
 
     extension.storage = callOrReturn(
--- a/tests/cypress/integration/core/extendExtensions.spec.ts
+++ b/tests/cypress/integration/core/extendExtensions.spec.ts
@@ -43,6 +43,30 @@ describe('extend extensions', () => {
     })
   })
 
+  it('should have a parent', () => {
+    const extension = Extension
+      .create({
+        addAttributes() {
+          return {
+            foo: {},
+          }
+        },
+      })
+
+    const newExtension = extension
+      .extend({
+        addAttributes() {
+          return {
+            bar: {},
+          }
+        },
+      })
+
+    const parent = newExtension.parent
+
+    expect(parent).to.eq(extension)
+  })
+
   it('should merge configs', () => {
     const extension = Extension
       .create({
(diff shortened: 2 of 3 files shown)


634750802:
I think it's same for Nodes and Marks?

nperez0111:
> I think it's same for Nodes and Marks?

Ah jeez, I missed that. I spent so much time trying to understand it I didn't realize it. Will follow up in another PR @634750802

tilgovi:
I think this has regressed on v3 with the move to `Extendable`, which re-uses the existing config during extension.

tilgovi:
My apologies, I believe the regression was not with the introduction of `Extendable`, but with this commit: 32958d6fa72a00ba657ecba0ac56de030107fa96

bdbch:
@tilgovi could you explain the regression a bit more so we can look into it? Not sure which issue you're referring to.

tilgovi:
I'm referring to this issue, which has regressed with Tiptap v3, due to the changes in the commit I mentioned above. If you `configure` a `Node` extension, for example, that has a function like `addProseMirrorPlugins` calls `this.parent()`, it results in the plugins being added twice.

bdbch:
Thanks, do you know if there's already an issue describing this problem somewhere?
If not - could you create one and link it here so I can track this?

smoores-dev:
@bdbch I believe the issue is here: https://github.com/ueberdosis/tiptap/issues/4257

Issue #5150, opened 2024-05-14
[Bug]: Content other than text can't be part of a simulated paste within an insertContent command


gabemagee-ev:
### Affected Packages

core

### Version(s)

2.3.2

### Bug Description

As of https://github.com/ueberdosis/tiptap/pull/5046, users can apply paste rules to the `insertContent` command, which is really helpful! Unfortunately, if the content the user inserts is an `Object` rather than a `string`, the paste rule fails because there is no `length` property of `Objects`.  Rather, this should probably check for `size` of an object.

### Browser Used

Firefox

### Expected Behavior

When `appendTransaction()` is called, it calls `processEvent()`, which within `run()` calls `state.doc.nodesBetween()`. In the case where non-string Content is passed to `insertContent`, no paste rules will be applied since `to` is `NaN` (see https://github.com/ueberdosis/tiptap/blob/ef7d195311746983bbe17b8f87c64a0be81b7ccd/packages/core/src/PasteRule.ts#L118). Thus, no paste rules will be applied. We want paste rules to be applied.

### Dependency Updates

- [X] Yes, I've updated all my dependencies.

PR #5152, opened 2024-05-14
fix: convert content to a string before passing it to `applyPasteRules` meta value in `insertContentAt.ts`

gabemagee-ev:
## Changes Overview

In `insertContentAt.ts` cast `content` as a string in `insertContentAt`, since `createNodeFromContent` can return an `Object` when the content is complex, which prevents paste and input rules from being applied.

## Implementation Approach

converting the `object` to a `string` prevents `text.length` from being `NaN` in `PasteRule.ts:appendTransaction`, which is necessary in order to process any paste or input events.

## Testing Done

`npm run test`

## Checklist

## Related Issues

https://github.com/ueberdosis/tiptap/issues/5150


packages/core/src/commands/insertContentAt.ts
```diff
@@ -127,11 +127,11 @@ export const insertContentAt: RawCommands['insertContentAt'] = (position, value,
     }
 
     if (options.applyInputRules) {
-      tr.setMeta('applyInputRules', { from, text: newContent })
+      tr.setMeta('applyInputRules', { from, text: newContent.toString() })
```
nperez0111, line 130: Hm, I see why you are doing this. But I think it is probably better to do it in the `else` branches on line :119 and :114 instead since those are the branches that does not attempt to convert it to a string (really 119 is the one that matters 114 should be a string)

nperez0111: Also tests would be greatly appreciated

gabemagee-ev: that what I originally had, but I opted for this since I think we want `newContent` to retain it's original assignment for the `tr.replaceWith(from, to, newContent)` in line :121.

gabemagee-ev: since this is my first contribution to this repo, I don't know where everything is. can you point me to where tests for a `command` or `pasteRule` would go, since I don't see any preexisting ones, or perhaps I am missing them. @nperez0111 thank you!

nperez0111: The tests would be in this file: https://github.com/ueberdosis/tiptap/blob/21aa96dee8deab1f439b7f655b8ed266a516a4cd/demos/src/Commands/InsertContent/React/index.spec.js#L1-L44

You can follow this guide to get your environment set up: https://github.com/ueberdosis/tiptap/blob/e597809c110011dd166e0cbb2488aee5a4434197/docs/overview/contributing.md#L22-L29


You are right about maintaining the call, I missed that.

I appreciate your contribution very much. If you need any help, I'm happy to 😄 !

gabemagee-ev:
after adding testing, I realized this wasn't a solution to the linked bug. Will work on a fix when I have time, likely next week. Sorry for the ping!

nperez0111:
> after adding testing, I realized this wasn't a solution to the linked bug. Will work on a fix when I have time, likely next week. Sorry for the ping!

Hey @gabemagee-ev, I actually ran into this myself sometime last week and I started working on something to fix it but did not get the time to finish it. Would be great if you could take it over the finish-line, here is my branch for it (pointed at the only additional commit I made branch name is `fix/simulated-paste-rules`): https://github.com/ueberdosis/tiptap/commit/fddfe3146fe2444884cfe7f135e423ac0a425ee8

The core of my idea here was to make the InputRules & PasteRules plugins do the work of accepting different content types and convert those to a format that can actually be used properly.

I think that my code works but I never got around to writing tests for it.

gabemagee-ev:
@nperez0111 hey, I'm looking back into this. have you done any further work other than the commit you linked? thanks

nperez0111:
> @nperez0111 hey, I'm looking back into this. have you done any further work other than the commit you linked? thanks

Nope just that branch (the one commit)

nperez0111:
I merged what I had

PR #4602, opened 2023-11-06
fix(extension-youtube) XSS risk with src tag

C-Hess:
## Please describe your changes

Fixes risks outlined in #4600 by verifying that any src urls are valid youtube URLs before rendering as HTML.

## How did you accomplish your changes

Added isValidUrl regex check in renderHTML method

## How have you tested your changes

Added unit tests to validate fix. Ran other extension-youtube cypress tests

## How can we verify your changes

Verify non-breaking on youtube extension, and ensure the new tests I've added passes

## Checklist

- [X] The changes are not breaking the editor
- [X] Added tests where possible
- [X] Followed the guidelines
- [X] Fixed linting issues

## Related issues

#4600

Issue #4600, opened 2023-11-06
[Bug]: Security vulnerability, XSS


ElizarBatin:
### Which packages did you experience the bug in?

extension-youtube

### What Tiptap version are you using?

2.1.12

### What’s the bug you are facing?

Vulnerability in `getEmbedUrlFromYoutubeUrl`. Attacker can send response to server to store youtube video with any url which contains `/embed/`. E.g. `javascript:alert(1)//embed/` (Executes arbitrary code in site origin).

### What browser are you using?

Chrome

### Code example

https://github.com/ElizarBatin/tiptap-poc

### What did you expect to happen?

1. Sandbox attributes for iframe must set for minimize possible iframe injection impact. 
2. Logic in `getEmbedUrlFromYoutubeUrl` function should fixed

### Did you update your dependencies?

- [X] Yes, I’ve updated my dependencies to use the latest version of all packages.

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

ElizarBatin:
I tried submit this vulnerability in email which mentions in README.md for contact, but i don't recieve answer :(

janthurau:
fixed by #4602

Issue #3673, opened 2023-01-30
Potential XSS vulnerability in link extension


fdoe:
### What’s the bug you are facing?

It is possible to execute JavaScript from the "href" attribute of a link. 

### Which browser was this experienced in? Are any special extensions installed?

Chrome 109.0.5414.119
Firefox 109.0
Safari 16.2
Edge 109.0.1518.70

### How can we reproduce the bug on our side?

1. Open code sandbox: https://codesandbox.io/s/tiptap-react-issue-template-nwvwck?file=/src/App.js
2. Replace content of App.js with:
 
```
import React from "react";
import { useEditor, EditorContent } from "@tiptap/react";
import {Link} from '@tiptap/extension-link';
import StarterKit from "@tiptap/starter-kit";

export default () => {
  const editor = useEditor({
    extensions: [
      StarterKit,
      Link.configure({
        validate: href => /^https?:\/\//.test(href),
      })
    ],
    content: {
      "type": "doc",
      "content": [
        {
          "type": "paragraph",
          "content": [
            {
              "type": "text",
              "marks": [
                {
                  "type": "link",
                  "attrs": {
                    "href": "javascript:alert(1)",
                    "target": "_blank",
                    "class": null
                  }
                }
              ],
              "text": "XSS attack link"
            }
          ]
        }
      ]
    }
  });

  return (
    <>
      <EditorContent editor={editor} />
    </>
  );
};
```

3. Click link displayed by the editor and wait for the JavaScript to open an alert popup

### Can you provide a CodeSandbox?

Yes. See "How can we reproduce the bug on our side?".

### What did you expect to happen?

Ideally, the editor would validate / sanitize the href attribute before rendering. 

### Anything to add? (optional)

Currently, the JSON content representation of the editor must be validated / sanitized separately in order to clear it from any malicious JavaScript. While this should be could practice anyways, the editor should perform such a sanitization by itself, too, as a safety measure. 

### Did you update your dependencies?

- [X] Yes, I’ve updated my dependencies to use the latest version of all packages.

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

bdbch:
I think this is related to #2169 which includes a more lengthy discussion about this topic. I'll keep this open for discussion and to keep track of what we want to do here.

fdoe:
Update: 

Apparently, TipTap treats JSON and HTML content differently:

[...]

PR #5153, opened 2024-05-15
fix(extension-link): make the javascript link detection case insensitive

julmud:
## Changes Overview
The fix for the XSS vulnerability in #4602 does not make the check on javascript URL in a case insentive way.

## Implementation Approach
Convert the href attribute to lowercase before testing if it starts with "javascript:"

## Testing Done
Increased the Cypress test coverage and ran it.

## Verification Steps
Verify non-breaking on link and youtube extension, and ensure the new tests pass.

## Checklist

## Related Issues
#3673 
#4600 
#4602
#5157

### DIFF
--- a/packages/extension-link/src/link.ts
+++ b/packages/extension-link/src/link.ts
@@ -161,9 +161,7 @@ export const Link = Mark.create<LinkOptions>({
   },
 
   renderHTML({ HTMLAttributes }) {
-    // False positive; we're explicitly checking for javascript: links to ignore them
-    // eslint-disable-next-line no-script-url
-    if (HTMLAttributes.href?.toLowerCase().trim().startsWith('javascript:')) {
+    if (HTMLAttributes.href?.substring(0, HTMLAttributes.href.indexOf(':')).toLowerCase().trim() === 'javascript') {
       // strip out the href
       return ['a', mergeAttributes(this.options.HTMLAttributes, { ...HTMLAttributes, href: '' }), 0]
     }
--- a/packages/extension-link/src/link.ts
+++ b/packages/extension-link/src/link.ts
@@ -163,7 +163,7 @@ export const Link = Mark.create<LinkOptions>({
   renderHTML({ HTMLAttributes }) {
     // False positive; we're explicitly checking for javascript: links to ignore them
     // eslint-disable-next-line no-script-url
-    if (HTMLAttributes.href?.toLowerCase().startsWith('javascript:')) {
+    if (HTMLAttributes.href?.toLowerCase().trim().startsWith('javascript:')) {
       // strip out the href
       return ['a', mergeAttributes(this.options.HTMLAttributes, { ...HTMLAttributes, href: '' }), 0]
     }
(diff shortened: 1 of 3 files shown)


packages/extension-link/src/link.ts
```diff
@@ -163,7 +163,7 @@ export const Link = Mark.create<LinkOptions>({
   renderHTML({ HTMLAttributes }) {
     // False positive; we're explicitly checking for javascript: links to ignore them
     // eslint-disable-next-line no-script-url
-    if (HTMLAttributes.href?.startsWith('javascript:')) {
+    if (HTMLAttributes.href?.toLowerCase().startsWith('javascript:')) {
```
nperez0111, line 166: oof good catch

nperez0111 (approved):
We should definitely release this

benja:
will this be released as soon as merged?

packages/extension-link/src/link.ts
nperez0111: I just tested this locally and it looks like you can fool it with leading white space too. I think that besides the protocol (e.g. `https://`) a colon is invalid so may this should be an `.includes` instead of a `.startsWith`

julmud: It gets tricky if the URL uses basic-auth and includes the username and password (i.e., `https://javascript:my_password@example.com/` for a user named javascript would be a false positive when using `.includes`). I think it'd be better to `.trim()` before calling `.startsWith()`

julmud: I think the correct way to do this would be to extract the scheme (everything up to the first colon), trim it, and then do a case insensitive match for `javascript`. I'll push something later today.

nperez0111 (approved):
lgtm

packages/extension-link/src/link.ts
chroth: Hello! Happy to see that you're looking into fixing this. ❤️ However I think there are a few issues with your approach here. 

- it would allow javascript protocol when it includes \x01-\x20 characters inside the `javascript:`
- it would allow javascript protocol when it includes whitespace characters before the colon `javascript\n:`.

It's really hard to get the blacklisting approach right. I made a PR where I try to use a whitelist approach instead, see here: https://github.com/ueberdosis/tiptap/pull/5160 It's still draft/WIP though.

I've taken inspiration from [DOMPurify](https://github.com/cure53/DOMPurify) (using it might in general be a good approach for tiptap).

I've also created test cases from PortSwiggers XSS cheat sheet: https://portswigger.net/web-security/cross-site-scripting/cheat-sheet#protocols

nperez0111: Thanks for chiming in @chroth we are looking into the right approach for this sensitive topic and your contribution made us re-think some things

There are three open PRs for XSS vuln fixes:
https://github.com/ueberdosis/tiptap/pull/5157
https://github.com/ueberdosis/tiptap/pull/5160
https://github.com/ueberdosis/tiptap/pull/5153

nperez0111:
We ended up resolving it with this https://github.com/ueberdosis/tiptap/pull/5160

But we appreciate your contribution, your work was helpful in the final fix.

We did this a bit strangely because of the multiple PRs and the urgency of it

Issue #4466, opened 2023-09-21
[Bug]: updateAttributes should only update attribute of current node, instead of all parent nodes with the same type


steven-qi:
### Which packages did you experience the bug in?

core

### What Tiptap version are you using?

2.1.8

### What’s the bug you are facing?

I'm trying to build a customized ordered list with different list item types (i.e., `a`, `A`, `i`, & `I`).

I was able to make it work by some trivial extending the default ordered list, which is great. 👍 

However, I found that when there are multiple levels of ordered lists, calling `commands.updateAttributes('orderedList', { type })` will always update the `type` attributes on both nested and parent levels.

From what I learned, this looks like a bug of `updateAttributes` itself, as it will find all nodes with the same type, and call `tr.setNodeMarkup` on each of them.

https://github.com/ueberdosis/tiptap/blob/develop/packages/core/src/commands/updateAttributes.ts#L48-L54

I was able to work around this by updating only the latest found node manually.

![image](https://github.com/ueberdosis/tiptap/assets/6123222/fe54d8d0-af0f-48d7-ae98-785a30020196)

### What browser are you using?

Chrome

### What did you expect to happen?

`updateAttributes` should only update attributes of the nearest node when it is wrapped into a node of the same type.

### Anything to add? (optional)

`resetAttributes` may have similar issue

### Did you update your dependencies?

- [X] Yes, I’ve updated my dependencies to use the latest version of all packages.

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

nperez0111:
Should be resolved with 2.10.0

Issue #3545, opened 2022-12-14
editor.updateAttributes() in nested ordered-list is also updating attributes of outer list.


ashu12chi:
### What’s the bug you are facing?

Hi,

I want to improve the [ordered-list plugin](https://tiptap.dev/api/nodes/ordered-list) to support different style-type for nested lists. I am doing this by adding `type` attribute on the ordered-list. I am doing it like this:

```javascript
addAttributes() {
    return {
     ...this.parent?.(),
      type: {
        default: '1',
        parseHTML: element => {
          return element.hasAttribute('type')
            ? element.getAttribute('type')
            : '1'
        },
      },
    }
  },
```
The issue I am facing is when we call the [sinkListItem command](https://tiptap.dev/api/commands/sink-list-item) to nest the ordered list, I need to update the `type` separately. I am doing this by calling the [updateAttributes() comand](https://tiptap.dev/api/commands/update-attributes) like this:

```javascript
const possibleTypes = ['1', 'A', 'a', 'I', 'i'];
const currentType = editor.getAttributes('orderedList').type;
const nextType = possibleTypes[(possibleTypes.findIndex(x => x === currentType) + 1)%5];
editor.chain().sinkListItem().updateAttributes('orderedList', {type: nextType}).run()
```
When we nest a list using `sinkListItem`, the editor schema will add `<ol><li><p></p></li></ol>`, which is a new ordered list and thus attribute of only this list should be updated but currently attribute of parent list is also getting updated due to which  the outcome looks like:
```
A. Hello
        A. 
```
But I am expecting it to be:
```
1. Hello
        A. 
```
I tried this same thing with the `start` attribute which is already defined in the ordered-list plugin, there also `attributes` of all parent lists are getting updated which should not happen ideally. This is shown in the following GIF, where I am setting `start` of child list equal to 5 times of parent list but `start` of parent list is also getting updated.
![Animation](https://user-images.githubusercontent.com/45908393/207613950-f990b9dd-3c12-4332-9f65-f1d30e363eaf.gif)

The alternative I considered here is to add `type` attribute on [list-item](https://tiptap.dev/api/nodes/list-item) which is not standard and thus I am not doing it. Another alternative is to use CSS to add [list-style-type](https://developer.mozilla.org/en-US/docs/Web/CSS/list-style-type), using something like this:

```css
/* -- Nested List  -- */

/* Set the first layer of list items to regular numeric decimals */
 ol {
    list-style-type: decimal;
}

[...]

PR #5154, opened 2024-05-15
updateAttributes: update current node only

silenius:
There is a bug (?) in updateAttributes, attributes of the node are updated but also all it's parent nodes of the same type. This is problematic with things like nested lists, but also nested paragraphs, etc

The fix updates attributes of the node ("last") only and leaves the parents intact

## Changes Overview
updateAttributes update node only (and not its parents)

## Implementation Approach
adapt updateAttributes so that it keep tracks of the "last" (current) node

## Testing Done
tested locally with bulletList nodes

## Verification Steps
try to update an attributes of a nested list for examplectionality of your changes. -->

## Checklist

## Related Issues
#3545 #4466

### DIFF
--- a/packages/core/src/commands/updateAttributes.ts
+++ b/packages/core/src/commands/updateAttributes.ts
@@ -51,37 +51,46 @@ export const updateAttributes: RawCommands['updateAttributes'] = (typeOrName, at
   }
 
   if (dispatch) {
+    let lastPos = null
+    let lastNode = null
+    let trimmedFrom = null
+    let trimmedTo = null
+
     tr.selection.ranges.forEach(range => {
       const from = range.$from.pos
       const to = range.$to.pos
 
       state.doc.nodesBetween(from, to, (node, pos) => {
         if (nodeType && nodeType === node.type) {
-          tr.setNodeMarkup(pos, undefined, {
-            ...node.attrs,
-            ...attributes,
-          })
-        }
-
-        if (markType && node.marks.length) {
-          node.marks.forEach(mark => {
-            if (markType === mark.type) {
-              const trimmedFrom = Math.max(pos, from)
-              const trimmedTo = Math.min(pos + node.nodeSize, to)
-
-              tr.addMark(
-                trimmedFrom,
-                trimmedTo,
-                markType.create({
-                  ...mark.attrs,
-                  ...attributes,
-                }),
-              )
-            }
-          })
+          trimmedFrom = Math.max(pos, from)
+          trimmedTo = Math.min(pos + node.nodeSize, to)
+          lastPos = pos
+          lastNode = node
         }
       })
     })
+
+    if (lastPos !== null && lastNode !== null) {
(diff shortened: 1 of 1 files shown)


nperez0111:
Appreciate your contribution @silenius but it looks like we need some tests resolved for this to be merged

silenius:
Sorry, I pressed the wrong button (:

nperez0111:
I don't think that is correct either. You were right to pull in the changes from main, it is your changes that are failing the build
@silenius

silenius:
I've fixed the lint issue, but I have no idea why it times out in one test..? (unfortunately Cypress is not supported on my OS, FreeBSD, so I can't check at the moment)

silenius:
I think I understand why it fails: the test in `demos/src/Extensions/TextAlign/React/` contains two `<p>` which were updated (as updateAttributes updated all nodes of specific type), but with this change only the _current_ (selected) node is updated. Could we simply remove a `<p>` from `demos/src/Extensions/TextAlign/React/index.jsx`?

nperez0111:
@silenius if you run the project locally `npm run dev` you'll see that when you go to: http://localhost:3000/src/Extensions/TextAlign/React/
that if you clear the editor completely, add some new text into the editor (which is what the test does on each run)
then select the content and click the `center` button that it does not work.

This is also seen on the netlify deployment: https://deploy-preview-5154--tiptap-embed.netlify.app/src/extensions/textalign/react/

So your change broke the ability to center text with the text align extension somehow


Hope this helps @silenius

silenius:
thanks, it should be fixed :)

packages/extension-code-block/src/code-block.ts
nperez0111, line 1: I think you still need to pull in the latest changes from main here, it is like you are trying to revert changes that have already been made and accepted

nperez0111 (changes requested):

packages/extension-code-block/src/code-block.ts
silenius: should be ok.. sorry about that

nperez0111 (approved):

nperez0111:
Appreciate your contribution here, it will go into the next release. Might take a while, so I will make a beta that includes it soon

svilen-ivanov-kubit:
Hi, I think this change caused the following problem: https://github.com/ueberdosis/tiptap/issues/5254

nperez0111:
> Hi, I think this change caused the following problem: #5254

Thanks for your feedback. I will write a test for this & see if I can resolve it or revert it

nperez0111:
Sorry @silenius I did not get around to handling this. If you have an idea on how to support this, that would very much be appreciated. I'm trying to get this new release out and so I had to revert it

silenius:
Hi @nperez0111,

Sorry for breakage, I'll try to find a little bit of time to review it.. but it could be reverted in the meantime

nperez0111:
Thanks for your contribution in the first place!

silenius:
FYI I've reworked this in #5738

PR #5157, opened 2024-05-15
Fix XSS potential

Nantris:
## Changes Overview
Fix XSS potential

## Implementation Approach
Trim and lowercase href before comparing. Rely on getAttrs instead of CSS selector.

## Testing Done
None

## Additional Notes
**Given that this potential was already overlooked, it's essential that these changes be tested and examined more vigorously by the core team than they were in the past.**

Sample case which previously allowed XSS (due to leading spaces and due to differences in capitalization): https://codesandbox.io/p/sandbox/sharp-hill-6k846v?file=%2Fsrc%2FApp.js%3A21%2C28

## Checklist
- [ ] I have renamed my PR according to the naming conventions. (e.g. `feat: Implement new feature` or `chore(deps): Update dependencies`)
- [ ] My changes do not break the library.
- [ ] I have added tests where applicable.
- [x] I have followed the project guidelines.
- [ ] I have fixed any lint issues.

## Related Issues
#3673

PR #5160, opened 2024-05-16
fix(extension-link): use whitelist for allowed href values

chroth:
## Changes Overview
Link href values needs to be sanitized to not result in XSS vulnerabilities.

## Implementation Approach
Link href values only allows whitelisted patterns for url, that are known to not be dangerous.

## Testing Done
The link extension spec has been updated to cover both positive and negative tests.

## Verification Steps
Link extension still functions as expected when testing with valid links, such as relative, absolute, with different schemes etc.

## Checklist

### DIFF
--- a/packages/extension-link/src/link.ts
+++ b/packages/extension-link/src/link.ts
@@ -166,10 +166,22 @@ export const Link = Mark.create<LinkOptions>({
   },
 
   parseHTML() {
-    return [{ tag: 'a[href]:not([href *= "javascript:" i])' }]
+    return [{
+      tag: 'a[href]',
+      getAttrs: dom => {
+        const href = (dom as HTMLElement).getAttribute('href')
+
+        // prevent XSS attacks
+        if (!href || !isAllowedUri(href)) {
+          return false
+        }
+        return { href }
+      },
+    }]
   },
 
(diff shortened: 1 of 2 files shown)


nperez0111:
Thanks for your contribution @chroth I've went ahead and added a couple more tests and merged some of the changes from https://github.com/ueberdosis/tiptap/pull/5157

Thank you as well @Nantris for your contribution here.

nperez0111 (approved):

bdbch (approved):

Nantris:
Thanks to all for their work on this. Based on the code I imagine that this properly prevents `   javascript:` hrefs. I'd submit a PR to add that test case but I'm not 100% clear that simply adding that line would work properly (eg after this line https://github.com/chroth/tiptap/blob/738c436a9ff4af39d1d6abd52208eb7d46616106/tests/cypress/integration/extensions/link.spec.ts#L186)

Is this backported to older versions?

Nantris:
Can anyone advise whether this will be backported? I understand the desire is probably not to continue maintaining older versions, but with this being a security issue and a number of outstanding issues blocking upgrade paths, I think it should be planned to be backported at least a few versions.

Issue #5001, opened 2024-03-21
[Bug]: useEditor initially returns null as of 2.2.2, breaking first-render usage


jtmclucas13:
### Which packages did you experience the bug in?

react

### What Tiptap version are you using?

2.2.4

### What’s the bug you are facing?

Immediately following the first render of a component using `useEditor`, I attempt to call `editor.commands.focus()` to give the user an autofocus UX. This worked in 2.1.12, but we recently upgraded to 2.2.4 where this autofocus behavior does not occur. 

I've traced the culprit to version [2.2.2 of @tiptap/react ](https://github.com/ueberdosis/tiptap/releases/tag/v2.2.2), and `useEditor` returning `null` for that first render. Is there a workaround for consumers who expect the prior behavior of an always-available editor?

### What browser are you using?

Chrome

### What did you expect to happen?

`useEditor` always returns an Editor instance.

### Anything to add? (optional)

It appears as though this solution might fix the problem: https://github.com/ueberdosis/tiptap/pull/4579.

### Did you update your dependencies?

- [X] Yes, I’ve updated my dependencies to use the latest version of all packages.

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

C-Hess:
Would be fixed by #4999 4579

hjoelh:
thanks for raising, thought i was going mad

hjoelh:
for anyone else seeing this, here is my work around https://github.com/ueberdosis/tiptap/pull/4579#issuecomment-2087132599

Aryan3212:
I think I've got a workaround for this. I needed to trigger a focus on a 2nd editor from my 1st editor. Successfully achieved this using `refs`.

We need to pass this ref into the innerRef prop of the EditorContent, then we can use that ref to check whether the editor exists and also do normal DOM manipulations with it.

```ts
export default function Editor({ onUpdate }: { onUpdate: Function }) {
  const customRef = useRef(null)
  const editor = useEditor({
    onUpdate: (e) => {
      onUpdate('body', e.editor.getJSON())
    },
    content: selectedNote?.body as Content
  })
  const titleEditor = useEditor({
    editorProps: {
      attributes: {
        class: 'h-[90%] prose lg:prose-lg xl:prose-2xl mx-auto focus:outline-none max-w-none',
      },
      handleKeyDown: (_, e) => {
        if (e.key === 'Enter') {
          console.log('focused', customRef.current.firstChild.focus())
[...]

PR #5161, opened 2024-05-16
fix: default to rendering the editor immediately, while staying backward compatible

nperez0111:
## Changes Overview
A third PR for performance of the useEditor hook :sweat_smile:.

What I've done here is combine a few of the different approaches and added a bit of my own spin to it:

Prior art:
 - https://github.com/ueberdosis/tiptap/pull/4579
 - https://github.com/ueberdosis/tiptap/pull/4858
 - https://github.com/ueberdosis/tiptap/pull/4453/files#diff-13acb5e551812e195c1140c10ba5ac298a0005996d07756cfd77c2d4194cb350

What I've done here is to take a spin on #4579 and base it on a `useSyncExternalStore` instead (for efficient re-rendering handled by react).
While adding the performance optimization made by #4453 (re-using existing editor instances, and just updating their options if they change).
It is backward-compatible to people who use SSR (printing a warning and telling them they should set the new flag explicitly for forward-compatible but only in dev mode).
It is forward-compatible, where if you don't specify the new flag, people will automatically have the editor on first render.
If the flag is specified, you get better TS types since it is now guaranteed that you get an editor instance.
If you include the `select` function, you can effectively choose when the component should re-render by returning something different than selected before (inspired by `useSelector` in react-redux)
If you include the `equalityFn` function, you can choose how to compare your `select` function's return value (again inspired by `useSelector` in react-redux)

## Implementation Approach

I've introduced 3 new options to the `useEditor` hook (and subsequently the editor provider).

 - `immediatelyRender` which is to control the new vs. the old behavior
 - `select` which is to give the consumer fine-grained control over rendering by selecting the values they care about in re-rendering the editor
 - `equalityFn` which defines how to compare the values that the consumer generates with the `select` function

When not specifying the `immediateRender` value, if it is detected to be in SSR mode (i.e. no window object) it will be assumed to be `false` to not break Next.js users (but printing a message that they should be explicit about this), otherwise it will be assumed to be `true` reducing flicker in existing applications.

Eventually, when we have confidence to do so (i.e. the next major version) we can throw an error instead and have Next.js users (and other ssr frameworks) have to explicitly opt out for their use case (or perhaps a new hook).

## Testing Done

I tested in the demos using all possible options and in a next application here is a table for reference:

| env | immediatelyRender | result | error message |
| --- | ----------------- | ------ | ------------- |
| vite | false | useEditor returns null on first render, an editor instance afterwards | none |
| vite | true  | useEditor returns an editor instance consistently | none |
| vite | *missing* AKA undefined | useEditor returns an editor instance consistently, since not in SSR mode | none |
| next | false | useEditor returns null on first render (client & server), an editor instance afterwards (client only, since server renders only once) | none |
| next | true  | **Error** next can't use DOM APIs on the server & will fail to rehydrate | [An error is thrown](https://github.com/ueberdosis/tiptap/blob/115998b13a9ad9315d9cccba097d0f457ee3a9d7/packages/react/src/useEditor.ts#L195) stating this is an invalid configuration in dev mode |
| next | *missing* AKA undefined | **Warning** in the console. useEditor returns null on first render (client & server), an editor instance afterwards (client only, since server renders only once) | [A warning is printed to the console](https://github.com/ueberdosis/tiptap/blob/115998b13a9ad9315d9cccba097d0f457ee3a9d7/packages/react/src/useEditor.ts#L180) stating that the use should explicitly set the flag in dev mode |

Testing this sort of thing in react can be a little complicated (since you need to control when react re-renders) but if someone knows of a way to test this I'm happy to add tests for it.

## Verification Steps

Run the demos and set different values for `immediatelyRender` and in a Next.js app.

## Additional Notes

I'm unsure what version tiptap claims to be compatible with, and I know older versions do not have `useSyncExternalStore` but I also know that `react-redux` uses a shim to make it compatible with older versions.

## Checklist

## Related Issues

https://github.com/ueberdosis/tiptap/issues/5001

### DIFF
--- a/packages/react/src/Context.tsx
+++ b/packages/react/src/Context.tsx
@@ -1,9 +1,8 @@
-import { EditorOptions } from '@tiptap/core'
 import React, { createContext, ReactNode, useContext } from 'react'
 
 import { Editor } from './Editor.js'
 import { EditorContent } from './EditorContent.js'
-import { useEditor } from './useEditor.js'
+import { useEditor, UseEditorOptions } from './useEditor.js'
 
 export type EditorContextValue = {
   editor: Editor | null;
@@ -15,17 +14,25 @@ export const EditorContext = createContext<EditorContextValue>({
 
 export const EditorConsumer = EditorContext.Consumer
 
+/**
+ * A hook to get the current editor instance.
+ */
 export const useCurrentEditor = () => useContext(EditorContext)
 
-export type EditorProviderProps = {
+export type EditorProviderProps<TSelectorResult> = {
   children?: ReactNode;
   slotBefore?: ReactNode;
   slotAfter?: ReactNode;
-} & Partial<EditorOptions>
+} & UseEditorOptions<TSelectorResult>
 
-export const EditorProvider = ({
+/**
+ * This is the provider component for the editor.
+ * It allows the editor to be accessible across the entire component tree
+ * with `useCurrentEditor`.
+ */
+export function EditorProvider<TSelectorResult>({
   children, slotAfter, slotBefore, ...editorOptions
-}: EditorProviderProps) => {
+}: EditorProviderProps<TSelectorResult>) {
   const editor = useEditor(editorOptions)
 
   if (!editor) {
--- a/packages/react/src/useEditor.ts
+++ b/packages/react/src/useEditor.ts
@@ -4,9 +4,143 @@ import {
   useEffect,
   useRef,
   useState,
+  useSyncExternalStore,
 } from 'react'
 
 import { Editor } from './Editor.js'
+
+const isDev = process.env.NODE_ENV !== 'production'
+const isSSR = typeof window === 'undefined'
+const isNext = isSSR || Boolean(typeof window !== 'undefined' && (window as any).next)
+
+/**
+ * The options for the `useEditor` hook.
+ */
+export type UseEditorOptions<TSelectorResult> = Partial<EditorOptions> & {
+  /**
+   * Whether to render the editor on the first render.
+   * If client-side rendering, set this to `true`.
+   * If server-side rendering, set this to `false`.
+   * @default true
+   */
+  immediatelyRender?: boolean;
+  /**
+   * A selector function to determine the value to compare for re-rendering.
+   * @default `({ transactionNumber }) => transactionNumber + 1`
+   */
+  selector?: (editor:Editor, options: {
+    /**
+     * The previous value returned by the selector.
+     */
+    previousValue: TSelectorResult | null;
+    /**
+     * The current transaction number. Incremented on every transaction.
+     */
+    transactionNumber: number
+  }) => TSelectorResult;
+  /**
+   * A custom equality function to determine if the editor should re-render.
+   * @default `(a, b) => a === b`
+   */
+  equalityFn?: (a: TSelectorResult, b: TSelectorResult | null) => boolean;
+}
+
+/**
+ * To synchronize the editor instance with the component state,
+ * we need to create a separate instance that is not affected by the component re-renders.
+ */
+function makeEditorInstance<TSelectorResult>({
+  immediatelyRender,
+  options: initialOptions,
+  selector = (_e, { transactionNumber }) => (transactionNumber + 1) as unknown as TSelectorResult,
+  equalityFn = (a: TSelectorResult, b: TSelectorResult | null) => a === b,
+}: Pick<UseEditorOptions<TSelectorResult>, 'selector' | 'equalityFn' | 'immediatelyRender'> & {
+  options: Partial<EditorOptions>;
+}) {
+
+  let editor: Editor | null = null
+  let transactionNumber = 0
+  let prevSnapshot: [Editor | null, TSelectorResult | null] = [editor, null]
+  const subscribers = new Set<() => void>()
+
+  const editorInstance = {
+    /**
+     * Get the current editor instance.
+     */
+    getSnapshot() {
+      if (!editor) {
+        return prevSnapshot
+      }
+
+      const nextSnapshotResult = selector(editor, { previousValue: prevSnapshot[1], transactionNumber })
+
+      if (equalityFn(nextSnapshotResult, prevSnapshot[1])) {
+        return prevSnapshot
+      }
+
+      const nextSnapshot: [Editor, TSelectorResult | null] = [editor, nextSnapshotResult]
+
+      prevSnapshot = nextSnapshot
+      return nextSnapshot
+    },
+    /**
+     * Always disable the editor on the server-side.
+     */
+    getServerSnapshot() {
+      return null
+    },
+    /**
+     * Subscribe to the editor instance's changes.
+     */
+    subscribe(callback: () => void) {
+      subscribers.add(callback)
+      return () => {
+        subscribers.delete(callback)
+      }
+    },
+    /**
+     * Create the editor instance.
+     */
+    create(options: Partial<EditorOptions>) {
+      if (editor) {
+        editor.destroy()
+      }
+      editor = new Editor(options)
+      subscribers.forEach(callback => callback())
+
+      if (editor) {
+      /**
+       * This will force a re-render when the editor state changes.
+       * This is to support things like `editor.can().toggleBold()` in components that `useEditor`.
+       * This could be more efficient, but it's a good trade-off for now.
+       */
+        editor.on('transaction', () => {
+          transactionNumber += 1
+          subscribers.forEach(callback => callback())
+        })
+      }
+    },
+    /**
+     * Destroy the editor instance.
+     */
+    destroy(): void {
+      if (editor) {
+        // We need to destroy the editor asynchronously to avoid memory leaks
+        // because the editor instance is still being used in the component.
+        const editorToDestroy = editor
+
+        setTimeout(() => editorToDestroy.destroy())
+      }
+      editor = null
+    },
+  }
+
+  if (immediatelyRender) {
+    editorInstance.create(initialOptions)
+  }
+
+  return editorInstance
+}
 
 /**
  * This hook allows you to create an editor instance.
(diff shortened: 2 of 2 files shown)


alii (approved):
LGTM, and would definitely be useful in our usage of tiptap. 

However, just out of curiosity, what might be a use-case of the selector? Was it added in this PR as a requirement you had, or just as a potentially-nice-to-have?

nperez0111:
> LGTM, and would definitely be useful in our usage of tiptap.
> 
> However, just out of curiosity, what might be a use-case of the selector? Was it added in this PR as a requirement you had, or just as a potentially-nice-to-have?

The use case of a selector would be two-fold. For preventing re-renders by allowing you to only re-render when that selected state changes (e.g. not on every transaction like it is by default), and to give a more convenient API for tiptap state.

I haven't updated the PR yet, but I envision an additional API like:

```ts
function MyComponent() {
  const { editor, state } = useEditorWithState({
    selector: (currentEditor) => {
      return { isBold: !currentEditor.can().setBold() };
    },
	equalityFn: deepEqual
  });

  return (
    <>
      <Toolbar>
        <button>{state.isBold ? "Bold" : "Not Bold"}</button>
      </Toolbar>
      <Editor editor={editor} />
    </>
  );
}
```

With the code above, the component only has to re-render when the editor switches from being able to bold the current selection and not being able to. Which drastically will reduce re-renders.
It also gives you a standard way to select state from the editor instead of relying on the instance at runtime which has been confusing to some users.

alii:
Understood, that is super! I think it would make sense to have a selector API on the `useCurrentEditor` hook as well, or another hook (perhaps "`useEditorState`" that would accept the editor and allow you to select state from it. e.g.

```tsx
function MyBubbleMenu() {
  const isBold = useCurrentEditor({
    selector: editor => !editor.can().setBold(),
  });
}

<EditorProvider>
  <MyBubbleMenu />
</EditorProvider>
```

Or, with a "useEditorState" hook

```tsx
function MyBubbleMenu({editor}: {editor: Editor}) {
  const isBold = useEditorState(editor, {
    selector: editor => !editor.can().setBold(),
  });
}

function App() {
  const editor = useEditor();
  
  return (
    <>
      <MyBubbleMenu editor={editor} />
    </>
  );
}
```

nperez0111:
> Understood, that is super! I think it would make sense to have a selector API on the `useCurrentEditor` hook as well, or another hook (perhaps "`useEditorState`" that would accept the editor and allow you to select state from it. e.g.
> 
> ```tsx
> function MyBubbleMenu() {
>   const isBold = useCurrentEditor({
>     selector: editor => !editor.can().setBold(),
>   });
> }
> 
> <EditorProvider>
>   <MyBubbleMenu />
> </EditorProvider>
> ```
> 
> Or, with a "useEditorState" hook
> 
> ```tsx
> function MyBubbleMenu({editor}: {editor: Editor}) {
>   const isBold = useEditorState(editor, {
>     selector: editor => !editor.can().setBold(),
>   });
> }
> 
> function App() {
>   const editor = useEditor();
>   
>   return (
>     <>
>       <MyBubbleMenu editor={editor} />
>       <EditorContent editor={editor} />
>     </>
>   );
> }
> ```

Yep, will be adding something like that to it. Just haven't gotten around to updating this PR yet

nperez0111:
This example proves the difference between the default behavior & the new behavior enabled by the new `shouldRerenderOnTransaction` field:
https://deploy-preview-5161--tiptap-embed.netlify.app/preview/Examples/Performance

bdbch (approved):

alii:
🎉Amazing work, thank you!!!🎉

alii:
Weirdly it seems that the overloads for `useEditorState` break in the published built declarations.

It seems to be fixed if the overload with `Editor | null` goes beneath, rather than above. e.g.:

CleanShot 2024-07-10 at 12 15 24@2x
(in the published version right now, the `| null` is above and so I assume typescript matches that first)

Otherwise the result from `useEditorState` is typed as `T | null`, even if the editor is always provided (which in my case, it is).

nperez0111:
Hm, I ran into this myself. I think you might need `useEditor({...options} as const)` because typescript by default interprets true to be a boolean. But I'm uncertain on that.

@alii

packages/react/src/useEditorState.ts
```diff
@@ -0,0 +1,125 @@
...
+
+  return editorInstance
+}
+
+export function useEditorState<TSelectorResult>(
+  options: UseEditorStateOptions<TSelectorResult, Editor | null>
+): TSelectorResult | null;
+export function useEditorState<TSelectorResult>(
+  options: UseEditorStateOptions<TSelectorResult, Editor>
+): TSelectorResult;
```
alii, line 102: Do you mean `useEditor` or `useEditorState`?

Because if I go into node_modules and just flip the order of these overloads (in the `.d.ts` file of `useEditorState`) it seems to be working all fine. I think this is because TypeScript will resolve the `Editor` overload first, and if it's null it will resolve the `Editor | null` second. But since `Editor` DOES extend `Editor | null`, it resolved the nullable one first. Sorry that's a bit of a mouthful to read😆

I'm wondering if just flipping the order of the overloads in the source code will fix it? Would it be possible to make another changeset & pre-release to test?

Happy to make a PR as well if I haven't explained it very well 😄

nperez0111: Oh I was talking useEditor, Hm interesting. I would happily take a PR for it!

alii: Gotcha. will make a PR

Fwiw, typescript IS able to infer the boolean to `true` exactly, but only because it was explicitly written out in the overloads. [Here's an example of that behaviour ](https://www.typescriptlang.org/play/?#code/JYOwLgpgTgZghgYwgAgKIBNhgPZWQbwF8BuAKFIHoLkBBEZCADzgFsAHAGxWxmRG2TYAbtA7Y46AM7I4YGRw4AaPtjlxkAcygRZyACIANUuggIOcbchgBXEAjDBs9a5IgYsuAOpYAFgDlsAHkRKDEJAApsNgcnSQB+AC4CYBYWCExZCA4ATwAlCBATKETkACNsbC44EEIASiT3HDwAHz5rBTJKam8wH0EQsKllOGkTFliwKEz0ZFBkB2i4NmNTc0sbOxjnV0avX2DRcXRI6McQSST8FLSMyBz8wugkyesIOobMJrITMwsUDfsZ2QLjcnz2vQOoSOJy28Uu13SwEy9wKRRK8A4rneaDBLTaHXIVDQzHYXEkXSolKp1JptLp9IZjKZ1PICAmyAA7r4ApDBjRkABeYE7XE9fxBAbQ2pkZCyokAPTirPZXIhkok-KFIN2UDFvKlMrl1EVyvOclVPn1EgAQoLhaCPLr9urjldUojkXlUU8rHBMW9paRZcgFUrTZJzc7DhIAMJ27WiqNQiJum5Iu5ex5QZ5QV51Q0h41K4NEgAqPmA0icKDZUG09hysxAMGg0hwZRQACsXHIdUA)

Nantris:
Thank you for the tremendous work!

I wonder if anyone can provide insight into possible circumstances where `immediatelyRender` is not compatible out of the box? Unfortunately it causes nothing to render for us. No errors either. It's too late for me to start tonight, but I'm not sure where I'd even begin to look.

nperez0111:
> Thank you for the tremendous work!
> 
> I wonder if anyone can provide insight into possible circumstances where `immediatelyRender` is not compatible out of the box? Unfortunately it causes nothing to render for us. No errors either. It's too late for me to start tonight, but I'm not sure where I'd even begin to look.

@Nantris Hm,

It causes nothing to render? That is strange, I mean I thought I tried all of the cases. If doing SSR it must be false, if doing CSR it can be true (for better typing). `undefined` will try to guess if you are doing CSR or SSR, better to be explicitly, this behavior was for backwards-compatibility.

Does the editor instance not get returned back to you? Are you on an older React version maybe?

andrictham:
I’m using Next.js. Which version will this be rolling out in, and will there docs for this change?

nperez0111:
Currently released in the latest beta version (which you can try this out with)

Definitely will be writing docs for it, just isn't on the main release yet so haven't gotten around to it.

We've been working on our release structure (moving to changesets and CI pipelines) so we just haven't been confident to do a stable release just yet. Would very interested here how it is working for you though.

For Next.js would amount to `immediatelyRender: false` and separately if you want more control over your re-rendering behavior:
 `shouldRerenderOnTransaction: false` and use the new `useEditorState` hook instead.

alii:
Chiming in as I'm also using Next.js, I've split the entire editor into a separate component, and then imported it dynamically, like this:

```tsx
// components/editor.tsx
import {useEditor} from '@tiptap/react';

export function Editor() {
  const editor = useEditor({immediatelyRender: true, shouldRerenderOnTransaction: false});

  // ...
}
```

```tsx
// pages/editor.tsx
import dynamic from 'next/dynamic';

const Editor = dynamic(
  () => import('../components/editor.tsx').then(mod => mod.Editor),
  {ssr: false},
);

<Editor />
```

I've had this setup from _before_ the changes in this PR, however there were no issues once I upgraded.

Worth noting that this, of course, changes when/where the code for the Editor is bundled and this may not be your desired behaviour. Though one advantage is that the editor only mounts in the browser anyway, and now you don't ever have to deal with null checks as the editor will always exist by the time you're rendering.

szymonchudy:
> > Thank you for the tremendous work!
> > I wonder if anyone can provide insight into possible circumstances where `immediatelyRender` is not compatible out of the box? Unfortunately it causes nothing to render for us. No errors either. It's too late for me to start tonight, but I'm not sure where I'd even begin to look.
> 
> @Nantris Hm,
> 
> It causes nothing to render? That is strange, I mean I thought I tried all of the cases. If doing SSR it must be false, if doing CSR it can be true (for better typing). `undefined` will try to guess if you are doing CSR or SSR, better to be explicitly, this behavior was for backwards-compatibility.
> 
> Does the editor instance not get returned back to you? Are you on an older React version maybe?

Hey @nperez0111, thanks for taking the time to prepare that improvement! 

We encountered a similar issue when upgrading from 2.3.1. It seems that the `content` we provide is not being rendered. Instead, there is only an empty `div` in TipTap. Our structure is very basic, with `text` nodes nested under the `doc`. The instance seems to be returned correctly. The problem occurs regardless of whether we set the `immediatelyRender` flag to true or not. We are using React 18.3.1. Let me know if you need any other details.

Nantris:
Thanks for your reply @nperez0111!

After having some time to look into this, I think the issue must be internal to TipTap somehow.

It's quite curious. I see you tested in Vite and that's where I do my development. There's no SSR involed. The `useEditor` hook does return the editor, but it seems that the DOM element is never attached to the page.

When I enter `editor.view.dom` in console and press enter it returns this DOM node:

```html
<div contenteditable=​"true" translate=​"no" class=​"tiptap ProseMirror" tabindex=​"0">​…​</div>​
```

But curiously, when I right-click it and "Reveal in Elements Panel" it tells me, `Node cannot be found in the current page.` - and indeed it is not part of the page's DOM. The detached DOM node does contains the expected content though.

React is on `18.3.1` and I ran `yarn upgrade` after installing the latest version just to be sure nothing was inadvertently out of date.

I wonder if you have any theory what the issue could be? Thank you again for the stellar work!

nperez0111:
Hm, nothing off the top of my head should have been different, could you double check with the previous beta (maybe something changed in another beta?) So maybe check @tiptap/react@2.5.0-pre.13?

Other than that, I've heard from others that it is working properly so I am unsure why it'd be different, could you share your code? If not maybe a reproducible example? Maybe just delete features until you have it working?

szymonchudy:
> > Thank you for the tremendous work!
> > I wonder if anyone can provide insight into possible circumstances where `immediatelyRender` is not compatible out of the box? Unfortunately it causes nothing to render for us. No errors either. It's too late for me to start tonight, but I'm not sure where I'd even begin to look.
> 
> @Nantris Hm,
> 
> It causes nothing to render? That is strange, I mean I thought I tried all of the cases. If doing SSR it must be false, if doing CSR it can be true (for better typing). `undefined` will try to guess if you are doing CSR or SSR, better to be explicitly, this behavior was for backwards-compatibility.
> 
> Does the editor instance not get returned back to you? Are you on an older React version maybe?

Hey! First, a big thank you for taking the time to prepare this PR! 

We observed similar behavior to that described by @Nantris while working in Strict Mode. Essentially, the editor displays an empty div. When Strict Mode is turned off, everything functions as expected.

nperez0111:
> > > Thank you for the tremendous work!
> > > I wonder if anyone can provide insight into possible circumstances where `immediatelyRender` is not compatible out of the box? Unfortunately it causes nothing to render for us. No errors either. It's too late for me to start tonight, but I'm not sure where I'd even begin to look.
> > 
> > 
> > @Nantris Hm,
> > It causes nothing to render? That is strange, I mean I thought I tried all of the cases. If doing SSR it must be false, if doing CSR it can be true (for better typing). `undefined` will try to guess if you are doing CSR or SSR, better to be explicitly, this behavior was for backwards-compatibility.
> > Does the editor instance not get returned back to you? Are you on an older React version maybe?
> 
> Hey! First, a big thank you for taking the time to prepare this PR, @nperez0111!
> 
> We observed similar behavior to that described by @Nantris while working in Strict Mode. Essentially, the editor displays an empty `div`. When Strict Mode is turned off, everything functions as expected.

Great hint, I put it in StrictMode and it is not working for me either. I will look into what is happening

Nantris:
I can confirm our issue is Strict Mode related as well. Great find!

nperez0111:
Great, I think I have a fix with: https://github.com/ueberdosis/tiptap/pull/5338

PR #5164, opened 2024-05-17
fix: revert `font-family` escaping introduced by #4545

nperez0111:
## Changes Overview

Using `CSS.escape` is the wrong tool for the job here:
 - it is meant for CSS selectors and does not handle CSS variables properly.
 - you can't use `var(--title)` as a font-family because it was getting escaped to `var\(--title\)`

Instead, this reverts back to the previous behavior and forces the caller to specify an actual valid CSS font-family value.

## Implementation Approach

I was notified that CSS variables don't work as the name.
CSS.escape did not look the right tool for the job.

I considered trying to quote the font-family using [this stack overflow mentioned how to do it](https://stackoverflow.com/questions/7638775/do-i-need-to-wrap-quotes-around-font-family-names-in-css), but I ended up going back on that decision because it is just over complicated and it really should be the caller's responsibility to input valid font-family values. But you can see my previous working example in [my previous commit](https://github.com/ueberdosis/tiptap/pull/5164/commits/b0e9336012fea137f01d3a34be8b5d4ad0bd2852)

## Testing Done

Added tests for font-family setting and new button for CSS variable

## Verification Steps
Try it with: http://localhost:3000/preview/Extensions/FontFamily

## Checklist

## Related Issues

PR: #4545
The original issue could have been avoided if they quoted properly

### DIFF
--- a/demos/src/Extensions/FontFamily/React/index.jsx
+++ b/demos/src/Extensions/FontFamily/React/index.jsx
@@ -73,6 +73,13 @@ export default () => {
       >
         CSS variable
       </button>
+      <button
+        onClick={() => editor.chain().focus().setFontFamily('"Comic Sans MS", "Comic Sans"').run()}
+        className={editor.isActive('textStyle', { fontFamily: '"Comic Sans"' }) ? 'is-active' : ''}
+        data-test-id="comic-sans-quoted"
+      >
+        Comic Sans quoted
+      </button>
       <button onClick={() => editor.chain().focus().unsetFontFamily().run()} data-test-id="unsetFontFamily">
         unsetFontFamily
       </button>
--- a/demos/src/Extensions/FontFamily/React/index.jsx
+++ b/demos/src/Extensions/FontFamily/React/index.jsx
@@ -1,3 +1,5 @@
+import './styles.scss'
+
 import Document from '@tiptap/extension-document'
 import FontFamily from '@tiptap/extension-font-family'
 import Paragraph from '@tiptap/extension-paragraph'
@@ -15,6 +17,7 @@ export default () => {
         <p><span style="font-family: serif">Serious people use serif fonts anyway.</span></p>
         <p><span style="font-family: monospace">The cool kids can apply monospace fonts aswell.</span></p>
         <p><span style="font-family: cursive">But hopefully we all can agree, that cursive fonts are the best.</span></p>
+        <p><span style="font-family: var(--title-font-family)">Then there are CSS variables, the new hotness.</span></p>
       `,
   })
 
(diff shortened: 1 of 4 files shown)


jonahallibone:
Hi @nperez0111, reporter for the bug here. 

I understand not wanting to just do a true revert, but IMO TipTap should not be responsible for user error. I don't think anybody would assume that TipTap is modifying their input when it get's attached to the DOM. In fact, I'd wager most people assume what they put in is what they'll get out. 

A true revert in this case should be a patch, because it's fixing a commit which broke previous behavior and introduced a bug (if you can classify the existing behavior as a bug, which I would). The breaking commit did not patch a bug rather it papered over a misunderstanding of the browser style API. I don't think there's any reason to continue to support that, and instead it should be encourage that people interact with the browser correctly and fix their mistakes instead of assuming a library will.

nperez0111:
> Hi @nperez0111, reporter for the bug here.
> 
> I understand not wanting to just do a true revert, but IMO TipTap should not be responsible for user error. I don't think anybody would assume that TipTap is modifying their input when it get's attached to the DOM. In fact, I'd wager most people assume what they put in is what they'll get out.
> 
> A true revert in this case should be a patch, because it's fixing a commit which broke previous behavior and introduced a bug (if you can classify the existing behavior as a bug, which I would). The breaking commit did not patch a bug rather it papered over a misunderstanding of the browser style API. I don't think there's any reason to continue to support that, and instead it should be encourage that people interact with the browser correctly and fix their mistakes instead of assuming a library will.

I totally agree with you, I was mulling it over after submitting this. This is just a weird edge case that I was trying to err towards backwards-compat. But honestly it will probably change content that shouldn't and that might be more breaking than this

jonahallibone:
> > Hi @nperez0111, reporter for the bug here.
> 
> > 
> 
> > I understand not wanting to just do a true revert, but IMO TipTap should not be responsible for user error. I don't think anybody would assume that TipTap is modifying their input when it get's attached to the DOM. In fact, I'd wager most people assume what they put in is what they'll get out.
> 
> > 
> 
> > A true revert in this case should be a patch, because it's fixing a commit which broke previous behavior and introduced a bug (if you can classify the existing behavior as a bug, which I would). The breaking commit did not patch a bug rather it papered over a misunderstanding of the browser style API. I don't think there's any reason to continue to support that, and instead it should be encourage that people interact with the browser correctly and fix their mistakes instead of assuming a library will.
> 
> 
> 
> I totally agree with you, I was mulling it over after submitting this. This is just a weird edge case that I was trying to err towards backwards-compat. But honestly it will probably change content that shouldn't and that might be more breaking than this

Yea, I find trying to maintain these kinds backwards compatibility patches can be sort of a time suck / painful as more edge cases start to be reported and you're expected to maintain it.

nperez0111:
@jonahallibone Unfortunately not, my colleague is going on vacation so likely won't see this for another couple weeks

Issue #5010, opened 2024-03-25
[Bug]: paste does not respect extension-text-align settings


volarname:
### Which packages did you experience the bug in?

core, extension-text-align

### What Tiptap version are you using?

2.2.4

### What’s the bug you are facing?

when you configure extension-text-align like this
```js
TextAlign.configure({
      alignments: ['left', 'center', 'right'],
      defaultAlignment: 'left',
}),
```
and you copy/paste justified text from html/word etc
it will be pasted as justified text in the document schema instead of default (left)
```json
"attrs": {
        "textAlign": "justify"
},
```

### What browser are you using?

Chrome

### What did you expect to happen?

copy/paste should respect extension settings and should set textAlign to default value instead of unsupported value

### Did you update your dependencies?

- [X] Yes, I’ve updated my dependencies to use the latest version of all packages.

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

PR #5169, opened 2024-05-20
fix: parse only allowed text align alignments

volarname:
## Changes Overview
TextAlign extension should respects alignments config options, so when you for example copy paste alignment that is not configured, it will insert it as default.

## Implementation Approach
I am just checking if provided value used in parseHTML is from allowed/configured list.

## Testing Done
I have not added test, because this is really a small change/fix.

## Verification Steps
Try to configure text aligns that exclude justify, copy/paste text from website where is justified. On paste it should be left, because justify is not supported by provided config.

## Checklist

## Related Issues
(https://github.com/ueberdosis/tiptap/issues/5010)

### DIFF
--- a/packages/extension-text-align/src/text-align.ts
+++ b/packages/extension-text-align/src/text-align.ts
@@ -63,11 +63,12 @@ export const TextAlign = Extension.create<TextAlignOptions>({
         attributes: {
           textAlign: {
             default: this.options.defaultAlignment,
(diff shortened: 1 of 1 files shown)


volarname:
@bdbch @svenadlung can someone please look at this, its really simple fix

nperez0111:
@volarname Appreciate the contribution, agreed this should be default behavior

PR #5178, opened 2024-05-22
feat: error handling of invalid content for a schema

nperez0111:
## Changes Overview

This PR introduces a basic primitive for handling schema errors. Giving the user a callback to know when the schema was given content that it cannot parse and should probably be read only.

## Implementation Approach

This change introduces two new top-level options to the editor: `enableContentCheck` & `onContentError` for dealing with content supplied that does not match the prosemirror schema generated by the set of tiptap extensions.

`enableContentCheck` allows the app developer to opt into the behavior to check for invalid schemas (this change is otherwise backwards compatible).
When true, this will try to parse the document, and any content that does not match the schema will emit a `contentError` which can be listened to via the `onContentError` callback.

## Testing Done

Added new tests for this, most tests were tested manually

## Checklist

## Related Issues

https://github.com/ueberdosis/tiptap/pull/4641

### DIFF
--- a/packages/core/src/Editor.ts
+++ b/packages/core/src/Editor.ts
@@ -294,6 +294,10 @@ export class Editor extends EventEmitter<EditorEvents> {
         { errorOnInvalidContent: this.options.enableContentCheck },
       )
     } catch (e) {
+      if (!(e instanceof Error) || ['[tiptap error]: Invalid JSON content', '[tiptap error]: Invalid HTML content'].includes(e.message)) {
+        // Not the content error we were expecting
+        throw e
+      }
       this.emit('contentError', {
         editor: this,
         error: e as Error,
--- a/packages/core/src/Editor.ts
+++ b/packages/core/src/Editor.ts
@@ -295,7 +295,20 @@ export class Editor extends EventEmitter<EditorEvents> {
       )
     } catch (e) {
       this.emit('contentError', { editor: this, error: e as Error })
-      return
+
+      // Remove the collaboration extension if the content is invalid, to not sync invalid content
+      this.options.extensions = this.options.extensions.filter(extension => extension.name !== 'collaboration')
+
+      // Recreate the extension manager without the invalid extensions
+      this.createExtensionManager()
+
+      // Content is invalid, but attempt to create it anyway, stripping out the invalid parts
+      doc = createDocument(
+        this.options.content,
+        this.schema,
+        this.options.parseOptions,
+        { errorOnInvalidContent: false },
+      )
     }
     const selection = resolveFocusPosition(doc, this.options.autofocus)
 
--- a/packages/core/src/Editor.ts
+++ b/packages/core/src/Editor.ts
@@ -294,14 +294,17 @@ export class Editor extends EventEmitter<EditorEvents> {
         { errorOnInvalidContent: this.options.enableContentCheck },
       )
     } catch (e) {
+      this.emit('contentError', {
+        editor: this,
+        error: e as Error,
+        disableCollaboration: () => {
+          // To avoid syncing back invalid content, reinitialize the extensions without the collaboration extension
+          this.options.extensions = this.options.extensions.filter(extension => extension.name !== 'collaboration')
 
-      // Remove the collaboration extension if the content is invalid, to not sync invalid content
-      this.options.extensions = this.options.extensions.filter(extension => extension.name !== 'collaboration')
-
-      // Recreate the extension manager without the invalid extensions
-      this.createExtensionManager()
-
-      this.emit('contentError', { editor: this, error: e as Error })
+          // Restart the initialization process by recreating the extension manager with the new set of extensions
+          this.createExtensionManager()
+        },
+      })
 
       // Content is invalid, but attempt to create it anyway, stripping out the invalid parts
       doc = createDocument(
--- a/packages/core/src/Editor.ts
+++ b/packages/core/src/Editor.ts
@@ -294,7 +294,7 @@ export class Editor extends EventEmitter<EditorEvents> {
         { errorOnInvalidContent: this.options.enableContentCheck },
       )
     } catch (e) {
-      if (!(e instanceof Error) || ['[tiptap error]: Invalid JSON content', '[tiptap error]: Invalid HTML content'].includes(e.message)) {
+      if (!(e instanceof Error) || !['[tiptap error]: Invalid JSON content', '[tiptap error]: Invalid HTML content'].includes(e.message)) {
         // Not the content error we were expecting
         throw e
       }
--- a/packages/core/src/Editor.ts
+++ b/packages/core/src/Editor.ts
@@ -1,5 +1,8 @@
 import {
-  MarkType, NodeType, Schema,
+  MarkType,
+  Node as ProseMirrorNode,
+  NodeType,
+  Schema,
 } from '@tiptap/pm/model'
 import {
   EditorState, Plugin, PluginKey, Transaction,
@@ -36,8 +39,10 @@ import { isFunction } from './utilities/isFunction.js'
 
 export * as extensions from './extensions/index.js'
 
-export interface HTMLElement {
-  editor?: Editor
+declare global {
+  interface HTMLElement {
+    editor?: Editor;
+  }
 }
 
 export class Editor extends EventEmitter<EditorEvents> {
@@ -69,6 +74,7 @@ export class Editor extends EventEmitter<EditorEvents> {
     enableInputRules: true,
     enablePasteRules: true,
     enableCoreExtensions: true,
+    enableContentCheck: false,
     onBeforeCreate: () => null,
     onCreate: () => null,
     onUpdate: () => null,
@@ -77,6 +83,7 @@ export class Editor extends EventEmitter<EditorEvents> {
     onFocus: () => null,
     onBlur: () => null,
     onDestroy: () => null,
+    onContentError: ({ error }) => { throw error },
   }
 
   constructor(options: Partial<EditorOptions> = {}) {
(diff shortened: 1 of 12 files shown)


packages/core/src/Editor.ts
```diff
@@ -276,7 +284,19 @@ export class Editor extends EventEmitter<EditorEvents> {
...
+    try {
+      doc = createDocument(
+        this.options.content,
+        this.schema,
+        this.options.parseOptions,
+        { errorOnInvalidContent: this.options.enableContentCheck },
+      )
+    } catch (e) {
+      this.emit('contentError', { editor: this, error: e as Error })
+      return
```
nperez0111, line 298: Unsure what to do with the editor when this happens? Should we keep trying to show it?

packages/core/src/commands/insertContentAt.ts
```diff
@@ -57,12 +70,21 @@ export const insertContentAt: RawCommands['insertContentAt'] = (position, value,
...
+        parseOptions: {
+          preserveWhitespace: 'full',
+          ...options.parseOptions,
+        },
+        errorOnInvalidContent: options.errorOnInvalidContent ?? editor.options.enableContentCheck,
+      })
+    } catch (e) {
+      // Only ever throws if enableContentCheck or errorOnInvalidContent is true
+      editor.emit('contentError', { editor, error: e as Error })
+      return false
```
nperez0111, line 86: Unsure here, if they supply invalid content what is expected in this command? Just to return false or to also call the onContentError callback? Maybe that only happens if `enabledContentCheck = true`

packages/core/src/helpers/createNodeFromContent.ts
```diff
@@ -46,18 +47,55 @@ export function createNodeFromContent(
...
+                getAttrs: () => {
+                  hasInvalidContent = true
+                  return null
+                },
+              },
+            ],
+          },
+        }),
+      })
+    }
```
nperez0111, line 86: Ended up finding the simplest way to express the catch-all node using prosemirror constructs

packages/core/src/types.ts
```diff
@@ -67,8 +68,20 @@ export interface EditorOptions {
   enableInputRules: EnableRules
   enablePasteRules: EnableRules
   enableCoreExtensions: boolean
+  /**
+   * If `true`, the editor will check the content for errors on initialization.
+   * Emitting the `contentError` event if the content is invalid.
+   * Which can be used to show a warning or error message to the user.
+   * @default false
+   */
+  enableContentCheck: boolean
```
nperez0111, line 77: I went with the `enable` prefix for context

demos/src/Examples/Images/React/index.jsx
```diff
@@ -19,8 +19,8 @@ export default () => {
     ],
     content: `
       <p>This is a basic example of implementing images. Drag to re-order.</p>
-      <img src="https://source.unsplash.com/8xznAGy4HcY/800x400" />
-      <img src="https://source.unsplash.com/K9QHL52rE2k/800x400" />
+      <img src="https://placehold.co/600x400" />
+      <img src="https://placehold.co/800x400" />
```
nperez0111, line 23: tests were failing because these images are no longer accessible

svenadlung: thx, fixed on main

nperez0111: rebased it out 👍

packages/core/src/Editor.ts
bdbch: Mhh. I think we should try to show it? Otherwise older documents may lock an editor when old unparseable nodes are loaded? Maybe we could just disable collaborative editing so users don't sync broken documents up and down?

packages/core/src/Editor.ts
```diff
@@ -276,7 +284,33 @@ export class Editor extends EventEmitter<EditorEvents> {
...
+
+    try {
+      doc = createDocument(
+        this.options.content,
+        this.schema,
+        this.options.parseOptions,
+        { errorOnInvalidContent: this.options.enableContentCheck },
+      )
+    } catch (e) {
+
```
svenadlung, line 297: does it need that empty line?

packages/core/src/Editor.ts
```diff
@@ -276,7 +284,33 @@ export class Editor extends EventEmitter<EditorEvents> {
...
+      // Remove the collaboration extension if the content is invalid, to not sync invalid content
```
svenadlung, line 298: I wonder if this is always the desired behavior. In any case, we should document it.

svenadlung: + I wonder if it's breaking because the behavior was different before.

packages/core/src/helpers/createNodeFromContent.ts
```diff
@@ -46,18 +47,58 @@ export function createNodeFromContent(
 
       return schema.nodeFromJSON(content)
     } catch (error) {
+
```
svenadlung, line 50: is this line on purpose?

packages/core/src/helpers/createNodeFromContent.ts
```diff
@@ -46,18 +47,58 @@ export function createNodeFromContent(
...
+    let hasInvalidContent = false
+
+    // Only ever check for invalid content if we're supposed to throw an error
+    if (options.errorOnInvalidContent) {
+      schemaToUse = new Schema({
+        topNode: schema.spec.topNode,
+        marks: schema.spec.marks,
+        // Prosemirror's schemas are executed such that: the last to execute, matches last
+        // This means that we can add a catch-all node at the end of the schema to catch any content that we don't know how to handle
+        nodes: schema.spec.nodes.append({
```
svenadlung, line 72: It is enough to have this as the last node, as it also captures any unknown element, even marks, right?

packages/core/src/Editor.ts
```diff
@@ -276,7 +284,33 @@ export class Editor extends EventEmitter<EditorEvents> {
...
+      this.options.extensions = this.options.extensions.filter(extension => extension.name !== 'collaboration')
+
+      // Recreate the extension manager without the invalid extensions
+      this.createExtensionManager()
+
+      this.emit('contentError', { editor: this, error: e as Error })
+
+      // Content is invalid, but attempt to create it anyway, stripping out the invalid parts
```
svenadlung, line 306: Above, we prevent a sync to other collaborative users, but regardless of this, the doc is changed and `onUpdate` is triggered. In a setup without collaboration, as a developer, I would have to hook in accordingly when saving (e.g. through the changed doc or onUpdate) and check whether an error was previously emitted. Is this correct?

packages/core/src/Editor.ts
nperez0111: I think I need help setting up my linter properly.

packages/core/src/helpers/createNodeFromContent.ts
nperez0111: yea, from my understanding since this is trying to parse HTML any thing that does not match should match this. I can add a couple more test cases to verify this

packages/core/src/Editor.ts
nperez0111: I put it there more as a conversation starter, that it is at least possible to do here. I wanted to talk more about what our options should be. Perhaps we can give the developer choice in how they want to handle this with some sort of a callback function to execute this maybe?

svenadlung: I like the callback idea. @bdbch what do you think?

packages/core/src/helpers/createNodeFromContent.ts
nperez0111: So I've looked further into this, yes it does catch marks but only those that aren't purely just a style rule. There does not seem to be anyway to capture those and this would not be detected by that.

So for example if you did not have the bold extension and `<p style="font-weight: bold">test</p>` came through, it would say that no error occurred and strip the content silently.

I tried to make a catch all StyleParseRule like this:
```
.append({
          __tiptap__private__unknown__catch__all__mark: {
            content: 'inline*',
            group: 'inline',
            parseDOM: [
              {
                style: '*',
                getAttrs: () => {
                  console.log('is called')
                  // If this is ever called, we know that the content has something that we don't know how to handle in the schema
                  hasInvalidContent = true
                  return null
                },
              },
            ],
          },
        })
```

But, [due to the way that prosemirror matches styles](https://github.com/ProseMirror/prosemirror-model/blob/d61616994c1907f6856aa2cf027a0e4944fc8023/src/from_dom.ts#L248), it has to be an exact match. If we wanted to be absolutely certain on this, then we would need to move upstream to prose-mirrors parser for this.

packages/core/src/commands/insertContentAt.ts
nperez0111: I'm unsure if your thumbs up means you like how it is or want to change it to what I sai.

packages/core/src/Editor.ts
svenadlung: Awesome. Thank you.

packages/core/src/commands/insertContentAt.ts
bdbch: Mh good question. I think just returning false is enough as long as invalid content is not inserted.

packages/core/src/helpers/createNodeFromContent.ts
bdbch: I think having this on node level first is good. We can try to discuss with Marjin how we could solve this problem with Marks and come back to it later.

bdbch (approved):
Just a few comments - otherwise looks good. nice job

Palanikannan1437:
Thanks for this @nperez0111 🫶

> As part of the error context, you are provided with a disableCollaboration function. Invoking this function reinitializes the editor without the collaboration extension, ensuring that any removed content is not synchronized with other users.

RomainLanz:
Looking at `onContentError`, how could I use it to rename a node type?

I only want to change the node type name in the JSON during parsing, something like:

```ts
onContentError(something) {
  if (something.nodeType === 'oldName') {
    something.nodeType = 'newName'
    return
  }
}
```

nperez0111:
@RomainLanz 
I would not rename node types if at all possible. `onContentError` Can reliably detect an issue, but it would be up to you on how to handle it. You would need to rewrite the JSON then unmount & remount the editor (since it's initial content changed)

PR #5206, opened 2024-06-03
fix: (vue-3) component rendering

Rirax:
## Changes Overview
This PR is a step forward in fixing performance issue with large documents migrating from vue 2 to vue 3.
https://github.com/ueberdosis/tiptap/issues/5031

## Implementation Approach
Our approach is to eliminate the use of `Teleport` to locate elements in the DOM and the need to render the app each time for each new node.
In the proposed changes we return to an implementation close to that of the vue-2 package.
`Vue.extends` no longer exists in vue-3, so we've added a `renderComponent` method to `VueRenderer.ts`, which reproduces its behaviour

## Testing Done
Create a document requiring the rendering of more than 2000 vue components

Capture d’écran 2024-06-03 à 22 03 27

## Checklist

## Related Issues
https://github.com/ueberdosis/tiptap/issues/5031


nperez0111 (approved):
Very excited to see this thank you @Rirax 

I'm not super confident in vue to know where the edge cases might be for something like this. Would it be alright to just put into a beta & see if we get any feedback on it from other users?

nperez0111:
I think that this approach could potentially help with our react integration as well. I'll need to think on how to actually do that

nperez0111:
Whoops meant to release on develop not main.
Reverted on main, and cherry-picked to develop

james-astalty:
It looks like this has broken a few things in our app.

I believe this is because `component.ref` is no longer exposed on the rendered component. This is an issue for the Mention functionality - this is evident in the docs as well.

https://tiptap.dev/docs/editor/extensions/nodes/mention

To reproduce, simply type `@` and then try to arrow-down within the mention menu to see that the keydown method does not get called at all.
```ts
      onKeyDown(props) {
        if (props.event.key === 'Escape') {
          popup[0].hide()

          return true
        }

        return component.ref?.onKeyDown(props)
      },
```

Rirax:
Hello @james-astalty 
Thanks for your message,
As you mentioned, we can now access exposed component methods via `component.renderedComponent.vNode?.component?.exposed`.

Perhaps a hotfix is useful to maintain compatibility with `ref`?

nperez0111:
> Hello @james-astalty Thanks for your message, As you mentioned, we can now access exposed component methods via `component.renderedComponent.vNode?.component?.exposed`.
> 
> Perhaps a hotfix is useful to maintain compatibility with `ref`?

This be great to have!

Rirax:
Hi @nperez0111 
There it is https://github.com/ueberdosis/tiptap/pull/5346

nperez0111:
@Rirax, I found an issue with this PR here: https://github.com/ueberdosis/tiptap/pull/5373#issuecomment-2242398445

Thanks for your support

PR #5207, opened 2024-06-04
feat: in a collab setting, disable transactions that are trying to sync invalid content

nperez0111:
## Changes Overview

This enables content checking for changes made by the collaboration plugin.
The changes made by the collaboration plugin are outside of the usual flow of content provided on initialization.
So, we need to handle it separately by filtering out the transactions where it occurs

## Checklist

### DIFF
--- a/packages/extension-collaboration/src/collaboration.ts
+++ b/packages/extension-collaboration/src/collaboration.ts
@@ -6,11 +6,12 @@ import {
   redo,
   undo,
   ySyncPlugin,
-  ySyncPluginKey,
   yUndoPlugin,
   yUndoPluginKey,
 } from 'y-prosemirror'
 import { UndoManager } from 'yjs'
+
+import { isChangeOrigin } from './helpers/isChangeOrigin.js'
 
 type YSyncOpts = Parameters<typeof ySyncPlugin>[1]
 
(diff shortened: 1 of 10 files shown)


nperez0111:
This isn't working, because when disabling the extension, it still seems to be bound to the y doc state (since it is synchronized outside the editor). I think I may have to destroy the editor instead of getting the plugin to be unregistered (since it is still holding onto references in the current editor).

Will need to think more on this later.

packages/extension-collaboration/src/collaboration.ts
```diff
@@ -173,6 +191,51 @@ export const Collaboration = Extension.create<CollaborationOptions>({
...
+            }
+
+            // Did the doc actually change as a result of this transaction?
+            if (tr.docChanged) {
+
+              // Attempt to parse the content of the step
+              try {
+                const content = tr.doc.toJSON()
+
+                this.editor.schema.nodeFromJSON(content)
```
nperez0111, line 220: Should also do a `content.check()` here to make the schema is valid in structure too

packages/extension-collaboration/src/collaboration.ts
```diff
@@ -173,6 +191,51 @@ export const Collaboration = Extension.create<CollaborationOptions>({
...
+              } catch (error) {
+                this.editor.emit('contentError', {
+                  error: error as Error,
+                  editor: this.editor,
+                  disableCollaboration: () => {
+                    this.storage.isCollaborationDisabled = true
+                    this.editor.unregisterPlugin(ySyncPluginKey)
+                    this.editor.unregisterPlugin(yUndoPluginKey)
```
nperez0111, line 228: I think that the best way to disable collaboration here is actually just to disconnect the provider, we cannot actually disconnect the editor from reading changes in Y.js because it may not be only this plugin that is connected to it.

If we were to provide the provider as an option to the collaboration plugin we could be more certain that the changes CANNOT be synchronized to other users. Of course the caller could also easily implement this too, but it is both consistent with the current API on initialization and makes for an easy-to-use API.

guarmo (approved):

PR #5252, opened 2024-06-19
fix(vue-3): fix editor.state updating too late during a transaction due to reactiveState

### LINKED ISSUE #4870: [Bug]: shouldShow broken in BubbleMenu extension from Vue 3 package

core, pm, starter-kit, vue-3

2.2.2

I'm noticing that the `shouldShow` callback is not functioning as expected in the Vue implementation of the Bubble Menu. When using `editor.isActive` as part of the condition, the value returned from `isActive` doesn't represent that state that I expect.

Chrome

### Code example

https://stackblitz.com/edit/vitejs-vite-kgmrcn?file=src%2Fcomponents%2FTiptap.vue,package.json

Consider `shouldShow: ({ editor }) => editor.isActive('bold')`. I expect on first click into a bolded word that the bubble menu would become visible. However, it isn't and logging out the value of `isActive` in the method at that point returns `false`. But upon clicking again on any text (bold or not), the value then returns `true`. It's as if the `editor` state inside the callback is from before the click is registered.

I do not see this same behavior in the vanilla JS implementation of Bubble Menu. I have not tested any other frameworks.

For comparison to the code example below, here is a working vanilla JS port of the same code.
[https://stackblitz.com/edit/vitejs-vite-djsnpl?file=index.html,main.js&terminal=dev](https://stackblitz.com/edit/vitejs-vite-djsnpl?file=index.html,main.js&terminal=dev)

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

segevfiner:
## Changes Overview

Fix the issue where `editor.state` is the previous state during callbacks called from `EditorView.updateState`.

## Implementation Approach

Update `reactiveState` earlier before calling `view.updateState` so that `editor.state` reflects the new state during the `updateState` call for plugin views, node views, and other stuff. e.g. Bubble menu, `Suggestion.items` callback. This is done by adding a new event `beforeTransaction`.

## Testing Done

Ran https://stackblitz.com/edit/vitejs-vite-kgmrcn?file=src%2Fcomponents%2FTiptap.vue,package.json

With `npm link @tiptap/core @tiptap/pm @tiptap/starter-kit @tiptap/vue-3`

## Verification Steps

You can try the same as I did in [Testing Done](#testing-done), before & after.

## Additional Notes

We have hit a similar issue with the suggestion plugin, which was what prompted me to find a fix.

## Checklist

## Related Issues

Fixes #4870


nperez0111:
Hi @segevfiner, we actually have a change for the vue-3 integration in the pipeline. Could you maybe double check whether it resolves your issue or not.

I don't totally understand vue's model so I wouldn't know exactly how to test this.

Either way, really appreciate your initiative & your contribution 😄

nperez0111:
Sorry, I failed to mention that the vue change was released as a beta under the npm tag: `2.5.0-beta.4`

segevfiner:
@nperez0111 The issue is still there in 2.5.0-beta.4

segevfiner:
Why was this closed?

nperez0111:
I'm sorry, I will talk to him about it later.

I think that this is valuable, but the only thing is that I don't like the idea of exposing a beforeTransaction as it feels like a leaky abstraction. I understand it is a common pattern in other places, but I tried to work on your code sandbox to see if I could get around it & couldn't after 30min

segevfiner:
We could have a private callback as well if you think this is something
that should be private.

בתאריך יום ה׳, 27 ביוני 2024, 9:45, מאת Nick Perez ‏<
***@***.***>:

packages/core/src/Editor.ts
```diff
@@ -406,6 +406,11 @@ export class Editor extends EventEmitter<EditorEvents> {
     const state = this.state.apply(transaction)
     const selectionHasChanged = !this.state.selection.eq(state.selection)
 
+    this.emit('beforeTransaction', {
+      editor: this,
+      transaction,
+      newState: state,
+    })
     this.view.updateState(state)
     this.emit('transaction', {
```
nperez0111, line 415: Or maybe, what if we emitted this event *before* updating the view's state. That should allow you to read things before the view's state actually applies

nperez0111:
I'd be interested to know what that looks like. Another approach that I'm thinking of is to maybe make this `beforeViewUpdate` which makes feel a little more comfortable

packages/core/src/Editor.ts
segevfiner: The current handlers for `transaction` expect it to already by applied, so this will just introduce the same issue in all framework bindings.

bdbch:
> Why was this closed?

Sorry did some restructuring of the branches and a few PR's were closed automatically due to this.

segevfiner:
> > Why was this closed?
> 
> Sorry did some restructuring of the branches and a few PR's were closed automatically due to this.

Figured as much.

nperez0111:
> I'd be interested to know what that looks like. Another approach that I'm thinking of is to maybe make this `beforeViewUpdate` which makes feel a little more comfortable

Any thoughts on this @segevfiner I would like to get this in

segevfiner:
This is how it could look like using a protected member instead:
```diff
diff --git a/packages/core/src/Editor.ts b/packages/core/src/Editor.ts
index 99aedb81e..9269c8e64 100644
--- a/packages/core/src/Editor.ts
+++ b/packages/core/src/Editor.ts
@@ -406,7 +406,7 @@ export class Editor extends EventEmitter<EditorEvents> {
     const state = this.state.apply(transaction)
     const selectionHasChanged = !this.state.selection.eq(state.selection)
 
-    this.emit('beforeTransaction', {
+    this.beforeTransaction?.({
       editor: this,
       transaction,
       newState: state,
@@ -453,6 +453,12 @@ export class Editor extends EventEmitter<EditorEvents> {
     })
   }
 
+  /**
+   * Internal callback used by some framework bindings to update their state
+   * before applying the transaction to the view
+   */
+  protected beforeTransaction?(props: { editor: Editor; transaction: Transaction, newState: EditorState }) : void;
+
   /**
    * Get attributes of the currently selected node or mark.
    */
diff --git a/packages/core/src/types.ts b/packages/core/src/types.ts
index 57fb48c47..4ce217f7d 100644
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -50,7 +50,6 @@ export interface EditorEvents {
   }
   update: { editor: Editor; transaction: Transaction }
   selectionUpdate: { editor: Editor; transaction: Transaction }
-  beforeTransaction: { editor: Editor; transaction: Transaction, newState: EditorState }
   transaction: { editor: Editor; transaction: Transaction }
   focus: { editor: Editor; event: FocusEvent; transaction: Transaction }
   blur: { editor: Editor; event: FocusEvent; transaction: Transaction }
diff --git a/packages/vue-3/src/Editor.ts b/packages/vue-3/src/Editor.ts
index e718eeff0..43625e521 100644
--- a/packages/vue-3/src/Editor.ts
+++ b/packages/vue-3/src/Editor.ts
@@ -52,11 +52,6 @@ export class Editor extends CoreEditor {
     this.reactiveState = useDebouncedRef(this.view.state)
     this.reactiveExtensionStorage = useDebouncedRef(this.extensionStorage)
 
-    this.on('beforeTransaction', ({ newState }) => {
-      this.reactiveState.value = newState
-      this.reactiveExtensionStorage.value = this.extensionStorage
-    })
-
     return markRaw(this) // eslint-disable-line
   }
 
@@ -68,6 +63,11 @@ export class Editor extends CoreEditor {
     return this.reactiveExtensionStorage ? this.reactiveExtensionStorage.value : super.storage
   }
 
+  protected override beforeTransaction({ newState }: { newState: EditorState }): void {
+    this.reactiveState.value = newState
+    this.reactiveExtensionStorage.value = this.extensionStorage
+  }
+
   /**
    * Register a ProseMirror plugin.
    */
```

If this looks better, I will push this to this branch (Squashing if you want) @nperez0111 ?

nperez0111:
@segevfiner Appreciate your effort here. On second thought, let's keep it how you had it though, and just rename `newState` to `nextState`

Really appreciate your solution here. I think that the event emitter might be useful for other bindings in the future

nperez0111 (approved):
LGTM now

segevfiner:
@nperez0111  Why was this reverted, it just returned the bug we had and we now will have to apply this fix manually again?

nperez0111:
Sorry @segevfiner I explained a bit more here: https://github.com/ueberdosis/tiptap/issues/5031#issuecomment-2238629762

Some people in the community had issues with vue, so I had reverted it to look into it more. I should be able to look more into this during the week (likely the first couple of days).

segevfiner:
Just to let you know, we are running this one in production over at https://swimm.io/, so my gut feeling it wasn't this change, but likely the other rendering related one?  Which is more risky... I do wish that Vue had a feature to do such a procedural teleport easily from a single Vue app instead of the hacks you need to do today..., not sure if there is a discussion or issue about this

PR #5257, opened 2024-06-20
fix: streamline html package generation functions to work as expected

### LINKED ISSUE #4089: [Bug]: Cannot parse generated html from YouTube extension

extension-youtube, html

^2.0.0

Using `generateHTML` & `generateJSON` utilities from `@tiptap/html` doesn't work with the YouTube extension. Sample easily replicable test case:

```ts
import { generateJSON, generateHTML } from '@tiptap/html';
import Document from '@tiptap/extension-document';
import Text from '@tiptap/extension-text';

it('can convert from & to html', async () => {
    const extensions = [Document, Text, Youtube];
    const html = `<p>Tiptap now supports YouTube embeds! Awesome!</p>
      <div data-youtube-video>
        <iframe src="https://www.youtube.com/watch?v=cqHqLQgVCgY"></iframe>
      </div>`;
    const json = generateJSON(html, extensions);
    expect(generateHTML(json, extensions)).to.equal(html);
});
```

Test fails with:
```
Unknown CSS selector type descendant div[data-youtube-video] iframe [
  { type: 'tag', name: 'div', namespace: null },
  {
    type: 'attribute',
    name: 'data-youtube-video',
    action: 'exists',
    value: '',
    namespace: null,
    ignoreCase: null
  },
  { type: 'descendant' },
  { type: 'tag', name: 'iframe', namespace: null }
]

TypeError: Cannot read properties of null (reading 'includes')
❯ getEmbedUrlFromYoutubeUrl ../../node_modules/@tiptap/extension-youtube/src/utils.ts:56:11
❯ Object.renderHTML ../../node_modules/@tiptap/extension-youtube/src/youtube.ts:138:22
❯ Object.schema.toDOM [as youtube] ../../node_modules/@tiptap/core/src/helpers/getSchemaByResolvedExtensions.ts:100:32
❯ DOMSerializer.serializeNodeInner ../../node_modules/prosemirror-model/dist/index.js:3280:100
❯ ../../node_modules/prosemirror-model/dist/index.js:3272:34
❯ Fragment.forEach ../../node_modules/prosemirror-model/dist/index.js:255:13
❯ DOMSerializer.serializeFragment ../../node_modules/prosemirror-model/dist/index.js:3246:18
❯ getHTMLFromFragment ../../node_modules/@tiptap/html/src/getHTMLFromFragment.ts:5:53
❯ Module.generateHTML ../../node_modules/@tiptap/html/src/generateHTML.ts:10:10
❯ src/index.spec.ts:47:12
```

Chrome

### Code example

_No response_

I would expect the above test to pass.

Looks like this could be an issue with `zeed-dom` and how it parses the html. I would happily open an issue there however since I couldn't find much information relating to a similar issue (only [this](https://github.com/ueberdosis/tiptap/issues/3539) seems close) it's hard to tell exactly what the issue is.

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

bdbch:
## Changes Overview
This PR removes logic that was specific to the `@tiptap/html` package and didn't work the same way the core `generateHTML` and `generateJSON` functions work. This caused issues with a few extensions (for example the Youtube extension not being able to be generated back to HTML from JSON, see #4089)

I kept the original HTML `getJSON` and `getHTML` functions to not break existing code for users but maybe we should add a deprecation message to it?

## Implementation Approach
I removed the logic of the functions of the html package and just used a reference to the core packages util functions.

## Testing Done
I tried running JSON and HTML generation steps before and after and checked if unexpected things were happening or the output was false.

## Verification Steps
1. Try to check out this PR
2. Create a demo and simply run the code example below
3. Verify output

```js
import { generateJSON, generateHTML } from '@tiptap/html'
const extensions = [Document, Paragraph, Text, Youtube]
const content `<p>Hello</p><div data-youtube-video><iframe src="..."></iframe></div>`

const json = generateJSON(content, extensions)
const html = generateHTML(json, extensions)

console.log(json, html)
```

## Checklist

## Related Issues
- fixes #4089


nperez0111:
Wouldn't this break being able to use `@tiptap/html` package on the server-side where this is no reference to a global document object 🤔

bdbch:
True, didn't think about this. I'll add back the zeed fallback if the document is not available but I guess this will also cause the issues from #4089 to pop back up.

nperez0111:
If you are on the server side anyway I think that we could use another package than zeed-dom. Perhaps happy-dom? It is a heavier package, but we should not be using this package in the browser. We could even give a warning to a user if in development mode and there is a global document object that they should not be bundling this into their application.

bdbch:
Closing this for now as we'd probably do this via another html parser.

Issue #3602, opened 2023-01-07
Text replacement feature in macOS is broken


craigcarlyle:
### What’s the bug you are facing?

There appears to be an issue when trying to use the text replacement feature built in to macOS with Tiptap. Attempting to do so will result in the following error:

![Screenshot 2023-01-06 at 4 22 29 PM](https://user-images.githubusercontent.com/1305500/211121400-ebe72750-2abe-4f91-95d1-8f1d6a27f298.png)

The error seems to be coming from [here](https://github.com/ueberdosis/tiptap/blob/62d42275c529aa3c819e85133b0ad0dd22cac934/packages/core/src/inputRules/markInputRule.ts#L64). This may be caused by an invalid `to` value 🤔 

![Screenshot 2023-01-06 at 4 02 58 PM](https://user-images.githubusercontent.com/1305500/211121548-65b5369d-939f-4c42-90ee-5b5d69231c8c.png)

### Which browser was this experienced in? Are any special extensions installed?

Chrome, Safari, Firefox

### How can we reproduce the bug on our side?

* Create a text replacement in macOS. For example, `[crd]` to `**[CARD]**`:
![Screenshot 2023-01-06 at 4 01 39 PM](https://user-images.githubusercontent.com/1305500/211121196-5f67115e-6c61-4f50-9e24-a998f5c0b9cb.png)
* Attempt to use the text replacement shortcut in Tiptap:

https://user-images.githubusercontent.com/1305500/211121301-086f4d9b-dda9-4ec7-8749-c4d8c9f49f7b.mov

### What did you expect to happen?

Expected text should be entered (like in this demo from another editor):

https://user-images.githubusercontent.com/1305500/211121065-3c909e12-f8c0-446f-9bda-e6462c55624f.mov

### Did you update your dependencies?

- [X] Yes, I’ve updated my dependencies to use the latest version of all packages.

### Are you sponsoring us?

- [x] Yes, I’m a sponsor. 💖

rfgamaral:
@bdbch This issue never ended up being fixed, can you please reopen? I have an upcoming PR to fix this 😊

bdbch:
Thanks! Reopened the issue

ElHex:
Same error, if you have the textStyle extension installed and using the ortho corrector you replace a word for the suggested one the marks dissapear.

nperez0111:
This has been released with 2.7.0-pre.0 and will be released in tiptap 2.7.0 when this beta has been tested by more users

PR #5261, opened 2024-06-21
fix: Broken text replacement feature (macOS-only)

rfgamaral:
## Changes Overview

Fixes https://github.com/ueberdosis/tiptap/issues/3602.

## Implementation Approach

This input rules feature currently works by handling the [`handleTextInput`](https://prosemirror.net/docs/ref/#view.EditorProps.handleTextInput) event, which is fired whenever the user directly inputs text. However, it's important to note that this handler is called **before** the input is applied, which is basically the source of the issue. Because of that, when a macOS text replacement happens, the positions used by the logic that handles input rules, don't match the editor state, and the editor breaks. This is an oversimplified version, but if you are interested to know more, I've written about it [here](https://discuss.prosemirror.net/t/wrong-state-in-handletextinput-when-using-macos-text-replacement/6461).

The solution implemented in this PR's patch replaces `handleTextInput` with [`handleDOMEvents`](https://prosemirror.net/docs/ref/#view.EditorProps.handleDOMEvents) for input rules, which allows us to handle the `keyup` event. Such event is better suitable here because it will fire **after** the user directly inputs text, which means the editor state will be as updated as it can be, and all position calculations will match that state.

## Testing Done

I've patched our app using Tiptap with the exact changes in this PR, and our millions of users have been using it for 3 weeks now, and we haven't received any tickets regarding editor issues since then. I'd say this change was a success.

## Verification Steps

See the reproduction steps in the referenced issue that this PR fixes.

## Checklist

### DIFF
--- a/demos/src/Examples/InteractivityComponentContent/React/index.spec.js
+++ b/demos/src/Examples/InteractivityComponentContent/React/index.spec.js
@@ -4,19 +4,19 @@ context('/src/Examples/InteractivityComponentContent/React/', () => {
   })
 
   it('should have a working tiptap instance', () => {
-    cy.get('.ProseMirror').then(([{ editor }]) => {
+    cy.get('.tiptap').then(([{ editor }]) => {
       // eslint-disable-next-line
       expect(editor).to.not.be.null
     })
   })
 
   it('should render a custom node', () => {
-    cy.get('.ProseMirror .react-component')
+    cy.get('.tiptap .react-component')
       .should('have.length', 1)
   })
 
   it('should allow text editing inside component', () => {
-    cy.get('.ProseMirror .react-component .content div')
+    cy.get('.tiptap .react-component .content div')
       .invoke('attr', 'contentEditable', true)
       .invoke('text', '')
       .type('Hello World!')
@@ -24,24 +24,27 @@ context('/src/Examples/InteractivityComponentContent/React/', () => {
   })
 
   it('should allow text editing inside component with markdown text', () => {
-    cy.get('.ProseMirror .react-component .content div')
+    cy.get('.tiptap .react-component .content div')
       .invoke('attr', 'contentEditable', true)
       .invoke('text', '')
-      .type('Hello World! This is **bold**.')
+      .click()
+    cy.get('.tiptap .react-component .content div')
+      .realType('Hello World! This is **bold**.')
+    cy.get('.tiptap .react-component .content div')
       .should('have.text', 'Hello World! This is bold.')
 
-    cy.get('.ProseMirror .react-component .content strong')
+    cy.get('.tiptap .react-component .content strong')
       .should('exist')
   })
 
   it('should remove node via selectall', () => {
-    cy.get('.ProseMirror .react-component')
+    cy.get('.tiptap .react-component')
       .should('have.length', 1)
 
-    cy.get('.ProseMirror')
+    cy.get('.tiptap')
       .type('{selectall}{backspace}')
 
-    cy.get('.ProseMirror .react-component')
+    cy.get('.tiptap .react-component')
       .should('have.length', 0)
   })
 })
--- a/tests/cypress/support/commands.js
+++ b/tests/cypress/support/commands.js
@@ -1,3 +1,5 @@
+import 'cypress-real-events'
+
 // ***********************************************
 // This example commands.js shows you how to
 // create various custom commands and overwrite
@@ -95,3 +97,11 @@ Cypress.Commands.add(
     return subject
   },
 )
+
+Cypress.Commands.add('resetEditor', () => {
+  cy
+    .get('.tiptap').then(([{ editor }]) => {
+      editor.commands.clearContent()
+    })
+    .click()
+})
--- a/demos/src/Commands/InsertContentApplyingRules/React/index.spec.js
+++ b/demos/src/Commands/InsertContentApplyingRules/React/index.spec.js
@@ -4,9 +4,7 @@ context('/src/Commands/InsertContentApplyingRules/React/', () => {
   })
 
   beforeEach(() => {
-    cy.get('.tiptap').then(([{ editor }]) => {
-      editor.commands.clearContent()
-    })
+    cy.resetEditor()
   })
 
   it('should apply list InputRule', () => {
--- a/demos/src/Examples/InteractivityComponentContent/Vue/index.spec.js
+++ b/demos/src/Examples/InteractivityComponentContent/Vue/index.spec.js
@@ -27,7 +27,10 @@ context('/src/Examples/InteractivityComponentContent/Vue/', () => {
     cy.get('.tiptap .vue-component .content')
       .invoke('attr', 'contentEditable', true)
       .invoke('text', '')
-      .type('Hello World! This is **bold**.')
+      .click()
+    cy.get('.tiptap .vue-component .content')
+      .realType('Hello World! This is **bold**.')
+    cy.get('.tiptap .vue-component .content')
       .should('have.text', 'Hello World! This is bold.')
 
     cy.get('.tiptap .vue-component .content strong')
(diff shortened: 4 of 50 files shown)


rfgamaral:
@bdbch A lot of tests are breaking because the new implementation doesn't seem to play well with `cy.type()`. Tried to replace with `realType()` ([ref](https://github.com/dmtrKovalenko/cypress-real-events)), and while I was able to fix some of them, not all of them were easily fixable with that library. Not exactly sure how to handle this, do you think you can have a look?

rfgamaral:
> This change does really feel like an edge case and I'm not super confident in the change.

For what it's worth, we've been using this change on [Todoist](https://todoist.com) for almost 2 months now, and so far we haven't had any bug reports related to the change. That said, I'd be happy if someone else could have a look and find a different solution, but personally, I have no idea how that looks like (I've already spent a lot of time on it, and this was my best shot).

rfgamaral:
@nperez0111 Any thoughts on how to better fix this issue?

nperez0111:
Hey @rfgamaral been looking at this now,

So, I think that we should be fixing the root cause instead of trying to make a workaround like this.

The root cause is that what got inputted & what exists in the editor are different. & I'm unsure that the macos text replacement feature let's you really do anything about this.

I'm not sure what a proper resolution is for this other than what you've come up with

Separately, I took a look at what remirror does (they seem to just not apply the text in this situation), I noticed that they are trying to preserve the initial stored marks https://github.com/remirror/remirror/blob/b3c8b02f7562afb7de08b4308a3302a500386d33/packages/remirror__core-utils/src/prosemirror-rules.ts#L213 and I think that we should be doing that too actually.

rfgamaral:
> Separately, I took a look at what remirror does (they seem to just not apply the text in this situation), I noticed that they are trying to preserve the initial stored marks https://github.com/remirror/remirror/blob/b3c8b02f7562afb7de08b4308a3302a500386d33/packages/remirror__core-utils/src/prosemirror-rules.ts#L213 and I think that we should be doing that too actually.

But going that route, the macOS text replacement feature will cease to work, right?

nperez0111:
> But going that route, the macOS text replacement feature will cease to work, right?

I think the storedMarks is completely separate & orthogonal to this, I just saw it in trying to understand how they implemented it.

In their editor, the text will replace, but not match the mark, which I honestly think is acceptable. I think that the main goal here should be to not crash the editor and a secondary goal can be to get it to be marked up properly.

The main issue that I have with your solution is:
 1. I'm unsure how it will work with someone using IME, as I'm unsure if that triggers keyup.
 2. it is pretty obscure why it needs to be handled like that & I don't like that it gates on there being a single character inputted, I'm unsure how that would work with other languages

rfgamaral:
> I think that the main goal here should be to not crash the editor and a secondary goal can be to get it to be marked up properly.

I can understand that, although, to be fair, the editor doesn't crash per se 😅, it continues to work despite the error on the console.

> The main issue that I have with your solution is

For what it's worth, I can tell you that we've been using this workaround for a couple of months now, and we have a huge user base where some users use IMEs, and we also [support lots of different languages](https://todoist.com/help/articles/change-languages-in-todoist-EJZMow). So far, we haven't had anyone complaining about issues in the editor related to this particular change.

> I'm unsure how it will work with someone using IME, as I'm unsure if that triggers keyup.

Firefox seems to fire it as expected ([ref](https://developer.mozilla.org/en-US/docs/Web/API/Element/keyup_event#keyup_events_with_ime)), and I'm sure Chromium-based browsers will as well.

> it is pretty obscure why it needs to be handled like that

I don't think it's that obscure... I mean, we've reached the conclusion that the main issue is "what got inputted & what exists in the editor are different", and that's an issue caused by handling input rules with `handleTextInput` which is called **before** the input is applied to the editor state. My solution only changes the handling of input rules to be called **after** the input is applied to the editor state. Why do you think that's obscure?

If we go to the ProseMirror source ([ref](https://github.com/ProseMirror/prosemirror-view/blob/master/src/input.ts#L141-L158)), we can see that `handleTextInput` is called within `keypress`, and I've only changed this to be handled on `keyup`. If ProseMirror had implemented a `handleTextInputBefore` (`keypress`) and `handleTextInputAfter` (`keyup`), it would've been just a matter of changing the name of the handler.

rfgamaral:
> Alright, I'm convinced by your argument.

I wasn't trying to convince you per se 😅, just sharing my point of view. Ultimately, this is your call, of course 👍

rfgamaral:
@nperez0111 The test pipeline is still running, but I should've fixed all Cypress tests. It required the usage of [`cypress-real-events`](https://github.com/dmtrKovalenko/cypress-real-events) for typing instead of using Cypress's default `type` (which are simulated and not real events, thus not all handlers are called correctly when expected).

nperez0111:
> @nperez0111 The test pipeline is still running, but I should've fixed all Cypress tests. It required the usage of [`cypress-real-events`](https://github.com/dmtrKovalenko/cypress-real-events) for typing instead of using Cypress's default `type` (which are simulated and not real events, thus not all handlers are called correctly when expected).

Thanks for this, didn't realize that you had a library in mind that worked for this.

rfgamaral:
> Thanks for this, didn't realize that you had a library in mind that worked for this.

It's what we also use internally to solve issues like this (i.e. when `.type` from Cypress is insufficient).

rfgamaral:
@nperez0111 @bdbch @svenadlung  Finally got all tests to pass, feel free to review.

tests/cypress/support/commands.js
```diff
@@ -95,3 +97,11 @@ Cypress.Commands.add(
     return subject
   },
 )
+
+Cypress.Commands.add('resetEditor', () => {
```
nperez0111, line 101: smart!

demos/src/Examples/InteractivityComponentContent/React/index.spec.js
```diff
@@ -4,44 +4,47 @@ context('/src/Examples/InteractivityComponentContent/React/', () => {
...
   it('should allow text editing inside component with markdown text', () => {
-    cy.get('.ProseMirror .react-component .content div')
+    cy.get('.tiptap .react-component .content div')
       .invoke('attr', 'contentEditable', true)
       .invoke('text', '')
-      .type('Hello World! This is **bold**.')
+      .click()
+    cy.get('.tiptap .react-component .content div')
+      .realType('Hello World! This is **bold**.')
+    cy.get('.tiptap .react-component .content div')
```
nperez0111, line 33: Just so I understand better, does realType not support chaining or is this a timing thing?

demos/src/Examples/InteractivityComponentContent/React/index.spec.js
```diff
@@ -4,44 +4,47 @@ context('/src/Examples/InteractivityComponentContent/React/', () => {
   })
 
   it('should have a working tiptap instance', () => {
-    cy.get('.ProseMirror').then(([{ editor }]) => {
+    cy.get('.tiptap').then(([{ editor }]) => {
```
nperez0111, line 7: Appreciate you adding consistency here

bdbch (approved):
lgtm! ty @rfgamaral

demos/src/Examples/InteractivityComponentContent/React/index.spec.js
rfgamaral: I'm no Cypress expert but:

- I initially tried with chaining, but it didn't work
- All our internal tests don't chain the `.real*` calls
- Examples on the [official repo](https://github.com/dmtrKovalenko/cypress-real-events) do use chaining
    - But like I said, I couldn't make it work...

Nantris:
@rfgamaral I'm sorry to report this change breaks our existing input rules, and further breaks `undo` behavior for our apparently still-functional input rules. See #5598

nperez0111:
This was not released with 2.7 due to the potential breaking change

Nantris:
Just FYI I noticed this is listed in the 2.7.0 change list despite not being included.

Sorry again to have been the bearer of bad news. It's a worthwhile goal to get `InputRules` working with the text replacement feature on macOS.

Issue #3976, opened 2023-04-19
[Bug]: Bad performence of ReactRenderer


fantasticit:
### Which packages did you experience the bug in?

react

### What Tiptap version are you using?

2.0.1

### What’s the bug you are facing?

The render method of ReactRenderer is called every time I type, which causes the editor to lag. Is there a way to improve this? Also, is it possible to add a viewport judgment inside the NodeViewContent to determine whether to render it or not to optimize performance?

```
import React from 'react';

import { Node } from '@tiptap/core';
import { NodeViewContent, NodeViewProps, NodeViewWrapper, ReactNodeViewRenderer } from '@tiptap/react';

export const CardView: React.FC<NodeViewProps> = ({ editor, node, getPos }) => {
  return (
    <NodeViewWrapper>
      {/* {some other element} */}
      <NodeViewContent />
    </NodeViewWrapper>
  );
};

export const Card = Node.create({
  name: 'card',
  group: 'card',
  content: 'block+',
  inline: false,
  isolating: true,
  marks: '',

  addNodeView() {
    return ReactNodeViewRenderer(CardView);
  },
});

Document.extend({ content: `card+` })
```

### What browser are you using?

Chrome

### What did you expect to happen?

improve performence 

### Did you update your dependencies?

- [X] Yes, I’ve updated my dependencies to use the latest version of all packages.

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

C-Hess:
> The render method of ReactRenderer is called every time I type, which causes the editor to lag

I don't think there's any safe optimizations to be made for this behavior that can be handled by the library itself, or without a breaking change. The `editor` object, which is in charge of maintaining TipTap's internal state, is passed down to all React node views. As a result, every change to the editor _should_ result in a re-render of node views. Otherwise, node views that depend on this behavior and have side effects that require a re-render will no longer be able to do so.

I wonder if a `useEditor` hook could be useful here. Node views may not need to re-render as frequently since only node views that use this `useEditor` hook would re-render when the editor content is changed outside of the node's "domain."

Anyways, if you are worried about performance when rendering a complicated node view, you can instead create an inner component that uses [React's `memo` higher-order-function](https://react.dev/reference/react/memo). Then, make it so your component only accepts props that, if changed, should trigger a re-render. You should get virtually the same performance benefits you're likely seeking.

>  is it possible to add a viewport judgment inside the NodeViewContent to determine whether to render it or not to optimize performance

[...]

Issue #4492, opened 2023-09-30
[Bug]: Usage of ReactNodeViewRender drastically reduces performance for large Text


wasular:
### Which packages did you experience the bug in?

core

### What Tiptap version are you using?

2.1.6

### What’s the bug you are facing?

If I use this simple React Component as Wrapper for a Custom Paragraph the performance drastically reduces for large text. Luckily in my use case I don't need to use React Render for the Paragraph but there seems to be some performance bug which could be critical for other users.

```js
import { NodeViewWrapper, NodeViewContent } from '@tiptap/react';
import DragIndicatorIcon from '@mui/icons-material/DragIndicator';
import styles from './DraggableNodeView.module.scss';
import React from 'react';

const DraggableNodeView = ({ contentComponent: ContentComponent, ...props }) => {
  return (
    <NodeViewWrapper className={styles.node__wrapper}>
      <DragIndicatorIcon draggable="true" data-drag-handle className={styles.drag__icon} />
      {ContentComponent ? (
        <ContentComponent {...props} />
      ) : (
        <NodeViewContent className={styles.node__content} />
      )}
    </NodeViewWrapper>
  );
};

export default DraggableNodeView;
```

I now switched to vanilla html and js for this and it works

```js
import { Paragraph } from '@tiptap/extension-paragraph';

const CustomParagraph = Paragraph.extend({
  draggable: true,
  addNodeView() {
    return () => {
      const dom = document.createElement('div');
      dom.className = 'node__wrapper';

      const dragIcon = document.createElement('div');
      dragIcon.draggable = true;
      dragIcon.setAttribute('data-drag-handle', '');
      dragIcon.className = 'drag__icon';
      dragIcon.innerHTML =
        'svg';
      dom.appendChild(dragIcon);

      const content = document.createElement('div');
      content.className = 'node__content';
      dom.appendChild(content);

      return {
        dom: dom,
        contentDOM: content,
      };
    };
  },
});

export default CustomParagraph;
```

### What browser are you using?

Chrome

### What did you expect to happen?

Using a simple React render should not have this big of influence on the performance.

### Did you update your dependencies?

- [X] Yes, I’ve updated my dependencies to use the latest version of all packages.

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

jackbb147:
I have the same issue. Terrible terrible performance with just a few react node views.

Zloka:
[...]

PR #5273, opened 2024-06-26
fix(react): resolves React NodeView performance issues

nperez0111:
## Changes Overview

This PR significantly improves the performance of React NodeViews in a couple of ways:

 - It uses `useSyncExternalStore` to synchronize changes between React & the editor instance
 - It dramatically reduces the number of re-renders by re-using instances of React portals that have already been initialized and unaffected by the change made in the editor

We were seeing performance problems with React NodeViews because a change to one of them would cause a re-render to all instances of node views, for an application that heavily relies on node views in React, this is expensive. This should dramatically cut down on the number of instances that have to re-render as well as making each of those re-renders much less costly.

## Implementation Approach

Similar to the performance updates for Tiptap 2.5, React has it's own lifecycle for components and it really does not like when you try to force it to be synchronous (since it would prefer to batch renders together). Now we can have the best of both worlds by forcing a synchronous render on initialization, but updates afterward can be done on React's normal lifecycle (i.e. without a flushSync)

## Verification Steps

Lots of manual testing with tons of custom react node views. Much faster than before, but hard to write tests for.

## Checklist

## Related Issues

https://github.com/ueberdosis/tiptap/issues/3976
https://github.com/ueberdosis/tiptap/issues/3580

### DIFF
--- a/packages/react/src/ReactRenderer.tsx
+++ b/packages/react/src/ReactRenderer.tsx
@@ -143,7 +143,7 @@ export class ReactRenderer<R = unknown, P = unknown> {
       }
     }
 
-    this.reactElement = <Component {...props } />
+    this.reactElement = React.createElement(Component, props)
 
     this.editor?.contentComponent?.setRenderer(this.id, this)
   }
--- a/packages/react/src/ReactRenderer.tsx
+++ b/packages/react/src/ReactRenderer.tsx
@@ -1,5 +1,6 @@
 import { Editor } from '@tiptap/core'
 import React from 'react'
+import { flushSync } from 'react-dom'
 
 import { Editor as ExtendedEditor } from './Editor.js'
 
@@ -71,6 +72,8 @@ type ComponentType<R, P> =
   React.FunctionComponent<P> |
   React.ForwardRefExoticComponent<React.PropsWithoutRef<P> & React.RefAttributes<R>>;
 
+let first = true
+
 /**
  * The ReactRenderer class. It's responsible for rendering React components inside the editor.
  * @example
@@ -121,7 +124,20 @@ export class ReactRenderer<R = unknown, P = unknown> {
       })
     }
 
-    this.render()
+    if (first) {
+      // TODO this is totally a hack, on the first render of the editor (parent editor)
+      // We need to delay the rendering of the ReactRenderer components (because it would render before the editor is ready)
+      // setTimeout(() => {
+      this.render()
+      first = false
+      // }, 100)
+    } else {
+      // On first render, we need to flush the render synchronously
+      // Renders afterwards can be async, but this fixes a cursor positioning issue
+      flushSync(() => {
+        this.render()
+      })
+    }
   }
 
   render(): void {
--- a/packages/react/src/ReactRenderer.tsx
+++ b/packages/react/src/ReactRenderer.tsx
@@ -72,8 +72,6 @@ type ComponentType<R, P> =
   React.FunctionComponent<P> |
   React.ForwardRefExoticComponent<React.PropsWithoutRef<P> & React.RefAttributes<R>>;
 
-let first = true
-
 /**
  * The ReactRenderer class. It's responsible for rendering React components inside the editor.
  * @example
@@ -124,19 +122,14 @@ export class ReactRenderer<R = unknown, P = unknown> {
       })
     }
 
-    if (first) {
-      // TODO this is totally a hack, on the first render of the editor (parent editor)
-      // We need to delay the rendering of the ReactRenderer components (because it would render before the editor is ready)
-      // setTimeout(() => {
-      this.render()
-      first = false
-      // }, 100)
-    } else {
+    if (this.editor.isInitialized) {
       // On first render, we need to flush the render synchronously
       // Renders afterwards can be async, but this fixes a cursor positioning issue
       flushSync(() => {
         this.render()
       })
+    } else {
+      this.render()
     }
   }
 
--- a/.changeset/rotten-beers-protect.md
+++ b/.changeset/rotten-beers-protect.md
@@ -1,0 +1,12 @@
+---
+"@tiptap/react": minor
+"@tiptap/core": minor
+---
+
+This PR significantly improves the performance of React NodeViews in a couple of ways:
+
+- It now uses useSyncExternalStore to synchronize changes between React & the editor instance
+- It dramatically reduces the number of re-renders by re-using instances of React portals that have already been initialized and unaffected by the change made in the editor
+
+We were seeing performance problems with React NodeViews because a change to one of them would cause a re-render to all instances of node views. For an application that heavily relies on node views in React, this was quite expensive.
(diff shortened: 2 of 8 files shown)


bdbch (approved):
LGTM - lets see how well it performs! :)

packages/react/src/ReactRenderer.tsx
```diff
@@ -121,7 +122,15 @@ export class ReactRenderer<R = unknown, P = unknown> {
       })
     }
 
-    this.render()
+    if (this.editor.isInitialized) {
```
fdintino, line 125: Wouldn't this call flushSync every time `setRenderer` was called after the editor has initialized?

nperez0111: This is in the constructor function, for the initial render.

Before this change, we were calling flushSync on every setRenderer call (with `maybeFlushSync`).

With this change, it will only flush sync on the first render and then every update after will be handled with a `useSyncExternalStore` (i.e. letting React "pick up" the changes whenever it feels like it)

bdbch (approved):

bartlomiej-korpus:
Isn't removing `flushSync` going to break the cursor in transactions that insert a new `NodeView` with `NodeViewContent` inside and try to set the selection inside of editable content in the new node? From my limited understanding of ProseMirror, in this case PM prosemirror-view is going to try to set the cursor to proper position synchronously - expecting the DOM to be there, but React will render sometime in the future so it's not yet available at this point in time.

I'm not sure, but looking at commit history it might've been added to account for this scenario in the first place?

nperez0111:
> Isn't removing `flushSync` going to break the cursor in transactions that insert a new `node` resulting in rendering a new React `NodeView` with `NodeViewContent` inside and try to set the selection inside of editable content in the new node? From my limited understanding of ProseMirror, in this case `prosemirror-view` is going to try to set the cursor to proper position synchronously - expecting the DOM to be there, but React will render sometime in the future so it's not yet available at this point in time.
> 
> I'm not sure, but looking at the commit history, it might've been added to account for this scenario in the first place.

That is what this code & comment is regarding: https://github.com/ueberdosis/tiptap/pull/5273/files#r1676838961

bartlomiej-korpus:
> > Isn't removing `flushSync` going to break the cursor in transactions that insert a new `node` resulting in rendering a new React `NodeView` with `NodeViewContent` inside and try to set the selection inside of editable content in the new node? From my limited understanding of ProseMirror, in this case `prosemirror-view` is going to try to set the cursor to proper position synchronously - expecting the DOM to be there, but React will render sometime in the future so it's not yet available at this point in time.
> > I'm not sure, but looking at the commit history, it might've been added to account for this scenario in the first place.
> 
> That is what this code & comment is regarding: https://github.com/ueberdosis/tiptap/pull/5273/files#r1676838961

I might be mistaken, but from what I understand this is regarding editor initialization, but a transaction like that might occur at any point in time after the editor is already initialized, isn't that the case?

nperez0111:
> bartlomiej-korpus

So this PR only skips flushSync if the editor has not yet initialized (because react can't render new nodes when the editor is just being initialized). So, in the case of an already initialized editor, it will use flushSync to force the node view to be created synchronously so that the selection can be moved into the element synchronously. Once the element is created though, this no longer is a problem because the DOM element is present (since it already is initialized and in the dom) so prosemirror can move the selection into that element.

I will definitely double check to verify this though

nperez0111:
> It is possible to reproduce the cursor issue with our in-house tiptap editor (it's why I never opened a PR for my "fix" from #4492). I can see whether I can reproduce it, and if the problem remains I'll try to share a reduced test case.

Alright, I think that this PR resolves that cursor issue (by `flushSync`ing when the node view is first constructed) but happy for you to test it!

fdintino:
I can confirm that we don't see the cursor placement issue with this patch.

nperez0111:
> I can confirm that we don't see the cursor placement issue with this patch.

That's great to hear, I just rebased it off the current develop (which has the latest useEditor impl) so this should be more accurate to how it actually will behave

bdbch:
Looks good from my side then and can be merged. Feel free @nperez0111

bartlomiej-korpus:
> > bartlomiej-korpus
> 
> So this PR only skips flushSync if the editor has not yet initialized (because react can't render new nodes when the editor is just being initialized). So, in the case of an already initialized editor, it will use flushSync to force the node view to be created synchronously so that the selection can be moved into the element synchronously. Once the element is created though, this no longer is a problem because the DOM element is present (since it already is initialized and in the dom) so prosemirror can move the selection into that element.
> 
> I will definitely double check to verify this though

that's right, now I see! sorry for confusion!

leoyli:
Hey @nperez0111, I have experienced some weird issue after this upgrade (after `@tiptap/react:2.5.9`).  Component menu rendered by `ReactRenderer` did not get the component rendered at all when I destroy one and quickly recreate another instance of `ReactRenderer`. For example, the `Mention` extension that uses with `Suggestion` extension depends on `ReactRenderer` will not show the menu at all.

https://github.com/ueberdosis/tiptap/blob/main/demos/src/Nodes/Mention/React/suggestion.js#L43-L63

I have a chat app that uses what similars to `Mention` to suggest user a text snippets. But when the user navigate to the next contact that has the same UI, the suggestion did not triggered. What I found after the inspection, the menu function component did not get called, and the `ReactRenderer` just attach an empty node there. See this video:

https://github.com/user-attachments/assets/49e60927-ef3b-4530-9d04-6ca3aa3ff1ba

I'm using React 19 RC, but it was fine previously... Do you have any clue why this is happening? Any help is much appreciated.

nperez0111:
> Hey @nperez0111, I have experienced some weird issue after this upgrade (after `@tiptap/react:2.5.9`).  Component menu rendered by `ReactRenderer` did not get the component rendered at all when I destroy one and quickly recreate another instance of `ReactRenderer`. For example, the `Mention` extension that uses with `Suggestion` extension depends on `ReactRenderer` will not show the menu at all.
> 
> 
> 
> https://github.com/ueberdosis/tiptap/blob/main/demos/src/Nodes/Mention/React/suggestion.js#L43-L63
> 
> 
> 
> I have a chat app that uses what similars to `Mention` to suggest user a text snippets. But when the user navigate to the next contact that has the same UI, the suggestion did not triggered. What I found after the inspection, the menu function component did not get called, and the `ReactRenderer` just attach an empty node there.  Even I force the editor to be a new instance, or force React to remount the component by `key` prop did not resolve the issue. See this video:
> 
> 
> 
> https://github.com/user-attachments/assets/49e60927-ef3b-4530-9d04-6ca3aa3ff1ba
> 
> 
> 
> I'm using React 19 RC, but it was fine previously... Do you have any clue why this is happening? Any help is much appreciated. 

Their were changes to how suggestions are triggered in this version too so it is more likely to do with that then an issue with React. Please look for an associated GitHub issue. Tracking a bug in a PR will be difficult

Issue #5298, opened 2024-07-01
[Bug]: getTextContentFromNodes not handling atoms correctly


artursvonda:
### Affected Packages

core

### Version(s)

2.4.0

### Bug Description

When using input rules on content that contains atoms (like mentions), the content of the node may be longer than 1 char which causes the `getTextContentFromNodes` to calculate position / length incorrectly.
That's mitigated by returning single char from `renderText`, but it also means copy/paste is not working properly.

### Browser Used

Chrome

### Expected Behavior

Atom content should be included in full. Or we need to change how we get content for atoms.

### Dependency Updates

- [X] Yes, I've updated all my dependencies.

nperez0111:
Should be resolved now

PR #5321, opened 2024-07-09
fix: take atom content in full

artursvonda:
## Changes Overview
Make sure that atoms are used in-full without cutting the content. Node size for atoms is 1 which causes text to be cut unexpectedly.

## Implementation Approach
Just made sure I handle the one exception for not cutting content – atom.

## Testing Done
Automated tests for new functionality. Unfortunately there are no existing tests so regression testing is manual.

## Verification Steps
Try adding atom nodes and copy text. Content should contain full atom content.

## Checklist

## Related Issues
https://github.com/ueberdosis/tiptap/issues/5298

### DIFF
--- a/.changeset/atom-text-content-in-full.md
+++ b/.changeset/atom-text-content-in-full.md
@@ -1,0 +1,5 @@
+---
+"@tiptap/core": patch
+---
+
+fix: take atom content in full
--- a/.changeset/atom-text-content-in-full.md
+++ b/.changeset/atom-text-content-in-full.md
@@ -2,4 +2,4 @@
 "@tiptap/core": patch
 ---
 
-fix: take atom content in full
+Make sure that atoms are used in-full without cutting the content. Node size for atoms is 1 which causes text to be cut unexpectedly.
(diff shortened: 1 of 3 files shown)


bdbch:
Looks nice. I guess before that atom nodes had the content slices off because of the expectation of Prosemirror nodes always having a padding of 2?

So an atom node with text content `test` was sliced to `tes`?

@nperez0111 LGTM - what do you think?

artursvonda:
> Looks nice. I guess before that atom nodes had the content slices off because of the expectation of Prosemirror nodes always having a padding of 2?
> 
> So an atom node with text content `test` was sliced to `tes`?
> 
> @nperez0111 LGTM - what do you think?

Yeah, essentially. I don't remember exact size but mismatch between node size and content size caused the issue. I think atom nodes internally are represented with size 1.

artursvonda:
Feedback on this? Could we move this forward perhaps? :)

nperez0111 (approved):

nperez0111:
Atoms are represented as 1 for sure. Yea the change looks correct

PR #5325, opened 2024-07-10
Flip the order of useEditorState's overloads

alii:
So that TypeScript will resolve the non-nullable one first, unless the editor is nullable, in which case the other one will resolve.

Related to discussion in #5161 @nperez0111


Looskie (approved):
LGTM

alii:
Awesome thank you! Would it be possible to get a beta release with these changes? I realised I didn't make a changeset, apologies.

nperez0111:
Ah yea, the changeset was why it didn't release. I'll get a version out for it

nperez0111:
Releasing now. Should be available in a couple of minutes

alii:
Got it! Thank you :)

Issue #5333, opened 2024-07-12
[Bug]: Race condition: the editor crashes if there is a frequent change in its dependency array


yyjlincoln:
### Affected Packages

react

### Version(s)

2.4.0

### Bug Description

When an editor is recreated due to dependency changes in its dependency, for example:

```
const Comp = ({editorOptions}: {editorOptions: any}) => {
  const editor = useEditor(
    {
      extensions: createExtensions(editorOptions),
    },
    [editorOptions],
  );

  return (
    <EditorContent editor={editor} />
  )
}
```

And if the same editorOptions is used somewhere in its parents, the editor will crash due to a potential race condition with:

```
TypeError: Cannot read properties of null (reading 'updateOuterDeco')
    at EditorView.updateStateInner (index.js:5306:34)
    at EditorView.update (index.js:5235:14)
    at EditorView.setProps (index.js:5249:14)
    at Editor.createNodeViews (Editor.ts:313:15)
    at PureEditorContent.init (EditorContent.tsx:83:14)
    at PureEditorContent.componentDidMount (EditorContent.tsx:58:10)
    at commitClassLayoutLifecycles (react-dom-client.development.js:22140:18)
    at commitLayoutEffectOnFiber (react-dom-client.development.js:22312:11)
    at recursivelyTraverseLayoutEffects (react-dom-client.development.js:23856:7)
    at commitLayoutEffectOnFiber (react-dom-client.development.js:22298:9)
    at recursivelyTraverseLayoutEffects (react-dom-client.development.js:23856:7)
... (omitted)
```

I have diagnosed it and here are my assumptions:

`useEditor` creates a new instance of the editor in the initial render. When it is passed to `EditorContent` (which wraps around `PureEditorContent`), on the first mount, it calls:

```
// EditorContent.tsx
componentDidMount() {
   this.init()
}

init() {
  // ...
  editor.createNodeViews()
  // ...
}
```

In `createNodeViews`, we have:

```
// Editor.ts
this.view.setProps({
      nodeViews: this.extensionManager.nodeViews,
})
```

Following the call stack, this eventually lands on (prosemirror):

```
// EditorView from ProseMirror
this.docView.updateOuterDeco(outerDeco);
```

However, if the dependency prop has changed, then:

```
// useEffect cleanup within useEditor.ts (dependency array has extensionOptions)
return () => {
      isMounted = false
      editor.destroy()
}

// Editor.ts
public destroy(): void {
    this.emit('destroy')

[...]

PR #5334, opened 2024-07-12
fix: check that view is not destroyed before creating node views in core Editor

yyjlincoln:
## Changes Overview
Adds a check before creating node views in the core editor so setProps is not called on a destroyed view.

## Implementation Approach
Added checks in `createNodeViews()` in `Editor.ts`.

## Additional Notes
This should be no-op for successful use cases. There should not be any reason to call `this.view.setProps` on a view that has been destroyed, and doing so should either be no-op or results in unexpected crashes. This PR fixes this behaviour.

Related to #5333.

## Checklist
- [ ] I have added tests where applicable. (needs contribution)

### DIFF
--- a/.changeset/sixty-gifts-dream.md
+++ b/.changeset/sixty-gifts-dream.md
@@ -1,0 +1,5 @@
+---
+"@tiptap/core": patch
+---
+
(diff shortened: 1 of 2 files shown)


nperez0111 (approved):

nperez0111:
Thank you for contribution!

PR #5357, opened 2024-07-16
fix(core): configuring extensions should add to the parent's options not replace them

nperez0111:
## Changes Overview
There was a bug where doing a `.configure` on an extension, node or mark would overwrite the extensions options instead of being merged with the default options.

## Testing Done

Added tests

## Checklist


bdbch (approved):
LGTM! 👍

PR #5366, opened 2024-07-18
fix: Address different behavior between `getHTML` and `generateHTML`

baseballyama:
## Changes Overview

Using `generateHTML` function from `@tiptap/html` package generates HTML like `<li data-checked data-type="taskItem">`.
However, in other parts of the code, the expected HTML is `<li data-checked="true" data-type="taskItem">`.
Due to this behavior, the selection state is lost when the HTML obtained using the `generateHTML` function is rendered again.

## Implementation Approach

The value set for the `data-checked` attribute can be either `true` or undefined.

## Testing Done

I manually tested both cases, using `generateHTML` and not using the `generateHTML` function, and confirmed that everything works correctly.

## Verification Steps

You can check out this repository to see how it works

https://github.com/baseballyama/tiptap-tasklist-bug-repl

## Additional Notes

N/A

## Checklist

## Related Issues

N/A

### DIFF
--- a/packages/extension-task-item/src/task-item.ts
+++ b/packages/extension-task-item/src/task-item.ts
@@ -68,7 +68,7 @@ export const TaskItem = Node.create<TaskItemOptions>({
         parseHTML: element => {
           const dataChecked = element.getAttribute('data-checked')
 
-          return dataChecked != null && dataChecked !== 'false'
+          return dataChecked == null || dataChecked === 'true'
         },
         renderHTML: attributes => ({
           'data-checked': attributes.checked,
--- a/packages/extension-task-item/src/task-item.ts
+++ b/packages/extension-task-item/src/task-item.ts
@@ -65,7 +65,11 @@ export const TaskItem = Node.create<TaskItemOptions>({
       checked: {
         default: false,
         keepOnSplit: false,
-        parseHTML: element => element.hasAttribute('data-checked') && element.getAttribute('data-checked') !== 'false',
+        parseHTML: element => {
+          const dataChecked = element.getAttribute('data-checked')
+
+          return dataChecked != null && dataChecked !== 'false'
+        },
         renderHTML: attributes => ({
           'data-checked': attributes.checked,
         }),
--- a/packages/extension-task-item/src/task-item.ts
+++ b/packages/extension-task-item/src/task-item.ts
@@ -65,9 +65,9 @@ export const TaskItem = Node.create<TaskItemOptions>({
       checked: {
         default: false,
         keepOnSplit: false,
-        parseHTML: element => element.getAttribute('data-checked') === 'true',
+        parseHTML: element => element.hasAttribute('data-checked') && element.getAttribute('data-checked') !== 'false',
         renderHTML: attributes => ({
-          'data-checked': attributes.checked ? 'true' : undefined,
+          'data-checked': attributes.checked,
         }),
       },
     }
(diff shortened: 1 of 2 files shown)


packages/extension-task-item/src/task-item.ts
```diff
@@ -67,7 +67,7 @@ export const TaskItem = Node.create<TaskItemOptions>({
         keepOnSplit: false,
         parseHTML: element => element.getAttribute('data-checked') === 'true',
```
nperez0111, line 68: I wonder if it is the `parseHTML` that should be more lenient instead I don't think that it needs to be checked against true, just that it actually exists right?

nperez0111: Yea so my testing shows that when data-checked is on the element (i.e. generateHTML outputted `<li data-checked>...`) that it gives an empty string to say that it was set but has no vlaue.

As opposed to being completely empty which is just null.

So we should be more liberal in our parsing rather than more strict in our output and the result should be the same:

image

baseballyama: Please see my REPL repo. https://github.com/baseballyama/tiptap-tasklist-bug-repl
And actually there is `data-checked='false'` attribute.

image

packages/extension-task-item/src/task-item.ts
```diff
@@ -67,7 +67,7 @@ export const TaskItem = Node.create<TaskItemOptions>({
...
         renderHTML: attributes => ({
-          'data-checked': attributes.checked,
+          'data-checked': attributes.checked ? 'true' : undefined,
```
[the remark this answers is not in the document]
baseballyama: Agreed: How about this? https://github.com/ueberdosis/tiptap/pull/5366/commits/72b887f6d4134b161262104ae96c2a7e9e309cba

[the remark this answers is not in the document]
baseballyama: Oh but I think my repl project doesn't work with this change...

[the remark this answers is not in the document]
baseballyama: I believe this works as expected and reflects our consensus.
https://github.com/ueberdosis/tiptap/pull/5366/commits/a5ebc622c06941f4a7b384b6d2eeeaaef4c57769

packages/extension-task-item/src/task-item.ts
```diff
@@ -65,7 +65,11 @@ export const TaskItem = Node.create<TaskItemOptions>({
       checked: {
         default: false,
         keepOnSplit: false,
-        parseHTML: element => element.getAttribute('data-checked') === 'true',
+        parseHTML: element => {
+          const dataChecked = element.getAttribute('data-checked')
+
+          return dataChecked != null && dataChecked !== 'false'
+        },
```
nperez0111, line 72: I think I would do `element.getAttribute('data-checked') === '' || element.getAttribute('data-checked') === 'true'` To be explicit about what we accept

baseballyama: I don't have a opinion for this. updated: https://github.com/ueberdosis/tiptap/pull/5366/commits/0f4c85d5eda54481a576c97c4df26335a56d4f70

nperez0111 (approved):
Looks good, thanks for the quick turn around

baseballyama:
@nperez0111 I didn't pay attention to the branch and this PR got merged into the main branch. I think the correct branch is develop, right?😅

baseballyama:
Thank you for the backmerge to develop @nperez0111!

nperez0111:
Yea, I saw that too. My mistake, it is on me to handle that.

I will wait to get the release out though, I've been releasing a lot lately and have had several complaints of how many releases have been going out lately.

Issue #5363, opened 2024-07-18
[Bug]: `splitBlock` returns `true` when it was unsuccessful


mgreystone:
### Affected Packages

core, extension-document, extension-paragraph, extension-text

### Version(s)

2.5.4

### Bug Description

The `splitBlock` command returns `true` even when it was unsuccessful.

If we have a `Document` with a content expression that only allows a single node, that node can never be split, as expected. However, the `splitCommand` still returns true even though the command failed. This causes `commands.first` not to attempt the next command in the list.

```ts
editor.commands.first(({ commands }) => [
  () => commands.splitBlock(), // always returns true
  () => commands.setHardBreak(), // we will never reach here, even if `splitBlock` fails
])
```

We can inspect the transaction & assert that the block was not split, even though the command returns true.
```ts
editor.commands.command(({ commands, tr }) => {
  const result = commands.splitBlock() // always true

  // If `splitBlock` was successfull, then `tr.steps` is
  // an array with a single `ReplaceStep`.
  // If `splitBlock` was unsuccesful, then `tr.steps` is
  // an empty array.
  console.log(tr.steps)

  return result
})
```

### Browser Used

Chrome

### Code Example URL

https://codesandbox.io/s/dank-flower-6scjk3?file=/src/index.js

### Expected Behavior

If `splitBlock` does not succeed in splitting a block, then it should return `false`.

### Dependency Updates

- [X] Yes, I've updated all my dependencies.

nperez0111:
It would be helpful to know what branch you are hitting here to see what was incorrect: https://github.com/ueberdosis/tiptap/blob/b941eea6daba09d48a5d18ccc1b9a1d84b2249dd/packages/core/src/commands/splitBlock.ts#L1-L123

mgreystone:
It looks like `canSplit` returns `false` [here](https://github.com/ueberdosis/tiptap/blob/b941eea6daba09d48a5d18ccc1b9a1d84b2249dd/packages/core/src/commands/splitBlock.ts#L84)

The [following conditional](https://github.com/ueberdosis/tiptap/blob/b941eea6daba09d48a5d18ccc1b9a1d84b2249dd/packages/core/src/commands/splitBlock.ts#L86-L90) is false and is skipped.

[...]

PR #5371, opened 2024-07-19
fix: Adjust the `splitBlock` command to return `false` when it was unsuccessful

mgreystone:
## Changes Overview

Fix the `splitBlock` command so that it returns `true` when it successfully split a block & `false` when it does not. Allows for consumers to determine if the command was successful & act accordingly.

## Implementation Approach

`splitBlock` already determined if a split was possible & saved those results to a `can` variable. This change returns the `can` variable, both when `dispatch` is defined & undefined.

## Testing Done

[I altered the `CustomDocument` example](https://github.com/mgreystone/tiptap/commit/3d762be990c08eeae0970b4f1d9c2a0948817154) to only allow a single block by setting `content: heading block`. I then added a custom Enter keyboard shortcut to log if `splitBlock` returned false. Pushing the enter key anywhere in the document will now log "success".

## Verification Steps

* Create a schema where splitting a block is impossible--for example, setting `content: 'block'` on a document. Call `editor.commands.splitBlock()` & ensure it returns `false`.
* Adjust the schema to allow for blocks to be split--for example, setting `content: 'block+'`. Ensure that `editor.commands.splitBlock()` now returns `true`.

## Checklist

## Related Issues
Fixes: https://github.com/ueberdosis/tiptap/issues/5363

### DIFF
--- a/packages/core/src/commands/splitBlock.ts
+++ b/packages/core/src/commands/splitBlock.ts
@@ -61,18 +61,34 @@ export const splitBlock: RawCommands['splitBlock'] = ({ keepMarks = true } = {})
     return false
   }
 
-  if (dispatch) {
-    const atEnd = $to.parentOffset === $to.parent.content.size
+  const atEnd = $to.parentOffset === $to.parent.content.size
 
-    if (selection instanceof TextSelection) {
-      tr.deleteSelection()
-    }
+  if (selection instanceof TextSelection) {
+    tr.deleteSelection()
+  }
 
-    const deflt = $from.depth === 0
-      ? undefined
-      : defaultBlockAt($from.node(-1).contentMatchAt($from.indexAfter(-1)))
+  const deflt = $from.depth === 0
+    ? undefined
+    : defaultBlockAt($from.node(-1).contentMatchAt($from.indexAfter(-1)))
 
-    let types = atEnd && deflt
+  let types = atEnd && deflt
+    ? [
+      {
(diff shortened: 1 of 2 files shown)


packages/core/src/commands/splitBlock.ts
```diff
@@ -61,44 +61,44 @@ export const splitBlock: RawCommands['splitBlock'] = ({ keepMarks = true } = {})
...
-  if (dispatch) {
-    const atEnd = $to.parentOffset === $to.parent.content.size
+  const atEnd = $to.parentOffset === $to.parent.content.size
 
-    if (selection instanceof TextSelection) {
-      tr.deleteSelection()
-    }
+  if (selection instanceof TextSelection) {
+    tr.deleteSelection()
+  }
```
nperez0111, line 68: Sorry this is not correct, it should only ever manipulate the transaction if `dispatch` is true. This would break the behavior of commands being allowed to fail silently

mgreystone: Good call. This made me find another issue, where if the block cannot be split, the selection is erroneously deleted. So i've moved this to only occur when both `dispatch` and `can` are truthy.

Not sure why/if we'd want to `ensureMarks` or `scrollIntoView` if we cannot split the block, but that behavior was already present, & i don't see how it breaks anything?

nperez0111 (approved):

PR #5373, opened 2024-07-22
Vue 3 Perf updates

### LINKED ISSUE #4870: [Bug]: shouldShow broken in BubbleMenu extension from Vue 3 package

nperez0111:
- **revert: "revert: "fix(vue-3): fix editor.state updating too late during a transaction due to reactiveState fixes #4870 (#5252)""**
- **revert: "revert: "fix(vue-3): faster component rendering (#5206)""**
- **fix: backwards compatibility ref exposed methods**
- **fix(vue-3): provide app global (#5354)**
- **fix(vue-3): fix(vue-3): vue's VueNodeViewRenderer should return `null` for `contentDOM` (#3966)**

## Checklist

### DIFF
--- a/.changeset/shy-clouds-smoke.md
+++ b/.changeset/shy-clouds-smoke.md
@@ -1,0 +1,5 @@
+---
+"@tiptap/vue-3": patch
+---
+
+VueNodeViewRenderer should return `null` for `contentDOM` for a non-leaf node with no `NodeViewContent`
--- a/packages/core/src/Editor.ts
+++ b/packages/core/src/Editor.ts
@@ -404,6 +404,11 @@ export class Editor extends EventEmitter<EditorEvents> {
     const state = this.state.apply(transaction)
     const selectionHasChanged = !this.state.selection.eq(state.selection)
 
+    this.emit('beforeTransaction', {
+      editor: this,
+      transaction,
+      nextState: state,
+    })
     this.view.updateState(state)
     this.emit('transaction', {
       editor: this,
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -50,6 +50,7 @@ export interface EditorEvents {
   }
   update: { editor: Editor; transaction: Transaction }
   selectionUpdate: { editor: Editor; transaction: Transaction }
+  beforeTransaction: { editor: Editor; transaction: Transaction, nextState: EditorState }
   transaction: { editor: Editor; transaction: Transaction }
   focus: { editor: Editor; event: FocusEvent; transaction: Transaction }
   blur: { editor: Editor; event: FocusEvent; transaction: Transaction }
--- a/packages/vue-3/src/Editor.ts
+++ b/packages/vue-3/src/Editor.ts
@@ -1,15 +1,13 @@
 import { Editor as CoreEditor, EditorOptions } from '@tiptap/core'
 import { EditorState, Plugin, PluginKey } from '@tiptap/pm/state'
 import {
+  AppContext,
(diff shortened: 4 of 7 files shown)


nperez0111:
Hi @Rirax, So I looked into this further, and found a bug with this implementation that I can't quite figure out how to resolve & I hope you would.

If you go to [this mention preview](https://deploy-preview-5373--tiptap-embed.netlify.app/src/nodes/mention/vue/) and try to add a mention with `@` and then use the arrow down key to move the selection down the list, the selection is not handled and instead the selection is moved within the editor when the custom vue component is supposed to be handling it: https://github.com/ueberdosis/tiptap/blob/aba6657d7026cf59a41b9532b767c8e3058b1b03/demos/src/Nodes/Mention/Vue/MentionList.vue#L46-L63

I did a git bisect & the first commit that had an issue was: 6ad77da961a9307cf8fc1bca3f67a285e78a9a19 which is regarding this PR https://github.com/ueberdosis/tiptap/pull/5206 that you made.

So for now I can't merge this as it is a regression in functionality.

I would very much appreciate your help here as I'm not seeing anything obvious for why events would not be handled by vue-3 properly here. Thanks for your contribution and support in this matter.

nperez0111:
@Rirax  I was able to resolve this by exposing the proxy instead, is this correct? I'm unfamiliar with vue 3, much less it's internals: https://github.com/ueberdosis/tiptap/pull/5373/commits/960236855a01a52189dc65539103e8965381cef4

Rirax:
Hi @nperez0111 ,
Thank you for your message.

I did my tests in vue-3 `Composition API` [where it is necessary to expose the methods you want to make public](https://vuejs.org/api/sfc-script-setup.html#defineexpose).

In the shared example it's vue-3 `Option API` and so `exposed` doesn't exist.

I've just tested in my case (Composition API),`this.renderedComponent.vNode?.component?.proxy` doesn't do the trick

Rirax:
@nperez0111 

In my case `renderedComponent.vNode.component.proxy` exists but doesn't contain the required methods.

That way it works for me:
`  get ref() {
    // Composition API
    if (this.renderedComponent.vNode.component.exposed) {
      return this.renderedComponent.vNode.component.exposed;
    }

    // Option API
    return this.renderedComponent.vNode.component.proxy;
  }`

PR #5385, opened 2024-07-24
Update prosemirror-view & prosemirror-tables to most recent version

bdbch:
## Changes Overview
Updated prosemirror-views & prosemirror-tables to most recent version

## Implementation Approach
Actually I was looking for the prosemirror-model update because there was an XSS issue with version 1.22.0 and below (see https://discuss.prosemirror.net/t/heads-up-xss-risk-in-domserializer/6572) - but it seems we already updated it.

While we're at it why not just bump the other prosemirror versions too.

## Testing Done
I ran the tests locally + went through the demos manually to see if any functionality is broken (specially tables).

## Verification Steps
See above

## Checklist

### DIFF
--- a/.changeset/cool-steaks-clean.md
+++ b/.changeset/cool-steaks-clean.md
@@ -1,0 +1,14 @@
+---
+"@tiptap/pm": patch
+---
+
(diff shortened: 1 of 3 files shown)


bdbch:
Looking into the typescript issues

nperez0111 (approved):

Issue #4929, opened 2024-02-28
[Bug]: `generateJSON` can return links `href` as number


ChidVanid:
### Which packages did you experience the bug in?

extension-link, html

### What Tiptap version are you using?

tiptap: 2.2.4, extension-link: 2.2.4, html: 2.2.4

### What’s the bug you are facing?

We have historically links like `<p><a href="123">Test</a></p>` in our data.
Before inserting this content in our tiptap editor we use `generateJSON` from the html package. The JSON this function returns is later used in the editors `setContent` command.
After `setContent` we get the following error: `HTMLAttributes.href.startsWith is not a function`.
After some inspection i could pinpoint it to the link extensions `renderHTML` function. The if statement there which checks if the href starts with `javascript:` fails since the href arrives in this function as a `Number` instead of a `String`.
For whatever reason `generateJSON` parses values like `123` as Numbers.

To fix this we extended the link extension and replaced the `renderHTML` function to handle actual numbers but I guess it would be the better way to fix this in the `generateJSON` function?

### What browser are you using?

Chrome

### Code example

https://codesandbox.io/p/sandbox/tender-rhodes-jhhpgg

### What did you expect to happen?

`href` to be parsed as String.

### Did you update your dependencies?

- [X] Yes, I’ve updated my dependencies to use the latest version of all packages.

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

denego:
Got the same issue, had to workaround like this:
```
import { mergeAttributes } from '@tiptap/core';
import LinkBase from '@tiptap/extension-link';

export const Link = LinkBase
  .extend({
    renderHTML({ HTMLAttributes }) {
      // force HTMLAttributes.href to be a string
      return ['a', mergeAttributes(this.options.HTMLAttributes, {
        ...HTMLAttributes, href: String(HTMLAttributes.href),
      }), 0];
    },
  });
```

yurtsiv:
To prevent numbers from entering your doc in the first place, you can do the following:

```typescript
const Link = LinkBase.extend({
  addAttributes() {
    return {
      ...(this.parent?.() ?? {}),
      href: {
        default: null,
        parseHTML(element) {
          return element.getAttribute('href');
        },
      }
    }
  },
})
```

[...]

PR #5391, opened 2024-07-24
fix(link): Do not convert link `href` to number or boolean

yurtsiv:
## Changes Overview

Link `href` attribute [doesn't have parseHTML](https://github.com/ueberdosis/tiptap/blob/develop/packages/extension-link/src/link.ts#L165) provided, so TipTap uses a default implementation that [converts strings to numbers/booleans](https://github.com/ueberdosis/tiptap/blob/341b01d384995aa429593a0842b817516f4111f6/packages/core/src/helpers/injectExtensionAttributesToParseRule.ts#L32). This leads to crashes in cases like:

```typescript
editor.commands.setContent('<a href="123">...</a>')
editor.commands.setContent('<a href="true">...</a>')

// TypeError: uri.replace is not a function
//   at isAllowedUri
```
[Render function](https://github.com/ueberdosis/tiptap/blob/develop/packages/extension-link/src/link.ts#L197) assumes href is `string | null`.

## Implementation Approach

Provide a custom `parseHTML` implementation.

## Checklist

## Related Issues

https://github.com/ueberdosis/tiptap/issues/4929


nperez0111 (approved):
This looks good to me, thanks for your contribution

PR #5395, opened 2024-07-26
fix(extension-suggestion): get client rect by anchor.

YonLJ:
## Changes Overview
Previously, clientRect was only obtained through decorationNode. If decorationNode could not be obtained, the value of clientRect was assigned to null. This commit adds a new method to obtain clientRect.

## Implementation Approach
If decorationNode cannot be obtained, the coordinates are obtained according to the editor's cursor position, and the cursor position is used to generate a DOMRect.

## Testing Done
In scenarios where Chinese IME is used, the value of clientRect is almost always null, which causes the "suggestion" to not render. This commit fixes this issue.

## Verification Steps
Just type @ when using the Chinese input method and the result will be verified.

## Additional Notes

before

after

## Checklist


YonLJ:
> Just to clarify a little bit more - which decoration node are we talking about here? For suggestions I usually used the `posToDOMRect` util function that is included in `@tiptap/core`, `@tiptap/react` and `@tiptap/vue-*` to get the clientRect by editor position (for example the cursor position)

The decoration node discussed here is the HTML tag that wraps the "@" character (like this `<span data-decoration-id="id_xxxxxxx" class="suggestion">@</span>`), which determines the display position of the "suggestion" component. If it work properly, the html in the editor will look like this
`<p>What do you all think about the new <span class="mention" data-type="mention" data-id="Winona Ryder" contenteditable="false">@Winona Ryder</span> movie? <span data-decoration-id="id_3457874414" class="suggestion">@</span></p>`

codeingforcoffee:
I have same problem, hope this can fix

bdbch:
@YonLJ could you rebase this to the current develop branch as we released V3, deprecating the next branch & would need to update most of our PR's anyway? Thanks a lot!

YonLD:
> @YonLJ could you rebase this to the current develop branch as we released V3, deprecating the next branch & would need to update most of our PR's anyway? Thanks a lot!

@bdbch https://github.com/ueberdosis/tiptap/pull/6884

PR #5397, opened 2024-07-26
fix(vue-3): Correctly set editor's appContext.provide to forward full inject chain

### LINKED ISSUE #5394: [Bug]: Fails to resolve a value via `provide/inject` in Vue node view components
### Affected Packages

@tiptap/vue-3

### Version(s)

2.5.6

### Bug Description

Attempting to resolve a value via `provide/inject` in Vue node view components produces `undefined` result. 

[Vue will keep parent's provides on object's prototype if current component itself has its own `provide` call](https://github.com/vuejs/core/blob/91112520427ff55941a1c759d7d60a0811ff4a61/packages/runtime-core/src/apiInject.ts#L19-L28). It's achieved by using `Object.create()`. When `EditorContent` constructs `appContext` it uses spread to forward parent chain provides. Spread operator won't copy prototype resulting in missing provided parent values.
https://github.com/ueberdosis/tiptap/blob/b0124717559071729a365d67b9d4747c5fb2a1cf/packages/vue-3/src/EditorContent.ts#L47-L54

Related to #5323, #5373.

`2.4.0` is not affected.

### Browser Used

Edge

### Code Example URL

https://stackblitz.com/edit/vitejs-vite-f212pf?file=src%2FApp.vue

### Expected Behavior

All `provide/inject` values should be properly available in editor component node views.

### Dependency Updates

romansp:
Fixes #5394.

## Changes Overview
Replace usage of `Object.assign` and set `appContext.provides` as-is from current instance `provides`. 

## Implementation Approach
Vue internally uses prototype chain to preserve injects across the entire component chain. Thus should avoid `Object.assign` or spread operator as it won't copy the prototype. All correct provides will be already present on `instance.provides`.

## Testing Done
Added new example with cypress test coverage.

## Checklist

### DIFF
--- a/packages/vue-3/src/EditorContent.ts
+++ b/packages/vue-3/src/EditorContent.ts
@@ -46,11 +46,10 @@ export const EditorContent = defineComponent({
           if (instance) {
             editor.appContext = {
               ...instance.appContext,
-              provides: Object.assign(
-                instance.appContext.provides,
-                // @ts-ignore
-                instance.provides,
-              ),
+              // Vue internally uses prototype chain to forward/shadow injects across the entire component chain
+              // so don't use object spread operator or 'Object.assign' and just set `provides` as is on editor's appContext
+              // @ts-expect-error forward instance's 'provides' into appContext
+              provides: instance.provides,
             }
           }
 
--- a/.changeset/odd-paws-tap.md
+++ b/.changeset/odd-paws-tap.md
@@ -1,0 +1,5 @@
+---
+"@tiptap/vue-3": patch
+---
+
+Correctly set editor's appContext.provide to forward full inject chain
--- a/demos/setup/vue.ts
+++ b/demos/setup/vue.ts
@@ -14,7 +14,12 @@ export default function init(name: string, source: any) {
 
   import(`../src/${demoCategory}/${demoName}/${frameworkName}/index.vue`)
     .then(module => {
-      createApp(module.default).mount('#app')
+      const app = createApp(module.default)
+
+      if (typeof module.configureApp === 'function') {
+        module.configureApp(app)
+      }
+      app.mount('#app')
       debug()
     })
 }
--- a/demos/src/Examples/InteractivityComponentProvideInject/Vue/Component.vue
+++ b/demos/src/Examples/InteractivityComponentProvideInject/Vue/Component.vue
@@ -1,0 +1,50 @@
+<template>
+  <node-view-wrapper class="vue-component">
+    <label>Vue Component</label>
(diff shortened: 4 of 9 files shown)


romansp:
> Each example has associated tests
> 
> https://github.com/ueberdosis/tiptap/blob/eaee9c7177ccd86796b8c051b525acf70fb72034/demos/src/Examples/Savvy/Vue/index.spec.js#L1
> 
> but I don't see an existing one that would be "natural" for this. Maybe just a new folder in `Examples/Integrations/Vue` that copies one of the other instances maybe?

Let's continue discussion in this PR @nperez0111 if you don't mind. I can add an example and tests as you said. But wouldn't it get picked up by the docs automatically? I'm sure we want to avoid that.

nperez0111:
> Let's continue discussion in this PR @nperez0111 if you don't mind. I can add an example and tests as you said. But wouldn't it get picked up by the docs automatically? I'm sure we want to avoid that.

If it is a new example, it wouldn't get picked up by the docs. The docs have to import the example, so it is fine to include new ones here

romansp:
Added test coverage, this is ready to review. Had to slightly adjust `demos/setup/vue.ts` to hook into app creation routine. Let me know if that's not ideal.

packages/vue-3/src/EditorContent.ts
```diff
@@ -46,11 +46,10 @@ export const EditorContent = defineComponent({
...
-              provides: Object.assign(
-                instance.appContext.provides,
-                // @ts-ignore
-                instance.provides,
-              ),
+              // Vue internally uses prototype chain to forward/shadow injects across the entire component chain
+              // so don't use object spread operator or 'Object.assign' and just set `provides` as is on editor's appContext
+              // @ts-expect-error forward instance's 'provides' into appContext
+              provides: instance.provides,
             }
```
nperez0111, line 53: Just to make sure, does this work for both the options API & composition API. I'm not familiar enough with vue to know

nperez0111 (approved):
Really appreciate the additional tests & example. Just the one question & we can get this into the next release

packages/vue-3/src/EditorContent.ts
romansp: Yes this will work for both Options and Composition API. Tbh I wrote tests in Composition first and then converted to Options for consistency with other examples. 

Also no need to worry about composition vs options at all: Options API is literally implemented on top of Composition primitives.

nperez0111: Thanks for the confirmation!

PR #5398, opened 2024-07-26
Remove tippy.js and replace with Floating UI

bdbch:
## Changes Overview
This PR completely removes tippy.js from all Tiptap packages and replaces the implementation with [Floating UI](https://www.floating-ui.com).

Tippy was already outdated and often caused rendering issues for example with React or conditional rendering. The new implement Floating UI implementation now can be conditionally rendered, customized and adjusted.

## Implementation Approach
Instead of letting Tippy handle the element creation and mounting, we now do this on our own and use Floating UI to compute the position of the floating element using [virtual elements](https://floating-ui.com/docs/virtual-elements).

The positions are calculated from the cursor or selection position.

Instead of `tippyOptions` we now pass down an `option` object which can be used to control the Floating UI middlewares like `shift`, `flip` or `offset`.

## Testing Done
I updated all tests that were previously using tippy and made sure all run successfully.

## Verification Steps

1. Try out this change manually on our demos locally
2. Check the test runs

## Additional Notes

Breaking change incoming. 😬

## Checklist
- [x] Add documentation in https://github.com/ueberdosis/tiptap-docs (See https://github.com/ueberdosis/tiptap-docs/pull/16)

## Related Issues

- #3170 (now we bundle Floating UI into the project - we could optionally use peerDependencies here, @nperez0111, @svenadlung  what do you think?)

### DIFF
--- a/demos/src/Examples/Community/React/suggestion.js
+++ b/demos/src/Examples/Community/React/suggestion.js
@@ -1,7 +1,28 @@
-import { ReactRenderer } from '@tiptap/react'
-import tippy from 'tippy.js'
+import {
+  computePosition,
+  flip,
+  shift,
+} from '@floating-ui/dom'
+import { posToDOMRect, ReactRenderer } from '@tiptap/react'
 
 import { MentionList } from './MentionList.jsx'
+
+const updatePosition = (editor, element) => {
+  const virtualElement = {
+    getBoundingClientRect: () => posToDOMRect(editor.view, editor.state.selection.from, editor.state.selection.to),
+  }
+
+  computePosition(virtualElement, element, {
+    placement: 'bottom-start',
+    strategy: 'absolute',
+    middleware: [shift(), flip()],
+  }).then(({ x, y, strategy }) => {
+    element.style.width = 'max-content'
+    element.style.position = strategy
+    element.style.left = `${x}px`
+    element.style.top = `${y}px`
+  })
+}
 
 export default {
   items: ({ query }) => {
@@ -12,7 +33,6 @@ export default {
 
   render: () => {
     let reactRenderer
-    let popup
 
     return {
       onStart: props => {
@@ -26,15 +46,11 @@ export default {
           editor: props.editor,
         })
 
-        popup = tippy('body', {
-          getReferenceClientRect: props.clientRect,
-          appendTo: () => document.body,
-          content: reactRenderer.element,
-          showOnCreate: true,
-          interactive: true,
-          trigger: 'manual',
-          placement: 'bottom-start',
-        })
+        reactRenderer.element.style.position = 'absolute'
+
+        document.body.appendChild(reactRenderer.element)
+
+        updatePosition(props.editor, reactRenderer.element)
       },
 
       onUpdate(props) {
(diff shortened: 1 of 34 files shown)


bdbch:
Here are some links you can try this on:

https://deploy-preview-5398--tiptap-embed.netlify.app/src/examples/menus/vue/
https://deploy-preview-5398--tiptap-embed.netlify.app/src/examples/menus/react/
https://deploy-preview-5398--tiptap-embed.netlify.app/src/examples/community/react/
https://deploy-preview-5398--tiptap-embed.netlify.app/src/examples/community/vue/

packages/extension-floating-menu/package.json
```diff
@@ -29,16 +29,15 @@
     "dist"
   ],
   "devDependencies": {
+    "@floating-ui/dom": "^1.6.0",
     "@tiptap/core": "^2.5.7",
     "@tiptap/pm": "^2.5.7"
   },
   "peerDependencies": {
+    "@floating-ui/dom": "^1.6.0",
```
nperez0111, line 37: This is a dep not peer dep though? You call its functions so it has to be included, no?

demos/src/Nodes/Mention/Vue/suggestion.js
```diff
@@ -63,8 +78,8 @@ export default {
       },
 
       onExit() {
-        popup[0].destroy()
         component.destroy()
+        component.element.remove()
```
nperez0111, line 82: I find this a bit strange, why did we not have to do this before? Shouldn't destroying the component also unmount the element? Should we fix this somehow?

nperez0111 (approved):
This looks great!

packages/extension-floating-menu/package.json
rfgamaral: This is a library that consumers of Tiptap might be using themselves in their apps for other purposes, so in a way, I think it makes sense to be a peer dependency. However, maybe a better solution would be `@tiptap/floating`, similar to `@tiptap/pm`?

nperez0111: It seems I had a misunderstanding of what peer deps do. My mistake it does make sense to have it as a peer dep. Though the specified range should be much wider probably just `^1` or whenever the functionality that we use in the package was introduced. This would conflict less with library consumer's versions.

demos/src/Nodes/Mention/Vue/suggestion.js
bdbch: Nice find - yeah we don't need that one.

Nantris:
@bdbch it seems tippy is still included as a dependency here: https://github.com/ueberdosis/tiptap/blob/49402f61045c36595d1bb5c498773fecb94607e6/packages/extension-bubble-menu/package.json#L32

nperez0111:
> @bdbch it seems tippy is still included as a dependency here:


This only applied to the next version (v3), not version 2 https://github.com/ueberdosis/tiptap/blob/f731cd977383118aa4e0f53e30480638678eb41a/packages/extension-bubble-menu/package.json#L44

I'm not sure why this is saying that it went into develop, it should not have as it was a breaking change

demos/src/Examples/Community/React/suggestion.js
```diff
@@ -26,15 +46,11 @@ export default {
...
-          appendTo: () => document.body,
-          content: reactRenderer.element,
-          showOnCreate: true,
-          interactive: true,
-          trigger: 'manual',
-          placement: 'bottom-start',
-        })
+        reactRenderer.element.style.position = 'absolute'
+
+        document.body.appendChild(reactRenderer.element)
```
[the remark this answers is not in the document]
bdbch: @michelmar sorry for the late answer. Could you create an issue and link this discussion so I can pick it up for #6351 ?

PR #5414, opened 2024-07-29
fix(react): always cleanup old editor instances

nperez0111:
## Changes Overview
React in strict mode can re-render the editor very quickly without unmounting, this makes it so that there cannot be two initalized editor instances active at the same time.

This was causing issues with the collaboration cursor which would infinitely send awareness updates while the editor was focused
## Checklist

### DIFF
--- a/.changeset/red-suns-allow.md
+++ b/.changeset/red-suns-allow.md
@@ -1,0 +1,5 @@
+---
+"@tiptap/react": patch
+---
+
+Resolve a bug an editor could be instantiated but not destroyed. This was causing issues with multiple instances of plugins still being active and interfering with each other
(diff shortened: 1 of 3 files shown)


fuadsaud:
Hi @nperez0111! We've recently bumped tiptap to the latest version in our app and we have observed a behavior where the editor options are not updated even when dependencies change. After looking at recent changes to Tiptap, we believe that this behavior might be related to this PR. It seems that the `mostRecentOptions` ref never gets set to the updated options passed to the hook.

nperez0111:
Yep @fuadsaud  would definitely be caused by that, I'd suggest a downgrade to 2.5.7 if you can (or just wait until my fix for this lands to upgrade)

PR #5421, opened 2024-07-30
chore: Use proper types for `configure` methods (Extension/Mark/Node)

rfgamaral:
## Changes Overview

Our [Typist](https://typist.doist.dev/) package (based on Tiptap) is still using `v2.3.0` for now, and while attempting a quick upgrade to the latest version, some type checks started to fail ([ref](https://github.com/Doist/typist/actions/runs/10148455201/job/28061425917?pr=773)).

One of the issues is that calling `Extension.create().configure()` returns `Node<any, any>`, whereas before it would return `Node<Options, Storage>`. This PR fixes that by making sure that when `.configure()` is called for an `Extension`, `Node`, or `Mark`, the correct type is returned.

Would love it if this could be reviewed, merged, and release soon, so that we can upgrade our Tiptap version.

## Testing Done

Built Tiptap locally and checked the generated `.d.ts` files, the expected types were there.

## Verification Steps

Same as above. Build Tiptap, and check that the `.d.ts` files are generated with the currect types for the `.configure()` method.

## Checklist

### DIFF
--- a/packages/core/src/Extension.ts
+++ b/packages/core/src/Extension.ts
@@ -457,7 +457,7 @@ export class Extension<Options = any, Storage = any> {
   configure(options: Partial<Options> = {}) {
     // return a new instance so we can use the same extension
     // with different calls of `configure`
-    const extension = this.extend({
+    const extension = this.extend<Options, Storage>({
       ...this.config,
       addOptions: () => {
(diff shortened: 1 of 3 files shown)


nperez0111 (approved):
Good catch. Type inference let us down here.

bdbch (approved):

PR #5426, opened 2024-07-31
fix: allow task items to be parsed when only having `<li data-checked` instead of only when `<li data-checked="true"` (re-fix of #5366)

baseballyama:
## Changes Overview

When I modified the code in response to [the comments](https://github.com/ueberdosis/tiptap/pull/5366#discussion_r1682594145) in PR #5366, somehow I made the wrong modification. I am very sorry.🙏🙏🙏 @nperez0111 

This PR will be correctly amended according to this comment.
I have confirmed that it works correctly with both the REPL created in #5366 and my own product code.

I will take care to avoid similar carelessness.🙏

## Testing Done
I tested it by manual. 

## Checklist

## Related Issues
https://github.com/ueberdosis/tiptap/pull/5366

### DIFF
--- a/packages/extension-task-item/src/task-item.ts
+++ b/packages/extension-task-item/src/task-item.ts
@@ -68,7 +68,7 @@ export const TaskItem = Node.create<TaskItemOptions>({
         parseHTML: element => {
           const dataChecked = element.getAttribute('data-checked')
 
-          return dataChecked == null || dataChecked === 'true'
+          return dataChecked === '' || dataChecked === 'true'
         },
         renderHTML: attributes => ({
           'data-checked': attributes.checked,
--- a/.changeset/wise-beers-reflect.md
+++ b/.changeset/wise-beers-reflect.md
@@ -1,0 +1,5 @@
+---
+"@tiptap/extension-task-item": patch
+---
+
+allow task items to be parsed when only having `<li data-checked` instead of only when `<li data-checked="true"` (re-fix of #5366)


baseballyama:
@bdbch 

Did you force pushed some commit to https://github.com/baseballyama/tiptap/commits/develop/ ? (This is my forked repo)
image

Maybe did you intend to push it to [ueberdosis:develop](https://github.com/ueberdosis/tiptap/tree/develop)?

packages/extension-task-item/src/task-item.ts
```diff
@@ -68,7 +68,7 @@ export const TaskItem = Node.create<TaskItemOptions>({
         parseHTML: element => {
           const dataChecked = element.getAttribute('data-checked')
 
-          return dataChecked == null || dataChecked === 'true'
```
bdbch, line 71: Should we still keep the null check here?

```ts
return dataChecked === null || dataChecked === '' || dataChecked === 'true'
```

bdbch (commented):
LGTM - just not sure if we need the null check or not.

bdbch:
@baseballyama I force pushed on our `develop` to move the commit you mentioned over to `next`.

You probably have to rebase your branch to the current develop here.

packages/extension-task-item/src/task-item.ts
baseballyama: No I believe we shouldn't.

Case1: `<li>`, `dataChecked` will be `null`,
Case2: `<li data-checked>`, `dataChecked` will be `''`,
Case3: `<li data-checked="true">`, `dataChecked` will be `'true'`,
Case3: `<li data-checked="false">`, `dataChecked` will be `'false'`,

Case1 and case4, `checked` attribute should be recognize as `false`,
Case 2 and case3, `checked` attribute should be recognize as `true`,

baseballyama:
> I force pushed on our develop to move the commit you mentioned over to next.
> You probably have to rebase your branch to the current develop here.

Understood.
I think current diff is looks fine.

https://github.com/ueberdosis/tiptap/compare/develop...baseballyama:fix/5366

bdbch (approved):

PR #5437, opened 2024-08-03
WIP: Decorations API

bdbch:
image

- **Decorations API**
    - Positions
        - Attach Decorations to specific Nodes
        - Attach Decorations to the cursor
        - Manually set positions of a decoration
    - We should use a DecorationManager that
        - Handles the state for decorations
        - Communicates with a DecorationPlugin to handle creation and rendering of Prosemirror Decorations
    - Decoration Plugin
        - Syncs it’s state with the decoration manager state
        - uses the attached positions of decorations to determine what type of decoration is needed
    - DecorationManager.create => editor.decorations.create
        - The create api is similar to how we create extensions
            - render: handles rendering and allows return of html or react / vue components
            - binding events, f.e. onUpdate, onTransaction, etc.
    - TiptapDecoration
        - attachTo(x)
            - Can take a position, a range or a node
            - The type of decoration is inferred from this attachTo value

### DIFF
--- a/demos/src/Examples/Default/React/index.jsx
+++ b/demos/src/Examples/Default/React/index.jsx
@@ -1,10 +1,10 @@
 import './styles.scss'
 
 import ListItem from '@tiptap/extension-list-item'
-import { Color , TextStyle } from '@tiptap/extension-text-style'
-import { EditorProvider, useCurrentEditor, useEditorState } from '@tiptap/react'
+import { Color, TextStyle } from '@tiptap/extension-text-style'
+import { EditorProvider, InlineDecoration, useCurrentEditor, useEditorState, WidgetDecoration } from '@tiptap/react'
 import StarterKit from '@tiptap/starter-kit'
-import React from 'react'
+import React, { useEffect } from 'react'
 
 const MenuBar = () => {
   const { editor } = useCurrentEditor()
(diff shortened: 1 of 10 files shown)


Gregoor:
I'm not sure if this helps, so here goes nothing: I'm currently using `addProseMirrorPlugins` to create inline decorations, but I'm struggling with having them be dynamic. This is what I'm doing:
<details><summary>View Code</summary>

```tsx
const buildTermHighlighter = (name: string) =>
  Extension.create<{
    terms: string[];
    className: string;
  }>({
    name: `termHighlighter-${name}`,
    addProseMirrorPlugins() {
      const findTerms = (doc: Node): DecorationSet => {
        const decorations: Decoration[] = [];
        for (let i = 0; i < doc.nodeSize - 2; i++) {
          for (const term of this.options.terms) {
            const end = Math.min(i + term.length, doc.nodeSize - 2);
            if (doc.textBetween(i, end, " ", " ") === term) {
              decorations.push(
                Decoration.inline(i, end, { class: this.options.className }),
              );
            }
          }
        }
        return DecorationSet.create(doc, decorations);
      };
      return [
        new Plugin({
          key: new PluginKey("termHighlighter"),
          state: {
            init: (_, { doc }) => findTerms(doc),
            apply: (transaction, oldState) =>
              transaction.docChanged ? findTerms(transaction.doc) : oldState,
          },
          props: {
            decorations(state) {
              return this.getState(state);
            },
          },
        }),
      ];
    },
  });
```

</details>

Unfortunately there's no update when `terms` change, so currently I'm literally react `key=`ing the editor component with the terms list to have it reload. So what I'm looking forward to with a tiptap-native Decorations API is to be able to do dynamic state-driven decoration changes.

spaceemotion:
@Gregoor here's how I solved it:

1.  Dispatch a signal to the editor (by using an effect or something, you should know when an update is needed):

    ```ts
    editor.view.dispatch(
      editor.state.tr.setMeta("updateTermHighlights", true)
    );
    ```

2.  Listen to the meta data in your plugin:
  
    ```ts
    const plugin = new Plugin({
      key: pluginKey,
      state: {
        init(_, { doc }) {
          return generateDecorations(doc, codex.value);
        },
        apply(transaction, oldState) {
          const needsUpdate = (
            transaction.docChanged ||
            transaction.getMeta("updateTermHighlights") === true
          );
  
          return needsUpdate
            ? generateDecorations(transaction.doc, codex.value)
            : oldState;
        },
      },
      props: {
        decorations(state) {
          return this.getState(state);
        },
      },
    });
    ```

Issue #5432, opened 2024-08-01
[Bug]: useEditor always creates at least 2 editors on mount


schontz:
### Affected Packages

react

### Version(s)

develop

### Bug Description

The editor is [destroyed and re-created whenever deps changes](https://github.com/ueberdosis/tiptap/blob/develop/packages/react/src/useEditor.ts#L155-L162). But this `useEffect` will run once on mount to start, before `deps` has actually changed. If `deps` are supplied, then the editor created in `useState` will get destroyed _immediately_ in `useEffect` and get re-created.

It may be generally useful to have a hook that only runs when things change, e.g.:

```javascript
function useChangeEffect(effect, deps) {
  const isFirstTime = useRef(true);
  useEffect(() => {
    if (isFirstTime.current) isFirstTime.current = false;
    else return effect();
  }, deps);
}
```

Then you can just replace [this useEffect](https://github.com/ueberdosis/tiptap/blob/develop/packages/react/src/useEditor.ts#L123) with `useChangeEffect`.

On a related note, [applying options instead of re-creating an editor is useless code](https://github.com/ueberdosis/tiptap/blob/develop/packages/react/src/useEditor.ts#L147-L153), because the effect only runs when `deps` change, not when `options` change. That means updated `options` are never injected into the editor. Furthermore, [mostRecentOptions](https://github.com/ueberdosis/tiptap/blob/develop/packages/react/src/useEditor.ts#L80) is a misnomer, because it is not actually updated on each render, so it is effectively useless.

### Browser Used

Chrome

### Expected Behavior

Editor is only created once if deps do not change.

### Dependency Updates

- [X] Yes, I've updated all my dependencies.

nperez0111:
Thanks for bringing it up, I was a bit too focused on another bug. I'll consider your some of your suggestions.

schontz:
Looking at `useEditor` further, it doesn't even look like [deleteUnusedEditor](https://github.com/ueberdosis/tiptap/blob/develop/packages/react/src/useEditor.ts#L124-L137) will ever work, because `mostRecentEditor.current` will always point to the value of `editor`. 

1. Create `editor` via `useState`
2. `useEffect` will [destroy right away](https://github.com/ueberdosis/tiptap/blob/develop/packages/react/src/useEditor.ts#L124-L137) and re-create on mount (this bug)
3. If deps changes, `deleteUnusedEditor` will run again, but the editor is current 
[...]

PR #5445, opened 2024-08-05
fix(react): optimize `useEditor` and `useEditorState` to reduce number of instances created while being performant #5432

nperez0111:
## Changes Overview
This addresses an admittedly minor problem where sometimes the editor would reinitialize itself when it did not have to. This now can reuse editor instances that have been created before & re-apply options continously
## Checklist

### DIFF
--- a/.changeset/smooth-rice-obey.md
+++ b/.changeset/smooth-rice-obey.md
@@ -1,0 +1,12 @@
+---
+"@tiptap/react": patch
+---
+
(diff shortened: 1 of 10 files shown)


schontz:
I like this approach a lot as it depends on the react hooks lifecycle as little as possible.

We were seeing memory leaks in our application because the replaced editor wasn’t destroyed, fwiw.

bdbch (approved):
Looks good to me

Issue #5435, opened 2024-08-02
[Bug]: <col> element missing min-width (or width) style when column hasn't been resized


Ragnar-Oock:
### Affected Packages

@tiptap/extension-table, @tiptap/extension-table-cell, @tiptap/extension-table-header, @tiptap/extension-table-row

### Version(s)

2.5.8

### Bug Description

The generated `<col>` elements don't have a width or min-width set on them unless the corresponding column is manually resized. For columns with wrapping text that can lead to layout jump when other column are resized (you can try resizing an empty column in the repro to provoke it) and columns can end up with a width smaller than the configured `cellMinWidth` (both in the editor and in the HTML output)

While the "in editor" `<col>` are controled by `prosemirror-table` (I will open an issue there and link it to this one later), the output is handled by tiptap and it can easily be fixed by changing this line :
https://github.com/ueberdosis/tiptap/blob/84ebd511d29db566e66e8f9b617179cf19ed2531/packages/extension-table/src/utilities/createColGroup.ts#L40
into :
```ts
cols.push([
  'col', 
  { style: hasWidth ? `width: ${ hasWidth }px` : `min-width: ${ cellMinWidth }px` }
]);
```

### Browser Used

Firefox

### Code Example URL

[...]

PR #5464, opened 2024-08-09
fix(table): set min-width to <col>

Ragnar-Oock:
## Changes Overview
Add `min-width` to the style attribute of `<col>` elements when `width` is not available (the user didn't resize column)

## Implementation Approach
Added a helper to generate the style of the `<col>` elements from the stored column width and the configured `cellMinWidth`

## Testing Done
- Updated tests for `editor.getHTML` in Table related suites
- Added a tests case to check if `<col>` elements were generated with the appropriate style attribute

## Verification Steps
- Open the table demo
- remove the `width: 100%` set on the table
- create an empty column (remove existing content or insert a column)

=> The empty columns (or those with not enough content to push them above the minimum) respect the configured cellMinWidth

## Additional Notes
I took the liberty of fixing up the types in the `TableView.ts` file after removing the `// @ts-nocheck` present at the top of the file. I have no objection to reverting the type changes if they are deemed out of scope for the PR.

## Checklist

## Related Issues
#5435


packages/extension-table/src/TableView.ts
```diff
@@ -1,49 +1,55 @@
...
-        colgroup.appendChild(document.createElement('col')).style.width = cssWidth
-      } else {
-        if (nextDOM.style.width !== cssWidth) {
-          nextDOM.style.width = cssWidth
+        if (!hasWidth) {
+          fixedWidth = false
         }
 
-        nextDOM = nextDOM.nextSibling
+        if (nextDOM === null) {
```
nperez0111, line 33: Sorry for the nit, would prefer to keep this as
```suggestion
        if (!nextDOM) {
```
Just in case it is undefined, and to reduce differentiation from the prior behavior

packages/extension-table/src/utilities/colStyle.ts
```diff
@@ -0,0 +1,14 @@
...
+  }
+
+  // set the minimum with on the column if it has no stored width
+  return ['min-width', `${minWidth}px`]
+
+}
+
+export function getColStyle(minWidth: number, width: number): string {
+  return getColStyleDeclaration(minWidth, width).join(': ')
+}
```
nperez0111, line 14: This is abstracting over a couple of if statements, I'm not sure it warrants the complexity it adds.

I would honestly prefer for their to be a single function here, either stick to the string version or the array version, both just adds complexity and indirection.

nperez0111 (approved):
Other than the slight nits, this looks look to be a high quality PR. Thanks for your contribution

packages/extension-table/src/utilities/colStyle.ts
Ragnar-Oock: I extracted them because the logic was originally copy-pasted in both the create and the update code, I can remove the `getColStyle` function and do the concatenation in the table creation function directly instead of having a function in this file

Ragnar-Oock:
I'll commit the changes when I get the time

Ragnar-Oock:
*I forgot to run the tests and I had fuck up something in my code*
something else came up with the tests but I don't think it's related to my code, the test suite for `Extensions/History/React` failed one test when running the full test suite and up to 4 when running them via the UI. Some other tests failed when run with `npm run test` but succeeded when running them manually via `npm run test:open`

Ragnar-Oock:
Is there anything I can do to get this PR merged ? It would allow me to remove some rather unsavory code from my codebase (i.e. copy-paste of the files changed by the PR)

nperez0111:
Hi @Ragnar-Oock, I hadn't realized this was ready yet. It seemed to me that some of your tests you wanted fix: https://github.com/ueberdosis/tiptap/pull/5464#issuecomment-2289724335

nperez0111:
Thanks for your contribution.

PR #5469, opened 2024-08-11
feat(table): make all table packages be in one package, add a new TableKit extension to configure them

nperez0111:
## Changes Overview
This represents a departure from our previous packaging approach. Instead of having a package for every single possible functionality that goes into the editor, we can reduce the number of packages by grouping them by functionality.
We believe that Tiptap's strength comes from it's modularity and this is no exception, each extension can still be installed separately, just under a new package namespace. But, it will definitely clean up your package.json to not have each component of a table as a package.
## Implementation Approach

Following the approach set forth by `@tiptap/starter-kit` this groups and installs the extensions for table all in a single package and import, greatly reducing duplication and reducing error of forgetting to install a package.

Once we decide that we like this approach, I can move forward with grouping:
 - text formatting: (bold, code, highlight, italic, strike, subscript, superscript, underline)
 - lists: (bullet list, ordered list, list item, list item keymap, task list, task item)
 - text styles: (color, font-family, text-align, text-style)

And, any other groupings that seem obvious to make.

## Checklist


nperez0111:
I've had a change of heart since implementing this, it may be worthwhile to keep the extraneous packages for at least one major version longer just to not break people on them changing their version. They should remove the packages but let's have a nicer upgrade experience. Also allows people to beta test

nperez0111:
@bdbch & @svenadlung  this is ready for review now.

I decided that this PR should only try to do the grouping for tables. It also keeps compatibility with the existing `@tiptap/extension-table-cell` package by doing a re-export of what `@tiptap/extension-table` has.

bdbch (approved):
LGTM

bdbch (approved):
LGTM

Issue #5468, opened 2024-08-11
[Bug]: isAllowedURI disregards custom protocols


Nantris:
### Affected Packages

extension-link

### Version(s)

2.5.9

### Bug Description

The `link` extension accepts custom protocols, but then it rejects them anyway since the XSS fix: https://github.com/ueberdosis/tiptap/blob/84febb29c7092ba0393af78fb44be29e0befa3f4/packages/extension-link/src/link.ts#L112C9-L114C1

### Browser Used

Chrome

### Expected Behavior

Custom protocols should be allowed - eg `file://` is necessary in our case.

### Dependency Updates

- [X] Yes, I've updated all my dependencies.

Nantris:
Suggest change: Make `isAllowedURI` accept a custom checker function to use in addition to the default logic (not instead of it.)

nperez0111:
Good catch, I think that we need to respect the `protocols` that have been allowed.
Maybe, if there is need for a custom checker function we can add it later, but the current bug is that it is not respecting the `protocols` option.

rfgamaral:
@nperez0111 Same as this one, latest releases says this is fixed, but issue is still opened.

Nantris:
I think a custom checker function would be better. This approach still disallows our custom links if they don't start with a protocol (they're transformed to include protocols elsewhere in our code, since we control the execution environment) and it forces inclusion of unsupported protocols - we don't support `cid`, `xmpp`, etc

We still have to patch in ~50 lines, essentially copied from `@tiptap/link-extension`, to allow our links to work.

It's not feasible for us to adapt to the current approach. I understand if you feel this is not worth implementing, but it's disappointing to have to duplicate 50 lines of code just to add an ability to pass a custom checker function (2 lines of code.)

nperez0111:
Appreciate the concern here @Nantris if you like could you create a PR to see that this is addressed?

Obviously my main concern was fixing the regression so I prioritized that. Now that we have that fixed we can figure out the better solution. It is already strange with a validate function and custom protocol handling I do feel a custom checker is in order with some of our changes as a default that can be overriden or something

Nantris:
Thanks for the reply @nperez0111! I'll be happy to do so. It's not a high priority right now since I'm running behind on other things, but I'd love to eliminate the code duplication so I'll definitely loop back on this!

PR #5470, opened 2024-08-12
fix(link): respect custom protocols #5468

nperez0111:
## Changes Overview
When [we fixed a XSS vuln](https://github.com/ueberdosis/tiptap/pull/5160), we inadvertently broke the ability to use custom protocols, this PR resolves that by allowing additional custom protocols to be considered valid and not stripped out
## Checklist

## Related Issues
https://github.com/ueberdosis/tiptap/issues/5468

### DIFF
--- a/.changeset/fluffy-bears-remember.md
+++ b/.changeset/fluffy-bears-remember.md
@@ -1,0 +1,5 @@
+---
+"@tiptap/extension-link": patch
+---
+
+Respect custom protocols for links again, custom protocols are supported in additional to the default set #5468
--- a/packages/extension-link/src/link.ts
+++ b/packages/extension-link/src/link.ts
@@ -106,11 +106,24 @@ declare module '@tiptap/core' {
 
 // From DOMPurify
 // https://github.com/cure53/DOMPurify/blob/main/src/regexp.js
-const ATTR_WHITESPACE = /[\u0000-\u0020\u00A0\u1680\u180E\u2000-\u2029\u205F\u3000]/g // eslint-disable-line no-control-regex
-const IS_ALLOWED_URI = /^(?:(?:(?:f|ht)tps?|mailto|tel|callto|sms|cid|xmpp):|[^a-z]|[a-z+.\-]+(?:[^a-z+.\-:]|$))/i // eslint-disable-line no-useless-escape
+// eslint-disable-next-line no-control-regex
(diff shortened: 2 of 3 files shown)


rfgamaral:
@nperez0111 We've finally integrated the latest Tiptap version with our product, and unfortunately this change (actually, the main culprit is the [XSS vulnerability fix](https://github.com/ueberdosis/tiptap/pull/5160) change) is causing issues for us.

We've started getting loads of support tickets of people complaining that custom links copied from the most various apps could no longer be used to link text in the editor. To give a few examples, apps such as Airmail, Bear, Obsidian, give you links with custom protocols (i.e. `airmail://`, `bear://`, `obsidian://`), and these stopped working. As you can imagine, figuring out a list of all supported custom protocols for all the apps out there will be nearly impossible.

I'm wondering if it would make sense to rethink the approach here, and introduce a disallow list (with sane defaults) instead of an allow list? The downside is that `registerCustomProtocol` from linkify.js needs an allow list, but maybe we could refactor it like this:

- `autolinkProtocols`: allow list of protocols for linkify.js `registerCustomProtocol`
- `disallowedProtocols`: disallow list of protocols for link validation

What do you think?

Alternatively, here are a few more ideas (mutually exclusive):

- Provide an option to completely disable link validation
- Provide an option to fully customize link validation

As an aside, I think there's a bit of a disconnect between the default `allowedProtocols` list, and the `registerCustomProtocol` function. For instance, you are allowed to have links with the `tel`, `callto`, `sms`, `cid`, and `xmpp` protocols, but these are not auto-linked, because they are not registered as custom protocols (the other one seem to be accepted by linkify.js by default, despite not being directly registered either).

nperez0111:
Hey @rfgamaral can you make a new issue for this? I think it warrants some discussion on the tradeoffs & I'm not sure that this is the best place for it.

I don't have the bandwidth right now to give my specific thoughts on it but I know it is important.

My not very well thought out thought is that we should just expose an API to let you choose how you want to handle this sort of a thing (I think there is an existing validate method that we could probably re-use for this). But I do think that the default behavior is quite valid & safe for a web-based editor that isn't really going to deal with custom protocols as much, but I of course understand your concern & it should be accounted for.

rfgamaral:
@nperez0111 Done: https://github.com/ueberdosis/tiptap/issues/5564. Let's continue the discussion there.

PR #5473, opened 2024-08-12
switch from npm to bun for faster dependency resolving

bdbch:
## Changes Overview
This PR moves away from `npm` in favor for `bun` to speed up package resolvement & caching.

## Implementation Approach
I replaced the package-lock with a new bun.lockb and updated all scripts & Github actions accordingly

## Testing Done
I ran builds, the dev setup & linting locally.

## Additional Notes
This is not super crucial and needs a bit of testing because of the Github Actions changes.

## Checklist


nperez0111:
Love bun, but I have to be honest that it isn't quite there yet in terms of NPM compatibility, you still will need to use NPM for publishing for example.

Could we maybe stick with pnpm instead, it is 90% the speed of bun, but is a much more mature project

bdbch:
I don't mind pnpm - will update this PR accordingly then

nperez0111:
I think this will be superceded by https://github.com/ueberdosis/tiptap/pull/5978

PR #5344, opened 2024-07-15
feat: Support type attribute for ordered lists

bastianjoel:
## Changes Overview
Adds `type` attribute to ordered lists which is used to change the list type to for example letters (https://www.w3.org/TR/2012/WD-html-markup-20121025/ol.html#ol.attrs.type).

## Checklist
- [x] I have created a [changeset](https://github.com/changesets/changesets) for this PR if necessary.
- [x] My changes do not break the library.
- [ ] I have added tests where applicable.
- [ ] I have followed the project guidelines.
- [ ] I have fixed any lint issues.

PR #5491, opened 2024-08-14
fix: Use undefined for type attribute default

mhemmings:
## Changes Overview

#5344 introduced a type attribute for ordered lists. The default value is set to `null` which causes a change in the outputted document when using `editor.getJSON()`. If the user is using strict backend validation of the document, this will cause validation to fail.

Changing to `undefined` removes it from the JSON document, matching previous behaviour.

#### Pre v2.6.1

```json
{
  "type": "orderedList",
  "attrs": {
    "start": 1,
  },
  "content": [],
}
```

#### Since v2.6.1

```json
{
  "type": "orderedList",
  "attrs": {
    "start": 1,
    "type": null,
  },
  "content": [],
}
```

## Checklist

## Related Issues

#5344

### DIFF
--- a/.changeset/ordered-list-type-attr-undefined.md
+++ b/.changeset/ordered-list-type-attr-undefined.md
@@ -1,0 +1,5 @@
+---
+"@tiptap/extension-ordered-list": patch
+---
+
+fix: Use undefined for type attribute default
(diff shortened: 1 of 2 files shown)


nperez0111 (approved):
This is a great catch. Thank you for your contribution. Will release it right away

Issue #4873, opened 2024-02-10
[Documentation]: Broken CDN import from https://esm.sh


mbovel:
### What’s the URL to the page you’re sending feedback for?

https://tiptap.dev/docs/editor/installation/cdn

### What part of the documentation needs improvement?

Copy and pasting the example and trying to type a line break in the editor throws: "RangeError: Can not convert <> to a Fragment (looks like multiple versions of prosemirror-model were loaded)".

Indeed, both versions of 1.32.4 and 1.32.7 of `prosemirror-view`, and versions 1.19.3 1.19.4 of `prosemirror-model` are loaded for example.

Current version of `@tiptap/core` and `@tiptap/starter-kit` is 2.2.2. I get the same problem with 2.2.1 and 2.2.0.

The same code seems to work fine when imported from jsdeliver:

```js
import { Editor } from 'https://cdn.jsdelivr.net/npm/@tiptap/core@2.2.2/+esm'
import { StarterKit} from 'https://cdn.jsdelivr.net/npm/@tiptap/starter-kit@2.2.2/+esm'
const editor = new Editor({
    element: document.querySelector('.element'),
    extensions: [
        StarterKit,
    ],
    content: '<p>Hello World!</p>',
})
```

### What is helpful about that part?

-

### What is hard to understand, missing or misleading?

-

### Anything to add? (optional)

-

mattbarrry:
Okay this is a better explanation of what I was experiencing in #4780, thanks for the tip - switching CDNs resolved it for me, but the suggested CDN should definitely be reconsidered.

lekoala:
I'm having this error as well when pressing return on the last line of the editor
RangeError: Can not convert <> to a Fragment (looks like multiple versions of prosemirror-model were loaded)
tiptap is wrapped in a custom element, but i don't see why it would break anything

nperez0111:
So the root of this issue is that some prosemirror packages need to be imported only once into a build, because it relies on `instanceof` checks which would fail if imported from different versions or even CJS vs. ESM builds of the package.

Specifically the issue is that the CDN is not resolving the packages according to their declared peerdeps, because it cannot know what other packages you've imported, just the one that you have requested.

I think that ultimately what we can do here is to specifically bundle in the prosemirror packages in our `@tiptap/pm` package without having to actually resolve versions and leave it up to a bundler.

nperez0111:
[...]

PR #5499, opened 2024-08-15
feat(pm): bundle prosemirror packages into the pm package

nperez0111:
## Changes Overview
This is more of an experiment than anything, if we bundle all of the prosemirror packages instead of relying on dependencies to be resolved properly, we would fix CDN importing of tiptap as well as guaranteeing that a known set of good prosemirror packages work together.
The only issue is when people would want to use prosemirror packages separately from tiptap, because it is now a completely different instance things can break if they don't import from @tiptap/pm.
## Checklist

## Related Issues
https://github.com/ueberdosis/tiptap/issues/4873

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4310,7 +4310,8 @@
     "node_modules/@remirror/core-constants": {
       "version": "2.0.2",
       "resolved": "https://registry.npmjs.org/@remirror/core-constants/-/core-constants-2.0.2.tgz",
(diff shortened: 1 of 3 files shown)


nperez0111:
After some more attempts at this, I think we will probably want to bundle the prosemirror packages but somehow have them still resolve to the `@tiptap/pm/*` packages, I tried with tsup & there is just not much flexibility. I think switching to rollup may be the only option. Probably would also have to build in dep order to have the packages that point to each other resolve properly, luckily there are no cycles between them

Issue #5490, opened 2024-08-14
[Bug]: The "clearDocument" plugin in the core keymap extension can cause issues with custom node change commands.


AlansCodeLog:
### Affected Packages

core

### Version(s)

2.6.2

### Bug Description

I'm writing a block based with a list like schema. The blocks can be different types and I've written a command to switch between them while preserving content. Now, some of the blocks have content and other are contentless.

The command works fine *except* if there is only one contentless node in the document and it's empty. Instead of turning into the block type I want, it turns into a paragraph.

It took some digging to find out why this was, my transaction doc was showing the change but the document wasn't. Turns out there's a clearDocument plugin in the Keymap core extension that was doing this. It can't tell the difference between a transaction that changes the node type like I'm doing and the clearing of the document (am not sure if it's possible) so it clears the document.

https://github.com/ueberdosis/tiptap/blob/da76972998e36a7c13c085d5cd1957f6e0ad8a90/packages/core/src/extensions/keymap.ts#L104

I don't really think this should be core behavior (no other core plugin appends transactions), but if it must be, could it be possible to have the plugin check for some ignore meta key to ignore certain transactions? 

I've gotten around this by disabling core extensions and then re-adding all except a modified version of the Keymap extension without the plugin, but that's a bit messy and annoying. If I published the extensions that make up my editor as a library, downstream users would have to do this as well. 

### Browser Used

Chrome

### Code Example URL

https://codesandbox.io/p/sandbox/tiptap-issue-v7mm9r

### Expected Behavior

In the reproduction I've recreated a tiny bit of my schema and onMounted it tries to change the contentless block to a code block, but instead it's a paragraph. 

If you uncomment the fix part that disables core extensions and adds all except the keymap, you can see it properly changes to a code block like I would expect.

### Dependency Updates

- [X] Yes, I've updated all my dependencies.

nperez0111:
> but if it must be, could it be possible to have the plugin check for some ignore meta key to ignore certain transactions?

I have to be honest, in most situations, I would think that the behavior of how it works now is ideal (most people wouldn't clear the doc just to have a codeblock stick around.

That said, I think your alternate solution of having a meta key to disable this is a good compromise & I'd be happy to accept a PR for that

nperez0111:
https://github.com/ueberdosis/tiptap/blob/c52a6026fcdd6e3e0867a395ef85a28aa91cd90e/packages/core/src/extensions/keymap.ts#L122
[...]

PR #5514, opened 2024-08-17
Fix for #5490 - Adds `preventClearDocument` meta + makes it easier to disable specific core plugins.

AlansCodeLog:
## Changes Overview
- Adds `preventClearDocument` to the clearDocument plugin to fix the main issue in #5490.
- Allows `enableCoreExtensions` to be an object disabling certain core extensions.
- Replaces the nodesBetween empty doc check with isNodeEmpty as requested.

## Implementation Approach
Regarding the disabling of core extensions, @nperez0111, you suggested using `coreExtensionOptions`, but I thought this made a bit more sense. It's only a bit weird in that passing something like `enableCoreExtensions: { keymap: true }` still enables everything. Only passing `false` will disable an extension (leaving others enabled). But I think that would be the most common use case.

Otherwise what sort of interface did you have in mind?

## Testing Done
I'm having a bit of issues running cypress locally. But I see the PRs have github actions.

I also tested it with my app to see everything was getting properly ignore/disabled.

## Verification Steps
- clearDocument should still work as usual unless one sets `preventClearDocument` on the transaction. My reproduction in #5490 shows the issue clearly.
- enableCoreExtensions should work normally, unless an object is passed, and then only core extensions set to false are disabled. Others stay enabled.

## Checklist
Am awaiting the github actions results, but otherwise no.

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -79,7 +79,24 @@ export interface EditorOptions {
   };
   enableInputRules: EnableRules;
   enablePasteRules: EnableRules;
-  enableCoreExtensions: boolean;
+  /**
+   * Determines whether core extensions are enabled.
+   *
+   * If set to `false`, all core extensions will be disabled.
+   * To disable specific core extensions, provide an object where the keys are the extension names and the values are `false`.
+   * Extensions not listed in the object will remain enabled.
+   *
+   * @example
+   * // Disable all core extensions
+   * enabledCoreExtensions: false
+   *
+   * @example
+   * // Disable only the keymap core extension
+   * enabledCoreExtensions: { keymap: false }
+   *
+   * @default true
+   */
+  enableCoreExtensions: boolean | Record<'editable' | 'clipboardTextSerializer' | 'commands' | 'focusEvents' | 'keymap' | 'tabindex', false>;
   /**
    * If `true`, the editor will check the content for errors on initialization.
    * Emitting the `contentError` event if the content is invalid.
--- a/.changeset/chatty-pumas-cross.md
+++ b/.changeset/chatty-pumas-cross.md
@@ -1,0 +1,5 @@
+---
+"@tiptap/core": minor
+---
+
+Fixes #5490. The `preventClearDocument` meta tag can now be used to prevent the `clearDocument` plugin in the core keymap extension from modifying transactions that appear to clear the document (but might be clearing it for other reasons).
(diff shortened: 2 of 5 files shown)


packages/core/src/types.ts
```diff
@@ -79,7 +79,13 @@ export interface EditorOptions {
...
   enableInputRules: EnableRules;
   enablePasteRules: EnableRules;
-  enableCoreExtensions: boolean;
+  /**
+   * If `false`, the core extensions will not be loaded.
+   * You can pass an object to disable specific extensions (e.g. `{ keymap: false }`).
+   * Extensions not listed in the object will still be enabled.
+   * @default true
+   */
+  enableCoreExtensions: boolean | Record<'editable' | 'clipboardTextSerializer' | 'commands' | 'focusEvents' | 'keymap' | 'tabindex', any>;
```
nperez0111, line 88:
```suggestion
  enableCoreExtensions: boolean | Record<'editable' | 'clipboardTextSerializer' | 'commands' | 'focusEvents' | 'keymap' | 'tabindex', false>;
```

Since `false` is the only valid option

packages/core/src/types.ts
```diff
@@ -79,7 +79,13 @@ export interface EditorOptions {
   };
   enableInputRules: EnableRules;
   enablePasteRules: EnableRules;
-  enableCoreExtensions: boolean;
+  /**
+   * If `false`, the core extensions will not be loaded.
+   * You can pass an object to disable specific extensions (e.g. `{ keymap: false }`).
+   * Extensions not listed in the object will still be enabled.
+   * @default true
+   */
```
nperez0111, line 87:
```suggestion
  /**
   * Determines whether core extensions are enabled.
   *
   * If set to `false`, all core extensions will be disabled.
   * To disable specific core extensions, provide an object where the keys are the extension names and the values are `false`.
   * Extensions not listed in the object will remain enabled.
   *
   * @example
   * // Disable all core extensions
   * enabledCoreExtensions: false
   *
   * @example
   * // Disable only the keymap core extension
   * enabledCoreExtensions: { keymap: false }
   *
   * @default true
   */
```

packages/core/src/types.ts
AlansCodeLog: Oops, meant to put false.

AlansCodeLog:
Incorporated both suggested changes.

AlansCodeLog:
Um, I just realized there is a slight bug with the types... The record should be a partial record. Should I open a new PR?

nperez0111:
Ah right, good catch. I'll fix it

AlansCodeLog:
:+1:  Thanks.

PR #5528, opened 2024-08-20
feat: allowing specifying the content of ReacNodeViewContent via a React Context

nperez0111:
## Changes Overview
This is to set ourselves up to be able to do static rendering in react with support for node views
## Checklist

### DIFF
--- a/packages/react/src/NodeViewContent.tsx
+++ b/packages/react/src/NodeViewContent.tsx
@@ -1,15 +1,14 @@
-import React from 'react'
+import React, { ComponentProps } from 'react'
 
 import { useReactNodeView } from './useReactNodeView.js'
 
-export interface NodeViewContentProps {
-  [key: string]: any,
-  as?: React.ElementType,
-}
+export type NodeViewContentProps<T extends keyof React.JSX.IntrinsicElements = 'div'> = {
+  // eslint-disable-next-line no-undef
+  as?: NoInfer<T>;
+} & ComponentProps<T>
 
-export const NodeViewContent: React.FC<NodeViewContentProps> = props => {
-  const Tag = props.as || 'div'
-  const { nodeViewContentRef } = useReactNodeView()
+export function NodeViewContent<T extends keyof React.JSX.IntrinsicElements = 'div'>({ as: Tag = 'div', ...props }: NodeViewContentProps<T>) {
+  const { nodeViewContentRef, nodeViewContentChildren } = useReactNodeView()
 
   return (
     // @ts-ignore
@@ -21,6 +20,8 @@ export const NodeViewContent: React.FC<NodeViewContentProps> = props => {
         whiteSpace: 'pre-wrap',
         ...props.style,
       }}
-    />
+    >
+      {nodeViewContentChildren}
+    </Tag>
   )
 }
--- a/packages/react/src/useReactNodeView.ts
+++ b/packages/react/src/useReactNodeView.ts
@@ -1,12 +1,23 @@
-import { createContext, useContext } from 'react'
+import { createContext, ReactNode, useContext } from 'react'
 
 export interface ReactNodeViewContextProps {
   onDragStart: (event: DragEvent) => void,
   nodeViewContentRef: (element: HTMLElement | null) => void,
+  /**
+   * This allows you to add children into the NodeViewContent component.
+   * This is useful when statically rendering the content of a node view.
+   */
+  nodeViewContentChildren: ReactNode,
 }
 
-export const ReactNodeViewContext = createContext<Partial<ReactNodeViewContextProps>>({
-  onDragStart: undefined,
+export const ReactNodeViewContext = createContext<ReactNodeViewContextProps>({
+  onDragStart: () => {
+    // no-op
+  },
+  nodeViewContentChildren: undefined,
+  nodeViewContentRef: () => {
+    // no-op
+  },
 })
 
 export const useReactNodeView = () => useContext(ReactNodeViewContext)


packages/static-renderer/src/pm/react.tsx
```diff
@@ -0,0 +1,432 @@
...
+  //   },
+  //   markMapping: {},
+  // }
+  generateMappings(extensions),
+)
+
+const schema = getSchema([StarterKit])
+
+console.log(
+  renderToStaticMarkup(
```
nperez0111, line 219: This is just a demo

bdbch (approved):
LGTM! Awesome idea @nperez0111

nperez0111:
Just as a note, if we render to something like [jsx-dom](https://github.com/alex-kinokon/jsx-dom/tree/master) then we can render directly to HTMLElements with pure JS & prosemirror

nperez0111:
> @nperez0111 I am super excited about this PR (finally took a look at it over the weekend). I tried building it locally today to test it out, and it failed because there was no `src/index.ts` file in the new `static-renderer` folder. Is there a pre-build step I was missing? I just added one locally to get past that.
> 
> Edit: Ah, its the same error currently failing the build CI step. You are probably already aware of this then!

Yep, this isn't completely prime-time yet. Needs a lot more tests to ensure things actually work

bdbch:
@nperez0111 I'll remove my review - let me check if I should review in the future.

masylum:
I'm running this branch to test this functionality (just what I needed!) and I see some weird behaviour. I have this imageNode:

```js
import { Node, nodeInputRule } from "@tiptap/core"

export const IMAGE_NODE = "image"

export const ImageNode = Node.create({
	name: IMAGE_NODE,
	group: "block",
	inline: false,
	draggable: true,

	addOptions() {
		return { HTMLAttributes: {} }
	},

	addAttributes() {
		return {
			src: {
				default: null,
				parseHTML: (element) =>
					element.querySelector("img")?.getAttribute("src"),
			},
			alt: {
				default: null,
				parseHTML: (element) =>
					element.querySelector("img")?.getAttribute("alt"),
			},
			title: {
				default: null,
				parseHTML: (element) =>
					element.querySelector("img")?.getAttribute("title"),
			},
		}
	},

	parseHTML() {
		return [{ tag: `figure[data-type="${IMAGE_NODE}"]` }]
	},

		const { title, alt, src } = HTMLAttributes
		const imgAttrs = { alt, src, title }
		const figureAttrs = { "data-type": IMAGE_NODE }
		if (!title) return ["figure", figureAttrs, ["img", imgAttrs]]

		return ["figure", figureAttrs, ["img", imgAttrs], ["figcaption", title]]
	},
})
```

pretty standard stuff, not actually that different from the original image node. But, for some reason, the `extensionRenderer` calls `getHTMLAttributes`, which in turn, calls `getAttributes` with that flag set to true. That flag discards src alt and title, since they are not rendered. Why that behaviour?

packages/static-renderer/src/helpers.ts
```diff
@@ -0,0 +1,65 @@
...
+): Record<string, any> {
+  const nodeOrMarkAttributes = nodeOrMark.attrs
+
+  if (!nodeOrMarkAttributes) {
+    return {}
+  }
+
+  return extensionAttributes
+    .filter(item => {
+      if (item.type !== nodeOrMark.type) {
```
masylum, line 25: on my testing, I had to use `nodeOrMark.type.name`, since otherwise it would not match anything.

nperez0111: Ah, yea. Since this is shared between JSON and prosemirror objects, I added a check to see if it is a string or not first

nperez0111:
Welp, didn't know about that feature, I'll have to get back to you on this

nperez0111:
@bdbch & @svenadlung I believe this is ready for review & merging into next

Richard87:
Hi! Very excited about this feature! But I was under the impression that custom React components/extensions required to be saved as HTML (to properly serialize the attributes), and it seems in the example above that you require `JSONConten` or `TNode`)

Has that changed, can we static-render html code with react components, 
or, can we store react components with attributes as JSON now? 

Awesome work! 🙏

nperez0111:
> Hi! Very excited about this feature! But I was under the impression that custom React components/extensions required to be saved as HTML (to properly serialize the attributes), and it seems in the example above that you require `JSONConten` or `TNode`)
> 
> Has that changed, can we static-render html code with react components, or, can we store react components with attributes as JSON now?
> 
> Awesome work! 🙏

Static rendering is for rendering JSON into React components, HTML strings, or markdown strings, (with the possibility of more target formats in the future like PDF & DOCX).

It was never meant to be for parsing HTML strings, if it is already in HTML, can't you just put it into a `dangerouslySetInnerHTML` to render that content to the page? The whole point of static rendering is that it does not require an editor instance, whereas to parse the html properly would require one, so if that is what you are looking to do, this isn't it. However, you can [use this to generate JSON from the HTML](https://tiptap.dev/docs/editor/api/utilities/html#generating-json-from-html), and then statically render that if you want

Richard87:
Thanks for the quick response! :)

Not directly, since custom react extensions would look like `<photo-node src="data:aksjdhf"></photo-node>`, so using `dangerouslySetInnerHTML` wont work.

_But_ I found yesterday after posting, that `JSON.serialize` retained the `attr` when executing in the browser (but it was lost when sending the content object to a Next Server Action for serializing there), so my test might be flawed :)

packages/static-renderer/src/pm/html-string/html-string.example.ts
bdbch, line 1: See above

bdbch (approved):
Looks good to me. I'm just not sure about all the example files. Is there a reason why they're here?

Afaik we don't have any example files in any of our other packages and we usually did this kind of thing in our demos.

nperez0111:
Yep, yea the .example files were just so I can test things while running it in bun.

I'll see if it makes sense to make demos out of them

volarname:
what about vue?

guqing:
Hi team 👋,

I’m trying to test this PR using `"@tiptap/static-renderer": "3.0.0-beta.3"` that includes a custom `NodeView` using `@tiptap/react`. I followed the example provided and integrated it into a Next.js project (App Router, `app/` directory structure). However, I’m encountering the following error when I run the project:

```
⨯ TypeError: Class extends value undefined is not a constructor or null

This might be caused by a React Class Component being rendered in a Server Component, React Class Components only work in Client Components.
Read more: https://nextjs.org/docs/messages/class-component-in-server-component

    at src/editor/extensions/nodes/MyCustomComponentWithContent.tsx:2:0
    at src/editor/extensions/nodes/index.ts:2:0
    at src/editor/extensions/index.ts:16:0
    at src/app/api/text/route.ts:9:0
    at Object.<anonymous> (.next/server/app/api/text/route.js:11:9)
```

The error is triggered by this import line in `MyCustomComponentWithContent.tsx`:

```ts
import { NodeViewWrapper, NodeViewProps, NodeViewContent } from '@tiptap/react';
```

From what I understand, this may be related to React Class Components not being compatible with Next.js Server Components. But since the extension is used in the editor context, I expected it to only run on the client.

### Repro context:
- Using `@tiptap/react`
- Running on Next.js 15.3.1 App Router
- The NodeView is part of a PR implementation (e.g. `< MyCustomComponentWithContent />`)

### Questions:
Is there a known issue when using `NodeViewWrapper` inside a Next.js app?

Let me know if you need a minimal repo—I’d be happy to prepare one.

Thanks for your awesome work!

nperez0111:
> what about vue? really dont want to use v-html

@volarname static renderer doesn't care what your output is. So you can implement it how you want.

guqing:
> > I’m trying to test this PR using `"@tiptap/static-renderer": "3.0.0-beta.3"` that includes a custom `NodeView` using `@tiptap/react`. I followed the example provided and integrated it into a Next.js project (App Router, `app/` directory structure). However, I’m encountering the following error when I run the project:
> 
> @guqing You are using the beta version of static renderer, but are you using the correct `@tiptap/react` version which also needs to be the beta version? This PR made changes to NodeViews as well so it needs the beta version too

Yes, tiptap/react is also on the beta version. Here's a minimal reproducible demo: [static-renderer-demo.zip](https://github.com/user-attachments/files/20036099/static-renderer-demo.zip)  ，It uses the code from this PR example

The following error can be reproduced by accessing localhost:3000 in Node.js 18 and 22:
```
TypeError: Class extends value undefined is not a constructor or null

This might be caused by a React Class Component being rendered in a Server Component, React Class Components only works in Client Components. Read more: https://nextjs.org/docs/messages/class-component-in-server-component

src/app/page.tsx (4:1) @ [project]/src/app/page.tsx [app-rsc] (ecmascript)


  2 | import StarterKit from "@tiptap/starter-kit";
  3 | import { Node } from '@tiptap/core'
> 4 | import {  ReactNodeViewRenderer } from '@tiptap/react'
    | ^
  5 | import React from 'react'
```
If use NodeViewWrapper from tiptap/react, NodeViewContent will throw the same error

nperez0111:
I briefly looked into this @guqing, it seems to be that the `@tiptap/react` package uses `React.Component` which is not valid in an RSC. Unfortunately, while the static renderer with `ReactNodeViewRenderer` works in SSR (on node), it does not work within an RSC.

RSCs impose a few additional constraints that the react package relies on:
 - class components (which is an easy refactor)
 - React contexts and effects (not easy to refactor)
 
So, this will not be viable to use custom node views within a Next.js RSC at this point in time. It will require significant effort to refactor this in a way that works both for client and server components.
 
I no longer work for Tiptap & don't have the capacity to try make these changes myself. I'd suggest you to create a GitHub Issue and link back to my explanation here. Even contribute a PR if you can! Tagging @bdbch for visibility
 
Your best option at the moment is to remove any references to `ReactNodeViewRenderer` from reach of the RSC, and implement the `nodeMapping` yourself. See this example:
```tsx
import { renderToReactElement } from "@tiptap/static-renderer";
import { Node } from "@tiptap/core";
// import { ReactNodeViewRenderer } from "@tiptap/react";
import React from "react";

// This component does not have a NodeViewContent, so it does not render it's children's rich text content
function MyCustomComponentWithoutContent() {
  // RSCs do not support useState
  // const [count, setCount] = React.useState(200);

  return (
    <div
      className="custom-component-without-content"
      // onClick={() => setCount((a) => a + 1)}
    >
      {/* {count} This is a react component! */}
      This is a react component!
    </div>
  );
}

const CustomNodeExtensionWithoutContent = Node.create({
  name: "customNodeExtensionWithoutContent",
  atom: true,
  renderHTML() {
    return ["div", { class: "my-custom-component-without-content" }] as const;
  },
  // This extension cannot be used with RSCs if it uses the ReactNodeViewRenderer
  // addNodeView() {
  //   return ReactNodeViewRenderer(MyCustomComponentWithoutContent);
  // },
});

export default function Home() {
  return renderToReactElement({
    extensions: [StarterKit, CustomNodeExtensionWithoutContent],
    options: {
      nodeMapping: {
        // render the custom node with the intended node view React component
        customNodeExtensionWithoutContent: MyCustomComponentWithoutContent,
      },
    },
    content: {
      type: "doc",
      content: [
        {
          type: "customNodeExtensionWithoutContent",
          // rich text content
          content: [
            {
              type: "text",
              text: "Hello, world!",
            },
          ],
        },
      ],
    },
  });
}
```
This example will correctly render the React component in an RSC.

Also you specified your deps incorrectly it should be:
```json
    "@tiptap/react": "3.0.0-beta.4",
    "@tiptap/starter-kit": "3.0.0-beta.4",
    "@tiptap/static-renderer": "3.0.0-beta.4",
    "@tiptap/core": "3.0.0-beta.4",
    "@tiptap/pm": "3.0.0-beta.4"
```

For future reference:
 
 I asked an LLM to refactor the `React.Component` usage in the repo and ended up with this diff (which seems about right):
 ```diff
 diff --git i/packages/react/src/EditorContent.tsx w/packages/react/src/EditorContent.tsx
index 39f220b5c..5a152784a 100644
--- i/packages/react/src/EditorContent.tsx
+++ w/packages/react/src/EditorContent.tsx
@@ -83,124 +83,107 @@ function getInstance(): ContentComponent {
   }
 }
 
-export class PureEditorContent extends React.Component<
-  EditorContentProps,
-  { hasContentComponentInitialized: boolean }
-> {
-  editorContentRef: React.RefObject<any>
+export const PureEditorContent: React.FC<EditorContentProps> = props => {
+  const { editor: currentEditor, innerRef, ...rest } = props
+  const editorContentRef = React.useRef<HTMLDivElement>(null)
+  const initialized = React.useRef(false)
+  const unsubscribeToContentComponent = React.useRef<(() => void) | undefined>(undefined)
 
-  initialized: boolean
+  const [hasContentComponentInitialized, setHasContentComponentInitialized] = React.useState(
+    Boolean((currentEditor as EditorWithContentComponent | null)?.contentComponent),
+  )
 
-  unsubscribeToContentComponent?: () => void
+  React.useEffect(() => {
+    const editor = currentEditor as EditorWithContentComponent | null
 
-  constructor(props: EditorContentProps) {
-    super(props)
-    this.editorContentRef = React.createRef()
-    this.initialized = false
+    const init = () => {
+      if (editor && !editor.isDestroyed && editor.options.element) {
+        if (editor.contentComponent) {
+          return
+        }
 
-    this.state = {
-      hasContentComponentInitialized: Boolean((props.editor as EditorWithContentComponent | null)?.contentComponent),
+        const element = editorContentRef.current
+
+        if (element) {
+          element.append(...editor.options.element.childNodes)
+
+          editor.setOptions({
+            element,
+          })
+
+          editor.contentComponent = getInstance()
+
+          // Has the content component been initialized?
+          if (!hasContentComponentInitialized) {
+            // Subscribe to the content component
+            unsubscribeToContentComponent.current = editor.contentComponent.subscribe(() => {
+              setHasContentComponentInitialized(prev => {
+                if (!prev) {
+                  return true
+                }
+                return prev
+              })
+
+              // Unsubscribe to previous content component
+              if (unsubscribeToContentComponent.current) {
+                unsubscribeToContentComponent.current()
+              }
+            })
+          }
+
+          editor.createNodeViews()
+          initialized.current = true
+        }
+      }
     }
-  }
 
-  componentDidMount() {
-    this.init()
-  }
+    init()
 
-  componentDidUpdate() {
-    this.init()
-  }
+    return () => {
+      // This is the cleanup function, equivalent to componentWillUnmount
+      const editorUnmount = currentEditor as EditorWithContentComponent | null
 
-  init() {
-    const editor = this.props.editor as EditorWithContentComponent | null
-
-    if (editor && !editor.isDestroyed && editor.options.element) {
-      if (editor.contentComponent) {
+      if (!editorUnmount) {
         return
       }
 
-      const element = this.editorContentRef.current
+      initialized.current = false
 
-      element.append(...editor.options.element.childNodes)
-
-      editor.setOptions({
-        element,
-      })
-
-      editor.contentComponent = getInstance()
-
-      // Has the content component been initialized?
-      if (!this.state.hasContentComponentInitialized) {
-        // Subscribe to the content component
-        this.unsubscribeToContentComponent = editor.contentComponent.subscribe(() => {
-          this.setState(prevState => {
-            if (!prevState.hasContentComponentInitialized) {
-              return {
-                hasContentComponentInitialized: true,
-              }
-            }
-            return prevState
-          })
-
-          // Unsubscribe to previous content component
-          if (this.unsubscribeToContentComponent) {
-            this.unsubscribeToContentComponent()
-          }
+      if (!editorUnmount.isDestroyed) {
+        editorUnmount.view.setProps({
+          nodeViews: {},
         })
       }
 
-      editor.createNodeViews()
+      if (unsubscribeToContentComponent.current) {
+        unsubscribeToContentComponent.current()
+      }
 
-      this.initialized = true
-    }
-  }
+      editorUnmount.contentComponent = null
 
-  componentWillUnmount() {
-    const editor = this.props.editor as EditorWithContentComponent | null
+      if (!editorUnmount.options.element?.firstChild) {
+        return
+      }
 
-    if (!editor) {
-      return
-    }
+      // TODO using the new editor.mount method might allow us to remove this
+      const newElement = document.createElement('div')
 
-    this.initialized = false
+      newElement.append(...editorUnmount.options.element.childNodes)
 
-    if (!editor.isDestroyed) {
-      editor.view.setProps({
-        nodeViews: {},
+      editorUnmount.setOptions({
+        element: newElement,
       })
     }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [currentEditor, hasContentComponentInitialized]) // Dependencies for the effect
 
-    if (this.unsubscribeToContentComponent) {
-      this.unsubscribeToContentComponent()
-    }
-
-    editor.contentComponent = null
-
-    if (!editor.options.element?.firstChild) {
-      return
-    }
-
-    // TODO using the new editor.mount method might allow us to remove this
-    const newElement = document.createElement('div')
-
-    newElement.append(...editor.options.element.childNodes)
-
-    editor.setOptions({
-      element: newElement,
-    })
-  }
-
-  render() {
-    const { editor, innerRef, ...rest } = this.props
-
-    return (
-      <>
-        <div ref={mergeRefs(innerRef, this.editorContentRef)} {...rest} />
-        {/* @ts-ignore */}
-        {editor?.contentComponent && <Portals contentComponent={editor.contentComponent} />}
-      </>
-    )
-  }
+  return (
+    <>
+      <div ref={mergeRefs(innerRef, editorContentRef)} {...rest} />
+      {/* @ts-ignore */}
+      {currentEditor?.contentComponent && <Portals contentComponent={currentEditor.contentComponent} />}
+    </>
+  )
 }
 
 // EditorContent should be re-created whenever the Editor instance changes

 ```

guqing:
Hi @nperez0111 ,
﻿
Thank you so much for taking the time to dig into this and write such a detailed explanation. 🙏
I’ll open a dedicated issue that links back to your comment (and tag @bdbch) so the core team can triage it properly. 
Really appreciate your guidance! 🚀

Thanks again

tasiotas:
please document how to use static renderer with Vue. Its great to see it with React, but its not clear how to make it work with Vue.

> > what about vue? really dont want to use v-html
> 
> @volarname static renderer doesn't care what your output is. So you can implement it how you want.

nperez0111:
> please document how to use static renderer with Vue. Its great to see it with React, but its not clear how to make it work with Vue.

Feel free to contribute, I don't know vue well enough to do that and have no need for this myself

PR #5539, opened 2024-08-22
feat: Add data attribute on ReactNodeView mounting div to assist styling

WilliamIPark:
## Changes Overview
The goal of this PR is to make the react mounting element a bit more identifiable. This should aid the ability to select and style this element via CSS.

## Implementation Approach
Added a data attribute to the element just after it is created, and is confirmed to exist.

## Checklist
* [ ]  I have created a [changeset](https://github.com/changesets/changesets) for this PR if necessary.
* [x]  My changes do not break the library.
* [x]  I have added tests where applicable.
* [x]  I have followed the project guidelines.
* [x]  I have fixed any lint issues.

### DIFF
--- a/packages/react/src/ReactNodeViewRenderer.tsx
+++ b/packages/react/src/ReactNodeViewRenderer.tsx
@@ -88,6 +88,7 @@ class ReactNodeView extends NodeView<
     }
 
     if (this.contentDOMElement) {
+      this.contentDOMElement.dataset.reactMount = '';
       // For some reason the whiteSpace prop is not inherited properly in Chrome and Safari
       // With this fix it seems to work fine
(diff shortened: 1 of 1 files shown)


packages/react/src/ReactNodeViewRenderer.tsx
```diff
@@ -88,6 +88,7 @@ class ReactNodeView extends NodeView<
     }
 
     if (this.contentDOMElement) {
+      this.contentDOMElement.dataset.reactMount = ''
```
nperez0111, line 91: Maybe `react-root` is more appropriate?

nperez0111: Or `node-view-content-react`

Issue #5533, opened 2024-08-21
[Bug]: Mention items loading state unavailable


regexyl:
### Affected Packages

extension-mention

### Version(s)

2.6.1

### Bug Description

Filing this as a bug instead of a feature as it impedes the instantaneous showing of the popover.

The `items` callback in the [suggestion utility](https://tiptap.dev/docs/editor/api/utilities/suggestion) sometimes returns a promise that is awaited, but there isn't a way to obtain the loading state. The showing of the popover is delayed until the promise is fulfilled, which stalls the UI while waiting for the promise to return a value. It would be nice to have a loading state returned in the popover props.

Thanks for your work on this library! Happy to contribute

### Browser Used

Chrome

### Expected Behavior

Have a loading state available for the popover element so that it doesn't just stall.

### Dependency Updates

- [X] Yes, I've updated all my dependencies.

nperez0111:
That is a great point. Would definitely take a contribution for this.

PR #5542, opened 2024-08-23
fix(suggestion): add isLoading state to suggestion popup

regexyl:
## Changes Overview
Fixes issue #5533 by exposing an `isLoading` prop for the suggestion popup.

## Implementation Approach
Not sure if this is the way to go, took the least resistant path. This re-renders the popup after resolving the `items` promise with the updated result.

## Testing Done
Here's a recording, though I'm not sure how it should be presented on the demo example.
https://www.loom.com/share/11fb168e6f0e4e6eb2853472b4f90842

## Verification Steps
As described in the video above.

## Checklist

### DIFF
--- a/packages/suggestion/src/suggestion.ts
+++ b/packages/suggestion/src/suggestion.ts
@@ -130,6 +130,11 @@ export interface SuggestionProps<I = any, TSelected = any> {
   items: I[]
 
   /**
+   * The state of loading items.
+   */
+  isLoading: boolean
+
+  /**
    * A function that is called when a suggestion is selected.
    * @param props The props object.
    * @returns void
@@ -215,6 +220,7 @@ export function Suggestion<I = any, TSelected = any>({
             query: state.query,
             text: state.text,
             items: [],
+            isLoading: false,
             command: commandProps => {
               return command({
                 editor,
@@ -247,9 +253,14 @@ export function Suggestion<I = any, TSelected = any>({
           }
 
           if (handleChange || handleStart) {
-            props.items = await items({
-              editor,
-              query: state.query,
+            props.isLoading = true
+            Promise.resolve(items({ editor, query: state.query })).then(result => {
+              if (props == null) {
+                return
+              }
+              props.isLoading = false
+              props.items = result
+              renderer?.onUpdate?.(props)
             })
           }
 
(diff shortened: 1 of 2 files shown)


packages/suggestion/src/suggestion.ts
```diff
@@ -247,9 +253,14 @@ export function Suggestion<I = any, TSelected = any>({
...
-              editor,
-              query: state.query,
+            props.isLoading = true
+            Promise.resolve(items({ editor, query: state.query })).then(result => {
+              if (props == null) {
+                return
+              }
+              props.isLoading = false
+              props.items = result
+              renderer?.onUpdate?.(props)
```
nperez0111, line 263: This is doing more than just exposing the loading state, it should still be awaited, you are changing the behavior of how this works.

nperez0111 (changes requested):

packages/suggestion/src/suggestion.ts
regexyl: Ah I see apologies - this is kind of a naive question, why does it need to be awaited?

nperez0111: By awaiting the function, you are suspending it's execution of the lines after it. If the goal truly is to only expose an `isLoading` flag, then you should not be messing with the order of execution. By doing a Promise.resolve & not awaiting it, you are immediately executing the next lines synchronously.

regexyl: That makes sense! Though wouldn't having `await` defeat the purpose of having an up-to-date `isLoading` value if the component is only going to be re-rendered after the promise returns?

I imagined it like this:
- update `isLoading` -> rerender (i.e. immediately executing next lines)
- when promise is resolved -> update `isLoading` -> rerender

and re: messing up the next lines: maybe something [like this](https://gist.github.com/regexyl/958aa74f8c77e6ae0176a311b407abd4#file-suggestion-ts-L255) might work?

Might be missing a bit of context and need a bit of hand-holding here

kalmigs: hey guys, is there plans to include this? or is there a work around?

bdbch:
@regexyl could you rebase this to the current develop branch as we released V3, deprecating the next branch & would need to update most of our PR's anyway? Thanks a lot!

Issue #4651, opened 2023-11-24
[Bug]: Aliases are not recognized in the editor


NorthLake:
### Which packages did you experience the bug in?

extension-code-block-lowlight

### What Tiptap version are you using?

@tiptap/vue-3: 2.1.12, @tiptap/extension-code-block-lowlight: 2.1.12, lowlight: 3.1.0

### What’s the bug you are facing?

Most languages have at least one alias baked in, and you can add additional ones yourself using `lowlight.registerAlias(...)`, but neither of these are checked when adding a codeblock in the editor.

### What browser are you using?

Chrome

### What did you expect to happen?

` ```js ` codeblocks are highlighted the same way as ` ```javascript ` codeblocks.

### Anything to add? (optional)

I tried to get the CodeSandbox to work, but I couldn't get any of them to load at all.

### Did you update your dependencies?

- [X] Yes, I’ve updated my dependencies to use the latest version of all packages.

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

NorthLake:
I did some more digging, and it gets really weird.
If I call `lowlight.registered('js')` from my own code, then everything runs fine. But when `highlight.getLanguage(aliasOrLanguage)` is called at `lowlight-plugin.ts:31`, the alias and language variables in `hightlght.js/lib/core.js` are just empty objects.
Maybe this helps with troubleshooting.

nperez0111:
This has been released with 2.7.0-pre.0 and will be released in tiptap 2.7.0 when this beta has been tested by more users

PR #5551, opened 2024-08-26
fix(extension-code-block-lowlight): support for lowlight v3 aliases

Ilya-Iskra:
## Changes Overview
Added support for lowlight v2+ method `registered`

## Implementation Approach
I tried not to remove support for earlier versions of lowlight that was implemented in #3155

## Testing Done
I checked that after entering ` ```javascript ` and ` ```jsx ` there was the same formatting result. It can be also checked on other aliases that are listed in [highlight.js Supported Languages](https://github.com/highlightjs/highlight.js/blob/main/SUPPORTED_LANGUAGES.md)

## Additional Notes
Before
After

## Checklist

## Related Issues
#4651

### DIFF
--- a/.changeset/sweet-ants-lick.md
+++ b/.changeset/sweet-ants-lick.md
@@ -1,0 +1,5 @@
+---
+"@tiptap/extension-code-block-lowlight": patch
+---
+
+Add support for lowlight v3 aliases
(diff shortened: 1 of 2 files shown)


nperez0111 (approved):
This looks right to me, thanks for your contribution!

PR #5558, opened 2024-08-27
feat(jsx): add `@tiptap/jsx` for more convenient rendering of Tiptap content

nperez0111:
## Changes Overview

This implements `@tiptap/jsx` which is a package that can render JSX in a node's `renderHTML` method to the format that Tiptap expects.
This makes for a more modern experience of describing the content of a Node as you might do in modern frameworks like React/Vue.
## Implementation Approach

Here is an example of what this allows for:
```tsx
/* @jsx jsxTiptap */
import { jsxTiptap, children } from "@tiptap/jsx";
import { Node } from "@tiptap/core";

const Nested = () => (
  <span>
    <span>{children}</span>
    <br />
    nested
  </span>
);

const Demo = Node.create({
  name: "demo",
  group: "block",
  content: "block*",
  draggable: true,
  addAttributes() {
    return {
      id: {
        default: null,
      },
    };
  },
    return <div {...HTMLAttributes}><Nested /></div>;
  },
  parseHTML() {
    console.log("parseHTML");
    return [
      {
        tag: "div",
      },
    ];
  },
});

export default Demo;
```
## Testing Done

Need to add some tests for ensuring that the arrays are all generated correctly. More of a POC than anything.
## Checklist


packages/jsx/README.md
```diff
@@ -0,0 +1,18 @@
+# @tiptap/static-renderer
```
bdbch, line 1: fix here

packages/jsx/src/jsx.tsx
```diff
@@ -0,0 +1,65 @@
...
+  | [string]
+  | [string, Attributes]
+  | [string, 0]
+  | [string, Attributes, 0]
+  | [string, Attributes, DOMOutputSpecArray | 0]
+  | [string, DOMOutputSpecArray];
+
+declare global {
+  // eslint-disable-next-line @typescript-eslint/no-namespace
+  namespace JSX {
```
bdbch, line 18: could this cause issues globally with the normal react typings (for example I get typescript issues while using normal React JSX?)

bdbch (approved):
LGTM - just a quick comment regarding naming and namespaces

bdbch:
What's the status of this one. Going through our approved PRs to see where we can move forward.

nperez0111:
I want to do this for v3, not relative to develop so I haven't worked on it in a minute

bdbch:
Moved the target branch to next (we'll probably need to sync next with develop soon too)

nperez0111:
This is ready now @bdbch

I'm unsure that it really is worth it for us to use the JSX transforms internally (although I did convert bold and blockquote as a proof of concept). But, it is a really nice API for consumers

PR #5588, opened 2024-09-04
fix(react): Fix incorrect extensionAttributes value

YaoKaiLun:
## Changes Overview
Prevent to call renderHTML function of all extensions.

## Checklist

### DIFF
--- a/packages/react/src/ReactNodeViewRenderer.tsx
+++ b/packages/react/src/ReactNodeViewRenderer.tsx
@@ -304,7 +304,9 @@ export class ReactNodeView<
       let attrsObj: Record<string, string> = {}
 
       if (typeof this.options.attrs === 'function') {
-        const extensionAttributes = this.editor.extensionManager.attributes
+        const extensionAttributes = this.editor.extensionManager.attributes.filter(
+          attribute => attribute.type === this.node.type.name,
+        )
         const HTMLAttributes = getRenderedAttributes(this.node, extensionAttributes)
 
         attrsObj = this.options.attrs({ node: this.node, HTMLAttributes })
--- a/.changeset/tricky-apricots-smash.md
+++ b/.changeset/tricky-apricots-smash.md
@@ -1,0 +1,5 @@
+---
+"@tiptap/react": patch
+---
+
+Fix incorrect extensionAttributes value


packages/react/src/ReactNodeViewRenderer.tsx
```diff
@@ -304,7 +304,9 @@ export class ReactNodeView<
       let attrsObj: Record<string, string> = {}
 
       if (typeof this.options.attrs === 'function') {
-        const extensionAttributes = this.editor.extensionManager.attributes
+        const extensionAttributes = this.editor.extensionManager.attributes.filter(
+          attribute => attribute.type === this.node.type.name,
+        )
```
nperez0111, line 309: Sorry, what did you run into that warrants this change? It is not clear to me why it is needed? It feels like something that is better placed as a responsibility of `getRenderedAttributes` instead

YaoKaiLun: The "attrs" function would accept all extensions' acttributes, it wasn't what i expected, it only need accpet relative extensions' acttributes. Maybe there's a better way to write it that I don't know?

nperez0111: Yep, you are right here. I just think it is better implemented in the function `getRenderedAttributes` instead.

nperez0111: I normally would just apply my patch on top of the branch, but some reason it was disabled, so here is the patch I would apply to this PR

```diff
diff --git a/packages/core/src/helpers/getRenderedAttributes.ts b/packages/core/src/helpers/getRenderedAttributes.ts
index 5dd0855a2..fd85e699a 100644
--- a/packages/core/src/helpers/getRenderedAttributes.ts
+++ b/packages/core/src/helpers/getRenderedAttributes.ts
@@ -8,6 +8,9 @@ export function getRenderedAttributes(
   extensionAttributes: ExtensionAttribute[],
 ): Record<string, any> {
   return extensionAttributes
+    .filter(
+      attribute => attribute.type === nodeOrMark.type.name,
+    )
     .filter(item => item.attribute.rendered)
     .map(item => {
       if (!item.attribute.renderHTML) {
diff --git a/packages/react/src/ReactNodeViewRenderer.tsx b/packages/react/src/ReactNodeViewRenderer.tsx
index 054b69130..4f16c0c8c 100644
--- a/packages/react/src/ReactNodeViewRenderer.tsx
+++ b/packages/react/src/ReactNodeViewRenderer.tsx
@@ -304,9 +304,7 @@ export class ReactNodeView<
       let attrsObj: Record<string, string> = {}
 
       if (typeof this.options.attrs === 'function') {
-        const extensionAttributes = this.editor.extensionManager.attributes.filter(
-          attribute => attribute.type === this.node.type.name,
-        )
+        const extensionAttributes = this.editor.extensionManager.attributes
         const HTMLAttributes = getRenderedAttributes(this.node, extensionAttributes)
 
         attrsObj = this.options.attrs({ node: this.node, HTMLAttributes })

```

YaoKaiLun: That's great! But i am not sure if it will break the previous features?

nperez0111 (approved):

PR #5590, opened 2024-09-05
Allow default behaviour in stopEvent/ignoreMutation overrides

alexkuz:
## Changes Overview
In my case, I needed to override `stopEvent`, but only for a specific case. Otherwise I'd like for it to have a default behaviour. To do this, I'd have to recreate all the checks from default `stopEvent` in my override. This PR addresses this inconvenience.

## Implementation Approach
`stopEvent` ignores override only if you explicitly return a `null`. This should minimize a breaking change for existing usages.

## Checklist

## Related Issues

#257

### DIFF
--- a/packages/core/src/NodeView.ts
+++ b/packages/core/src/NodeView.ts
@@ -121,7 +121,11 @@ export class NodeView<
     }
 
     if (typeof this.options.stopEvent === 'function') {
-      return this.options.stopEvent({ event })
+      const shouldStopEvent = this.options.stopEvent({ event })
+
+      if (shouldStopEvent !== null) {
+        return shouldStopEvent
+      }
     }
 
     const target = event.target as HTMLElement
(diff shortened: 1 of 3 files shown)


packages/core/src/NodeView.ts
```diff
@@ -119,7 +119,10 @@ export class NodeView<
     }
 
     if (typeof this.options.stopEvent === 'function') {
-      return this.options.stopEvent({ event })
+      const shouldStopEvent = return this.options.stopEvent({ event })
```
nperez0111, line 122: This syntax looks wrong

alexkuz: oh, I just saw this. Editing PR directly on Github was a mistake :sweat_smile:

bdbch:
@alexkuz could you take a look at those linting issues?

alexkuz:
@bdbch all fixed

bdbch (approved):

bdbch:
Sorry - one more thing. Could you rebase this PR into **next** so we can cherry-pick it down to main/develop? We don't use develop for active dev anymore.

alexkuz:
Done!

bdbch:
@alexkuz could you rebase this to the current develop branch as we released V3, deprecating the next branch & would need to update most of our PR's anyway? Thanks a lot! I will look into getting this merged afterwards!

PR #5616, opened 2024-09-13
fix(vue-3): late-registering plugins

### LINKED ISSUE #5259: [Bug]: Bubble Menu Not Working with Collaboration Extension
### Affected Packages

Bubble menu

### Version(s)

2.4.0

### Bug Description

We encountered an issue where the Bubble Menu extension, which previously worked fine alongside the Collaboration extension, is no longer functioning after updating all dependencies. The Bubble Menu works correctly when the Collaboration extension is not included, but fails to work when Collaboration is enabled. No errors are indicated.

We discovered that the Bubble Menu plugin gets removed during the execution of the updatePluginViews function in prosemirror-view/dist/index.js. Specifically, in the y-prosemirror/src/plugins/sync phase, the length of this.state.plugins decreases from 76 (including Bubble Menu) to 75, effectively removing the Bubble Menu plugin.

Below is the basic code from the Tiptap example, which replicates the issue:

```js
<template>
  <div class="container">
    <div class="control-group">
      <label>
        <input type="checkbox" :checked="isEditable" @change="() => (isEditable = !isEditable)" />
        Editable
      </label>
    </div>
    <bubble-menu :editor="editor" :tippy-options="{ duration: 100 }" v-if="editor">
      <div class="bubble-menu">
        <button @click="editor.chain().focus().toggleBold().run()" :class="{ 'is-active': editor.isActive('bold') }">
          Bold
        </button>
        <button
          @click="editor.chain().focus().toggleItalic().run()"
          :class="{ 'is-active': editor.isActive('italic') }"
        >
          Italic
        </button>
        <button
          @click="editor.chain().focus().toggleStrike().run()"
          :class="{ 'is-active': editor.isActive('strike') }"
        >
          Strike
        </button>
      </div>
    </bubble-menu>
    <editor-content :editor="editor" />
  </div>
</template>
<script>
import StarterKit from '@tiptap/starter-kit';
import { BubbleMenu, Editor, EditorContent } from '@tiptap/vue-3';
import Collaboration from '@tiptap/extension-collaboration';
import * as Y from 'yjs';
export default {
  components: {
    EditorContent,
    BubbleMenu
  },
  data() {
    return {
      editor: null,
      isEditable: true
    };
  },
  watch: {
    isEditable(value) {
      this.editor.setEditable(value);
    }
  },
  mounted() {
    this.editor = new Editor({
      extensions: [
        StarterKit,
        Collaboration.configure({
          document: new Y.Doc()
        })
      ],
      content: `
        <p>
          Hey, try to select some text here. There will popup a menu for selecting some inline styles. Remember: you have full control about content and styling of this menu.
        </p>
      `
    });
  },
  beforeUnmount() {
    this.editor.destroy();
  }
};
</script>
```
image

### Browser Used

Chrome

### Code Example URL

_No response_

### Expected Behavior

Bubble menu working with collaboration extension

_No response_

### Dependency Updates

svenadlung:
## Changes Overview
It seems that `y-prosemirror` 1.2.4 or 1.2.5 breaks late registering plugins, specifically in Vue 3. This might be related to a point Kevin mentioned [here](https://github.com/yjs/y-prosemirror/issues/155#issuecomment-2181395723).

The issue arises because `this.view.updateState(state)` doesn't immediately update `this.view.state`. This delay means the reconfigured state in `registerPlugin()` doesn't include newly registered plugins. 

## Implementation Approach
To address this, I've adjusted the code to use the state returned by `this.state.reconfigure` directly.

## Testing Done
I tested the changes, added Vue and React demos of a collaboration setup in combination with bubble and floating menu, and added tests.

## Verification Steps
Open the `CollaborationWithMenus` Vue demo and check if collaboration and menus are working properly.

## Checklist

## Related Issues
Closes #5259


packages/vue-3/src/Editor.ts
```diff
@@ -74,19 +74,39 @@ export class Editor extends CoreEditor {
...
+      ? handlePlugins(plugin, [...this.state.plugins])
+      : [...this.state.plugins, plugin]
+
+    const state = this.state.reconfigure({ plugins })
+
+    this.view.updateState(state)
+
     if (this.reactiveState) {
-      this.reactiveState.value = this.view.state
+      this.reactiveState.value = state
```
nperez0111, line 86: This is the only real difference from what it was doing before no? Why do we need to duplicate what the editor class is doing?

The same was done for the unregistering, so I'd like to understand better what the actual problem was

svenadlung: You are right. I’ve added more context to the PR's description. I'm using the reconfigured state directly since `this.view.state` is not updated at that point.

nperez0111 (approved):
I refactored it a little to not have to expose the internals to the vue package

Issue #3823, opened 2023-03-06
getPos returning undefined


michelmar:
### What’s the bug you are facing?

I'm trying to update to the last version of tiptap, but I'm facing a problem that was not happening before, the function getPos which should return only numbers, is now returning undefined as well, my solution, for now, is to treat this separately, but this isn't very pleasant. 

### Which browser was this experienced in? Are any special extensions installed?

Chrome

### How can we reproduce the bug on our side?

Should have a custom node view in react, and use getPos(), in the begin it'll return undefined

### What did you expect to happen?

getPos : ()=> number
should return only numbers

### Did you update your dependencies?

- [X] Yes, I’ve updated my dependencies to use the latest version of all packages.

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

michelmar:
Actually I think this function is expected to return undefined, as I see in the prosemirror view package;
export type NodeViewConstructor = (node: Node, view: EditorView, getPos: () => number | undefined,
                                   decorations: readonly Decoration[], innerDecorations: DecorationSource) => NodeView

So we could say that this function is wrongly annotated in tiptap .

The prosemirror documentation says this about the undefined case:
```
/// The third argument `getPos` is a
  /// function that can be called to get the node's current position,
  /// which can be useful when creating transactions to update it.
  /// Note that if the node is not in the document, the position
  /// returned by this function will be `undefined`.
```

PR #5623, opened 2024-09-17
Update `getPos` type inside `NodeViewRendererProps` to reflect actual functionality

WilliamIPark:
## Changes Overview
The type for `getPos` inside of `NodeViewRendererProps` was incorrectly typed to indicate that it always returns a number, when this is not the actual case. It can also return `undefined`.

As per the ProseMirror documentation:
> The third argument getPos is a function that can be called to get the node's current position, which can be useful when creating transactions to update it. **Note that if the node is not in the document, the position returned by this function will be undefined**.

Trusting this function to always return a number has lead to some stability issues in our application, so I believe it's best that the type reflects the actual functionality, even if this may cause some headaches to handle `undefined` cases. 

## Implementation Approach

Update the type to point towards `Parameters<NodeViewConstructor>[2]`, which I believe inherits off `pm/view`.

## Testing Done

- Ran test suite, including cypress tests. All passing.

## Verification Steps

- Try to use `getPos` inside of a `NodeView`.

## Additional Notes
I'd consider this a breaking change, as it could flare up people's type checks if they're not handling the possibility of the return value being `undefined`,  so I have marked it as needing a major version bump in the changeset.

## Checklist

## Related Issues
https://github.com/ueberdosis/tiptap/issues/3823

### DIFF
--- a/.changeset/green-wolves-arrive.md
+++ b/.changeset/green-wolves-arrive.md
@@ -1,0 +1,9 @@
+---
+"@tiptap/core": major
+---
+
+Fix `getPos` type in `NodeViewRendererProps` to potentially be `undefined`
+
+Breaking change: Types may flag uses of getPos where an `undefined` possibility isn't handled.
+Why this change was made: To ensure the type reflects the real functionality of this function.
+How to update: Ensure that the return value of `getPos` exists before making use of the value.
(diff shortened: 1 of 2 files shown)


nperez0111:
As mentioned in the comment, this has to go towards a new major version so this PR is going to sit for a while

WilliamIPark:
Yep sounds good :) Just wanted to make sure it wasn't left behind when that time comes!

nperez0111 (approved):
I'll merge this into `next` branch

PR #5674, opened 2024-09-27
feat(CharacterCount): Add counter function to `CharacterCount` extension.

ho991217:
## Changes Overview
Made `CharacterCount`'s text and word counting logic to accessible from outside, which can solve issues with string containing emoji.

## Implementation Approach
I've just added these two options
```tsx
export interface CharacterCountOptions {
  /**
   * The text counter function to use. Defaults to a simple character count.
   * @default (text) => text.length
   * @example (text) => [...new Intl.Segmenter().segment(text)].length
   */
  textCounter: (text: string) => number
  /**
   * The word counter function to use. Defaults to a simple word count.
   * @default (text) => text.split(' ').filter(word => word !== '').length
   * @example (text) => text.split(/\s+/).filter(word => word !== '').length
   */
  wordCounter: (text: string) => number
}
```
and default options
```ts
      textCounter: text => text.length,
      wordCounter: text => text.split(' ').filter(word => word !== '').length,
```

## Checklist

### DIFF
--- a/.changeset/calm-bottles-shout.md
+++ b/.changeset/calm-bottles-shout.md
@@ -1,0 +1,5 @@
+---
+"@tiptap/extension-character-count": minor
+---
+
+Make counter function accessible from outside character-count extension
--- a/packages/extension-character-count/src/character-count.ts
+++ b/packages/extension-character-count/src/character-count.ts
@@ -16,6 +16,18 @@ export interface CharacterCountOptions {
    * @example 'textSize'
    */
   mode: 'textSize' | 'nodeSize'
+  /**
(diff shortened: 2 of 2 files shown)


gethari:
just a thought !
> how about adding a flag that can ignore/consider emoji as a character ? user's can decided between the flags ?

bdbch (approved):
Nice idea - LGTM

nperez0111:
> how about adding a flag that can ignore/consider emoji as a character ? user's can decided between the flags ?
I think that a function gives the most flexibility to the end user on how they would want to implement this. Let's just go with this for now

Issue #5681, opened 2024-09-30
[Bug]: The PluginView be trigger destroy on create Tiptap instance.


yangfuhai:
### Affected Packages

core

### Version(s)

2.7.4

### Bug Description

```ts
export const MyExtension = Extension.create<ClickMenuOptions>({
  name: "MyExtension",
  addProseMirrorPlugins() {
    const destroy = () => {
      console.log("destroy!!!!")
    }
    return [
      new Plugin({
        key: new PluginKey(`MyExtension`),
        view: () => ({ destroy }),
      }),
    ];
  },
});
```

the method `destroy` will be invoked twice in Tiptap create.

>ps: This problem does not exist in tiptap v2.6x.

### Browser Used

Chrome

### Expected Behavior

the destory method not be trigger in tiptap create.

### Dependency Updates

- [X] Yes, I've updated all my dependencies.

nperez0111:
Thanks for bringing this up. Was the new onDrop & onPaste events that were always registering, I have a fix in this https://github.com/ueberdosis/tiptap/pull/5684

nperez0111:
This has been released with tiptap v2.8.0

PR #5684, opened 2024-09-30
fix(core): onDrop and onPaste only registered when defined #5681

nperez0111:
## Changes Overview

This updates the `onDrop` and `onPaste` event callbacks to only ever be registered if they exist on the editor instance at creation.
It also refactors them to be extensions rather than separate Prosemirror plugins which would causes plugins to be destroyed and recreated immediately
## Checklist


packages/core/src/Editor.ts
```diff
@@ -112,14 +111,8 @@ export class Editor extends EventEmitter<EditorEvents> {
...
-
-    if (this.options.onPaste) {
-      this.registerPlugin(PastePlugin(this.options.onPaste))
-    }
-
-    if (this.options.onDrop) {
-      this.registerPlugin(DropPlugin(this.options.onDrop))
-    }
+    this.on('drop', ({ event, slice, moved }) => this.options.onDrop(event, slice, moved))
+    this.on('paste', ({ event, slice }) => this.options.onPaste(event, slice))
```
nperez0111, line 115: I figured, that while I was here, I may as well promote these to be actual editor events that follow the same pattern as others. This allows flexibility to register multiple handlers to listen to the same event

bdbch (approved):
LGTM

packages/core/src/Editor.ts
svenadlung: @nperez0111 This resolves [this](https://github.com/ueberdosis/tiptap/discussions/5613), right?

nperez0111: @svenadlung it does 😉

Issue #5691, opened 2024-10-01
[Bug]: `@tiptap/extension-text-style` won't install in starter-kit


ydfzgyj:
### Affected Packages

starter-kit

### Version(s)

2.8.0

### Bug Description

 `@tiptap/extension-text-style` had been added to  `@tiptap/extension-bullet-list` and  `@tiptap/extension-ordered-list` as peerDependencies, so it should also be added to dependencies of `@tiptap/starter-kit`

### Browser Used

Chrome

### Expected Behavior

add `@tiptap/extension-text-style` to dependencies of `@tiptap/starter-kit`

### Dependency Updates

- [X] Yes, I've updated all my dependencies.

nperez0111:
Yea this is correct. Would accept a PR to fix it

Korywon:
Temporary workaround is to include `@tiptap/extension-text-style` as a dependency in our app. Not great, but it works for now.

nperez0111:
Dupe of https://github.com/ueberdosis/tiptap/issues/5707

nperez0111:
This is released with v2.9.0

PR #5693, opened 2024-10-02
fix: add extension-text-style as a dep of @tiptap/starter-kit #5691

nperez0111:
## Changes Overview

Adds @tiptap/extension-text-style to @tiptap/starter-kit deps but does not install the extension, since it is only to resolve a peer dep install for list-items
## Checklist

## Related Issues

#5691

### DIFF
--- a/.changeset/thirty-apes-boil.md
+++ b/.changeset/thirty-apes-boil.md
@@ -1,0 +1,5 @@
+---
+"@tiptap/starter-kit": patch
+---
+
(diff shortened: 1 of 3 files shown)


Nantris:
I haven't gotten to test 2.9.0 yet, but I think this will only resolve the issue for projects using the starter-kit, but not for those directly importing the numbered or bullet list extension, one of which iirc is the source of this issue.

nperez0111:
Right, should move this into a dep: https://github.com/ueberdosis/tiptap/blob/cbbb8a5d717932ab17047f8d8bb24a21e30c8f6b/packages/extension-bullet-list/package.json#L39

PR #5717, opened 2024-10-10
Fix getMarkRange not finding marks when at the start of a mark

### LINKED ISSUE #5715: [Bug]: getMarkRange returned an incorrect value
### Affected Packages

core

### Version(s)

2.8.0

### Bug Description

I recorded a video to demonstrate this bug.

https://github.com/user-attachments/assets/b0cfbcec-0da6-4920-8f1a-7400bc31e70e

As shown in the figure, clicking the link icon will select the entire link. This is achieved by using `editor.commands.extendMarkRange('link')`

截屏2024-10-10 11 52 33

It works as expected

截屏2024-10-10 11 55 04

But if the selection is like the one below, it will not work as expected

截屏2024-10-10 12 02 10

This problem only occurs when the `$from` of the selection is at the beginning of the link
After debugging I found out that this is because `getMarkRange` returns undefined

截屏2024-10-10 12 04 31
截屏2024-10-10 12 05 08

And if I change the first parameter of `getMarkRange` from `$from` to `$to` I can get it to work correctly.

I use the following code to meet my needs. But I prefer the simple `editor.commands.extendMarkRange('link')`

```js
      const range = getMarkRange(
        editor.state.selection.$to,
        editor.schema.marks.link,
      );

      if (range) {
        editor.commands.setTextSelection(range);
        setPopover('link');
      }
```

Hope this problem can be solved

### Browser Used

Chrome

### Code Example URL

_No response_

### Expected Behavior

`getMarkRange` returns correct range.

_No response_

### Dependency Updates

bdbch:
## Changes Overview

This pull request addresses an issue in the `@tiptap/core` package where the `getMarkRange` function did not return the correct range when the cursor was at the start of the specified mark. The changes involve both documentation and code modifications to ensure accurate behavior.

### Bug Fix:

* [`packages/core/src/helpers/getMarkRange.ts`](diffhunk://#diff-e265670b2c081e4c8486eb69ad5c34a81f449ba84871895b072de287ed55ab4bL32-R45): Updated the `getMarkRange` function to correctly handle cases where the cursor is at the start of a text node without the mark, ensuring it looks backward and returns the appropriate range.

## Implementation Approach
You can find out more about my implementation from the comments in code. I basically added a forward, then backward check for marks on the current nodes parent with the parent offset.

## Testing Done
Tested it locally on my machine in a demo I build. (See below)

## Verification Steps
My demo:

```
import './styles.scss'

import Link from '@tiptap/extension-link'
import { EditorContent, getMarkRange, useEditor } from '@tiptap/react'
import React, { useCallback } from 'react'

export default () => {
  const editor = useEditor({
    extensions: [
      Document,
      Paragraph,
      Text,
      Link.configure({ openOnClick: false }),
    ],
    content: `
      <p>
        This is a radically reduced version of Tiptap. It has support for a document, with paragraphs and text. That’s it. It’s probably too much for real minimalists though.
      </p>
      <p>
        This is a <a href="#">Link</a> with some text afterwards.
      </p>
      <p>
        The paragraph extension is not really required, but you need at least one node. Sure, that node can be something different.
      </p>
    `,
  })

  const handleClick = useCallback(() => {
    const range = getMarkRange(
      editor.state.selection.$from,
      editor.schema.marks.link,
    )

    console.log(editor.state.selection.from, range)
  }, [editor])

  return (
    <div>
      <div>
        <button onClick={handleClick}>Expand selection to mark</button>
      </div>
    </div>
  )
}
```

## Additional Notes
Nothing to add here

## Checklist

## Related Issues
Fixes #5715

### DIFF
--- a/tests/cypress/integration/core/getMarkRange.spec.ts
+++ b/tests/cypress/integration/core/getMarkRange.spec.ts
@@ -1,0 +1,87 @@
+/// <reference types="cypress" />
+
+import {
+  getMarkRange,
+  getSchemaByResolvedExtensions,
+} from '@tiptap/core'
+import Document from '@tiptap/extension-document'
+import Link from '@tiptap/extension-link'
+import Paragraph from '@tiptap/extension-paragraph'
+import Text from '@tiptap/extension-text'
+import { Node } from '@tiptap/pm/model'
+
+describe('getMarkRange', () => {
+  const document = {
+    type: 'doc',
+    content: [
+      {
+        type: 'paragraph',
+        content: [
+          { type: 'text', text: 'This is a ' },
+          { type: 'text', text: 'linked', marks: [{ type: 'link', attrs: { href: 'https://tiptap.dev' } }] },
+          { type: 'text', text: ' text.' },
+        ],
+      },
+    ],
+  }
+
+  const schema = getSchemaByResolvedExtensions([
+    Document,
+    Paragraph,
+    Text,
+    Link.configure({ openOnClick: false }),
+  ])
+
+  it('gets the correct range for a position inside the mark', () => {
+    const doc = Node.fromJSON(schema, document)
+    const $pos = doc.resolve(14)
+    const range = getMarkRange($pos, schema.marks.link)
+
+    expect(range).to.deep.eq({
+      from: 11,
+      to: 17,
+    })
+
+    // eslint-disable-next-line no-console
+    console.log(range)
+  })
+
+  it('gets the correct range for a position at the start of the mark', () => {
+    const doc = Node.fromJSON(schema, document)
+    const $pos = doc.resolve(11)
+    const range = getMarkRange($pos, schema.marks.link)
+
+    expect(range).to.deep.eq({
+      from: 11,
+      to: 17,
+    })
+
+    // eslint-disable-next-line no-console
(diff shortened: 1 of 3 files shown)


tests/cypress/integration/core/getMarkRange.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+
+  it('gets undefined if a mark is not found', () => {
+    const doc = Node.fromJSON(schema, document)
+    const $pos = doc.resolve(6)
+    const range = getMarkRange($pos, schema.marks.link)
+
+    expect(range).to.eq(undefined)
+
+    // eslint-disable-next-line no-console
+    console.log(range)
```
nperez0111, line 85: Can remove these

tests/cypress/integration/core/getMarkRange.spec.ts
nperez0111, line 1: Could we add a test to prove that the look back won't cross a node boundary?

Like two paragraphs with the first having the mark we are looking for but we use the position of the first text position in the second paragraph. 

The danger of looking backward is that it can misreport so I just want to prove that it doesn't in a test

nperez0111 (approved):
Other than the test file, looks good to me

Issue #5564, opened 2024-08-28
[Bug]: Expose an API to overwrite the default link validation


rfgamaral:
### Affected Packages

extension-link

### Version(s)

2.6.6

### Bug Description

We've finally integrated the latest Tiptap version with our product, and unfortunately [this](https://github.com/ueberdosis/tiptap/pull/5470) and [this](https://github.com/ueberdosis/tiptap/pull/5160) caused some issues for us.

We've started getting loads of support tickets of people complaining that custom links copied from the most various apps could no longer be used to link text in the editor. To give a few examples, apps such as Airmail, Bear, Obsidian, give you links with custom protocols (i.e. `airmail://`, `bear://`, `obsidian://`), and these stopped working. As you can imagine, figuring out a list of all supported custom protocols for all the apps out there will be nearly impossible.

To work around this issue (it was really starting to get critical), we had to put up a patch to remove link validation, but we would love to have a more clean implementation, and remove the patch as soon as possible.

### Expected Behavior

I think the best option is to keep the default behaviour as-is, but somehow allow the extension to be configured (or extended) in a way that allows to completely customize/overwrite the link validation behaviour in the most clean possible way.

### Additional Context (Optional)

> But I do think that the default behavior is quite valid & safe for a web-based editor that isn't really going to deal with custom protocols as much, but I of course understand your concern & it should be accounted for.

_(by @nperez0111 - [source](https://github.com/ueberdosis/tiptap/pull/5470#issuecomment-2315898498))_

I agree that the default behaviour is valid and safe for a web-based editor, but for more advanced and/or different usages, it would be useful to be able to customize the behaviour for different needs.

---

As an aside, I think there's a bit of a disconnect between the default `allowedProtocols` list, and the `registerCustomProtocol` function. For instance, you are allowed to have links with the `tel`, `callto`, `sms`, `cid`, and `xmpp` protocols, but these are not auto-linked, because they are not registered as custom protocols (the other ones seem to be accepted by linkify.js by default, despite not being directly registered either). Something to consider when reviewing/refactoring the implementation.

### Dependency Updates

- [X] Yes, I've updated all my dependencies.

rfgamaral:
@nperez0111 If I work on a PR for this, that gives the option to the consumer to have full control on link validation while having the default link validation as-is, would that be something that you be willing to accept? Or do you think this needs a different solution?

nperez0111:
[...]

PR #5726, opened 2024-10-14
Added optional custom uri validator in link extension

SanderLeenders:
## Changes Overview
Added custom url validation to link extension

## Implementation Approach
Added an optional custom url validation to provide full control over (in)valid uri's

## Testing Done
Added test 

## Checklist

## Related Issues
[[Bug]: Expose an API to overwrite the default link validation](https://github.com/ueberdosis/tiptap/issues/5564)

### DIFF
--- a/packages/extension-link/src/link.ts
+++ b/packages/extension-link/src/link.ts
@@ -78,6 +78,13 @@ export interface LinkOptions {
    * @returns - True if the url is valid, false otherwise.
    */
   validate: (url: string) => boolean
+
+  /**
+   * A custom validation function for uri's.
+   * @param url - The url to be validated.
+   * @returns - True if the url is valid, false otherwise.
+   */
+  customUriValidator?: (url: string | undefined) => boolean
 }
 
 declare module '@tiptap/core' {
@@ -124,6 +131,13 @@ function isAllowedUri(uri: string | undefined, protocols?: LinkOptions['protocol
 
   // eslint-disable-next-line no-useless-escape
   return !uri || uri.replace(ATTR_WHITESPACE, '').match(new RegExp(`^(?:(?:${allowedProtocols.join('|')}):|[^a-z]|[a-z+.\-]+(?:[^a-z+.\-:]|$))`, 'i'))
+}
+
+function validateUri(uri: string | undefined, protocols?: LinkOptions['protocols'], customUriValidator?: LinkOptions['customUriValidator']) {
+  if (customUriValidator) {
+    return customUriValidator(uri)
+  }
+  return isAllowedUri(uri, protocols)
 }
 
 /**
@@ -200,7 +214,7 @@ export const Link = Mark.create<LinkOptions>({
         const href = (dom as HTMLElement).getAttribute('href')
 
         // prevent XSS attacks
-        if (!href || !isAllowedUri(href, this.options.protocols)) {
+        if (!href || !validateUri(href, this.options.protocols, this.options.customUriValidator)) {
           return false
         }
         return null
@@ -210,7 +224,7 @@ export const Link = Mark.create<LinkOptions>({
 
   renderHTML({ HTMLAttributes }) {
     // prevent XSS attacks
-    if (!isAllowedUri(HTMLAttributes.href, this.options.protocols)) {
+    if (!validateUri(HTMLAttributes.href, this.options.protocols, this.options.customUriValidator)) {
       // strip out the href
       return ['a', mergeAttributes(this.options.HTMLAttributes, { ...HTMLAttributes, href: '' }), 0]
     }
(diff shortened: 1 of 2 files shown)


packages/extension-link/src/link.ts
```diff
@@ -78,6 +78,13 @@ export interface LinkOptions {
    * @returns - True if the url is valid, false otherwise.
    */
   validate: (url: string) => boolean
+
+  /**
+   * A custom validation function for uri's.
+   * @param url - The url to be validated.
+   * @returns - True if the url is valid, false otherwise.
+   */
+  customUriValidator?: (url: string | undefined) => boolean
```
rfgamaral, line 87: I like this, but now we have two validation options:

- `validate`
- `customUriValidator`

This could be confusing, especially since they don't apply to the same places. Maybe we should merge the implementation into a single validation function that is used everywhere?

SanderLeenders: I see what you mean, and I have thought about it. However, there must be a reason why the validate function was not used when this change was made: https://github.com/ueberdosis/tiptap/pull/5470/files#diff-1d1f108bd30f776ec9689dc1bb00a3cb3dafc89fde595994aa8fa400471a87d6. Therefore, I have decided to keep these functionalities separate.

rfgamaral: @nperez0111 What are your thoughts here? Also, would it be possible to have this one fully reviewed soon? 🙏

nperez0111: I, too, would not make them be separate functions since they are too similar in name.

I can't speak to why this what not done like this here: https://github.com/ueberdosis/tiptap/pull/5470/files#diff-1d1f108bd30f776ec9689dc1bb00a3cb3dafc89fde595994aa8fa400471a87d6 but, I think that if a custom validator is passed, it is expected to be used properly.


Ultimately, I'd have:
```ts
validate: (url: string, defaultValidate: (url: string) => boolean ) => boolean
```

This would allow the user to specify their own validation, and if they don't want to both writing their own rules, they can use our existing `defaultValidate` function (which currently maps to `isAllowedUri` with the protocols being prefilled.

SanderLeenders: Yes, they are similar in name, but they serve different purposes. The validate function is solely used for the autolink option, while the newly added one ensures that there are no malicious URLs. If these functionalities were handled by a single function, it would be impossible to differentiate between them, meaning all valid links would be autolinked. I don't think that's something you would want. For example, should relative links be autolinked? In my use case, this is something I do not want.

nperez0111: I see your point with auto linking but I feel that validation should be about validity. How about we add a new callback for shouldAutoLink where you could reject a relative link separately from validation?

SanderLeenders: Please correct me if I'm mistaken, but it sounds like you're suggesting a simple rename from `validate: (url: string) => boolean` to `shouldAutoLink(validate: (url: string) => boolean)`, right? From what I understand, the `validate` method is an optional/custom method meant to override the default behavior of the `autolink: boolean`. 

The main issue seems to be that the method’s name doesn't clearly reflect its purpose  —the validate function doesn’t actually validate links, but rather checks whether a link should be auto-linked. 

The `customUriValidator` method could be renamed to `validate`, but personally, I’d keep the name `customUriValidator` since it more clearly describes the function’s role."

nperez0111: To make my case, take this example:

```ts
Link.configure({
 validate: () => true,
 customUriValidator: () => true
})
```

This does not "more clearly describes the function’s role", it makes it confusing to know which applies to auto-linking and which applies to actual validation.

The current `validate` function, as I understand, is used in two cases, for auto-linking and matching against pasted text when looking for links. But, we are introducing a 3rd case where it will be useful for determining what we consider to be a "valid" URL (i.e. not malicious).

You wanted to support the ability to differentiate what would be considered for auto-linking from what is considered a valid link. I offered you a way to do this, a new function which would be specifically for auto-linking behavior to choose when something would be auto-linked (assuming it was already valid).

So, this is what I'm proposing:
```ts
Link.configure({
 validate: (url: string, defaultValidate: (url: string) => boolean) => boolean
 shouldAutoLink: (url: string) => boolean
})
```

If a link is not valid (according to the validate function), then it is not considered as a link that can be linked to and will be dropped.
If a link is considered valid (according to the validate function), then the `shouldAutoLink` function will be called to allow rejecting links that we may consider valid, but do not want to auto-link to.

Does this sound reasonable?

SanderLeenders:
The pull request below appears to contain the suggested change, so I’m closing this one: https://github.com/ueberdosis/tiptap/pull/5808

Issue #5547, opened 2024-08-24
[Bug]: Mod-z Not Blocked When editor.isEditable Is False


jan-schweiger:
### Affected Packages

core, react, @tiptap/extension-history

### Version(s)

2.6.6

### Bug Description
1. Edit the content
2. Run editor.setEditable(false)
3. Press CTRL + Z
4. The undo operation is performed

However, this only happens for Undo but not for Redo.

### Reproduce
I was also able to reproduce it by slightly adjusting the minimal setup: https://tiptap.dev/docs/examples/basics/minimal-setup

```
import Document from '@tiptap/extension-document'
import Paragraph from '@tiptap/extension-paragraph'
import Text from '@tiptap/extension-text'
import { EditorContent, useEditor } from '@tiptap/react'
import React from 'react'
import History from '@tiptap/extension-history'

function App() {
  const editor = useEditor({
    extensions: [
      Document,
      Paragraph,
      Text,
      History  // <-- Added History
    ],
    content: "<p>Tiptap content</p>",
  })

  return (
    <div className="App">
      <EditorContent editor={editor} />

      <button onClick={() => editor.setEditable(!editor.isEditable)}>
        Toggle isEdit
      </button>
    </div>
  );
}

export default App;
```

You just need to edit some content and click the button to experience the bug for yourself.

Thank you very much in advance for your help!

### Browser Used

Chrome

### Expected Behavior

Mod + Z shouldn't change the content when the editor is read-only.

### Dependency Updates

- [X] Yes, I've updated all my dependencies.

jan-schweiger:
I also tried the following quickfix. But strangely the Mod-z function is only called when the isEditable is true.

```
     History.extend({
        addKeyboardShortcuts() {
          return {
            'Mod-z': () => {
              console.log('Mod-z', this.editor.isEditable);
              return this.editor.isEditable ? this.editor.commands.undo() : false;
            },
            'Mod-Shift-z': () => {
              return this.editor.isEditable ? this.editor.commands.redo() : false;
            },
          };
        },
      }),
```

tanja-kovakic:
I can also reproduce the bug. Just tried it out in my application. Can anyone have a look on this one?

nperez0111:
Maybe this is a browser default behavior? I wonder if the element is still a contenteditable when editable=false

rosenbauerwillbacker:
I can also reproduce the bug.

[...]

PR #5731, opened 2024-10-16
Fix undo/redo still worked with isEditable false

Nantris:
## Changes Overview
Block undo/redo when `editable.isEditable` isn't truthy.

## Testing Done
None

## Verification Steps
None

## Checklist

## Related Issues
#5547

### DIFF
--- a/packages/extension-history/src/history.ts
+++ b/packages/extension-history/src/history.ts
@@ -55,10 +55,12 @@ export const History = Extension.create<HistoryOptions>({
 
   addCommands() {
     return {
-      undo: () => ({ state, dispatch }) => {
+      undo: () => ({ editor, state, dispatch }) => {
+        if (!editor.isEditable) return false
         return undo(state, dispatch)
       },
-      redo: () => ({ state, dispatch }) => {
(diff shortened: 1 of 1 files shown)


Nantris:
I'm not sure why this check is necessary. None of the other commands seem to need it. Something seems not right about this.

Nantris:
I don't understand how/why this breaks building the examples. Is that unrelated to the PR? Looking at other recent PRs I think it might be,

Nantris:
Friendly bump on this. Not sure this is the best approach, but the editor isn't production-ready with this bug present.

nperez0111:
We tried this solution & it does not actually works @Nantris which is why we've not taken it into the project. 

prosemirror-history has default keybinds that listen for the historyUndo event: https://github.com/ProseMirror/prosemirror-history/blob/master/src/history.ts#L409-L419 which is still listening even when the editor is not editable.

So this should be fixed upstream.


I can assure you that the editor is production ready, millions use it

Nantris:
Whoops I don't think I meant to re-open this. I agree this PR doesn't work.

Issue #5254, opened 2024-06-19
[Bug]: Text align is applied to a single paragraph only


svilen-ivanov-kubit:
### Affected Packages

core, extension-text-align

### Version(s)

2.5.0-beta.4

### Bug Description

If you select several paragraphs and apply an alignment, only one paragraph is updated. Here is an example from https://tiptap.dev/docs/editor/api/extensions/text-align

https://github.com/ueberdosis/tiptap/assets/92510182/387b6c8b-eb92-4e69-9d55-52038d20295b

See how _first paragraph_ never aligns.

The problem is not reproducible with 2.4.0. I was able to track the problem to a recent change of the `updateAttributes` command: https://github.com/ueberdosis/tiptap/pull/5154

### Browser Used

Chrome

### Expected Behavior

All paragraphs should align.

### Dependency Updates

- [X] Yes, I've updated all my dependencies.

nperez0111:
I've reverted this change for now. I could not spend the time at the moment to fix the original bug

nperez0111:
Ok, releeasing that with https://github.com/ueberdosis/tiptap/releases/tag/v2.5.0-beta.5

PR #5738, opened 2024-10-17
improve updateAttributes

silenius:
## Changes Overview
Currently `updateAttributes` updates attributes of all nodes (or marks) of specified `typeOrName` to new values.
It works well, except with nested nodes of the same type. This is because the implementation uses `node.nodesBetween` which return all nodes in the selection range but also **nodes that cover it** (and which are outside the selection range).

This is a second attempt (first was in #5154 ) to fix this. The first version was broken (#5254) because it didn't handled range selection properly.

## Implementation Approach
We first check if we have a selection range or if the selection is empty. If the selection is empty we simply loops over all nodes returned by `nodesBetween` and we keep the last one which match `typeOrName`.
If we have a selection range we also have to keep track of the last node which match `typeOrName` because it could be outside the `(from, to)` selection range, for example with:

```html
<div class="flex">
  <article>
    <p>foo</p>
    <p>bar</p>
  </article>
</div>
```
if the user text select 'ooba' and that we `updateAttributes` of an _article_ `typeOrName` we must be sure that the `<article>` node is included.

If the selection range is empty we also update nodes, but we check that it is in the selection range.

## Checklist

## Related Issues
#3545 #4466 #5154


silenius:
there is a single test failing .. but I don't know why

silenius:
:heavy_check_mark:

nperez0111 (approved):
Thanks @silenius for your contribution, I hope we've got it this time!

Issue #3229, opened 2022-09-24
.can() returns false for extension-list even when the command is runnable


C-Hess:
### What’s the bug you are facing?

The `.can()` command is a bit overzealous for the list extension. If the cursor/selection is on an existing list or heading element, the `can()` command returns false for `toggleBulletList()` and `toggleOrderedList()` commands. However, these commands are still executable, and can/will replace the current node with the new list node type selected.

### Which browser was this experienced in? Are any special extensions installed?

- Chrome
- Bullet lists
- Ordered lists
- Headings

### How can we reproduce the bug on our side?

1. Open the sandbox
    1. The sandbox is similar to the demo. The only difference is that the buttons are disabled if the `can()` command for the corresponding controls returns false
3. Create a bullet or ordered list with one item in the list
4. Note that the other list type and heading are disabled
5. Create another list item at the same level
6. Note that if the second list item in a list is selected with the cursor, they are enabled again
7. Nest the second list item with the tab key
8. Note that if two list items exist, but the cursor is on the first list item, **the header is disabled**  :confused:

### Can you provide a CodeSandbox?

https://codesandbox.io/s/frosty-goldwasser-visq8r?file=/src/App.tsx

### What did you expect to happen?

The `.can()` command should accurately reflect whether a command is possible to run given the editor state. In this case, because it is totally possible to switch between lists or convert a list to a heading, the `.can()` command should return true for all of the described cases above.

### Anything to add? (optional)

It's possible that the heading being disabled when the nested list item is selected is a separate issue, but I figured I'd leave it here in case it's the same. Otherwise, let me know, and I can open another issue

This issue is related to #3223, but that `pr` fixes issues with the `.can()` command for marks

### Did you update your dependencies?

- [X] Yes, I’ve updated my dependencies to use the latest version of all packages.

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

corbinschwartz:
Also running into this, repro steps in original description are accurate.

Goodosky:
The same problem occurs with `setHeading()` and `toggleHeading()`

If the selection is on an existing list, both commands are runnable, but `.can()` returns false.

gethari:
https://github.com/ueberdosis/tiptap/issues/4057#issuecomment-4021559366

Issue #3025, opened 2022-07-26
dispatch=undefined in createCan() results in dispatch attempts during can()


kivikakk:
### What’s the bug you are facing?

I've hit a weird issue that requires a combination of three things:

1. A custom node that nests another custom node, which in turn contains inlines.
   a. It needs to have higher priority than `Paragraph`.
2. `editor.can().toggleBulletList()` (or `editor.can().toggleOrderedList()`)
3.  A selection from within a paragraph _after_ the custom node into the inline in the custom node.

### Which browser was this experienced in? Are any special extensions installed?

Latest Chrome on macOS, but unlikely to be browser-specific.

### How can we reproduce the bug on our side?

Here's a sandbox those shows the issue: <https://codesandbox.io/s/tiptap-dispatch-repro-gc54w8?file=/src/App.js>

Try making a selection like this:

<img width="265" alt="image" src="https://user-images.githubusercontent.com/1915/180921841-0c4b669e-b227-4f2b-84bc-56cb9fff2d51.png">

`editor?.can().toggleBulletList()` will cause a `RangeError`, specifically from here:

https://github.com/ueberdosis/tiptap/blob/e2f8747e2b10f98d49798b15d64db5f0a36a3427/packages/core/src/commands/clearNodes.ts#L41-L45

The `tr.setNodeMarkup` call throws:

```
RangeError
Invalid content for node type mycontainer
```

See [relevant PM code](https://github.com/ProseMirror/prosemirror-transform/blob/1906d3698d92dfecfcd2c656d54bfe1ce2b391de/src/structure.ts#L146-L147). At the point of error, `type` is the NodeType for `MyContainer`, and `node.content` is a fragment containing TextNode(s).

### What did you expect to happen?

I'm surprised that a `can()` call could `throw` an error at all, regardless of the schema. `clearNodes` checks for whether it should dispatch:

https://github.com/ueberdosis/tiptap/blob/e2f8747e2b10f98d49798b15d64db5f0a36a3427/packages/core/src/commands/clearNodes.ts#L20-L22

But in this case, `dispatch` is `() => void 0`. That value comes from `CommandManager.buildProps`:

https://github.com/ueberdosis/tiptap/blob/e2f8747e2b10f98d49798b15d64db5f0a36a3427/packages/core/src/CommandManager.ts#L141-L143

`shouldDispatch` is true. This is surprising to me. Here's the relevant part of `CommandManager.createCan`:

[...]

Issue #5721, opened 2024-10-12
[Bug]: Using clearNodes in the can() command will return incorrect results


guanriyue:
### Affected Packages

core, extension-bullet-list

### Version(s)

2.8.0

### Bug Description

_My English isn't very good, so I use ChatGPT to help with translations. If there are any inaccuracies in my expressions, please let me know!_

The `can()` command returns false when the command can actually be executed.

https://github.com/user-attachments/assets/c55da5b6-7750-428d-859a-23586aabc078

I tried debugging and fixing this issue, but in the end, it seems like it's actually something that needs to be handled by tiptap.

I have some hypotheses, but I can't confirm if they are correct. They might introduce new issues or cause breaking changes.

There are several related issues in tiptap's issues section, which we will mention in the following explanation.
- https://github.com/ueberdosis/tiptap/issues/3229
- https://github.com/ueberdosis/tiptap/issues/4057
- https://github.com/ueberdosis/tiptap/issues/3743
- https://github.com/ueberdosis/tiptap/issues/3025

——————————————————————————————————————————————————————————————————
### The most immediate issue occurs in the `clearNodes` command
In the screen recording, we see that `editor.can().toggleHeading({ level: 1 })` and `editor.commands.toggleHeading({ level: 1 })` return different values, so let's first take a look at what's happening inside the `toggleHeading` command.

```js
toggleHeading: attributes => ({ commands }) => {
  if (!this.options.levels.includes(attributes.level)) {
    return false
  }

  return commands.toggleNode(this.name, 'paragraph', attributes)
},
```
`toggleNode` calls `setNode`, so let's directly take a look at setNode.
The original code comments use `//`.
The comments below are my own additions, marked with `////` for clarity.
We made two markers, A and B. The state at marker A is referred to as stateA, and the state at marker B is referred to as stateB (updatedState). As expected, stateA and stateB should be different.
```ts
export const setNode: RawCommands['setNode'] = (typeOrName, attributes = {}) => ({ state, dispatch, chain }) => {
  //// Skipping a small portion of unnecessary code.

  return (
    chain()
    // try to convert node to default node if needed
      .command(({ commands }) => {
        //// Mark A (StateA)
        const canSetBlock = setBlockType(type, attributes)(state)

        //// At this point, we attempt to determine whether it is allowed to directly modify list item 
        //// nodes into headings.
        //// Clearly, this is an invalid operation. So the value of canSetBlock is false
        if (canSetBlock) {
          return true
        }

[...]

Issue #4057, opened 2023-05-18
[Bug]: can command not working correctly with list


ashu12chi:
### Which packages did you experience the bug in?

extension-ordered-list, extension-unordered-list, extension-task-list, core

### What Tiptap version are you using?

2.0.3

### What’s the bug you are facing?

We have [can command](https://tiptap.dev/api/editor#can) in tiptap to check if a command can be executed - without actually executing it. I am using this command in my editor to check weather [toggleOrderedList](https://tiptap.dev/api/nodes/ordered-list#toggle-ordered-list), [toggleBulletList](https://tiptap.dev/api/nodes/bullet-list#toggle-bullet-list), and [toggleTaskList](https://tiptap.dev/api/nodes/task-list#toggle-task-list) can be executed or not.

These commands are returning `false` for the first item of an already present list, although we are able to execute these commands.

For example, if I have a bullet list, and I am at the first item of the list, then `can` command is returning I can't run `toggleOrderedList`, but actually we can run the command to change this list to an ordered list.

In below image, ordered-list is present and bullet-list button is `disabled` when we are at first item:
![image](https://github.com/ueberdosis/tiptap/assets/45908393/8d1732a6-5354-43d4-937e-53541c1359ea)

In below image, ordered-list is present and bullet-list button is `not disabled` when we are at second item:
![image](https://github.com/ueberdosis/tiptap/assets/45908393/2cc9a32f-85a4-4ecc-a768-40197d0d63b8)

We can execute the command in the both the cases can be verified here: https://tiptap.dev/examples/default

### What browser are you using?

Chrome

### Code example

https://codesandbox.io/s/sleepy-heyrovsky-fwddzb?file=/src/App.js

### What did you expect to happen?

`can` command should work correctly.

### Did you update your dependencies?

- [X] Yes, I’ve updated my dependencies to use the latest version of all packages.

### Are you sponsoring us?

- [X] Yes, I’m a sponsor. 💖

mejo-:
Seems like quite a few commands get reported as not supported even though they can be used. E.g. `editor.can().toggleBlockquote()` inside a bullet list returns false, while `editor.can().toggleCodeBlock()` returns true.

Slapbox:
Same issue. It seems to be related to #3743 as has already been mentioned by @mejo-

gethari:
@bdbch / @nperez0111 is this still an issue ? I could replicate this even in tiptap's website. Please check screenshot

[...]

PR #5744, opened 2024-10-21
Fixed issue where can() returned false when the command was executable

guanriyue:
## Changes Overview

Fix the issue where can() check returns false even when the command is actually executable. And added cypress test cases.

## Implementation Approach

When chainCommand is executed, multiple commands are run sequentially. The editor state used by later commands may depend on the preceding commands. Therefore, when executing can(), createChain must ensure that shouldDispatch is true to prevent skipping operations on tr.

This fix effectively reverts the code changes from the pull request https://github.com/ueberdosis/tiptap/pull/3026 and resolves the issue https://github.com/ueberdosis/tiptap/issues/3025 caused by the clearNodes error.

At the same time, it will also fix issue https://github.com/ueberdosis/tiptap/issues/3229 and issue https://github.com/ueberdosis/tiptap/issues/4057.

For more details, please refer to this issue https://github.com/ueberdosis/tiptap/issues/5721.

## Testing Done

First, add the disabled attribute to the heading and list action buttons in the example default editor.

Before the fix, if the selection was on a heading node, the list action buttons would remain disabled, even though the commands were actually available. After the change, the list action buttons will appear enabled.

Additionally, corresponding Cypress test cases have been added.

## Verification Steps

Access the default example editor at https://tiptap.dev/docs/examples/basics/default-text-editor or by executing cypress test

## Checklist

## Related Issues

- https://github.com/ueberdosis/tiptap/issues/5721
- https://github.com/ueberdosis/tiptap/issues/3229
- https://github.com/ueberdosis/tiptap/issues/4057
- https://github.com/ueberdosis/tiptap/issues/3025


nperez0111:
Hey @guanriyue I haven't reviewed this super thoroughly, because I think that this approach is wrong.

What `can` is meant for is to be able to allow commands to check whether they would be able to execute without actually modifying the document via transactions. At times, this can be sort of ambiguous, so sometimes a lazy implementation may not actually implement this properly. I think that `clearNodes` in this case is just implemented improperly where it should not have that early dispatch.

I do not believe that you need to actually change `can` in this case. I think that what needs to happen is there needs to be a better implementation of `clearNodes`.

packages/core/src/CommandManager.ts
```diff
@@ -101,7 +101,7 @@ export class CommandManager {
 
   public createCan(startTr?: Transaction): CanCommands {
     const { rawCommands, state } = this
-    const dispatch = false
+    const dispatch = true
```
nperez0111, line 104: I believe this is wrong.

If dispatch is true, commands will think that they can modify the `transaction` when they should not be able to.

`dispatch` should probably be named `shouldDispatch` where it being `false` means you should not, as a command, actually apply your change to the editor.

tests/cypress/integration/core/can.spec.ts
nperez0111, line 1: This test was correct. Please revert

packages/core/src/commands/clearNodes.ts
nperez0111, line 1: The actual problem in this file is:
```ts
  if (!dispatch) {
    return true
  }
```

nperez0111 (changes requested):

guanriyue:
Have you looked at the description of this issue https://github.com/ueberdosis/tiptap/issues/5721? I explained why I believe `dispatch` should be a function in `can.chain`.

Perhaps it is a mistake; in fact, setting `dispatch` to true was part of the code logic before the last bug fix https://github.com/ueberdosis/tiptap/pull/3026. It's actually a bit of a long story.

Four years ago, at the very beginning, the `dispatch` variable was set to `false`.
https://github.com/ueberdosis/tiptap/commit/7bfab4685f958be82a3519c6cbd377977c88794d#diff-ac93a7861b6cdca5f5dd83ec422e904ebb145fd00a39343834384c6d241965c6R103

About five months later, it was changed to `undefined` for unclear reasons. When passed into `buildProps`, JavaScript treated it as true by default. These two changes come from the same author.
https://github.com/ueberdosis/tiptap/commit/ca8d1a42450be9d41adf64697b607c3ca8697f4f#diff-ac93a7861b6cdca5f5dd83ec422e904ebb145fd00a39343834384c6d241965c6L87-R90

About a year and a half later, `clearNodes` caused an error, but the issue with `clearNodes` was not fixed, and executing `command.clearNodes` could still lead to problems, while the implementation of `createCan` was modified. `dispatch` was reset to `false`.
https://github.com/ueberdosis/tiptap/commit/40f4ea31efc59c1578b783a5b64572ac2846c5a6#diff-ac93a7861b6cdca5f5dd83ec422e904ebb145fd00a39343834384c6d241965c6L109-R111

After that fix, several new issues with can checks emerged.
- https://github.com/ueberdosis/tiptap/issues/3229
- https://github.com/ueberdosis/tiptap/issues/4057

I am now trying to change it back to `true` .  And fix the real issue in `clearNodes`. It runs all the tests successfully on my local machine, except for the `can.chain.dispatch.except.to.be.undefined`, which is a test case added after the fix https://github.com/ueberdosis/tiptap/pull/3026.

> If dispatch is true, commands will think that they can modify the transaction when they should not be able to.

Yes, you're right. However, in `can.chain`, it seems that we should not skip modifications on `tr` under any circumstances, as subsequent commands may depend on the state and `tr` after the preceding command is executed. Just like how skipping `clearNodes` caused issues. There are many commands similar to `clearNodes`. In our pursuit of higher performance and avoiding unnecessary code execution, we tend to favor lazy implementations.

> dispatch should probably be named shouldDispatch where it being false means you should not, as a command, actually apply your change to the editor.

Yes, that is an inappropriate naming; it should be renamed to `shouldDispatch`.

packages/core/src/commands/clearNodes.ts
guanriyue: Remove this section of code will resolve the current issue. However, it seems there is a deeper problem that needs to be addressed, which would involve a significant change.

I will take your advice and work to close this PR as soon as possible.

nperez0111:
I'm sorry, but, I still believe that this is all misguided. Please refer to the documentation of Prosemirror https://prosemirror.net/docs/guide/#commands It specifically says that:

> To figure out whether a selection can currently be deleted, you'd call deleteSelection(view.state, null), whereas to actually execute the command, you'd do something like deleteSelection(view.state, view.dispatch). A menu bar could use this to determine which menu items to gray out.

> In this form, commands do not get access to the actual editor view—most commands don't need that, and in this way they can be applied and tested in settings that don't have a view available. But some commands do need to interact with the DOM—they might need to [query](https://prosemirror.net/docs/ref/#view.EditorView.endOfTextblock) whether a given position is at the end of a textblock, or want to open a dialog positioned relative to the view. For this purpose, most plugins that call commands will give them a third argument, which is the whole view.

> function blinkView(_state, dispatch, view) {
>  if (dispatch) {
>    view.dom.style.background = "yellow"
>    setTimeout(() => view.dom.style.background = "", 1000)
>  }
>  return true
> }
> That (rather useless) example shows that commands don't have to dispatch a transaction—they are called for their side effect, which is usually to dispatch a transaction, but may also be something else, such as popping up a dialog.

It specifically says it should be called like `deleteSelection(view.state, null)`, with the implementation as:

```js
 function blinkView(_state, dispatch, view) {
  if (dispatch) {
    view.dom.style.background = "yellow"
    setTimeout(() => view.dom.style.background = "", 1000)
  }
  return true
 }
```

This is done purposefully, so that a command can be "tried" without actually committing the side-effect. If you pass `true` you defeat the purpose of it as you'll modify the editor content with the transaction or side-effect.


You are not solving the real problem which is the implementation of the command. You are attempting to change how can chains work to fit a wrong implementation, fix the implementation not how can chains work.

packages/core/src/commands/clearNodes.ts
nperez0111: This is what the PR should be addressing. Everything else around this PR is incorrect with the intention behind can chains and how they should work

Issue #4434, opened 2023-09-09
[Bug]: unsetBlockquote causes listItems to be lifted instead of blockquote


Slapbox:
### Which packages did you experience the bug in?

core

### What Tiptap version are you using?

2.1.8

### What’s the bug you are facing?

This is reproducible if you use the underlying command `commands.lift('blockquote')` from inside of a list inside of a `blockquote`. The `listItem` will be lifted and the `blockquote` will remain.

### What browser are you using?

Chrome

### What did you expect to happen?

Lift list out of blockquote, not lift paragraph out of listItem.

### Anything to add? (optional)

I assume this must be a mistake in the underlying `lift` command, but that leverages ProseMirror so I'm not sure what the true root cause might be.

### Did you update your dependencies?

- [X] Yes, I’ve updated my dependencies to use the latest version of all packages.

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

Slapbox:
Same with `toggleBlockquote`

The commands calls `toggleWrap` which then calls `lift` which lifts the nearest node, not the correct node.

I don't know how to fix this but I hope we can get some attention on this.

Slapbox:
I've discovered this issue can not only unwrap lists unexpectedly, but for nested `listItems` it can remove their content from the editor entirely, whereas I'd expect it would simply do nothing. Can anyone from the core team comment on this? Is there a plan to fix the blockquote commands? They're the only basic commands that don't seem to work.

Slapbox:
Friendly bump.

Issue #3743, opened 2023-02-14
Unable to convert list to blockqoute 


varun-raj:
### What’s the bug you are facing?

When I tried to convert a list into a blockquote it isnt working but when I tried the vice versa it is working. Even in the tiptap example.

https://user-images.githubusercontent.com/3992409/218647584-32f5653f-ce4c-438f-9ca3-02fe16d2613c.mov

### Which browser was this experienced in? Are any special extensions installed?

Happens in Chrome Version 109.0.5414.119 (Official Build) (arm64)

### How can we reproduce the bug on our side?

1. Open tiptap editor https://tiptap.dev/examples/default
2. Type a text and make it to ordered list
3. Click on blockquote menu and you see nothing happens
4. Now try making a blockquote and converting it to a list and it works

### What did you expect to happen?

Its happening all the time.

### Did you update your dependencies?

- [X] Yes, I’ve updated my dependencies to use the latest version of all packages.

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

bstro:
Experiencing this as well in my own project, 
```
"@tiptap/react": "^2.0.2",
"@tiptap/starter-kit": "^2.0.2"
```

mejo-:
This is most likely related to #4057.

bdbch:
I think the problem is that Prosemirror tries to turn the `paragraph` node inside the `listItem` into a blockquote which is not allowed in the schema while the user would expect the list item to turn into a blockquote. While this would be cool for simple lists - what happens, if you do this in a nested list?

Or just what happens when you do this on the 2nd item inside list? Would you split the list at that item and lift it out and turn it into a blockquote?

I'd say the blockQuote button should be disabled in this case because it's not usable inside the lists paragraph. For that we first need to fix the `can` command.

Slapbox:
Possibly related to #4434

Slapbox:
@bdbch can you or anyone else on the team comment further on this and on #4434?

I don't feel that patching `can` constitutes a real solution here.

With nested lists, if the user wraps the entire list in their selection the expected behavior in my mind would be to wrap that list in a blockquote but currently nothing happens. 

For a `listItem` with a single top-level list, I'd expect an `empty` selection in the top node to wrap the entire list in a blockquote.

For more complex use cases, I agree they do not need to be supported.

These behaviors can at least be meaningfully replicated today by including the paragraph after the list in the selection before calling `toggleBlockquote`.

[...]

PR #5746, opened 2024-10-22
Fix/blockquote commands

guanriyue:
## Changes Overview

Fix the issues with `setBlockquote` and `unsetBlockquote` commands.

## Implementation Approach

Tiptap checks `isNodeActive` by searching upwards from the current position, and based on the result, it executes either `wrapIn` or `lift`. However, ProseMirror's `wrapIn` and `lift` only evaluate the current `blockRange`, which can cause the command to not function as expected. 

To address this, add a `findWrappingUntil` helper, which will search upwards until it finds a valid wrapping position. This will resolve issue https://github.com/ueberdosis/tiptap/issues/3743.

Additionally, add a `nodeType` check when retrieving `blockRange` in the `lift` command, which will resolve issue https://github.com/ueberdosis/tiptap/issues/4434.

## Testing Done

In the example default editor, a `disabled` attribute was added to the blockquote action button. This will disable the button when the action cannot be performed. The `id` attribute was also added for better Cypress testing.

Three test cases were added:
1. When the cursor is on the first child element of a list item, `toggleBlockquote` will wrap the entire list.
2. When the cursor is on the second child element of a list item, `toggleBlockquote` will wrap only the second element.
3. For a list already wrapped in a blockquote, when the cursor is on the first child element of a list item, `toggleBlockquote` will remove the blockquote wrapping without altering the list.

## Verification Steps

Visit the default editor page or run the Cypress tests.

## Additional Notes

It seems we should directly modify `commands.wrapIn` instead of changing `setBlockquote` in the blockquote extension. The range of checks in `ProseMirror.wrapIn` and `tiptap.isNodeActive` are inconsistent, which may be the core issue. 

On the other hand, the situation with `commands.lift` is the opposite, as `lift` has a specific node type. We must search upwards from the current position to find the required type.

## Checklist

## Related Issues

- https://github.com/ueberdosis/tiptap/issues/3743
- https://github.com/ueberdosis/tiptap/issues/4434


bdbch:
@guanriyue could you rebase this to the current develop branch as we released V3, deprecating the next branch & would need to update most of our PR's anyway? Thanks a lot!

PR #5759, opened 2024-10-23
feat(core): add support for markviews

nperez0111:
## Changes Overview

This adds support for Prosemirror markviews to `Mark`s.

For this to actually become supported we will need to:

 - [x] Have a MarkView class that can manage the markview element instead of having to create your own elements
 - [x] Have a way to render React / Vue components like node views does

From there, we'd pretty much just support them!

Here is what the API looks like right now, but we will probably add a class to make these elements for you:

```ts
  addMarkView() {
    return ({ mark, HTMLAttributes }) => {
      const dom = document.createElement('b')
      const contentDOM = document.createElement('span')

      dom.appendChild(contentDOM)

      return {
        dom,
        contentDOM,
      }
    }
  },
```

I've implemented both React rendering for mark views and Vue 3 rendering (I couldn't be bothered to do vue-2, we should probably deprecate that package).

## Implementation Approach

Pretty much following the same sort of implementation as node views, I was able to add support for markviews.

## Verification Steps

I added a temporary markview to a mark and it worked just fine
## Checklist


svenadlung (approved):

bdbch (approved):
Awesome idea, LGTM!

bdbch:
Tests failed and we'd need to merge with develop again as we got a few merge conflicts.

Looking forward to get this merged.

bdbch:
Should we change the target branch to next then?

bdbch:
@nperez0111 let me know when you need another code review on this one.

nperez0111:
@bdbch close to a review here, would probably want to add some tests that it works properly. I also found out that prosemirror-dev-tools extension does not have support for markviews yet and causes it to render incorrectly, so disable the extension to get it to run properly. I made an issue so that they know to fix it: https://github.com/TeemuKoivisto/prosemirror-dev-toolkit/issues/84

nperez0111:
Alright, all done here @bdbch

Issue #5752, opened 2024-10-22
[Bug]: Vue - Editor view destruction before Transition end


Ericlm:
### Affected Packages

vue

### Version(s)

2.9.0

### Bug Description

The issue is very close to the one described in #5646, concerning the fact that the content of the DOM is destroyed before Vue finished a Transition, so the editor content disappears during the leaving transition.

But unlike the issue above, it is linked to the hierarchy of component. Basically, this is something like:
- Parent has `<Transition>` and a child with a v-if
- Child component has the `EditorContent` component

When parent turns off the condition, the `useEditor` in child calls the `onBeforeUnmount` hook that calls the `editor.destroy`, finally removing all DOM nodes. But this call to PM leads to the editor content being removed while the editor stills visible.

It seems more complicated than #5646 as it may needs to go to the core of tiptap. The linked fork of tiptap shows the explained issue by restructuring the Transition example with Vue.

I'd be willing to make a PR to fix this, but I don't really know what's the best way to solve this.

### Browser Used

Firefox

### Code Example URL

https://github.com/Ericlm/tiptap

### Expected Behavior

Transition on a parent should not cause a child component to lose it's tiptap content, and let Vue destroying all children nodes at the end.

### Additional Context (Optional)

This issue is directly linked to:
- #5646 
- #5648

### Dependency Updates

- [X] Yes, I've updated all my dependencies.

Ericlm:
Maybe the `destroy` method needs a way to avoid destroying DOM nodes (to let Vue or other framework do that).

nperez0111:
Not sure what the solution here would be, but I'd accept a PR to fix it.

Prosemirror does need to be the one to do the destroy though since it also needs to clean up nodeviews. 

This maybe have been the purpose of trying to copy that element before destruction that we saw here: https://github.com/ueberdosis/tiptap/pull/5648/files#diff-6b499b16726df128fdb6e3206be8b2f4cfc407ebf838ee23d98dd514a93efab4

Ericlm:
Yes, copying the DOM just before destroy seems to be the right thing to do. I'm working on it to make sure it'll work in all cases. Maybe after that Vue Transitions will be good in every case 😄

nperez0111:
Released with 2.10.0

Issue #5646, opened 2024-09-23
[Bug]: Vue 3 - editor destroyed before transitioning out is done 


Ericlm:
### Affected Packages

@tiptap/vue-3

### Version(s)

2.6.6^

### Bug Description

When using the Vue package, and wrapping the editor within a Transition with v-if makes the editor disappear before the end of the transition. The onDestroy hook is called before unmounting other Vue components.

### Browser Used

Firefox

### Code Example URL

https://codesandbox.io/p/devbox/dazzling-kate-rczsdc

### Expected Behavior

The editor should be destroyed only when the transition has ended.

### Additional Context (Optional)

Linked issue : #2296

### Dependency Updates

- [X] Yes, I've updated all my dependencies.

nperez0111:
I would totally accept a patch here, I don't have enough knowledge of vue to know what the right solution is here

nperez0111:
This is released with v2.9.0

PR #5772, opened 2024-10-26
fix(vue-3): Transition with editor destruction

Ericlm:
## Changes Overview

I changed the `useEditor` composable so that it supports transitions if the editor is in the same component as the Transition, and now also if the editor itself is a descendant.

## Implementation Approach

I made a DOM clone of the content before the destroy in the useEditor composable. Then Vue removes the nodes when transition is over.

## Testing Done

- I added both nested and not nested components to the Transition example.
- I fixed e2e tests for two editors instead of only one

## Verification Steps

Going to the examples section and to Transition. In the `index.vue` component, both examples are shown. One example where the editor is hosted in the same component as the transition, and one where it's nested.

## Checklist

## Related Issues

#5752 and #5646 

See also #5648


nperez0111 (approved):
This looks good to me, appreciate your contribution!

PR #5786, opened 2024-10-30
feature: add global node type name attributes to editor DOM

bdbch:
## Changes Overview
This pull request introduces a new feature to add type attributes to nodes and marks in the editor for better identification, which can be useful for testing and styling purposes. The changes span across multiple files to implement and support this feature.

This makes it easier to style specific HTML elements of the editor while not overwriting any styles for node view related elements or other decorations that may unwillingly be styled.

If enabled, a node could be rendered like this in HTML:

```html
<p data-tiptap="paragraph">
  Hello <strong data-tiptap"bold">World</strong>, this is a <em data-tiptap="italic">test</em> to test node-type data attributes.
</p>
```

## Implementation Approach
I added a new option called `addTypeAttributes` to the editor which is global and enabled by default. This `addTypeAttributes` option enables a `Typenames` extension that will dynamically add those type names to the editor DOM.

Those attributes are currently **stripped** from the editors output.

## Testing Done
I tested this all locally on any test of our demos pages.

## Verification Steps
Clone and check the branch for your self or check the deployed netlify page.

## Additional Notes
I don't think this is a breaking change as it's not really changing the general behavior of the editor and the `data-node-type` attribute is vague enough to not be used by the general user base.

## Checklist


packages/core/src/extensions/typenames.ts
```diff
@@ -0,0 +1,19 @@
+import { Extension } from '../Extension.js'
+
+export const Typenames = Extension.create({
```
svenadlung, line 3: maybe better `TypeNames`?

packages/core/src/Editor.ts
```diff
@@ -80,6 +81,7 @@ export class Editor extends EventEmitter<EditorEvents> {
     enablePasteRules: true,
     enableCoreExtensions: true,
     enableContentCheck: false,
+    addTypeAttributes: true,
```
nperez0111, line 84: Can we make this default `false` for backwards-compat?

v3 we can do it default on

bdbch: done

svenadlung: I have the feeling that this is not really clear. But I don't have any alternatives either 😂

`enableHelperAttributes`, `renderDOMHelperAttributes`, …

packages/core/src/extensions/typenames.ts
```diff
@@ -0,0 +1,19 @@
...
+  name: 'typenames',
```
nperez0111, line 4: typenames?

Maybe `namedAttributes`? Or `labeledAttributes`?

packages/core/src/extensions/typenames.ts
```diff
@@ -0,0 +1,19 @@
...
+  addGlobalAttributes() {
+    return this.extensions.filter(extension => {
+      return extension.name !== 'text' && extension.name !== 'doc'
+    }).map(extension => ({
+      types: [extension.name],
+      attributes: {
+        'data-tiptap': {
+          default: extension.name,
+          rendered: true,
+        },
```
nperez0111, line 15: Does this get stored to the Tiptap JSON? Kind of redundant if it does

packages/react/src/ReactNodeViewRenderer.tsx
```diff
@@ -150,6 +150,10 @@ export class ReactNodeView<
       className: `node-${this.node.type.name} ${className}`.trim(),
     })
 
+    if (this.editor.options.addTypeAttributes) {
+      this.renderer.element.setAttribute('data-node-view', '')
```
svenadlung, line 154: Maybe `data-node-view-wrapper`, which is more consistent to the usage with `data-node-view-content`?

packages/core/src/extensions/typenames.ts
bdbch: But this adds the nodetype names of nodes and marks? Wouldn't typenames make sense?

packages/core/src/helpers/getHTMLFromFragment.ts
```diff
@@ -3,6 +3,11 @@ import { DOMSerializer, Fragment, Schema } from '@tiptap/pm/model'
 export function getHTMLFromFragment(fragment: Fragment, schema: Schema): string {
   const documentFragment = DOMSerializer.fromSchema(schema).serializeFragment(fragment)
 
+  // remove the data-node-type attribute
```
nperez0111, line 6: comment is off now.

packages/react/src/ReactNodeViewRenderer.tsx
```diff
@@ -150,6 +150,10 @@ export class ReactNodeView<
...
+    }
```
nperez0111, line 155: `props` in ReactRenderer sets attributes to the element, we should use that. This can cause layout shift

packages/core/src/extensions/typenames.ts
nperez0111: typename just is super generic.

It should be clear that this is applying to the HTMLAttributes of nodes & marks, unsure of a good name for that.

packages/core/src/extensions/typenames.ts
bdbch: Don't think those are added to the Tiptap JSON.

packages/react/src/ReactNodeViewRenderer.tsx
bdbch: done

packages/core/src/Editor.ts
bdbch: Removed it in favor of just using the `enableCoreExtensions` options. So by default it's disabled via `enableCoreExtensions: { typenames: false }` - but can be enabled by just setting this to true.

This is how we do it for the other core extensions already.

bdbch:
Just figured - the `addGlobalAttributes` function will add all attributes to the Tiptap JSON which is what we don't want. I'll see if I can find a different solution again.

demos/src/Examples/Default/React/index.jsx
```diff
@@ -10,6 +10,8 @@ import React from 'react'
 const MenuBar = () => {
   const { editor } = useCurrentEditor()
 
+  window.editor = editor
```
StephanMeijer, line 13: I think you accidentally left this in?

PR #5808, opened 2024-11-06
refactor: adjust validate and add shouldAutoLink to improve URL handling

guarmo:
## Changes Overview
Added a new `shouldAutoLink` function to control the auto-linking of URLs based on specific criteria, such as disallowed domains, and enhanced the existing `validate` function to handle URLs more robustly.

## Testing Done
- Created tests to verify that shouldAutoLink does not auto-link URLs with disallowed domains and correctly auto-links valid URLs.
- Updated integration tests for validate to ensure URLs are correctly validated with or without protocols. Verified that URLs with disallowed protocols and domains are not linked.

## Checklist

## Related Issues
#5564

### DIFF
--- a/demos/src/Marks/Link/React/index.spec.js
+++ b/demos/src/Marks/Link/React/index.spec.js
@@ -136,7 +136,7 @@ context('/src/Marks/Link/React/', () => {
     disallowedProtocols.forEach(url => {
       cy.get('.tiptap').then(([{ editor }]) => {
         editor.commands.setContent(`<p><a href="${url}">Example Text</a></p>`)
-        expect(editor.getHTML()).to.not.include(`<a href="${url}">`)
+        expect(editor.getHTML()).to.not.include(url)
       })
     })
   })
@@ -147,7 +147,7 @@ context('/src/Marks/Link/React/', () => {
     disallowedDomains.forEach(url => {
       cy.get('.tiptap').then(([{ editor }]) => {
         editor.commands.setContent(`<p><a href="${url}">Example Text</a></p>`)
-        expect(editor.getHTML()).to.not.include(`<a href="${url}">`)
+        expect(editor.getHTML()).to.not.include(url)
       })
     })
   })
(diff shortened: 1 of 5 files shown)


packages/extension-link/src/link.ts
```diff
@@ -75,9 +75,21 @@ export interface LinkOptions {
   /**
    * A validation function that modifies link verification for the auto linker.
```
nperez0111, line 76: I guess this isn't just for the auto linker any more right

rfgamaral:
@guarmo I can't afford to look into this carefully right now, but does this PR implement the proposed solution by @nperez0111 mentioned [here](https://github.com/ueberdosis/tiptap/pull/5726#discussion_r1807933435) (an older PR that aims to fix the same issue).

demos/src/Marks/Link/React/index.spec.js
```diff
@@ -129,4 +129,32 @@ context('/src/Marks/Link/React/', () => {
...
+
+    disallowedDomains.forEach(url => {
+      cy.get('.tiptap').then(([{ editor }]) => {
+        editor.commands.setContent(`<p><a href="${url}">Example Text</a></p>`)
+        expect(editor.getHTML()).to.not.include(`<a href="${url}">`)
+      })
+    })
+  })
+
+  it('should not auto-link a URL from a disallowed domain', () => {
```
nperez0111, line 155: good test 👍

packages/extension-link/src/link.ts
```diff
@@ -75,9 +75,21 @@ export interface LinkOptions {
...
    * @param url - The url to be validated.
+   * @param ctx - An object containing:
+   *              - `defaultValidate`: A function that performs the default URL validation.
+   *              - `protocols`: An array of allowed protocols for the URL (e.g., "http", "https").
+   *              - `defaultProtocol`: A string that represents the default protocol (e.g. 'http')
```
nperez0111, line 81: You can do jsdoc on each value like this:

```
{
/** comment **/
defaultValidate: (url: string) => boolean;
/** another **/
protocols: Array<LinkProtocolOptions | string>;
/** and again **/
defaultProtocol: string
}
```

This allows typescript to show you what each key of that object does

nperez0111 (approved):
Once you've got that. This is gtg

SanderLeenders:
What about backward compatibility? For example, what if someone previously only wanted to autolink URLs starting with `https://`. They may have added a validation function like:
``` validate: url => url.startsWith('https://),```
With the current change, only URLs starting with https:// would be valid, which is not the same behavior as before.

PR #5810, opened 2024-11-07
List Keymap: Fix backspace behavior when selection is not collapsed

### LINKED ISSUE #4368: [Bug]: ListKeymap Uncaught RangeError: Position -1 out of range

Isolated to only list-keymap, happens in tiptap.dev example as well

2.1.6

Steps to reproduce:
* Open https://tiptap.dev/api/extensions/list-keymap, scroll to live editor
* Add bullet list (item) as first block
* Add plain paragraph as second block
* Select all and press `Backspace`

Get the error: `Uncaught RangeError: Position -1 out of range`.
Content is removed and editor seems to be usable after the error without having to reload.

Observations:
* Doesn't happen with bullet nor ordered list if `list-keymap` is not present
* Described issue occurs only if there's 1 bullet list item
* Doesn't happen when you press `Delete` key
* Bullet list (item) nor following paragraph need to contain any content, error is thrown even if both are empty
* Don't see the issue if bullet list (item) is not the first block in the document

---

#### Just mentioning, possibly related/same cause:
Same situation as before, only difference is that there are >1 bullet list items

Issue: last list item is removed and content is shuffled around.

```html
Starting doc:
<ul>
  <li>A</li>
  <li>B</li>
</ul>
<p>C</p>

*Select all and hit Backspace*

End up with doc:
<ul>
  <li>BA</li>
</ul>
```

Chrome

### Code example

https://tiptap.dev/api/extensions/list-keymap

Paragraph and bullet list are deleted and replaced with empty enforced paragraph.

_No response_

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

juraj98:
## Changes Overview
Fixes #4368

## Implementation Approach
I figured that just checking if selection's `to` and `from` doesn't match, default backspace behavior will result in expected behavior. So I added check to [handleBackspace.ts](https://github.com/ueberdosis/tiptap/pull/5810/files#diff-15ca0926de038c0b181f99b43c4af4614ecb69b3fe559da219a9001d606bf750R17).

## Testing Done
N/A

## Verification Steps
1. Start development environment and navigate to `/preview/Extensions/ListKeymap`. 
2. Create a list with one or multiple list items, and paragraph underneath. 
3. Select all and hit backspace.

If you created list with one list item there will be no error in console. If you created list with multiple list items, the behavior will be as expected. Compare this with previous behavior either of tiptap's `develop` or [live example](https://tiptap.dev/docs/editor/extensions/functionality/listkeymap)

## Checklist

## Related Issues
#4368

### DIFF
--- a/.changeset/chatty-monkeys-hear.md
+++ b/.changeset/chatty-monkeys-hear.md
@@ -1,0 +1,5 @@
+---
+"@tiptap/extension-list-keymap": patch
+---
+
+Fix backspace behavior when selection is not collapsed
--- a/packages/extension-list-keymap/src/listHelpers/handleBackspace.ts
+++ b/packages/extension-list-keymap/src/listHelpers/handleBackspace.ts
@@ -12,6 +12,12 @@ export const handleBackspace = (editor: Editor, name: string, parentListTypes: s
     return true
   }
 
+  // if the selection is not collapsed
+  // we can rely on the default backspace behavior
+  if (editor.state.selection.from !== editor.state.selection.to) {
+    return false
+  }
+
   // if the current item is NOT inside a list item &
   // the previous item is a list (orderedList or bulletList)
   // move the cursor into the list and delete the current item


nperez0111 (approved):
Looks correct to me

Issue #5813, opened 2024-11-07
[Bug]: editor.commands.setContent does not set attributes on Document node


StephanMeijer:
### Affected Packages

core

### Version(s)

2.91

### Bug Description

On the Document Node, attributes were not set properly when using `editor.commands.setContent`.

### Browser Used

Chrome

### Code Example URL

https://codesandbox.io/p/sandbox/adoring-elion-vvydqz

### Expected Behavior

Attributes on Document to be set, like it was set on Paragraph.

### Dependency Updates

- [X] Yes, I've updated all my dependencies.

Manaskhurana:
some times bug heppen with new React and React-Dom version

1. try to downgrade react virsion and if your working with next.js try to downgrade it also to fix the issue
2. due new updates of the frame works make difficult solve the issue
3. delete node_modules and reinstall may fix this issue
4. install latest version of tiptap

StephanMeijer:
> some times bug heppen with new React and React-Dom version
> 
> 1. try to downgrade react virsion and if your working with next.js try to downgrade it also to fix the issue
> 2. due new updates of the frame works make difficult solve the issue
> 3. delete node_modules and reinstall may fix this issue
> 4. install latest version of tiptap

Not the case.

StephanMeijer:
Closed. See https://github.com/ueberdosis/tiptap/pull/5814

PR #5814, opened 2024-11-07
fix(core): editor.commands.setContent setting document attributes

StephanMeijer:
Related to #5813. Fixes editor.command.setContent not setting attribute on document.

## Changes Overview

Related to #5813. Fixes editor.command.setContent not setting attribute on document.

## Implementation Approach

As ProseMirror doesn't seem to be replacing the document attributes, we now do that manually.

## Testing Done

Manually, but feel free to recommend me. Also see #5813

## Verification Steps

```ts
import { Editor } from '../tiptap/packages/core'
import Document from '../tiptap/packages/extension-document'
import Paragraph from '../tiptap/packages/extension-paragraph'
import Text from '../tiptap/packages/extension-text'


const editor = new Editor({
    element: document.querySelector('.element'),
    extensions: [
        Document.extend({
            addAttributes() {
                return {
                    ...this?.parent?.(),
                    testAttr1: { default: 'DEFAULT' }
                }
            }
        }),
        Paragraph.extend({
            addAttributes() {
                return {
                    ...this?.parent?.(),
                    testAttr2: { default: 'DEFAULT' }
                }
            }
        }),,
        Text
    ],
    content: '<p>Default content</p>',
})

editor.commands.setContent({ 
    type: "doc", 
    attrs: { testAttr1: "Not default" },
    content: [
        {
            type: "paragraph",
            attrs: { testAttr2: "Not default" },
            content: [
                { type: "text", text: "Updated content" }
            ]
        }
    ]
})

console.log(editor.state.doc.attrs); // { testAttr1: "Not default" }
console.log(editor.state.doc.content.firstChild.attrs); // { testAttr2: "Not default" }
```

## Checklist

## Related Issues
#5813

### DIFF
--- a/packages/core/src/commands/setContent.ts
+++ b/packages/core/src/commands/setContent.ts
@@ -58,6 +58,10 @@ export const setContent: RawCommands['setContent'] = (content, emitUpdate = fals
 
     if (dispatch) {
       tr.replaceWith(0, doc.content.size, document).setMeta('preventUpdate', !emitUpdate)
+
+      Object.entries(document.attrs).forEach(([key, value]) => {
+        tr.setDocAttribute(key, value)
+      })
     }
     return true
   }
--- a/.changeset/wise-oranges-deny.md
+++ b/.changeset/wise-oranges-deny.md
@@ -1,0 +1,5 @@
+---
+"@tiptap/core": patch
+---
+
+Fixes editor.command.setContent not setting attributes on the Document Node.


packages/core/src/commands/setContent.ts
```diff
@@ -58,6 +58,10 @@ export const setContent: RawCommands['setContent'] = (content, emitUpdate = fals
 
     if (dispatch) {
       tr.replaceWith(0, doc.content.size, document).setMeta('preventUpdate', !emitUpdate)
+
+      Object.entries(document.attrs).forEach(([key, value]) => {
+        tr.setDocAttribute(key, value)
+      })
```
nperez0111, line 64: We are planning to remove this branch in the next version (and already have).

So, I don't think that this is the right place for it.

I'm not sure that a method named `setContent` _should_ update the doc's attributes. I think you may want to implement this in your application instead. I don't think it belongs in this command

nperez0111: We tend not to promote doc attributes because they are not synchronized in collaboration with y-prosemirror because of a long standing bug.

So, we don't want to promote something that would not work in the Tiptap ecosystem

StephanMeijer: Alright, I do understand.

nperez0111:
Appreciate the contribution though & your understanding. Will close this as a won't do for now

PR #5838, opened 2024-11-17
fix emptyTextBlock detection to handle leaf nodes too

### LINKED ISSUE #4327: [Bug]: `renderText` does not propagate to `NodeSpec.leafText` and `extension-floating-menu` uses `Node.textContent` which uses `NodeSpec.leafText`

bdbch:
## Changes Overview

This pull request includes a small but significant change to the `FloatingMenuView` class in the `packages/extension-floating-menu/src/floating-menu-plugin.ts` file. The change ensures that the `isEmptyTextBlock` condition also checks if the parent node has no children (`childCount === 0`).

This fixes a bug (see #4327) causing leaf nodes not to be counted as content of a node. It also deprecates the PR #4444

Changes in `FloatingMenuView`:

* Updated the `isEmptyTextBlock` condition to include a check for `childCount === 0` to ensure the block is truly empty.

## Checklist

## Related Issues
- Closes #4327


nperez0111:
I wonder if this fix is still valid https://github.com/ueberdosis/tiptap/pull/4444

bdbch:
Do we need to be that specific in that case though? I thought about it and I couldn't find a reason why a node should be considered empty even when it has any child nodes.

nperez0111:
> Do we need to be that specific in that case though? I thought about it and I couldn't find a reason why a node should be considered empty even when it has any child nodes. Do you have any thing in mind that would cause a node to have child content but still should be considered empty for the floating menu?

I think you understood me reverse, if a node is something that can serialize to text (e.g. a mention), which has no child nodes, then to the floating menu it will be considered empy.

I did this in the current version to show what I mean
image


You can apply this diff in your project to see if your fix addresses this too, but I'm pretty sure the other fix addresses this:

```diff
diff --git a/demos/src/Nodes/Mention/React/index.jsx b/demos/src/Nodes/Mention/React/index.jsx
index 8d40a8d4d..ef78a95b6 100644
--- a/demos/src/Nodes/Mention/React/index.jsx
+++ b/demos/src/Nodes/Mention/React/index.jsx
@@ -4,7 +4,7 @@ import Document from '@tiptap/extension-document'
 import Mention from '@tiptap/extension-mention'
 import Paragraph from '@tiptap/extension-paragraph'
 import Text from '@tiptap/extension-text'
-import { EditorContent, useEditor } from '@tiptap/react'
+import { EditorContent, FloatingMenu, useEditor } from '@tiptap/react'
 import React from 'react'
 
 import suggestion from './suggestion.js'
@@ -35,5 +35,28 @@ export default () => {
     return null
   }
 
-  return <EditorContent editor={editor} />
+  return <div>
+  {editor && <FloatingMenu editor={editor} tippyOptions={{ duration: 100 }}>
+    <div className="floating-menu">
+      <button
+        onClick={() => editor.chain().focus().toggleHeading({ level: 1 }).run()}
+        className={editor.isActive('heading', { level: 1 }) ? 'is-active' : ''}
+      >
+        H1
+      </button>
+      <button
+        onClick={() => editor.chain().focus().toggleHeading({ level: 2 }).run()}
+        className={editor.isActive('heading', { level: 2 }) ? 'is-active' : ''}
+      >
+        H2
+      </button>
+      <button
+        onClick={() => editor.chain().focus().toggleBulletList().run()}
+        className={editor.isActive('bulletList') ? 'is-active' : ''}
+      >
+        Bullet list
+      </button>
+    </div>
+  </FloatingMenu>}
+  <EditorContent editor={editor} /></div>
 }
```

nperez0111:
If yours works, then I'd go with the simpler one which is yours

PR #5853, opened 2024-11-21
fix: #5851 - While setting content directly while using CharacterCount with limit is not obeyed

### LINKED ISSUE #5851: [Bug]: While setting `content` directly while using CharacterCount with limit is not obeyed
### Affected Packages

core,extension-character-count,react

### Version(s)

2.10.x

### Bug Description

I thought I messed up something, my editor was not recognizing the content, it was just white screen. After investigating I found out that, I was using `CharacterCount` extension with limit set to `1000`, my content length was longer than that.

- Can we have a console warning or something to solve the above nightmare for future users ?

Now there's another bug that I found is, my setup use the below way to set the content,

```ts
const handleCreate = React.useCallback(
		(editor: Editor) => {
			if (content && editor.isEmpty) {
				editor.commands.setContent(content);
			}
		},
		[content],
	);
```

```ts
	onCreate: ({ editor }) => handleCreate(editor),
```

While the limit is obeyed and no content is set in the above way, but by directly passing the `content` to the `useEditor` hook sets the content in the editor bypassing the limit.

I have added a [CodeSandbox](https://codesandbox.io/p/sandbox/tiptap-react-qidlsv) to reproduce this issue, please let me know if I can be of help

### Browser Used

Firefox

### Code Example URL

### Expected Behavior

The `content` should not be set when the `limit` is set & the size of the content is longer than the allowed limit.

_No response_

### Dependency Updates

gethari:
Fixes #5851 

https://github.com/user-attachments/assets/eec097c3-0e35-4c67-95f5-9e44452bd190

### DIFF
--- a/packages/extension-character-count/src/character-count.ts
+++ b/packages/extension-character-count/src/character-count.ts
@@ -28,6 +28,11 @@ export interface CharacterCountOptions {
    * @example (text) => text.split(/\s+/).filter(word => word !== '').length
    */
   wordCounter: (text: string) => number
+  /**
+   * Whether to trim initial content that exceeds the limit
+   * @default true
+   */
+  trimInitialContent: boolean
 }
 
 export interface CharacterCountStorage {
@@ -60,6 +65,7 @@ export const CharacterCount = Extension.create<CharacterCountOptions, CharacterC
       mode: 'textSize',
       textCounter: text => text.length,
       wordCounter: text => text.split(' ').filter(word => word !== '').length,
+      trimInitialContent: true,
     }
   },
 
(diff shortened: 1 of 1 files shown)


packages/extension-character-count/src/character-count.ts
```diff
@@ -92,6 +98,32 @@ export const CharacterCount = Extension.create<CharacterCountOptions, CharacterC
...
+    // If content exceeds limit, trim it
+    if (currentCharCount > limit) {
+      const text = doc.textBetween(0, doc.content.size, undefined, ' ')
+      const trimmedContent = text.slice(0, limit)
+
+      // Replace the entire document content
+      editor.commands.setContent(trimmedContent)
+
+      // Add a console warning
+      console.warn(`[CharacterCount] Initial content exceeded limit of ${limit} characters. Content was automatically trimmed.`)
```
nperez0111, line 123: I'd prefer to leave this to the plugin, by issuing a transaction or allowing a transaction through which would execute the existing logic.

Prevents drift between the two implementations. Especially since this implementation does not even do the same thing as the existing implementation, you are slicing based on text alone while the existing implementation is slicing the whole range by counted chars

PR #5862, opened 2024-11-22
fix: #5851 - While setting content directly while using CharacterCount with limit is not obeyed

### LINKED ISSUE #5851: [Bug]: While setting `content` directly while using CharacterCount with limit is not obeyed

gethari:
Fixes #5851

https://github.com/user-attachments/assets/48554ff9-308e-482a-a083-1164d3ffe4ef

### DIFF
--- a/.changeset/chilled-pigs-drum.md
+++ b/.changeset/chilled-pigs-drum.md
@@ -1,0 +1,5 @@
+---
+"@tiptap/extension-character-count": major
+---
+
+fix: #5851 - While setting `content` directly while using character-count
--- a/packages/extension-character-count/src/character-count.ts
+++ b/packages/extension-character-count/src/character-count.ts
@@ -93,9 +93,39 @@ export const CharacterCount = Extension.create<CharacterCountOptions, CharacterC
   },
 
   addProseMirrorPlugins() {
+    let initialEvaluationDone = false
+
     return [
       new Plugin({
         key: new PluginKey('characterCount'),
+        appendTransaction: (transactions, oldState, newState) => {
+          if (initialEvaluationDone) {
+            return
+          }
+
+          const limit = this.options.limit
+
+          if (limit === null || limit === undefined || limit === 0) {
+            initialEvaluationDone = true
+            return
+          }
+
+          const initialContentSize = this.storage.characters({ node: newState.doc })
+
+          if (initialContentSize > limit) {
+            const over = initialContentSize - limit
+            const from = 0
+            const to = over
+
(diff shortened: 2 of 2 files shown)


gethari:
@nperez0111 is this ok ? I have made changes as suggested, I could'nt able to run `npm run changeset` earlier as I was taking my base branch from `develop` it was showing , hence I took branch from `main`, Am i correct ?

```
Failed to find where HEAD diverged from main. Does main exist ?
```

gethari:
Hey @nperez0111 , I think I might have caused a side effect because of this, when the charLimit is set, whenever we try to call `setContent` , its not getting reflected 😢

nperez0111:
That doesn't sound right to me, can you reproduce in a sandbox & prove that it this.

Most likely, you are running into something a lot of people do, trying to put html in setContent and thinking it will work exactly as you input it. As a test, could you try doing insertContent which takes a different codepath than setContent

gethari:
@nperez0111 please find the [Sanbox](https://codesandbox.io/p/sandbox/fragrant-morning-mlpffd), 

The size of the content inside `setContent` is 321, so if we set the limit >= the size, this works, but when we use limit of 300, the content is not getting overridden, I guess previously it was just setting the content all-together without obeying the limit, after my `appendTransaction`, i think this is getting skipped

nperez0111:
Yea, not sure what is going on here & I don't have time to fix it. Could you open a new issue? What is strange to me is that it feels like appendTransaction should always run so I'm not sure that we need to run the modification in the filter transaction anymore, & just let it be in the appendTransaction instead

PR #5880, opened 2024-11-27
fix: font family extesion demo

alexvcasillas:
## Changes Overview

This PR aims to fix the [Font Family extension](https://tiptap.dev/docs/editor/extensions/functionality/fontfamily#page-title) demo

## Implementation Approach

Correctly make use of the returned value from the `textStyle` API

## Testing Done

Demo is correctly working

## Additional Notes

Will update the docs with additional information for the users about the returned value from `textStyle` API so they don't get confused it this happens to them while setting the plugin

## Checklist

### DIFF
--- a/demos/src/Extensions/FontFamily/React/index.jsx
+++ b/demos/src/Extensions/FontFamily/React/index.jsx
@@ -43,7 +43,7 @@ export default () => {
           <button
             onClick={() => editor.chain().focus().setFontFamily('Comic Sans MS, Comic Sans').run()}
             className={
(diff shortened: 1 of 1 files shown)


demos/src/Extensions/FontFamily/React/index.jsx
```diff
@@ -79,13 +79,6 @@ export default () => {
           >
             CSS variable
           </button>
-          <button
```
alexvcasillas, line 82: I have removed this `button` since I don't see any utility into it and does not apply for any of the cases that we're showcasing

nperez0111: 👍 yep

Issue #578, opened 2020-01-09
Menu bubble with link input active won't hide


laurensiusadi:
**Describe the bug**
If you use link input within menu bubble, after clicking the input, menu bubble won't hide when clicking outside the menu bubble like usual.

**Steps to Reproduce / Codesandbox Example**
1. Go to https://tiptap.scrumpy.io/links
2. Select some text without links
3. Menu bubble appears, click Add Link
4. Input for link will appear on menu bubble
5. Click outside of editor, menu bubble won't hide

**Expected behavior**
Menu bubble hide if click outside editor

**Screenshots**
After link input active, click outside editor trying to dismiss menu bubble (mark with red arrow), but it won't close. Only able to close after placing cursor or selecting on other text.
![Annotation 2020-01-09 105134](https://user-images.githubusercontent.com/5661040/72036641-09f62700-32ce-11ea-87bf-53dd872e79f9.png)

**Environment**
 - OS: Windows 10 build 19041.1
 - Browser: Mozilla Firefox
 - Version: 71.0
 - Desktop

hanspagel:
Thanks @laurensiusadi! We’re in the process of rewriting the whole interface for tiptap v2 and I added this to the list of things we need to consider! 🙌

cairon666:
I have the same problem, is there a plan to solve it?

nperez0111:
Feel free to investigate @cairon666

PR #5882, opened 2024-11-27
fix(bubble-menu): add focusout handler for floating element

cairon666:
## Changes Overview

Fix the bug where the bubble menu does not hide

## Implementation Approach

Add an additional focusout event handler to the floating element.

Also, set tabIndex to -1 for the bubble menu to ensure that clicking on an empty block in the bubble menu does not result in the parentNode being null.

## Testing Done

Test on a local

## Verification Steps

Add input into bubble-menu demo
Open BubbleMenu
Click on the input
Click outside the editor

## Checklist

## Related Issues
#578

### DIFF
--- a/.changeset/twelve-turtles-cry.md
+++ b/.changeset/twelve-turtles-cry.md
@@ -1,0 +1,8 @@
+---
+"@tiptap/extension-bubble-menu": patch
+"@tiptap/react": patch
+"@tiptap/vue-2": patch
+"@tiptap/vue-3": patch
+---
+
+Fix the bug where the bubble menu does not hide
--- a/.changeset/twelve-turtles-cry.md
+++ b/.changeset/twelve-turtles-cry.md
@@ -5,4 +5,4 @@
 "@tiptap/vue-3": patch
 ---
 
-Fix the bug where the bubble menu does not hide
+Added a `focusout` event listener to handle blur events on the bubble menu element, ensuring the menu closes properly when blurred.
--- a/packages/react/src/BubbleMenu.tsx
+++ b/packages/react/src/BubbleMenu.tsx
@@ -50,7 +50,7 @@ export const BubbleMenu = (props: BubbleMenuProps) => {
   }, [props.editor, currentEditor, element])
 
   return (
-    <div ref={setElement} className={props.className} style={{ visibility: 'hidden' }}>
+    <div ref={setElement} tabIndex={-1} className={props.className} style={{ visibility: 'hidden' }}>
       {props.children}
     </div>
   )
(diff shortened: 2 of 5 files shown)


.changeset/twelve-turtles-cry.md
```diff
@@ -0,0 +1,8 @@
+---
+"@tiptap/extension-bubble-menu": patch
+"@tiptap/react": patch
+"@tiptap/vue-2": patch
+"@tiptap/vue-3": patch
+---
+
+Fix the bug where the bubble menu does not hide
```
bdbch, line 8: I think a more precise description would help here (don't specify the bug but explain more what changed as this will be generated into the changelog and helps people to understand what's new. This wouldn't tell a user looking at the changelog nothing really).

bdbch:
```suggestion
Added a `focusout` event listener to handle blur events on the bubble menu element, ensuring the menu closes properly when blurred.
```

packages/react/src/BubbleMenu.tsx
```diff
@@ -50,7 +50,7 @@ export const BubbleMenu = (props: BubbleMenuProps) => {
   }, [props.editor, currentEditor, element])
 
   return (
-    <div ref={setElement} className={props.className} style={{ visibility: 'hidden' }}>
+    <div ref={setElement} tabIndex={-1} className={props.className} style={{ visibility: 'hidden' }}>
```
bdbch, line 53: Did you test if a negative tab index still allows the client to focus elements inside the div via the keyboard?

bdbch (changes requested):
Looks good to me - would like to change the changeset a bit to be more clear about what changed.

packages/react/src/BubbleMenu.tsx
cairon666: Yes, you can still focus on other elements inside

This is necessary so that `event.relatedTarget` is not `null` if an empty space was clicked inside the bubble-menu

cairon666: Yes, you can still focus on other elements inside using tab

This is necessary so that `event.relatedTarget` is not `null` if an empty space was clicked inside the bubble-menu

cairon666:
The `focusout` event handler cannot affect elements that are not displayed inside the bubble-menu and intentionally close the bubble-menu when interacting with such elements, for example, as with `Popover` or `Tooltip` in MUI. Is it necessary to handle this behavior?

bdbch:
@cairon666 could you rebase this to the current develop branch as we released V3, deprecating the next branch & would need to update most of our PR's anyway? Thanks a lot!

PR #5891, opened 2024-11-30
Bdbch/fix text default alignment

bdbch:
Original PR here - took it over to bring this close to be mergable:

#3460


packages/extension-text-align/src/text-align.ts
```diff
@@ -20,7 +20,7 @@ export interface TextAlignOptions {
    * @default 'left'
    * @example 'center'
    */
-  defaultAlignment: string,
+  defaultAlignment: string | null,
```
nperez0111, line 23: Should we be stricter on the types here, can only be left, right & center, correct?

Same for alignments too

nperez0111 (approved):
Ah, looking at the code now I see what you were saying, yea as long as the default moves too, this looks good to me

PR #5945, opened 2024-12-18
Fixed Link extension's commands not respecting XSS prevention via unallowed protocols

bdbch:
This pull request includes changes to improve error handling and validation for links in the `Link` extension. The most important changes include adding error handling in the React component, exporting the `isAllowedUri` function, and validating link protocols before setting or toggling links.

## Implementation Approach
I added checks inside the `setLink` and `toggleLink` commands to see if the link being tried to set is actually valid. If not, we throw an error so the developer can implement error handling outside of the extension.

```js
try {
  editor.chain().setLink('javascript:alert("Hello world")').focus().run()
} catch (e) {
  if (e.message === 'Invalid protocol') {
    alert('Please don\'t put XSS injections into our content please.')
  }
}
```

## Testing Done
1. Tested it locally on the Link demo (where I also added said check + error handler)
2. Used the already existing test that will check if insertion of unwanted protocols works or not

## Verification Steps
See above

## Additional Notes
Because people can potentially also override said commands, I added an export for the `isValidUri` function to make it easy for people to add said checks themselves.

## Checklist

### DIFF
--- a/packages/extension-link/src/link.ts
+++ b/packages/extension-link/src/link.ts
@@ -160,7 +160,7 @@ declare module '@tiptap/core' {
 // eslint-disable-next-line no-control-regex
 const ATTR_WHITESPACE = /[\u0000-\u0020\u00A0\u1680\u180E\u2000-\u2029\u205F\u3000]/g
 
-function isAllowedUri(uri: string | undefined, protocols?: LinkOptions['protocols']) {
+export function isAllowedUri(uri: string | undefined, protocols?: LinkOptions['protocols']) {
   const allowedProtocols: string[] = [
     'http',
     'https',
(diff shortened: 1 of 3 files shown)


packages/extension-link/src/link.ts
```diff
@@ -322,11 +322,23 @@ export const Link = Mark.create<LinkOptions>({
     return {
       setLink:
         attributes => ({ chain }) => {
+          const { href } = attributes
+
+          if (!isAllowedUri(href, this.options.protocols)) {
+            throw new Error('Invalid protocol')
+          }
```
nperez0111, line 329: I don't think throwing an error is appropriate here. This makes the caller have to deal with the error.

The more common pattern for commands would be to return false, i.e. I refuse to execute this command

packages/extension-link/src/link.ts
```diff
@@ -322,11 +322,23 @@ export const Link = Mark.create<LinkOptions>({
...
           return chain().setMark(this.name, attributes).setMeta('preventAutolink', true).run()
         },
 
       toggleLink:
         attributes => ({ chain }) => {
+          const { href } = attributes
+
+          if (!isAllowedUri(href, this.options.protocols)) {
+            throw new Error('Invalid protocol')
+          }
```
nperez0111, line 340: Same here

packages/extension-link/src/link.ts
nperez0111: Also, we should be calling the user's `isAllowedUri` https://github.com/ueberdosis/tiptap/pull/5945/files#diff-1d1f108bd30f776ec9689dc1bb00a3cb3dafc89fde595994aa8fa400471a87d6R368

Because our `isAllowedUri` does not accept things like custom protocols which some may link to for example: `ftp://, obsidian://` etc.

bdbch:
@nperez0111 done

nperez0111 (approved):
Looks good. I think to release we got to make something not a major. I would look into it but I have an appointment right now

Issue #4521, opened 2023-10-10
[Bug/Missing feature]: OnReadOnlyChecked


Tagaishi:
### Which packages did you experience the bug in?

Extension-task-item

### What Tiptap version are you using?

2.1.11

### What’s the bug you are facing?

The function OnReadOnlyCheck doesn't update the editor, only the checkbox visibly.
It has the node and value of the tick in it's parameters to try and offer the possibility to manually update the editor but running the following function:
```tsx
onReadOnlyChecked: (node, checked) => {
  editor!.state.doc.descendants((subnode, pos) => {
    if (node.eq(subnode)) {
      const { tr } = editor!.state;
      tr.setNodeMarkup(pos, undefined, {
        ...node.attrs,
        checked: checked,
      });
      editor!.view.dispatch(tr);
    }
  });
  return widget.properties.allowReadOnlyCheck;
},
```
The nodes never match using ```node.eq(subnode)```. <= This is what I think is a bug unless I am doing something wrong.

Now that test would not be needed if the position was also given in the parameters. <= Missing feature?

### What browser are you using?

Other

### What did you expect to happen?

For ```node.eq(subnode)``` to return "true" once.
I actually tried to isolate those nodes values and found that they should in fact be matching but yet aren't.w

### Did you update your dependencies?

- [X] Yes, I’ve updated my dependencies to use the latest version of all packages.

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

Tagaishi:
I've noticed #3676 also refers to the same problem but isn't solved nor does it offer a simple solution that isn't "patchwork".

nicksellen:
I posted my kludgey fix here 👉 https://github.com/ueberdosis/tiptap/issues/3676#issuecomment-2555606541 - maybe helps!

nicksellen:
And a PR that should hopefully resolve it 👉 https://github.com/ueberdosis/tiptap/pull/5952

PR #5952, opened 2024-12-21
fix(task-item): fix read only checked handling

### LINKED ISSUE #3676: update event not triggered on checkbox when onReadOnlyChecked returns true and editable is false

Hello & thank you for this awesome lib :+1: 

The `'update'` event is not not triggered on checkbox elements when the editor is not in its editable state but `onReadOnlyChecked` returns `true` (tested with `react`)

latest firefox version on ubuntu

with this sandbox : https://codesandbox.io/s/update-not-trigger-on-checkbox-when-editable-is-false-v87v22?file=/src/App.js:468-485
you can see in the console that the update event is not trigger when editable is false, but it is when editable is true.

here https://codesandbox.io/s/update-not-trigger-on-checkbox-when-editable-is-false-v87v22?file=/src/App.js:468-485

I would expect the update event to be trigger when the checkbox state changes, even if editable is set to false. Is this is not a bug but a API design choice, do I have the ability to trigger manually an update event?

_No response_

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

nicksellen:
Fixes #3676

## Changes Overview

`onReadOnlyChecked` allows you to determine whether the checking/unchecking the task item should apply the change, or not.

It did not work because:
- it only ever passed the original node to the function, not the current node
- even if you returned true it would not apply the change to the document

## Implementation Approach

Now it runs the logic inside the transaction where we have access to the current node, and we can either abort it as before if it returns `false`, or allow it to proceed with the transaction if it returns `true`.

## Testing Done

Only manual testing whilst clicking in a browser.

## Verification Steps

Configure your editor like this:

```js

new Editor({
    editable: false,
    // + other options
    extensions: [
        TaskItem.configure({
            // Allow checking/unchecking when in read only mode
            onReadOnlyChecked: () => true,
        }),
        // + other extensions
    ]
})    
```

Then, ideally, open a couple of windows with a collaborative session running, and notice that you can check/uncheck the box in one window, but the change is not propagated to the other.

Additionally, if you want to have more logic (e.g. the kind of logic in https://github.com/ueberdosis/tiptap/issues/4521 - where the node descendent are checked for matching), it should be possible to observe the node is now the correct node, and can be matched if desired:

```js
onReadOnlyChecked: (node, checked) => {
  editor.state.doc.descendants((subnode, pos) => {
    if (node.eq(subnode)) {
      const { tr } = editor!.state;
      tr.setNodeMarkup(pos, undefined, {
        ...node.attrs,
        checked: checked,
      });
      editor.view.dispatch(tr);
    }
  });
  return true;
},
```

## Additional Notes

There is also another bug, in the update method because it uses `checkbox.setAttribute('checked', 'checked')` and `checkbox.removeAttribute('checked')` which is not the correct way to change the state of a checkbox, should just be `checkbox.checked = checked`.

Should I add a fix to this PR? Maybe best in another one....

Update: I submitted a separate PR for that https://github.com/ueberdosis/tiptap/pull/5953

## Checklist

## Related Issues

https://github.com/ueberdosis/tiptap/issues/3676
https://github.com/ueberdosis/tiptap/issues/4521

### DIFF
--- a/packages/extension-task-item/src/task-item.ts
+++ b/packages/extension-task-item/src/task-item.ts
@@ -163,7 +163,7 @@ export const TaskItem = Node.create<TaskItemOptions>({
 
               if (!editor.isEditable && this.options.onReadOnlyChecked) {
                 // Reset state if onReadOnlyChecked returns false
-                if (!this.options.onReadOnlyChecked(currentNode, checked)) {
+                if (!currentNode || !this.options.onReadOnlyChecked(currentNode, checked)) {
                   checkbox.checked = !checkbox.checked
                   return false
                 }
--- a/packages/extension-task-item/src/task-item.ts
+++ b/packages/extension-task-item/src/task-item.ts
@@ -149,7 +149,7 @@ export const TaskItem = Node.create<TaskItemOptions>({
 
         const { checked } = event.target as any
 
-        if (editor.isEditable && typeof getPos === 'function') {
+        if (typeof getPos === 'function') {
           editor
             .chain()
             .focus(undefined, { scrollIntoView: false })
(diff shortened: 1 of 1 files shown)


nperez0111 (approved):
This looks good code-wise. Would love to have a test for this added though to make sure that we don't regress

nilpatel42 (approved):

bdbch:
@nicksellen could you rebase this to the current develop branch as we released V3, deprecating the next branch & would need to update most of our PR's anyway? Thanks a lot!

PR #5953, opened 2024-12-21
fix(task-item): fix updating checked status

nicksellen:
## Changes Overview

The code currently uses `setAttribute` / `removeAttribute` to update the checked status of the checkbox, this is the incorrect method to use. Checkboxes have a `checked` property that can be set.

See https://aileenrae.co.uk/blog/programatically-check-uncheck-checkbox-javascript/

## Implementation Approach

Switch to setting `checkbox.checked` property

## Testing Done

Manually in browser.

## Verification Steps

Open a collaborative session with task item extension available.

In one window check/uncheck the box a few times, then in the other window do the same.

Notice that the update _appears_ to no longer propagate, the checkbox state does not change in the other window.

It _is_ actually changing the attribute on the document node correctly, which you can see if have styling applied such as this (on the task item):

```scss
li[data-checked='true'] {
    text-decoration: line-through;
}
```

With the fix, it works again!

## Checklist

## Related Issues

I couldn't find an existing issue for it, which I thought was surprising, but maybe I didn't look well enough.

### DIFF
--- a/packages/extension-task-item/src/task-item.ts
+++ b/packages/extension-task-item/src/task-item.ts
@@ -202,12 +202,7 @@ export const TaskItem = Node.create<TaskItemOptions>({
             return false
           }
 
-          listItem.dataset.checked = updatedNode.attrs.checked
-          if (updatedNode.attrs.checked) {
-            checkbox.setAttribute('checked', 'checked')
-          } else {
-            checkbox.removeAttribute('checked')
-          }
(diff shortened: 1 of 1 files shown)


nperez0111 (approved):
Thanks for the fix, looks like we were setting this inconsistently (using checked in some places and setting the attribute in others).

I've updated this to now only ever use the checked property

Issue #5870, opened 2024-11-26
[Bug]: Uncaught TypeError: Cannot read properties of undefined (reading 'props')  at ReactNodeView.handleSelectionUpdate


rushillshah:
### Affected Packages

tiptap/react

### Version(s)

2.9.1

### Bug Description

We are encountering an issue where `this.renderer` is initialized correctly but becomes undefined during certain interactions. This typically occurs when repositioning floating variables in our editor. Occasionally, this issue also arises in production during regular text editing.

This issue appears to be related to a lifecycle problem, where the `renderer` is not properly cleaned up or initialized under specific circumstances. Using null coalescing operators (e.g., `this.renderer?.props?.selected`) in @tiptap/react/dist/index.js:1367 seems to mitigate the issue, but this feels like a workaround rather than a proper solution.

Stacktrace:

```
index.js:1376 Uncaught TypeError: Cannot read properties of undefined (reading 'props')
    at ReactNodeView.handleSelectionUpdate (index.js:1376:1)
    at eval (index.js:159:1)
    at Array.forEach (<anonymous>)
    at Editor.emit (index.js:159:1)
    at Editor.dispatchTransaction (index.js:4601:1)
    at EditorView.dispatch (index.js:5611:1)
    at Object.method [as focus] (index.js:66:1)
    at eval (variable.js:161:1)
    at invokePassiveEffectCreate (react-dom.development.js:23487:1)
    at HTMLUnknownElement.callCallback (react-dom.development.js:3945:1)
```

Environment

- Tiptap version: @tiptap/react v2.9.1
-  Prosemirror version: prosemirror-view v1.28.0
-  Node.js version: 20.13.1
-  Browser: Chrome 118.0.5993.89
-  React version: 17.0.2
- Webpack 5.94.0

Steps to Reproduce

	1.	Use @tiptap/react with a custom addNodeView setup, particularly one involving floating elements.
	2.	Interact with floating variables (e.g., reposition or edit them).
	3.	Observe the error during these interactions or general text editing

### Expected Behavior

No TypeError and for the `renderer` to handle intermittent undefined values, or not take undefined values

### Additional Context (Optional)

The issue seems to stem from the ReactNodeView lifecycle. Using null coalescing operators for accessing this.renderer.props.selected appears to resolve the issue temporarily but does not address the root cause.

### Dependency Updates

- [X] Yes, I've updated all my dependencies.

nperez0111:
[...]

PR #5982, opened 2025-01-03
fix: handleSelectionUpdate callback is triggered before the renderer is initialized.

rushillshah:
## Changes Overview
This PR addresses a crash in the ReactNodeView implementation that occurs when the handleSelectionUpdate callback is triggered before the renderer is initialized. The issue arises due to the asynchronous nature of JavaScript, where the subscription to the selectionUpdate event (editor.on) is executed before the renderer initialization.

The proposed fix ensures that the renderer is initialized before subscribing to the selectionUpdate event, maintaining the intended execution order.

## Implementation Approach
The subscription to the selectionUpdate event was moved to occur after the initialization of the renderer. This ensures the callback does not execute prematurely and prevents the crash. The implementation adds no new functionality or logic but rearranges the code to avoid the race condition.

## Testing Done
The fix was tested by replicating the scenario that caused the crash:
	1.	Adding a new line before a tag.
	2.	Observing the behavior when the old node is destroyed, and a new node is created.

The changes successfully prevent the crash in these scenarios. Additionally, basic functionality was verified to ensure no regressions.

This normally occurs with editors built with tiptap, tippy and prosemirror. 

## Verification Steps
	1.	Open a document with the affected editor.
	2.	Add a new line before an existing tag and observe that no crash occurs.
	3.	Verify that the selectionUpdate events are properly handled without regressions in functionality.

## Additional Notes
This fix does not alter the functionality of the library; it only ensures proper execution order, and does not affect any other functionality, but avoids this condition that causes the crash

## Checklist

## Related Issues
https://github.com/ueberdosis/tiptap/issues/5870

### DIFF
--- a/.changeset/early-parrots-carry.md
+++ b/.changeset/early-parrots-carry.md
@@ -1,0 +1,5 @@
+---
+"@tiptap/react": patch
+---
+
+Resolves #5870 by re-ordering executation of selectionUpdate handler
--- a/packages/react/src/ReactNodeViewRenderer.tsx
+++ b/packages/react/src/ReactNodeViewRenderer.tsx
@@ -141,7 +141,6 @@ export class ReactNodeView<
     const { className = '' } = this.options
 
     this.handleSelectionUpdate = this.handleSelectionUpdate.bind(this)
-    this.editor.on('selectionUpdate', this.handleSelectionUpdate)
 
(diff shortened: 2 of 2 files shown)


nperez0111 (approved):
This is great, simple solution

PR #5991, opened 2025-01-06
Support `validate` options in node and mark attribute definitions.

bobthecow:
## Changes Overview

ProseMirror allows node and mark specs to define a `validate` function or type, which is used when deserializing JSON, parsing from the DOM, etc.

See https://prosemirror.net/docs/ref/#model.AttributeSpec.validate

## Implementation Approach

Currently, `default` is the only attribute spec option passed through to ProseMirror when resolving the schema. This change updates `getAttributesFromExtensions` and `getSchemaByResolvedExtensions` (and relevant types) to also pass through any defined `validate` attribute option.

## Testing Done

Manual testing (plus using the patch in production!)

## Verification Steps

To use this, add a `validate` type or function option in any extension's `addAttributes` or `addGlobalAttributes` definition, e.g.:

```
export const NodeId = Extension.create({
  name: 'nodeId',

  addGlobalAttributes() {
    return [
      {
        types: [...],
        attributes: {
          nodeId: {
            default: '__node_id__',
            validate: 'string',
            ...
          }
        }
      }
    ]
  }
});
```

A more complex validation could ensure that the ID is shaped like a UUID, or allow it to be `null` or `undefined`.

## Checklist


nperez0111 (approved):
This is a fantastic PR, I'm going to merge this to be part of the next branch, as we are in a sort of feature freeze for Titpap v2

PR #6006, opened 2025-01-09
fix: add missing node view props to props type object

GYHHAHA:
## Changes Overview

I was trying to follow the step from [render-a-vue-component](https://tiptap.dev/docs/editor/extensions/custom-extensions/node-views/vue#render-a-vue-component) section. Everything works except that there was a type error when I pass the customized component to `VueNodeViewRenderer`. After some searchs, now I think some props are missing. Thus adding them on could be a fix if the absence is not on purpose.

image

## Implementation Approach

add 3 missing props declaration to the `nodeViewProps` object

## Testing Done & Verification Steps

error resolved after the change

image

## Checklist

### DIFF
--- a/.changeset/silly-wolves-attack.md
+++ b/.changeset/silly-wolves-attack.md
@@ -1,0 +1,5 @@
+---
+"@tiptap/vue-3": patch
+---
+
+update node view props to match `NodeViewProps` type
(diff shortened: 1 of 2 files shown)


nperez0111 (approved):
Thanks for the PR, yea this looks right

Issue #3700, opened 2023-02-03
Event on node deleted


gianLuigiL:
### What problem are you facing?

Would it be possibile for tiptap to emit events when a node is deleted altogether? I'm trying to track image deletions but it looks like i have to observe the markup and match tags to discern whether a tag has been deleted by the user.

### What’s the solution you would like to see?

it would be cool to have a "deleted" or "removed" event that only takes into account deleted nodes.

### What alternatives did you consider?

Matching generated markup by regex, but it's cumbersome.

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

holdenmatt:
I'm also interested in this. I have some interactive node views where I need to perform some cleanup when they're deleted.

Currently planning to track all node ids myself and diff them onUpdate, so I can detect node deletions, but that's a bit cumbersome.

If tiptap had an `onNodeDeleted` event, that would be perfect. Or if someone finds another way to implement this, please share here!

Dhalsimzhao:
> I'm also interested in this. I have some interactive node views where I need to perform some cleanup when they're deleted.
> 
> Currently planning to track all node ids myself and diff them onUpdate, so I can detect node deletions, but that's a bit cumbersome.
> 
> If tiptap had an `onNodeDeleted` event, that would be perfect. Or if someone finds another way to implement this, please share here!


Same problem.

And I found a response from this discuss [react to node adding/removing](https://discuss.prosemirror.net/t/reacting-to-node-adding-removing-changing/676). Maybe we can just pick the first approach before tiptap support that event (if they are going to implement this).

Btw, when going with the first approach, remember tiptap export some useful helpers as well, like `findChildren`, which can save us some time.

holdenmatt:
I came up with a solution that works well for me, happy to share here.

I use an `onUpdate` handler to diff nodes and emit `onNodeDeleted` events that can handle whatever cleanup you like. I wrote a separate extension to add unique `id` attrs to nodes, which are used here. You could probably use the TipTap UniqueId extension to do this, or find some other way to compare nodes across edits.

Hopefully helpful to someone!

<img width="585" alt="image" src="https://user-images.githubusercontent.com/764380/229245577-61d50d77-0478-48a8-ae63-6d49a5e91dba.png">

stevecastaneda:
> I came up with a solution that works well for me, happy to share here.

Here's a gist of the above to make it easier to copy/paste, including imports.   This works great for me.

[...]

PR #6022, opened 2025-01-14
feat(on-delete): add a new extension `OnDelete` which can detect node deletions

nperez0111:
## Changes Overview

This adds a new extension `OnDelete` which can detect node deletions.

I'm not certain about the naming here, will probably rename it.
## Checklist

## Related Issues

https://github.com/ueberdosis/tiptap/issues/3700


nperez0111:
Maybe we should make this a core extension, given that has performance implications, we may only ever want to register this if someone is actually listening.

`coreExtensionOptions: {onDelete:{async: boolean}}`

PR #6030, opened 2025-01-17
feat(docx): add export to docx support using the static renderer

nperez0111:
## Changes Overview

This is an example implementation of a Tiptap JSON -> DOCX renderer that is completely static, and isomorphic between client & server.

image

## Checklist


packages/static-renderer/src/json/docx/docx.example.ts
nperez0111, line 1: You can run this file in bun & it should just work & output a docx file

PR #6061, opened 2025-01-27
Modify the TextAlign extension documentation so that the default value of the defaultAlignment option matches the code

arnaugomez:
The default value of `defaultAlignment` is `null` in the code. Modify the documentation so that it matches the value in the code.

## Changes Overview
In the code of the TextAlign extension, the default value of the `defaultAlignment` property is `null`.

However, the documentation of the property says that its default value is `"left"`. The documentation does not match the code. By default, if the `defaultAlignment` property is not configured, the text alignment of a block is unset.

One might argue that, if the text alignment of a block is unset, its alignment is left in practice, but this is not always true. For example, if the parent element of the text editor has right text alignment, the text blocks will have right alignment too.

## Implementation Approach
Modify the documentation

## Testing Done
There were no changes to the code, only in the documentation. The testing consists in verifying that the documentation has changed.

## Verification Steps
Read the documentation of the `defaultAlignment` property

## Additional Notes
None

## Checklist

## Related Issues
None

### DIFF
--- a/.changeset/yellow-hairs-search.md
+++ b/.changeset/yellow-hairs-search.md
@@ -1,0 +1,5 @@
+---
+"@tiptap/extension-text-align": patch
+---
+
+Modify the documentation of the default value of defaultAlignment so that it matches the code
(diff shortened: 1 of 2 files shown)


nperez0111 (approved):

nperez0111:
Makes sense, thanks for contributing

PR #6068, opened 2025-01-29
fix(extension-link): update regex to allow numbers in URI

timoisik:
## Changes Overview
This PR allows numbers in links, especially in emails. Before something like test1@example.com did not work.

### DIFF
--- a/.changeset/three-crabs-beg.md
+++ b/.changeset/three-crabs-beg.md
@@ -1,0 +1,5 @@
+---
+"@tiptap/extension-link": patch
+---
+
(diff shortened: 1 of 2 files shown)


nperez0111 (approved):
That looks right to me. Is it valid for URLs to have numbers in them too?

alexvcasillas (approved):

bdbch (approved):

bdbch:
Looks good to me too.

> Is it valid for URLs to have numbers in them too?

Pretty sure they are

Issue #4777, opened 2024-01-11
[Bug]: Links with marks not working


amit-verma-qatalog:
### Which packages did you experience the bug in?

core

### What Tiptap version are you using?

2.1.3

### What’s the bug you are facing?

If I style the links with marks such as bold or change colors, the links stop working. 
the `clickHandler` requires the `event.target` to be the `A` tag but in case of added marks, the current target is often a `span`

```
handleClick: (view, pos, event) => {
        if (event.button !== 0) {
          return false
        }

        const eventTarget = event.target as HTMLElement

        if (eventTarget.nodeName !== 'A') {
          return false
        }

        const attrs = getAttributes(view.state, options.type.name)
        const link = (event.target as HTMLLinkElement)

        const href = link?.href ?? attrs.href
        const target = link?.target ?? attrs.target

        if (link && href) {
          if (view.editable) {
            window.open(href, target)
          }

          return true
        }

        return false
      },
    },
```

### What browser are you using?

Chrome

### What did you expect to happen?

Links to open even if the text is styled

### Did you update your dependencies?

- [X] Yes, I’ve updated my dependencies to use the latest version of all packages.

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

nperez0111:
Hm, depending on how marks are being wrapped, we can instead check that what was clicked has a parent that is the `a` tag instead

GormanDesign:
This is still an issue on v2.9.1, we addressed it by forking and fixing the clickHandler logic. The current logic always uses the `event.target` as the `link`, but when marks are present the `event.target` is the mark itself (ie `<u />`, `<strong />`). The change below uses the `els` array to find and assign the `link` instead of relying on the `event.target`

**Original Click Handler**
``` ts
let a = event.target as HTMLElement
const els = []

while (a.nodeName !== 'DIV') {
  els.push(a)
  a = a.parentNode as HTMLElement
}

if (!els.find(value => value.nodeName === 'A')) {
  return false
}

const attrs = getAttributes(view.state, options.type.name)
const link = (event.target as HTMLAnchorElement)

const href = link?.href ?? attrs.href
const target = link?.target ?? attrs.target
```

**Forked Click Handler**
``` ts
let link: HTMLAnchorElement | null = null;

if (event.target instanceof HTMLAnchorElement) {
  link = event.target;
} else {
  let a = event.target as HTMLElement;
  const els = [];

[...]

PR #6078, opened 2025-01-31
fix(link): fixes links with marks

GormanDesign:
## Changes Overview
The current logic always uses the event.target as the link, but when marks are present the event.target is the mark itself (ie <u />, <strong />). The change below uses the els array to find and assign the link instead of relying on the event.target

## Implementation Approach
1. Check if the event target is in fact a HTMLAnchorElement
2. If it is not, use the els array to find and assign the link instead of relying on the event.target

## Testing Done
Tested in application with dynamic user content. 

## Verification Steps
Manually verification and Cypress tests run in application.

## Checklist

## Related Issues
https://github.com/ueberdosis/tiptap/issues/4777


nperez0111 (approved):
Looks great to me

alexvcasillas:
Hey @GormanDesign can we have a changeset here? :)

alexvcasillas (approved):

bdbch:
The linter failed, @GormanDesign could you look into this?

PR #6083, opened 2025-02-04
Add option to set rel on youtube embed link

teamclouday:
## Changes Overview
Add `sameChannelRel` option to youtube extension, according to the embed url parameter [description](https://developers.google.com/youtube/player_parameters#rel).

## Implementation Approach
Following the same pattern as other parameters.

## Testing Done
Not sure how to write a test for this.

## Checklist


teamclouday:
Btw I also think this code is wrong, because it skips the rest of url parameters configurations if the copied link has `youtu.be`, which is true in current youtube website when click on the share button
https://github.com/ueberdosis/tiptap/blob/b7a7b2ad85cce18449bf856fac8a8b6a301f502c/packages/extension-youtube/src/utils.ts#L71

packages/extension-youtube/src/utils.ts
```diff
@@ -50,6 +51,7 @@ export const getEmbedUrlFromYoutubeUrl = (options: GetEmbedUrlOptions) => {
     playlist,
     progressBarColor,
     startAt,
+    sameChannelRel,
```
alexvcasillas, line 54: I'd call this `rel: boolean` or `relatedVideos: boolean` and allow to pass `rel={Number(rel)}` so the user can customize it and would be following more the youtube documentation for the embedded player parameters.

alexvcasillas:
> Also probably better to first parse the url using `new URL(url)`, and then merge the new search params based on the configuration, with the existing ones, to avoid having multiple `?` in the link.
> 
> I can create new PRs if necessary

Feel free to contribute with as what you think it would significantly improve the overall experience of the user for this particular extension :)

packages/extension-youtube/src/utils.ts
teamclouday: Sounds good! I've changed it to `rel: number`

teamclouday:
> Feel free to contribute with as what you think it would significantly improve the overall experience of the user for this particular extension :)

Sounds good! I'll create another PR after this one is merged!

alexvcasillas (approved):
Looking great! Thanks for addressing the changes 🙇🏻

.changeset/thirty-forks-share.md
```diff
@@ -0,0 +1,5 @@
+---
+"@tiptap/extension-youtube": major
```
alexvcasillas, line 2: This should be `patch` instead of `major`

alexvcasillas (changes requested):
Just saw that you applied a `major` change instead of a `patch` change, please, switch it to `patch` instead, we don't want a major bump in all of the packages for a `patch` :)

.changeset/thirty-forks-share.md
teamclouday: Got it! I've changed to `patch`

bdbch (approved):
Thanks, looks good to me!

alexvcasillas (approved):
Thanks for addressing the requested changes! Let's rebased and we're good to go 🚀

teamclouday:
I noticed there were some lint issues and test failures from the automated workflows, and tried to fix it in the latest commit. But I'm not able to verify it locally. Let me know if it works. Thanks!

bdbch:
I started another run for our tests/linting. Will see if it runs through successfully!

PR #6088, opened 2025-02-06
upgrade prosemirror-tables to 1.6.4 (fixes #6074)

### LINKED ISSUE #6074: [Bug]: Dropping text in front of a table produces a broken, empty table
### Affected Packages

core

### Version(s)

Any version released in the last year (at least)

### Bug Description

When dropping text, prosemirror [attempts to calculate the insert position](https://github.com/ProseMirror/prosemirror-view/blob/0954f523ee0a193f7db3992702e6a3979ade89b8/src/input.ts#L709C3-L709C54) via posAtCoords. In this case, [it uses](https://github.com/ProseMirror/prosemirror-view/blob/0954f523ee0a193f7db3992702e6a3979ade89b8/src/domcoords.ts#L225) the element's bounding box. So if the cursor gets there, rather than adjusting the position, it assumes that the insertion occurs at the depth level of the table rows and looks for suitable neighbors for the new element and finds tableRow. 

A table that is created in this way is invisible and broken. It cannot be deleted via Backspace (throws error), properly selected, etc. To reproduce it, you can just go to the example from the tiptap website 

There are many places where it is possible to handle this case both in prosemirror modules and outside the prosemirror logic, although it should probably be considered as a bug of prosemirror itself. 

https://github.com/user-attachments/assets/2675da7f-8d04-4f2b-8dd6-757a443a8245

### Browser Used

Chrome

### Code Example URL

https://tiptap.dev/docs/examples/basics/tables

### Expected Behavior

The text is inserted before the table and no new tables are created.

_No response_

### Dependency Updates

mr1name:
## Changes Overview

Just a dependency upgrade to fix a bug with tables - #6074.

PR in prosemirror-tables:
https://github.com/ProseMirror/prosemirror-tables/pull/267

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15398,19 +15398,6 @@
         "prosemirror-view": "^1.27.0"
       }
     },
-    "node_modules/prosemirror-tables": {
-      "version": "1.6.3",
-      "resolved": "https://registry.npmjs.org/prosemirror-tables/-/prosemirror-tables-1.6.3.tgz",
-      "integrity": "sha512-8B0X6PjAkXaHKntKndetNquxLIhWDDTybON1N4flKMY9Bq8/rm5k2ddW6X6LvFpqJBQeiKRp4yG3FqI/zOyQuA==",
-      "license": "MIT",
-      "dependencies": {
-        "prosemirror-keymap": "^1.2.2",
-        "prosemirror-model": "^1.24.1",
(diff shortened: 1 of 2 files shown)


alexvcasillas (approved):
Thanks for checking, upstreaming the fix to `prosemirror` and applying the patch here :)

bdbch (approved):

PR #6089, opened 2025-02-07
[RFC] Add vitest integration for package unit testing

alexvcasillas:
> [!NOTE]  
> RFC
> We would like to discuss if there's actually a need to introduce Vitest to have another testing step along with Cypress or if with Cypress would be enough

This pull request introduces several changes to the project, primarily focusing on adding unit testing capabilities using Vitest, updating dependencies, and configuring the testing environment. The most important changes are summarized below:

### Testing Enhancements:

* Added new unit test scripts using Vitest in `package.json` and `packages/extension-bold/package.json`. [[1]](diffhunk://#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R23-R24) [[2]](diffhunk://#diff-73fd1cafcadb2a6d81a4bde99e631eac6d647e0fe5dc8afd243b6a6c586f125dL44-R49)
* Created a new test configuration file `vitest.config.ts` to set up the testing environment with JSDOM and alias resolutions.
* Added a specific Vitest configuration file for the `extension-bold` package to include JSDOM environment setup and alias resolutions.

### Dependency Updates:

* Added `jsdom` and `vitest` as new dependencies in `package.json` and `packages/extension-bold/package.json`. [[1]](diffhunk://#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R68) [[2]](diffhunk://#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R80) [[3]](diffhunk://#diff-73fd1cafcadb2a6d81a4bde99e631eac6d647e0fe5dc8afd243b6a6c586f125dL32-R35)

### New Tests:

* Added a new test file `bold.spec.ts` for the `extension-bold` package to test the initialization and commands of the Bold extension.
* Created a corresponding HTML file `test.dom.html` to serve as the test document for the Bold extension tests.


vitest.config.ts
```diff
@@ -0,0 +1,19 @@
...
+import { defineConfig } from 'vitest/config'
+
+export default defineConfig({
+  test: {
+    environment: 'jsdom',
+    include: ['packages/**/*.{test,spec}.{js,ts,jsx,tsx}'],
+  },
+  resolve: {
+    alias: {
+      '@tiptap/extension-bold': path.join(__dirname, 'packages/extension-bold/src'),
```
alexvcasillas, line 12: The alias resolution is needed so the `Vitest` runner knows how to resolve the `@tiptap/extension-bold`, `@tiptap/extension-document`, and other import aliases correctly

packages/extension-bold/test.dom.html
```diff
@@ -0,0 +1,11 @@
+<!DOCTYPE html>
```
alexvcasillas, line 1: Since we rely on `JSOM` to mount an in-memory `DOM` so we can bind the `Editor` to an `HTMLElement` this file is needed as it serves `JSDOM` for the initial `DOM` structure

packages/extension-bold/src/bold.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+let dom
+let document: Document
+
+const getRootElement = () => {
+  return document.querySelector('#editor') as Element
+}
+
+describe('Bold Extension', () => {
+  beforeAll(async () => {
+    dom = await JSDOM.fromFile(path.resolve(__dirname, '..', 'test.dom.html'), {
```
alexvcasillas, line 21: `JSDOM` initialization, maybe this can be extracted into a separate utility

packages/extension-bold/src/bold.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+        element: getRootElement(),
+        content: '<p>Example Text</p>',
+        extensions: [Document, Paragraph, Text, Bold],
+      })
+
+      const boldExtension = editor.extensionManager.extensions.find(
+        extension => extension.name === 'bold',
+      )
+
+      expect(boldExtension).toBeTruthy()
```
alexvcasillas, line 49: You can test for an extension's presence within the editor

packages/extension-bold/src/bold.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+
+  describe('Commands', () => {
+    it('defines commands within the editor', () => {
+      const editor = new Editor({
+        element: getRootElement(),
+        content: '<p>Example Text</p>',
+        extensions: [Document, Paragraph, Text, Bold],
+      })
+
+      expect(editor.commands.setBold).toBeDefined()
```
alexvcasillas, line 61: You can test that commands inserted by extensions are present in the editor

packages/extension-bold/src/bold.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+    it('allows command extensions to be called via editor commands', () => {
+      const editor = new Editor({
+        element: getRootElement(),
+        content: '<p>Example Text</p>',
+        extensions: [Document, Paragraph, Text, Bold],
+      })
+
+      editor.commands.setBold()
+
+      expect(editor.getText()).toContain('Example Text')
```
alexvcasillas, line 74: This is very naive, but we could get the specific node from the editor and check for it's `textStyle` here, or their marks to check that the `bold` mark is actually being applied

alexvcasillas:
> Should this only live on `extension-bold`? I guess every extension would have it's own test run which we can run through with turbo?

As this is more like the initial state of the Vitest setup I wanted to make sure that it works fine with one extension and now we could build up from there to have it in all extensions, either way, there's no need from `turbo` unless required, since running `npm run test:unit` in the root, will run all of the unit tests in the `packages` folder.

packages/extension-bold/package.json
```diff
@@ -29,7 +29,10 @@
     "dist"
   ],
   "devDependencies": {
-    "@tiptap/core": "^2.11.5"
+    "@tiptap/core": "^2.11.5",
+    "@types/jsdom": "^21.1.7",
+    "jsdom": "^26.0.0",
+    "vitest": "^3.0.5"
```
nperez0111, line 35: I'd not make this a dev dep of the package. It should only be for testing & having it in the root should be enough

packages/extension-bold/src/bold.spec.ts
nperez0111, line 1: I'm not sure about this, the sorts of tests that you can assert here are not anything that cannot be exercised by cypress in an e2e fashion.

E2E sort of is the best approach for editor testing because it exercises actual paths people would interface with rather than just that commands exist or not

vitest.config.ts
nperez0111: Seems verbose, I'd have thought there was a way to resolve using a tsconfig

turbo.json
```diff
@@ -14,6 +14,10 @@
     "dev": {
       "persistent": true,
       "cache": false
+    },
+    "test": {
+      "dependsOn": ["^build"],
```
nperez0111, line 19: Does it depend on build? based on your path resolution it is using the source files & not the built files

nperez0111 (approved):
This PR should have been done on the `next` branch since that will be the next major version. v2 should be in maintenance only mode right now

nperez0111 (changes requested):
Sorry meant to reject

vitest.config.ts
bdbch: I also ran into a similar issue with a private project and didn't find a quick way to resolve from a tsconfig. There are some posts out there talking about getting it from the tsconfig and also a vite plugin.

However I think what we could do is just parse package names and paths via the filesystem to automate this and have control over it.

Issue #6087, opened 2025-02-06
[Bug]: YouTube Extension accepting music.youtube.com/playlist links, but failing on them


bjhess:
### Affected Packages

@tiptap/extension-youtube

### Version(s)

2.11.2

### Bug Description

We are on version 2.11.2, but I see the same basic issue on [the extension's page in the Tiptap docs](https://tiptap.dev/docs/editor/extensions/nodes/youtube). Here's what happens when pasting the YouTube Music playlist URL `https://music.youtube.com/playlist?list=PLrggHaKS2xueclxIYbhkSiPai3da7cdWB&si=Xgxani5KWk3ieFnc` into the editor on the docs site:

https://github.com/user-attachments/assets/16a1d070-02b8-4236-bb4b-0922249dae85

YouTube Music links to albums (e.g. `https://music.youtube.com/playlist?list=OLAK5uy_mrF_EHJJul_9cUfE-snfFgdEY_nggl9c0&feature=shared`) also fail in a similar way. Albums appear to be implemented as playlists by YouTube music.

### Browser Used

Chrome

### Expected Behavior

I would be fine if such links did not get caught by the YouTube extension and thus did not resolve at all. 

Unless there is a way to provide a "video player" like what happens when YouTube Music links directly to songs (e.g. `https://music.youtube.com/watch?v=FPrHuDvXEBw`) are embedded. 

This could potentially be found by going from the YouTube _Music_ link:

https://music.youtube.com/playlist?list=OLAK5uy_mrF_EHJJul_9cUfE-snfFgdEY_nggl9c0

To the embed code similar to what is found at the _YouTube_ link (e.g.  replace the `music` subdomain with `www`):

https://www.youtube.com/playlist?list=OLAK5uy_mrF_EHJJul_9cUfE-snfFgdEY_nggl9c0

In this example, the embed code is:

```html
<iframe width="560" height="315" src="https://www.youtube.com/embed/videoseries?si=GoeLuM76GMGRt4LG&amp;list=OLAK5uy_mrF_EHJJul_9cUfE-snfFgdEY_nggl9c0" title="YouTube video player" frameborder="0" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture; web-share" referrerpolicy="strict-origin-when-cross-origin" allowfullscreen></iframe>
```

### Dependency Updates

- [x] Yes, I've updated all my dependencies.

bdbch:
[...]

PR #6096, opened 2025-02-12
feat(youtube): enhance embed URL generation to support video lists

Simon-He95:
## Checklist

## Related Issues

close: #6087


packages/extension-youtube/src/utils.ts
```diff
@@ -26,7 +26,10 @@ export interface GetEmbedUrlOptions {
   startAt?: number;
 }
 
-export const getYoutubeEmbedUrl = (nocookie?: boolean) => {
+export const getYoutubeEmbedUrl = (nocookie?: boolean, withV?:boolean) => {
```
bdbch, line 29: Not sure what `withV` means exactly - video lists? Could we find a better name for the variable

bdbch: If it's about the V in the URL, how about we describe what the V will actually do instead of just that a V was supplied?

bdbch (changes requested):
Just a small nitpick

packages/extension-youtube/src/utils.ts
Simon-He95: I change it to isPlayList，it should be more expressive.

bdbch (approved):
Looks good, thanks!

PR #6132, opened 2025-02-27
Fix/enforce-type-imports-so-that-bundler-ignores-types

arnaugomez:
## Changes Overview
Solves a bug where the index.js file generated in the dist folder during the build command contained imports of types, even though they do not exist at runtime. This caused certain bundlers like esbuild to throw an error when using the library. This is a blocking error, as it makes impossible to use our library in frameworks like Vite. 

The solution was to enforce type imports so that the bundler ignores types at the build step

## Implementation Approach
- Add an eslint rule to enforce type imports
- Apply eslint rule

## Testing Done
Run automated tests. No new tests were created

## Verification Steps
Check that the index.js files in the dist folder no longer contain type imports.

## Checklist


nperez0111 (approved):

bdbch (approved):
LGTM

PR #6174, opened 2025-03-12
fix: clone node before dragging for better dispaly

JinForYoda:
## Changes Overview
Using editor's `dom` caused dragging with overlaying elements. Instead we need to copy editor's dom to purify draggable node

## Testing Done
Tested with `@tiptap/react` using `ReactNodeViewRenderer`

## Verification Steps
1. Implement `addNodeView` with `ReactNodeViewRenderer`
2. Make it draggable
3. Some elements should be on another layer above the dragged node
4. Try dragging it

## Additional Notes

1. Before fix 
CleanShot 2025-03-12 at 18 59 51

2. After fix
CleanShot 2025-03-12 at 18 58 23

## Checklist


packages/core/src/NodeView.ts
```diff
@@ -98,7 +98,9 @@ export class NodeView<
       y = handleBox.y - domBox.y + offsetY
     }
 
-    event.dataTransfer?.setDragImage(this.dom, x, y)
+    const clonedNode = this.dom.cloneNode(true) as HTMLElement
```
bdbch, line 101: Do we need to clean this up after cloning the node?

bdbch (approved):
Looks good in general as we only use it for the drag image - however I wonder if we need to manually clean the cloned DOM node.

packages/core/src/NodeView.ts
JinForYoda: We don't insert it into the DOM, so I believe we don't need manual cleanup

bdbch: Alright, thanks for the clarification. I'll just double check that myself and will merge when I find it's fine like that. Thanks for your contribution!

PR #6182, opened 2025-03-16
fix(core): add even more strict types for Nodes, Marks, and Extensions

nperez0111:
## Changes Overview

- **fix: update the `AnyExtension` type to use the new base type `Extendable`**
- **fix: use the more specific config sub-types for Marks, Nodes and Extensions**
- **feat: add the ability to unmount an editor instance, without having to destroy and recreate it**
- **fix: return more specific types for configure and extend on Marks, Nodes, and Extensions**

This also introduces a new `unmount` method, which allows for mounting and remounting using the same editor options
## Checklist


bdbch:
Could you add the necessary changeset files for all changes? I'll take a look at the PR in the meanwhile.

packages/core/src/Editor.ts
```diff
@@ -162,6 +165,29 @@ export class Editor extends EventEmitter<EditorEvents> {
...
+      if (dom && dom.editor) {
+        delete dom.editor
+      }
+      this.editorView.destroy()
+    }
+    this.editorView = null
+    this.isInitialized = false
+
+    if (this.css) {
+      this.css.remove()
```
bdbch, line 186: I think we could just use optional chaining here removing the unneeded if-wrapping.

bdbch (approved):
Looks good to me - just some minor improvement, otherwise good to go.

nperez0111:
Updated! Thanks @bdbch!

btw, you could get the pkg pr releases working again if you were to release another `next` version.

I'm also going to make a PR into the docs for this too

bdbch:
Sure thing.

Will look into why the releases are not going out right now.

bdbch:
For the docs, could you work off the `feature/versioning` branch and add it to the 3.x docs? I'm currently finalizing the documentation versioning and am a bit worried about merging getting too crazy after that. 😂

nperez0111:
> Will look into why the releases are not going out right now.

It's just that there is a new package that has not yet been released to NPM yet. So would just need to make a new `next` version by releasing

bdbch:
I'll probably kick off a new `next` release today so they should work again after that.

PR #6204, opened 2025-03-25
refactor: export focusEvents plugin key

jomifepe:
## Changes Overview

This PR exports the `PluginKey` of the `FocusEvents` extension plugin. 

### Why?

In order to access the `meta` property using `getMeta`, we need to have access the plugin key (see example of [`prosemirror-tables`](https://github.com/ProseMirror/prosemirror-tables/blob/b0130d3bea42c368d38c47dcd5ff09f40dfdd33e/src/columnresizing.ts#L17)). This is useful to identify focus transactions and perform actions upon them. The solution is either to accept a `pluginKey` prop that can be passed with `configure` or export the defined key. I chose the latter because this is an extension used within `Editor`, and the users don't usually configure it.

We do have the `focus` and `blur` events, but we need a separate listener for them. With this exported, we can have a single `transaction` listener and we can check the `meta` to distinguish the type.

## Implementation Approach

Extract an existing `PluginKey` to a `const` and export it.

## Testing Done

May not be applicable for such a change.

## Checklist


jomifepe:
I see that the meta property is set by prosemirror, not by tiptap. Sorry about that, closing 🙂

bdbch (approved):

PR #6205, opened 2025-03-25
Rename focus extension default export from FocusClasses to Focus

arnaugomez:
## Changes Overview
The Focus extension is exported as FocusClasses instead of Focus. This breaks autocomplete, so when I type "Focus" my code editor does not autocomplete the import as Focus extension. Instead, it autocompletes it as FocusClasses. Which contradicts the documentation.

## Implementation Approach
Rename the imports in the extension code.

## Testing Done
Run tests. Does not introduce new tests.

## Verification Steps
The default export is Focus instead of FocusClasses

## Additional Notes
The bug was found by user  @sparrow-chik-chrk in this PR: https://github.com/ueberdosis/tiptap-docs/pull/92

## Checklist

## Related Issues
https://github.com/ueberdosis/tiptap-docs/pull/92

### DIFF
--- a/.changeset/purple-wasps-jam.md
+++ b/.changeset/purple-wasps-jam.md
@@ -1,0 +1,5 @@
+---
+"@tiptap/extension-focus": patch
+---
+
+Rename focus extension default export from FocusClasses to Focus
--- a/packages/extension-focus/src/focus.ts
+++ b/packages/extension-focus/src/focus.ts
@@ -27,7 +27,7 @@ export interface FocusOptions {
  * This extension allows you to add a class to the focused node.
  * @see https://www.tiptap.dev/api/extensions/focus
  */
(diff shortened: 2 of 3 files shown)


bdbch (changes requested):
Not sure if we should rename the default exports in a minor/patch version as people may rely on the named exports. 

I see two solutions:

1. Move this into the next branch so we include it in the next major version.
2. Keep the old export, but make it re-export the `Focus` export

Example:

```js
import { Focus } from './focus.js'

export * from './focus.js'

export default Focus

export const FocusClasses = Focus
```

arnaugomez:
@bdbch True, that's a breaking change. I'll move it to the next branch because there is no hurry to merge this now, we can wait until editor v3.

Issue #6237, opened 2025-04-04
[Bug]: Adding empty td tag inside table throws RangeError


its-me-dhina:
### Affected Packages

table

### Version(s)

@tiptap/extension-table

### Bug Description

const insert = `<table><tr><td></td></tr></table>`;

This is the content I am try to set into editor using 

```
 editor?.chain()
        .focus()
        .insertContentAt(cursorPosition, content, {
          updateSelection: true,
          parseOptions: {
            preserveWhitespace: false,
          },
        })
        .run();
```

I am getting the following error. 

`Error inserting content: RangeError: Invalid content for node tableCell: <>`

Dependencies
```
    "@tiptap/extension-table": "^2.6.6",
    "@tiptap/extension-table-cell": "^2.6.6",
    "@tiptap/extension-table-header": "^2.6.6",
    "@tiptap/extension-table-row": "^2.6.6",
    "@tiptap/react": "^2.6.6",
    "@tiptap/starter-kit": "^2.6.6",
```

### Browser Used

Chrome

### Expected Behavior

Should support the empty <td></td> insertion

### Dependency Updates

- [x] Yes, I've updated all my dependencies.

m-idler:
Are there any news regarding this issue as the existing MR was closed without any followup?

BickelLukas:
We are running across this issue when inserting AI generated content. So we don't have much control over the content that will be inserted. But right now if it contains a table with an empty row the insert fails. Any updates here?

vBenTec:
It's a pity that this couldn't be resolved. However, as a temporary solution for those who encounter the same issue (v^3.18.0), you can work around it by parsing the empty table cell and inserting, for example, a non-breaking space (&nbsp;).

PR #6241, opened 2025-04-06
Fix: Adding empty td tag inside table throws RangeError

Meyazhagan:
## Changes Overview
Closes: #6237 

By default we using `DOMParser.parseSlice` (this function is not parsing the full node.), and no option to disable it 
so, I created an option `useParseSlice`,  

https://discuss.prosemirror.net/t/domparser-parseslice-not-parsing-correctly-empty-table-cell/5510 

## Checklist

### DIFF
--- a/packages/core/src/commands/insertContentAt.ts
+++ b/packages/core/src/commands/insertContentAt.ts
@@ -74,6 +74,7 @@ export const insertContentAt: RawCommands['insertContentAt'] = (position, value,
 
     try {
       content = createNodeFromContent(value, editor.schema, {
+        slice: false,
         parseOptions: {
           preserveWhitespace: 'full',
           ...options.parseOptions,
--- a/packages/core/src/commands/insertContentAt.ts
+++ b/packages/core/src/commands/insertContentAt.ts
@@ -50,6 +50,12 @@ declare module '@tiptap/core' {
            * Whether to throw an error if the content is invalid.
            */
           errorOnInvalidContent?: boolean
+
+          /**
+           * Whether to use parseSlice function to parse the content.
+           * @default true
+           */
+          slice?: boolean
         },
       ) => ReturnType
     }
@@ -67,6 +73,7 @@ export const insertContentAt: RawCommands['insertContentAt'] = (position, value,
       updateSelection: true,
       applyInputRules: false,
       applyPasteRules: false,
+      slice: true,
       ...options,
     }
 
@@ -74,7 +81,7 @@ export const insertContentAt: RawCommands['insertContentAt'] = (position, value,
 
     try {
       content = createNodeFromContent(value, editor.schema, {
-        slice: false,
+        slice: options.slice,
         parseOptions: {
           preserveWhitespace: 'full',
           ...options.parseOptions,
--- a/.changeset/warm-yaks-speak.md
+++ b/.changeset/warm-yaks-speak.md
@@ -2,4 +2,4 @@
 "@tiptap/core": minor
 ---
 
-fix parsing of empty html table in insertContentAt func
+add slice option to insertContentAt and insertContent function
--- a/.changeset/warm-yaks-speak.md
+++ b/.changeset/warm-yaks-speak.md
@@ -1,0 +1,5 @@
+---
+"@tiptap/core": minor
+---
+
+fix parsing of empty html table in insertContentAt func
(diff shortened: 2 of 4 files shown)


its-me-dhina:
Verified the changes on local and the fix seems to be working fine. Can we prioitize merging this request?

Thanks!

packages/core/src/helpers/createNodeFromContent.ts
```diff
@@ -10,7 +10,7 @@ import { Content } from '../types.js'
 import { elementFromString } from '../utilities/elementFromString.js'
 
 export type CreateNodeFromContentOptions = {
-  slice?: boolean
```
arnaugomez, line 13: Isn't it possible to disable `DOMParser.parseSlice` already by specifying the parameter `slice: false`? @Meyazhagan . I believe your PR only changes the name of the parameter from `slice` to `useParseSlice`. It also changes its default value from `true` to `false`, thus introducing a breaking change.

Therefore, I believe you could achieve the desired behavior of disabling `DOMParser.parseSlice` with the current code, by setting the `slice` parameter to `false`. Could you check that this is true?

arnaugomez (changes requested):

packages/core/src/helpers/createNodeFromContent.ts
arnaugomez: If it is already possible to disable `DOMParser.parseSlice` by calling `createNodeFromContent` with the `slice` parameter set to `false`, then this PR is likely not necessary. Or is there anything I'm missing? Correct me if I'm wrong.

hafeezulkareem-yellow: @arnaugomez is it possible to disable it while inserting the content into the editor?

Ref: https://github.com/ueberdosis/tiptap/issues/6237

arnaugomez: I think it's possible to achieve the same behavior of disabling `DOMParser.parseSlice` by calling `createNodeFromContent` with the slice parameter set to `false`. Have you tried it? @hafeezulkareem-yellow  @Meyazhagan

Meyazhagan: @arnaugomez yeah I tried this changes, it is working!

quick question: `createNodeFromContent` function is being used in two places,  `createDocument` and in `insertContentAt` both the place also we setting `slice` to `false` to use `parse` function, instead we can change in  `createNodeFromContent`, by default to using `parse` function!

is there any good reason to have `parseSlice` as default parser ?

arnaugomez: @Meyazhagan Yes, sure, if you think we can refactor it so that the code is simpler, create another PR with your changes and we'll review it. Thanks for your suggestion!

arnaugomez:
@bdbch Could you review this? To me it looks good but I'm not familiar with this code and I want to make sure it does not break things.

bdbch (changes requested):
I don't like that we hard-code `slice: false` into the `insertContentAt` function as it can potentially break the behavior for existing implementations.

Could we please have that as an option for `insertContent` and `insertContentAt` if this needs to be changed depending on what kind of content you want to insert and how you want to do it?

packages/core/src/commands/insertContentAt.ts
```diff
@@ -67,13 +73,15 @@ export const insertContentAt: RawCommands['insertContentAt'] = (position, value,
...
       applyPasteRules: false,
+      slice: true,
       ...options,
     }
 
     let content: Fragment | ProseMirrorNode
 
     try {
       content = createNodeFromContent(value, editor.schema, {
+        slice: options.slice,
```
arnaugomez, line 84: I am concerned this is a breaking change. Before these changes, the `createNodeFromContent` was called with the `slice` value being false. But now it is called with a default value of `true`. If the default value of options.slice was `false` then this would not be a breaking change. But if the value should be true by default, then you have my approval. Just wanted to point this out.

arnaugomez (approved):
LGTM

packages/core/src/commands/insertContentAt.ts
bdbch: I would also rather keep the previous default to not cause breaking changes.

bdbch:
@Meyazhagan could you rebase this to the current develop branch as we released V3, deprecating the next branch & would need to update most of our PR's anyway? Thanks a lot!

PR #6283, opened 2025-04-20
Fixed mergeAttributes utility for correct merge of inline styles

lyohaplotinka:
## Changes Overview
The style attribute merge does not work correctly in some cases. For example, if the inline style contains a link to a background image, a simple colon split of the string is not suitable for splitting the string into keys and values. This results in invalid styles. 

I've updated the test for the mergeAttributes utility, it clearly shows which cases will not work correctly. 

## Implementation Approach
The styles property key is obtained by the first colon index, anything after it is considered a value. Only non-empty keys and values are included in the result.

## Testing Done
Updated the test for the mergeAttributes utility, added several cases to check for corrupted styles, as well as styles with background images and font families.

## Verification Steps
The problem is noticeable when adding custom attributes based on styles. In my work, I need to preserve styles from the input HTML as much as possible, since I'm working on a visual email editor. 

## Checklist


bdbch:
@lyohaplotinka could you rebase this to the current develop branch as we released V3, deprecating the next branch & would need to update most of our PR's anyway? Thanks a lot!

lyohaplotinka:
@bdbch of course I will. I'd like to see it in 2.x though, as I use it in a project and am unlikely to upgrade anytime soon. Is that possible?

bdbch:
@lyohaplotinka we could merge it backwards to the v2 branch and make a release if needed

lyohaplotinka:
@bdbch that would be great, thanks a lot

bdbch:
Link me the new PR here - I'll take a look and will back-merge it to V2.

PR #6233, opened 2025-04-03
fix: use `editorView` for `isDestroyed` check for SSR

nperez0111:
## Changes Overview

This should resolve this issue: https://github.com/ueberdosis/tiptap/discussions/5793#discussioncomment-12706934
## Checklist
- [ ] I have created a [changeset](https://github.com/changesets/changesets) for this PR if necessary.
- [ ] My changes do not break the library.
- [ ] I have added tests where applicable.
- [ ] I have followed the project guidelines.
- [ ] I have fixed any lint issues.

PR #6284, opened 2025-04-20
Fix broken React unmount because of non-existing view on unmount

bdbch:
## Changes Overview

This pull request addresses a bug fix in the `EditorContent` component of the `@tiptap/react` package. The fix prevents a crash in Firefox browsers caused by attempting to set properties on an uninitialized editor view when the editor is unmounted, such as during React's Strict Mode.

### Bug Fixes:

* [`.changeset/shaggy-hats-rescue.md`](diffhunk://#diff-7ad48e49078f41227d545319e8c06e2d578d95efa8e230312797b7fbc20be5c6R1-R5): Added a patch-level changeset entry describing the bug fix for the `EditorContent` component crash in Firefox browsers.
* [`packages/react/src/EditorContent.tsx`](diffhunk://#diff-c1d83c0fa32f3b18f2d8e5b3eb81f9d951c0d6de903191d495d169d248188ceeR167-R175): Wrapped the `editor.view.setProps` call in a `try-catch` block to handle cases where the editor view is not initialized, preventing errors during unmounting.

## Implementation Approach

Added a simple try-catch block around the `editor.view.setProps()` call as the view could not be available yet.

## Testing Done

Tested on Gecko and Chromium browsers if the error still occurs.

## Verification Steps

See above

## Additional Notes

N/A

## Checklist

### DIFF
--- a/packages/react/src/EditorContent.tsx
+++ b/packages/react/src/EditorContent.tsx
@@ -164,10 +164,14 @@ export class PureEditorContent extends React.Component<
 
     this.initialized = false
 
-    if (!editor.isDestroyed) {
-      editor.view.setProps({
-        nodeViews: {},
-      })
+    try {
+      if (!editor.isDestroyed) {
+        editor.view.setProps({
+          nodeViews: {},
+        })
+      }
+    } catch {
+      // ignore error as we can't set props on a non-existing view
     }
 
     if (this.unsubscribeToContentComponent) {
--- a/.changeset/shaggy-hats-rescue.md
+++ b/.changeset/shaggy-hats-rescue.md
@@ -1,0 +1,5 @@
+---
+'@tiptap/react': patch
+---
+
+Fixed a bug in the EditorContent component that caused a crash in Firefox based browsers because of the editor view not being available when an uninitialized editor is unmounted (for example via Strict mode)


packages/react/src/EditorContent.tsx
```diff
@@ -164,10 +164,14 @@ export class PureEditorContent extends React.Component<
 
     this.initialized = false
 
-    if (!editor.isDestroyed) {
-      editor.view.setProps({
-        nodeViews: {},
-      })
+    try {
```
arnaugomez, line 167: I wonder if, instead of handling the error with a try-catch block, an extra check could be performed on the editor view to determine if the `setProps` method can be called.

For example, this property: https://prosemirror.net/docs/ref/#view.EditorView.isDestroyed

But the solution of the try-catch block works too.

arnaugomez (approved):

packages/react/src/EditorContent.tsx
bdbch: The problem with that was that there's no actual editor view to check on in the first place. The thrown error actually comes from a check if the editor view is accessible or not (that's used for using Tiptap on the server where you would not have an EditorView).

https://github.com/ueberdosis/tiptap/blob/next/packages/core/src/Editor.ts#L302

Maybe there's another way but I think just catching the error to proceed with the editor component logic should be fine.

nperez0111: This PR would have actually fixed this issue: https://github.com/ueberdosis/tiptap/pull/6233

`this.editorView` stored the actual view whereas `this.view` is the proxied one. If `editor.isDestroyed` was based on the `editor.editorView` instead of the `editor.view`, you would not have this issue.

nperez0111: I think this try catch should be reverted, as it does not even check what kind of error was thrown & may silence a real error that occured

PR #6300, opened 2025-04-25
Multiple triggers on mention component

arnaugomez:
## Changes Overview
Support multiple triggers in mention component.

Example: `@` is used to mention people and `#` is used to mention a place.

## Implementation Approach
- Expose a new "suggestions" option where multiple suggestion triggers can be attached to the Mention extension.

## Testing Done
Manually, with the demos. No automated tests. Let me know if you need automated tests.

## Verification Steps
1. Open the Mention demo.
2. Verify the behaviour of the Mention extension is the same as before. Try creating and deleting a mention.
3. Open the MultiMention demo.
4. Verify you can use both @ and # to trigger the mention extension. Verify that the list of options is different depending on the trigger character.

## Additional Notes
The motivation of this PR is to enable a common workflow of AI assistant applications, where different resources (files, urls, users) are mentioned in the same prompt. It is also a requested feature in discussion #1768.

## Checklist

## Related Issues
#1768


packages/extension-mention/src/mention.ts
```diff
@@ -240,30 +265,63 @@ export const Mention = Node.create<MentionOptions>({
...
-        ...this.options.suggestion,
-      }),
-    ]
+        overrideSuggestionOptions: suggestion,
+        extensionName: this.name,
+        char: suggestion.char,
+      })
+    })
+
+    this.storage.getSuggestionFromChar = char => {
```
bdbch, line 315: No problem to do something like this but it brings me back to an idea I had before. It feels weird to store internal functions in storage as I'd expect storage to contain data useful for the users.

In the future in another breaking change we could maybe change the API of the `Extension.create` call to accept a callback function like this:

```js
// we could maybe even have editor/other context related things as arguments
const MyNode = Node.create((editor) => {
  function internalHelperFunction() {

  }

  const internalVar = "my-internal-var"

  return {
    // ... your extension config as usual
  }
})
```

what are your thoughts?

bdbch: But no need to do anything right now

bdbch (approved):
LGTM

bdbch:
It would be nice if we could have another test case that handles multiple suggestion triggers?

packages/extension-mention/src/mention.ts
arnaugomez: I agree with this. Sometimes a helper function can help you reuse duplicated code.

bdbch: In general having an enclosed scope for an extension can be really helpful. We could make this change retroactive by allowing both a callback function & the old way of creating extensions.

arnaugomez:
Ok, I'll add some tests on Monday

packages/extension-mention/src/mention.ts
bdbch: Created a PR for this:
https://github.com/ueberdosis/tiptap/pull/6301

Maybe we can merge this and use this instead of using storage?

arnaugomez:
@bdbch I just added some tests

bdbch (approved):
LGTM! We'd just need to make sure the docs are up to date. But feel free to merge.

arnaugomez:
I'll update the `next` docs then

PR #6301, opened 2025-04-25
Extension, Mark, and Node classes allow functions for creation

bdbch:
## Changes Overview

This pull request enhances the flexibility of the `create` and `extend` methods in the `Extension`, `Mark`, and `Node` classes by allowing configuration objects to be provided as functions. It also improves code readability by adding detailed comments and JSDoc annotations.

### Enhancements to `create` methods:
* Updated the `create` method in the `Extension` class to accept either a configuration object or a function that returns a configuration object. Added JSDoc annotations for better documentation. (`packages/core/src/Extension.ts`, [packages/core/src/Extension.tsL18-R27](diffhunk://#diff-5646f4773d47ddc5a0ff04bd8aa67213e7457fbac3b135e3237ffda72737b323L18-R27))
* Applied the same enhancement to the `create` method in the `Mark` class, allowing for functional configuration and adding comments for clarity. (`packages/core/src/Mark.ts`, [packages/core/src/Mark.tsL149-R156](diffhunk://#diff-c17789a64b74bdd437c357796b400a72ba9e777d4df45d7654c1e0164b67dcf6L149-R156))
* Similarly updated the `create` method in the `Node` class to support functional configuration and added JSDoc annotations. (`packages/core/src/Node.ts`, [packages/core/src/Node.tsL343-R350](diffhunk://#diff-9ad329d37c15a745e13b71b29c5b558533d8a9fd30f50cf0acd681c1ccbfc6feL343-R350))

### Enhancements to `extend` methods:
* Modified the `extend` method in the `Extension` class to handle functional configurations, improving flexibility. (`packages/core/src/Extension.ts`, [packages/core/src/Extension.tsL30-R41](diffhunk://#diff-5646f4773d47ddc5a0ff04bd8aa67213e7457fbac3b135e3237ffda72737b323L30-R41))
* Updated the `extend` methods in both the `Mark` and `Node` classes to support functional configurations, maintaining consistency across the codebase. (`packages/core/src/Mark.ts`, [[1]](diffhunk://#diff-c17789a64b74bdd437c357796b400a72ba9e777d4df45d7654c1e0164b67dcf6L189-R198); `packages/core/src/Node.ts`, [[2]](diffhunk://#diff-9ad329d37c15a745e13b71b29c5b558533d8a9fd30f50cf0acd681c1ccbfc6feL355-R364)

This change allows developers to create extensions like this:

```js
const MyCustomNode = Node.create(() => {
  // this encapsulated scope can be used for extension internal utility functions, variables and more

  return {
    name: 'my-custom-node',
    content: 'block+',
    // ... and your further configuration
  }
})
```

## Checklist


bdbch:
Added the documentation for this new option here:
https://github.com/ueberdosis/tiptap-docs/pull/176

arnaugomez (changes requested):
I like the idea. Before we merge, I'd like to point out that this solution does have some limitations. It's not possible to access the extension options and storage from the method. So the amount of helper functions, variables and methods you can initialize there is fairly limited. Most of them, you will need to initialize in the `onBeforeCreate` lifecycle method, where you have access to the editor state, extension options and storage (via the `this` object).

So, although this initializer function is a good step forward, I think the developer ergonomics can still be improved.

Perhaps a successful outcome would be to:
- Be able to define an internal storage object of the extension, which includes helper functions that are not exposed externally.
- Make the internal storage straightforward to initialize/update.

bdbch:
> I like the idea. Before we merge, I'd like to point out that this solution does have some limitations. It's not possible to access the extension options and storage from the method. So the amount of helper functions, variables and methods you can initialize there is fairly limited. Most of them, you will need to initialize in the `onBeforeCreate` lifecycle method, where you have access to the editor state, extension options and storage (via the `this` object).
> 
> So, although this initializer function is a good step forward, I think the developer ergonomics can still be improved.
> 
> Perhaps a successful outcome would be to:
> 
>     * Be able to define an internal storage object of the extension, which includes helper functions that are not exposed externally.
> 
>     * Make the internal storage straightforward to initialize/update.

The problem with accessing storage in that callback is that the logic is happening before any extension is even registered to the editor (and therefore no storages to access).

I think it would still be possible to access storage in functions defined in the `Create` callback - you'd basically pass through storage values as arguments in your normal event lifecycles and other extension related configuration option.

```js
const MyExtension = Extension.create(() => {
  const initialName = 'Jane Doe'

  function sayHello(name) {
    console.log(name)
  }

  // writing into storage wouldn't really work here - but you could split up logic into smaller chunks of methods that you could use below
  function createInitials(name) {
    return name.split(' ')
      .map(s => s[0].toUpperCase())
      .join('')
  }

  return {
    name: 'myExtension',

    addStorage() {
      name: initialName,
      initials: createInitials(initialName),
    },
    
    onUpdate() {
      sayHello(this.storage.name)
    },
  }
})
```

bdbch:
@arnaugomez anything we could do here to move this forward in your opinion? I think this would for now be a big improvement as you could actually initialize (for example) resizeObservers, mutationObservers, custom calculation classes, etc. without having them in storage and use them in all kind of places around the extension.

arnaugomez:
Ok tomorrow I'll have a look at it

Issue #6315, opened 2025-05-01
[Bug]: flushSync error on dynamic rendering of EditorContent


romarybi:
### Affected Packages

core, react

### Version(s)

2.11.9

### Bug Description

Warning: flushSync was called from inside a lifecycle method. React cannot flush when React is already rendering. Consider moving this call to a scheduler task or micro task.

The above error occurs when the <EditorContent /> is rendered conditionally. 

### Browser Used

Chrome

### Code Example URL

https://codesandbox.io/p/sandbox/tqc8hx

### Expected Behavior

No error to occur.

### Additional Context (Optional)

I have a use case where the editor instance needs to be saved in memory and its content should be rendered conditionally. In the sandbox provided, open up the console. After clicking the show button, you should see the error occur.

### Dependency Updates

- [x] Yes, I've updated all my dependencies.

PR #6318, opened 2025-05-03
fix: avoid flushSync call during <EditorContent /> lifecycle

romarybi:
## Changes Overview
Fixed the bug where rendering the EditorContent dynamically would throw a flushSync error.

## Implementation Approach
Changed ReactRenderer flushSync condition to use a new boolean isEditorContentInitialized instead of isInitialized. This change makes sure that flushSync is not called in the EditorContent component lifecycle. 

## Testing Done
All tests passed locally.

## Verification Steps
[Sandbox](https://codesandbox.io/p/sandbox/tqc8hx) that shows the error that is now fixed. Create the tiptap editor that is initialized with content that includes a node that uses a react component. Create a conditional boolean to render EditorContent. Toggle the boolean and you should see the error in the console.

## Additional Notes
Previously, isInitialized was used and should probably be removed now. The reason I kept it in is it might be a breaking change for users who rely on it in their codebase.

## Checklist
- [] I have added tests where applicable.

## Related Issues
[Issue](https://github.com/ueberdosis/tiptap/issues/6315)

### DIFF
--- a/.changeset/three-peaches-kick.md
+++ b/.changeset/three-peaches-kick.md
@@ -1,0 +1,5 @@
+---
+"@tiptap/react": patch
+---
+
+Fixed the error where flushSync would run in EditorContent lifecycle


germs12:
Would love to see this merged sooner than later.  Thank you!

teo-dellamico (approved):

bdbch:
@romarybi could you rebase this to the current develop branch as we released V3, deprecating the next branch & would need to update most of our PR's anyway? Thanks a lot!

PR #6375, opened 2025-05-20
Prevent overriding default renderToMarkdown options

DrewJohnsonGT:
## Changes Overview

The `renderToMarkdown` function options are being spread in a way that will override the default `nodeMapping` or `markMapping` options when provided

So there is no way to simply override one node or mark mapping without overriding all of them.

## Implementation Approach

Simply destructuring the options and spreading specifically to `nodeMapping` and `markMapping` to prevent the final `...options` from overriding the defaults 

## Verification Steps

Add a single custom or default override to `nodeMapping`
Ex: 
```
renderToMarkdown({
  content: editor.getJSON(),
  extensions: editor.extensionManager.extensions,
  options: {
    nodeMapping: {
      templateInput: (props) => {
        return `{{${props.node.attrs.inputName}}}`;
      },
    },
  },
});
```

See that other default `nodeMappings` like listItems, headings etc are still rendered to markdown as expected

## Checklist


packages/static-renderer/src/pm/markdown/markdown.ts
```diff
@@ -134,9 +135,9 @@ export function renderToMarkdown({
         highlight({ children }) {
           return `==${serializeChildrenToHTMLString(children)}==`
         },
-        ...options?.markMapping,
+        ...markMapping,
       },
-      ...options,
```
DrewJohnsonGT, line 139: This overrides `nodeMapping` and `markMapping` entirely as opposed to allow extending each

DrewJohnsonGT:
@bdbch should this just be closed per
`This is not a full implementation of a markdown renderer..`
For people wanting to extend from this starting point they'll just need to copy the source code and adjust (which is what I have done)
I also know there is more direct markdown support coming in the future so understand if this doesn't make sense to change at this time

On the other hand, this is a really small change that does allow extending this function without copying the full source code 🤷

PR #6402, opened 2025-05-31
fix: preserve table cell selection when opening context menu (#6396)

### LINKED ISSUE #6396: [Bug]: table context menu focus issue
### Affected Packages

packages/extension-table/

### Version(s)

2.7.2

### Bug Description

When you drag to select several table cells and right-click any but the last cell you selected, the table loses selection.

Pretty small issue which I think most users will work out pretty quick but I'm putting all my table controls in the context menu so it would be good if the focus was consistent. Thanks. Great product. 

### Browser Used

Chrome

### Code Example URL

https://tiptap.dev/docs/editor/extensions/nodes/table

### Expected Behavior

Select 2 or more cells in a row and right click the last you selected, the focus remains, but if you right-click any other cell, the selection is lost. Same behaviour in react and vue demos. 

_No response_

### Dependency Updates

ay690:
When right‐clicking on a cell that’s part of a multi‐cell selection, prevent
the default mousedown handler from collapsing the selection. This ensures the
table context menu can open without losing focus or clearing the existing
cell highlight. Closes #6396.

### DIFF
--- a/demos/src/Examples/Tables/React/index.jsx
+++ b/demos/src/Examples/Tables/React/index.jsx
@@ -1,31 +1,86 @@
 import './styles.scss'
 
-import { TableCell, TableKit } from '@tiptap/extension-table'
+import { Table as TableExtension, TableCell as DefaultTableCell, TableHeader, TableRow } from '@tiptap/extension-table'
 import { EditorContent, useEditor } from '@tiptap/react'
 import StarterKit from '@tiptap/starter-kit'
+import { Plugin } from 'prosemirror-state'
+import { CellSelection } from 'prosemirror-tables'
 import React from 'react'
 
-const CustomTableCell = TableCell.extend({
+/**
+ * CustomTableCell
+ *
+ * 1) Extends the default TableCell so we can add a `backgroundColor` attribute.
+ * 2) Installs a tiny ProseMirror plugin that intercepts right-clicks (button === 2)
+ *    inside any cell that is already part of a multi-cell selection, and prevents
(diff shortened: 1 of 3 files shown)


demos/src/Examples/Tables/Vue/index.vue
```diff
@@ -80,29 +80,71 @@
...
-        },
+        renderHTML: attributes => ({
+          'data-background-color': attributes.backgroundColor,
+          style: `background-color: ${attributes.backgroundColor}`,
+        }),
       },
     }
   },
+
+  addProseMirrorPlugins() {
```
bdbch, line 110: Shouldn't this be part of the extension instead of the demo?

demos/src/Examples/Tables/React/index.jsx
```diff
@@ -1,31 +1,86 @@
...
-        },
+        renderHTML: attributes => ({
+          'data-background-color': attributes.backgroundColor,
+          style: `background-color: ${attributes.backgroundColor}`,
+        }),
       },
     }
   },
+
+  addProseMirrorPlugins() {
```
bdbch, line 36: Same here

bdbch (changes requested):
Cool idea - however I don't think we should implement that in the demo but in the extension?

bdbch:
@ay690 could you rebase this to the current develop branch as we released V3, deprecating the next branch & would need to update most of our PR's anyway? Thanks a lot!

Issue #5846, opened 2024-11-19
[Bug]: React 19 messes with "Mentions" example


CHE1RON:
### Affected Packages

extension-mention

### Version(s)

2.9.1

### Bug Description

When React v19 is installed, the example over at https://tiptap.dev/docs/editor/extensions/nodes/mention stops working (= `ref` not being passed correctly) due to changes to the way refs are handled in general it would seem - and `ref` cannot be passed to `ReactRenderer` 🤔 

When using `forwardRef` everything works as expected though.

### Browser Used

Chrome

### Expected Behavior

Example works just as before

### Additional Context (Optional)

Clicking on a list entry still works

### Dependency Updates

- [X] Yes, I've updated all my dependencies.

nperez0111:
Thanks for pointing this out @CHE1RON I think I've resolved it & will push a new release

CHE1RON:
That's great to hear, and let me say I'm really impressed by your speed ⚡ 🏆

nperez0111:
Released with 2.10.0

CHE1RON:
I will test & report back!

PR #6405, opened 2025-06-03
add support for react 19 ref props

### LINKED ISSUE #6110: [Bug]: ReactRenderer is broken on React 19
### Affected Packages

react

### Version(s)

2.11.5

### Bug Description

I upgraded my project to React.js 19. I expected `ref` to be part of the props when using `ReactRenderer` for my mentions list extension. According to the documentation, I need to use `ReactRenderer` to render my mentions list. However, if I don't wrap my component with `forwardRef`, the `ref` in the props will be `undefined`.

### Browser Used

Chrome

### Code Example URL

_No response_

### Expected Behavior

I expected that when I pass `ref` as a prop, it would be available on the component rendered by `ReactRenderer` for the mentions list, since the React team suggested not using `forwardRef` anymore.

Temporary solution: Keep using `forwardRef` for this part until the TipTap team fixes this issue.
I believe this issue happens here because the condition checks for `'Symbol(react.forward_ref)'`
```js
function isForwardRefComponent(Component: any) {
  return !!(
    typeof Component === 'object'
    && Component.$$typeof?.toString() === 'Symbol(react.forward_ref)'
  )
}
```
https://github.com/ueberdosis/tiptap/blob/main/packages/react/src/ReactRenderer.tsx
However, if we pass ref as a prop, this condition is not valid anymore. As a suggestion, if you want to be backward compatible you can check React version and check different conditions for React 19

### Dependency Updates

bdbch:
## Changes Overview

This pull request includes updates to migrate the codebase to React 19 and introduces changes to improve type handling for React components. The migration involves updating dependencies and refactoring components to accommodate new React features, including the use of `ref` in `ReactNodeViewProps`.

### React 19 Migration

* Updated React and React DOM dependencies to version 19 in `package.json`, `pnpm-lock.yaml`, and `demos/package.json`. This includes updating their corresponding type definitions (`@types/react` and `@types/react-dom`) to version 19.0.0. [[1]](diffhunk://#diff-2759f93cee6a28002b544bc4f705b4890d72bbb85eea0d7ff670da00277e70f4L59-R60) [[2]](diffhunk://#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L46-R47) [[3]](diffhunk://#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L69-R69) [[4]](diffhunk://#diff-32824c984905bb02bc7ffcef96a77addd1f1602cff71a11fbbfdd7f53ee026bbL40-R46) [[5]](diffhunk://#diff-32824c984905bb02bc7ffcef96a77addd1f1602cff71a11fbbfdd7f53ee026bbL111-R112) [[6]](diffhunk://#diff-32824c984905bb02bc7ffcef96a77addd1f1602cff71a11fbbfdd7f53ee026bbL163-R163) [[7]](diffhunk://#diff-32824c984905bb02bc7ffcef96a77addd1f1602cff71a11fbbfdd7f53ee026bbR2731-R2741) [[8]](diffhunk://#diff-32824c984905bb02bc7ffcef96a77addd1f1602cff71a11fbbfdd7f53ee026bbR5579-R5583) [[9]](diffhunk://#diff-32824c984905bb02bc7ffcef96a77addd1f1602cff71a11fbbfdd7f53ee026bbR5604-R5607) [[10]](diffhunk://#diff-32824c984905bb02bc7ffcef96a77addd1f1602cff71a11fbbfdd7f53ee026bbR5788-R5790) [[11]](diffhunk://#diff-32824c984905bb02bc7ffcef96a77addd1f1602cff71a11fbbfdd7f53ee026bbL8221-R8241) [[12]](diffhunk://#diff-32824c984905bb02bc7ffcef96a77addd1f1602cff71a11fbbfdd7f53ee026bbL8241-R8263) [[13]](diffhunk://#diff-32824c984905bb02bc7ffcef96a77addd1f1602cff71a11fbbfdd7f53ee026bbL8587-R8615) [[14]](diffhunk://#diff-32824c984905bb02bc7ffcef96a77addd1f1602cff71a11fbbfdd7f53ee026bbR8686-R8698)

### Type Handling Improvements

* Introduced `ReactNodeViewProps` type in `packages/react/src/types.ts` to extend `NodeViewProps` with support for `ref`. Updated components and functions across the codebase to use this new type. [[1]](diffhunk://#diff-84f256c9526675be4ce59318afe5f494f4ef705bc836911864ff1f4f6159602dR1-R7) [[2]](diffhunk://#diff-1cee8646d2cfba37d6ce6a6e9a8d16f8caba0b99fc3a1ad0cb997ed8c7384d2eR8)
* Refactored `ReactNodeViewRenderer` and related classes to use `ReactNodeViewProps`, enabling better type safety and support for `ref`. [[1]](diffhunk://#diff-412f21f9d1fcee64a72af483ab1e8358936b96c7496f66b74622a6a44506e7fbL1-R1) [[2]](diffhunk://#diff-412f21f9d1fcee64a72af483ab1e8358936b96c7496f66b74622a6a44506e7fbR10) [[3]](diffhunk://#diff-412f21f9d1fcee64a72af483ab1e8358936b96c7496f66b74622a6a44506e7fbL48-R57) [[4]](diffhunk://#diff-412f21f9d1fcee64a72af483ab1e8358936b96c7496f66b74622a6a44506e7fbL79-R82) [[5]](diffhunk://#diff-412f21f9d1fcee64a72af483ab1e8358936b96c7496f66b74622a6a44506e7fbL99-R102) [[6]](diffhunk://#diff-412f21f9d1fcee64a72af483ab1e8358936b96c7496f66b74622a6a44506e7fbL305-R309) [[7]](diffhunk://#diff-412f21f9d1fcee64a72af483ab1e8358936b96c7496f66b74622a6a44506e7fbL317-R320)
* Updated React components in `demos/src` to use `ReactNodeViewProps` and handle `ref` for improved compatibility with React 19. [[1]](diffhunk://#diff-fb44a68dcde65822fc805522b2d8fe027ab60b026d50bf4a2e90c3ed2529bef6L1-R5) [[2]](diffhunk://#diff-fb44a68dcde65822fc805522b2d8fe027ab60b026d50bf4a2e90c3ed2529bef6L14-R14) [[3]](diffhunk://#diff-685cd0cf58c390122f7af132ad1e2d6e6d6491c3067f550ca8ab0bad18302406L1-R7) [[4]](diffhunk://#diff-492658f195c89579fa60e09f795bf78a8423a64a41994ec554d997cb96724d59L3-R5) [[5]](diffhunk://#diff-492658f195c89579fa60e09f795bf78a8423a64a41994ec554d997cb96724d59L30-R30) [[6]](diffhunk://#diff-492658f195c89579fa60e09f795bf78a8423a64a41994ec554d997cb96724d59L68-R68)

### React 19-Specific Adjustments

* Modified `ReactRenderer` to handle `ref` initialization for components that are not class-based or do not use `forwardRef`. This ensures compatibility with React 19's ref handling.

## Additional Notes

I'm not sure if we should push this as a major or minor release. In theory this should still work with React 18 and should not lead to issues with the existing implementations.

## Checklist

## Related Issues

- #5846
- closes #6110


demos/src/Nodes/Mention/React/MentionList.jsx
```diff
@@ -27,7 +27,7 @@ export default forwardRef((props, ref) => {
 
   useEffect(() => setSelectedIndex(0), [props.items])
 
-  useImperativeHandle(ref, () => ({
+  useImperativeHandle(props.ref, () => ({
```
arnaugomez, line 30: Will Tiptap 3 be compatible with react 18 or will it require React 19?

packages/react/src/ReactNodeViewRenderer.tsx
```diff
@@ -76,7 +78,8 @@ export class ReactNodeView<
       getPos: () => this.getPos(),
       updateAttributes: (attributes = {}) => this.updateAttributes(attributes),
       deleteNode: () => this.deleteNode(),
-    } satisfies NodeViewProps
+      ref: React.createRef<HTMLElement>(),
```
arnaugomez: Consider importing entities from the React module instead of importing the whole React namespace. `import { createRef } from 'react'` insead of `import React from 'react'`

demos/src/Nodes/Mention/React/MentionList.jsx
arnaugomez: I am worried that this change will make it incompatible with React 18, which does not support ref as props.

Also, React 19 still supports `forwardRef` (they plan to deprecate it in the next versions).

https://react.dev/blog/2024/12/05/react-19#ref-as-a-prop

arnaugomez (changes requested):
I am worried that some changes in this PR could make Tiptap 3 incompatible with React 18

demos/src/Nodes/Mention/React/MentionList.jsx
```diff
@@ -1,8 +1,8 @@
 import './MentionList.scss'
 
-import React, { forwardRef, useEffect, useImperativeHandle, useState } from 'react'
+import React, { useEffect, useImperativeHandle, useState } from 'react'
 
-export default forwardRef((props, ref) => {
+export default props => {
```
arnaugomez, line 5: I am worried that this change will make it incompatible with React 18, which does not support ref as props.
Will Tiptap 3 be compatible with react 18 or will it require React 19?

Also, React 19 still supports forwardRef (they plan to deprecate it in the next versions).

https://react.dev/blog/2024/12/05/react-19#ref-as-a-prop

bdbch: This is just the demo that I upgraded to React 19. The important code is in the react package code where I added a check for forwarded ref components

packages/react/src/ReactRenderer.tsx
```diff
@@ -120,11 +148,27 @@ export class ReactRenderer<R = unknown, P extends Record<string, any> = object>
...
+        }
+      } else if (isClassComp || isForwardRefComp) {
+        // React 18 and prior: only set ref for class components and forwardRef components
+        // @ts-ignore - Setting ref prop for React 18 class/forwardRef components
+        props.ref = (ref: R) => {
+          this.ref = ref
+        }
       }
+      // For function components in React 18, we can't use ref - the component won't receive it
+      // This is a limitation we have to accept for React 18 function components without forwardRef
```
arnaugomez, line 173: Should we throw an error or log a warning in this case? If the developer tries to send a ref but it's a React 18 function component.

arnaugomez (approved):
I liked this message by Copilot, you can review it: https://github.com/ueberdosis/tiptap/pull/6405/files#r2126227325

packages/react/src/ReactRenderer.tsx
bdbch: Not sure as this is only used as a function to check if we should move forward with the ref creation or not so throwing or logging could be unwanted here.

bdbch: Also I think the only thing that would actually be catched is the `parseInt` call that could theoretically break. In case we can't parse the version string we should safely assume it's not React 19.

packages/react/src/ReactRenderer.tsx
```diff
@@ -120,14 +148,32 @@ export class ReactRenderer<R = unknown, P extends Record<string, any> = object>
...
+        elementProps.ref = (ref: R) => {
+          this.ref = ref
+        }
       }
+      // For function components in React 18, we can't use ref - the component won't receive it
+      // This is a limitation we have to accept for React 18 function components without forwardRef
     }
 
-    this.reactElement = <Component {...props} />
+    this.reactElement = <Component {...elementProps} />
```
davidlukerice, line 176: Something still seems off here in React 18. I just bumped TipTap from `beta.5` to `beta.6` on a project, and my custom TipTap React nodes are logging the following ref warning regardless of whether I wrap my `TestComponent` render function in a `forwardRef` or not.

Warning:
```
Warning: Function components cannot be given refs. Attempts to access this ref will fail. Did you mean to use React.forwardRef()? Error Component Stack
    at ReactNodeView (<anonymous>)
    ...
```

`reactVersion` is logging as `18.3.1`

One of my components' render functions is
```
  addNodeView() {
    return ReactNodeViewRenderer(TestComponent)
  },
  ```

bdbch: Thanks, will take a look into that

cdeutsch: Yeah, I'm getting the same thing with upgrading from 2.12 to 2.14

I'm on React 18.

bdbch: @cdeutsch could you re-verify with the latest version?

cdeutsch: @bdbch I'm no longer seeing the warning. Thanks!

PR #6434, opened 2025-06-10
feat(static-renderer): mark `react` and `react-dom` as optional

hywax:
Currently, `static-renderer` always imports `react`, even though not all projects use it.

I removed `react` from the main index so it’s no longer always imported in `static-renderer`. I also made `react` an optional dependency.

Those who need the `react`-related functionality can use the specific import paths:

```
@tiptap/static-renderer/pm/react  
@tiptap/static-renderer/json/react
```

## Changes Overview

This commit marks `react` and `react-dom` as optional peer dependencies in the `static-renderer` package. This change allows the package to be installed without requiring these dependencies to be present, which is useful in environments where they are not needed.

## Checklist

### DIFF
--- a/demos/vite.config.ts
+++ b/demos/vite.config.ts
@@ -41,12 +41,12 @@ const getPackageDependencies = () => {
 
             paths.push({
               find: `@tiptap/${name}/${subPkgName}`,
-              replacement: resolve(`../${path}/${name}/src/${subPkgName}/index.ts`),
+              replacement: resolve(`../${path}/${name}/src/${subPkgName}`),
             })
           })
-          paths.push({ find: `@tiptap/${name}`, replacement: resolve(`../${path}/${name}/src/index.ts`) })
+          paths.push({ find: `@tiptap/${name}`, replacement: resolve(`../${path}/${name}/src`) })
         } else {
-          paths.push({ find: `@tiptap/${name}`, replacement: resolve(`../${path}/${name}/src/index.ts`) })
+          paths.push({ find: `@tiptap/${name}`, replacement: resolve(`../${path}/${name}/src`) })
         }
       })
   }


hywax:
@arnaugomez hey! Could you please look up PR? It's a real pain to work with.

arnaugomez:
Hi @hwax, sure, tomorrow I take a look and let you know my thoughts

hywax:
@arnaugomez updated the body of the PR to make it clear what I did.

arnaugomez (changes requested):
First of all thanks @hywax for spotting this issue and opening the PR.

Though I agree that the static renderer should be framework-agnostic, I am not favorable to forcing developers to import entities with nested import paths, like this:

```
import { renderToReactElement } from '@tiptap/static-renderer/pm/react'
```

I believe deeply nested import paths like `@tiptap/static-renderer/pm/react` have confusing DX and make it look like `renderToReactElement` is a low-level API that you are not meant to be using.

Instead, let me suggest: would it be possible to remove the `renderToReactElement` function from the `@tiptap/static-renderer` package altogether, and create a new `@tiptap/static-renderer-react` package that contains `renderToReactElement` ?

I think this would be more consistent with Tiptap's package structure, where, for example, you have `@tiptap/core` for framework-agnostic code, and `@tiptap/react` for framework-specific code.

@bdbch Let me get your thoughts on this, do you like the solution I suggested, or do you prefer @hywax 's ? Or do you have another idea?

alexvcasillas:
> First of all thanks @hywax for spotting this issue and opening the PR.
> 
> 
> 
> Though I agree that the static renderer should be framework-agnostic, I am not favorable to forcing developers to import entities with nested import paths, like this:
> 
> 
> 
> ```
> 
> import { renderToReactElement } from '@tiptap/static-renderer/pm/react'
> 
> ```
> 
> 
> 
> I believe deeply nested import paths like `@tiptap/static-renderer/pm/react` have confusing DX and make it look like `renderToReactElement` is a low-level API that you are not meant to be using.
> 
> 
> 
> Instead, let me suggest: would it be possible to remove the `renderToReactElement` function from the `@tiptap/static-renderer` package altogether, and create a new `@tiptap/static-renderer-react` package that contains `renderToReactElement` ?
> 
> 
> 
> I think this would be more consistent with Tiptap's package structure, where, for example, you have `@tiptap/core` for framework-agnostic code, and `@tiptap/react` for framework-specific code.
> 
> 
> 
> @alexvcasillas  Let me get your thoughts on this, do you like the solution I suggested, or do you prefer @hywax 's ? Or do you have another idea?

I agree with @arnaugomez here, I believe we should split the packages to be single-responsibility, so if you want the react one, you get the react one and if you don't want react you don't get react :)

arnaugomez:
@hywax We've discussed it with the team and we've agreed to move the `renderToReactElement` and other React-related exports from `@tiptap/static-renderer` to the `@tiptap/react` package.

This way, the react-related exports are in our framework-specific package and you don't need to install them unless you're using React.

If you make this change, we'll review it again and merge the PR. 

Thanks for your contribution. I hope we can merge it soon 😄

nperez0111:
As the guy who wrote the static renderer, I disagree with this decision @arnaugomez 

I explicitly designed the package to be based on package exports, so that the user does not have to worry about compatibility between packages. Reducing the number of packages was a core tenet for Tiptap V3, this is not an exception.

How I'd approach this is to:
1. Mark react & react-dom as optionalPeerDeps
2. Keep the top-level package export the same as it was
3. Allow anyone who wants to be _certain_ that they are not accidentally including React to their bundle, should use the sub-exports, but keep things simple for everyone else.

Modern bundlers & ESM would be able to tree-shake the React dependency away if they do not use the React static renderer. I explicitly designed it to work like this. and it should not be a concern for most users, they can safely use the top-level API and not include React into their bundle.

bdbch:
@hywax Could you rebase this to `develop` and reopen your PR? We're deprecating the next branch.

PR #5812, opened 2024-11-07
fix(link): add backwards compat by deprecating validate and using isAllowedUri instead

nperez0111:
## Changes Overview

As a follow up to this: https://github.com/ueberdosis/tiptap/pull/5808
Adds `isAllowedUri` which handles XSS protections,
deprecates `validate` in favor of the new `shouldAutoLink` which is a better name for what it does
## Checklist
- [ ] I have created a [changeset](https://github.com/changesets/changesets) for this PR if necessary.
- [ ] My changes do not break the library.
- [ ] I have added tests where applicable.
- [ ] I have followed the project guidelines.
- [ ] I have fixed any lint issues.

PR #6448, opened 2025-06-14
Improve Link Extension's `isAllowedUri()` Function

justinbeaty:
## Changes Overview

While working with TipTap, I noticed two strange behaviors with the Link extension:

1. `<a href="">empty link</a>` resulted in the non-linked text `empty link`
2. `<a href="relative/index.html">relative</a>` resulted in the non-linked text `relative`

## Implementation Approach

The first problem was with `parseHTML()`, which checks if the `href` attribute is empty. However, we don't have that check with `renderHTML()`, `setLink()`, or `toggleLink()`, so it would seem we allow empty `href` tags, we just weren't parsing them from the input HTML.

```diff
           // prevent XSS attacks
           if (
-            !href ||
             !this.options.isAllowedUri(href, {
               defaultValidate: url => !!isAllowedUri(url, this.options.protocols),
               protocols: this.options.protocols,
               defaultProtocol: this.options.defaultProtocol,
             })
           ) {
             return false
           }
```

---

The second problem was with the `isAllowedUri()` function's regex. Here's what the regex matches for a few examples:
```
"relative.html"         // match = ["relative.html"]
"relative/index.html"   // match = null

"./relative.html"       // match = ["."]
"./relative/index.html" // match = ["."]

"/absolute.html"        // match = ["/"]
"/absolute/index.html"  // match = ["/"]
```

According to [RFC 3986 Section 3.1.](https://datatracker.ietf.org/doc/html/rfc3986#section-3.1):
>   Scheme names consist of a sequence of characters beginning with a
>   letter and followed by any combination of letters, digits, plus
>   ("+"), period ("."), or hyphen ("-").  Although schemes are case-
>   insensitive, the canonical form is lowercase and documents that
>   specify schemes must do so with lowercase letters.  An implementation
>   should accept uppercase letters as equivalent to lowercase in scheme
>   names (e.g., allow "HTTP" as well as "http") for the sake of
>   robustness but should only produce lowercase scheme names for
>   consistency.

So I changed the regex to `/^[a-z]+[a-z0-9+.-]*:/i`, which extracts the scheme, if any, from the URL [(demo)](https://regex101.com/r/oIWzOJ/4). Then, I just check it against our `allowedProtocols` array. To me, this seems much more straightforward.

## Testing Done

I added a few more valid URLs into the test file. Without this fix, `relative/index.html` was failing, but the other two were okay. With this fix, all three pass.

## Verification Steps

Run the demos with `<a href="">empty link</a> <a href="relative/index.html">relative</a>`, and both should be successfully turned into links.

Before: 
image

After:
image

## Additional Notes

Could it be possible that some browsers parse schemes that aren't valid per RFC 3986? I also looked at the [WHATWG URL spec](https://url.spec.whatwg.org/#url-parsing), and it seems to require the same thing.

## Checklist

## Related Issues

#5160
#5470
#5812

### DIFF
--- a/packages/extension-link/src/link.ts
+++ b/packages/extension-link/src/link.ts
@@ -261,7 +261,6 @@ export const Link = Mark.create<LinkOptions>({
 
           // prevent XSS attacks
           if (
-            !href ||
             !this.options.isAllowedUri(href, {
               defaultValidate: url => !!isAllowedUri(url, this.options.protocols),
               protocols: this.options.protocols,


bdbch:
@justinbeaty could you rebase this to the current develop branch as we released V3, deprecating the next branch & would need to update most of our PR's anyway? Thanks a lot!

PR #6479, opened 2025-06-20
fix(drag-handle-react): prevent plugin re-registration

Aslam97:
- Use refs to store unstable props (`onNodeChange`, `tippyOptions`)
- Remove unstable dependencies from plugin registration useEffect

### DIFF
--- a/packages/extension-drag-handle-react/src/DragHandle.tsx
+++ b/packages/extension-drag-handle-react/src/DragHandle.tsx
@@ -1,22 +1,28 @@
+import type { DragHandlePluginProps } from '@tiptap/extension-drag-handle'
 import {
   DragHandlePlugin,
   dragHandlePluginDefaultKey,
-  DragHandlePluginProps,
 } from '@tiptap/extension-drag-handle'
-import { Node } from '@tiptap/pm/model'
-import { Plugin } from '@tiptap/pm/state'
-import { Editor } from '@tiptap/react'
-import React, {
-  ReactNode, useEffect, useRef, useState,
-} from 'react'
+import type { Node } from '@tiptap/pm/model'
+import type { Plugin } from '@tiptap/pm/state'
+import type { Editor } from '@tiptap/react'
+import type { ReactNode } from 'react'
+import React, { useEffect, useRef, useState } from 'react'
 
-type Optional<T, K extends keyof T> = Pick<Partial<T>, K> & Omit<T, K>;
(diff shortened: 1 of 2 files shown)


packages/extension-drag-handle-react/src/DragHandle.tsx
```diff
@@ -30,36 +36,41 @@ export const DragHandle = (props: DragHandleProps) => {
   const [element, setElement] = useState<HTMLDivElement | null>(null)
   const plugin = useRef<Plugin | null>(null)
 
+  const onNodeChangeRef = useRef(onNodeChange)
```
bdbch, line 47: I'm not sure about those changes.

I'd assume that the developer integrating the DragHandle has to properly memoize the props passed to our components.

The re-rendering problems occur because those options - when provided - will never deepEqual as they will technically always be new objects or functions.

What do you think @Aslam97 ?

Aslam97: You're right @bdbch. I tested it and removing the ref works perfectly when consumers properly memoize their callbacks with `useCallback`. The tradeoff is that users will get the _'Maximum update depth exceeded'_ error when they pass unmemoized functions.

This causing infinite re-renders. For example, devs might naturally want to use the simple approach:

```javascript
onNodeChange={(data: NodeChangeData) => {
  if (data.node) setNode(data.node)
  setNodePos(data.pos)
}}
```

Rather than the memoized version:

```javascript
const handleNodeChange = React.useCallback((data: NodeChangeData) => {
  if (data.node) setNode(data.node)
  setNodePos(data.pos)
}, [])

onNodeChange={handleNodeChange}
```

The inline approach feels simpler but breaks our component. Should we go with the cleaner implementation and expect proper memoization?

bdbch: I think we should go for the clean version and maybe use JSDoc's to document the prop so they can get in-editor documentation. We could add a warning in there to make sure they should memoize their objects before passing.

bdbch: @svenadlung what's your thought on this. Cleaner code but more overhead for users or more code to handle unmemoized code from users?

Aslam97: This PR moved to https://github.com/ueberdosis/tiptap/pull/6582 @bdbch @svenadlung

PR #6487, opened 2025-06-21
fix(NodePos): add validation for target position within document range

### LINKED ISSUE #6367: [Bug]: NodePos.children out of range
### Affected Packages

core

### Version(s)

3.0.0-beta.5 but I believe it affects v2 as well

### Bug Description

NodePosition.children throws an error if the position it was initialized with is at the end of a node (with a lot of content) with children. This only happens if the NodePosition.node has enough content so that when you add the child's offset, the sum is greater than the entire doc's nodeSize. 

Here's the exact line I think is incorrect: https://github.com/ueberdosis/tiptap/blob/next/packages/core/src/NodePos.ts#L138

I believe it should be:
```
const targetPos = (this.resolvedPos.inside + 1) + offset + (isNonTextAtom ? 0 : 1)
```

### Browser Used

Chrome

### Code Example URL

https://codesandbox.io/p/sandbox/adoring-hopper-8jvdxg

### Expected Behavior

Hovering over long text and using `nodePos.children` should not error

_No response_

### Dependency Updates

AdeboyeDN:
## Changes Overview
Added validation for target position within document range in the ⁠ NodePos ⁠ class to prevent out-of-range errors when accessing child nodes.

## Implementation Approach
The fix adds a simple boundary check in the children getter method of the NodePos class. Before resolving a position with this.resolvedPos.doc.resolve(targetPos)

## Testing Done
All tests passed locally.

## Additional Notes
issue Reproduction link according to #6367
https://codesandbox.io/p/sandbox/adoring-hopper-8jvdxg

Previous behaviour:

https://github.com/user-attachments/assets/365420af-94ca-402a-86a4-968245de3377

Current Behaviour after fix is applied:

https://github.com/user-attachments/assets/fc4918fa-8a1c-4448-b72b-055afd0835fa

## Checklist

## Related Issues
Fixes #6367


bdbch:
@AdeboyeDN could you run prettier and eslint in your project and fix the linting issues?
Afterwards I'll merge and make another release.

AdeboyeDN:
@bdbch  i think the checks passed now.
there were other files that were not linted too

PR #6580, opened 2025-07-08
feat: enhance suggestion plugin with decoration content and empty state styling

Aslam97:
## Changes Overview
Enhanced the suggestion plugin with improved decoration handling and better support for empty query states. Added new configuration options for decoration content and empty state styling, while refactoring the import structure and callback signatures.

## Implementation Approach
- **Import optimization**: Reorganized imports to use type imports where appropriate and removed unused dependencies
- **Interface enhancement**: Extended `SuggestionOptions` to include `decorationContent` and `decorationEmptyClass` properties
- **Decoration system refactor**: 
  - Added support for custom decoration content with fallback to empty string
  - Implemented dynamic CSS class handling based on query state (empty vs populated)
  - Modified decoration creation to use an array of class names instead of a single class


arnaugomez (approved):

.changeset/fast-countries-thank.md
```diff
@@ -0,0 +1,5 @@
+---
+"@tiptap/suggestion": major
```
bdbch: Is this a major change?

.changeset/fast-countries-thank.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+support empty query class and content in decorations
```
bdbch:
```suggestion
Added support for empty query messages via decorations
```

packages/suggestion/src/suggestion.ts
```diff
@@ -66,6 +68,20 @@ export interface SuggestionOptions<I = any, TSelected = any> {
    */
   decorationClass?: string
 
+  /**
+   * The content of the decoration node.
+   * @default null
+   * @example 'Type to search...'
+   */
+  decorationContent?: string
```
bdbch, line 77: So if there's no content, we don't create a decoration?
Maybe we could explain this in the JSDoc

bdbch (changes requested):
A few things here

PR #6581, opened 2025-07-08
chore(drag-handle): add keyboard support and optimize TypeScript imports

Aslam97:
## Changes Overview

This pull request introduces keyboard event handling to the drag handle plugin and optimizes TypeScript imports for better build performance and code organization. The changes improve user experience by allowing keyboard-based popup dismissal while maintaining existing drag functionality.

### DIFF
--- a/packages/extension-drag-handle/src/drag-handle-plugin.ts
+++ b/packages/extension-drag-handle/src/drag-handle-plugin.ts
@@ -281,6 +281,12 @@ export const DragHandlePlugin = ({
 
     props: {
       handleDOMEvents: {
+        keydown() {
+          if (popup && popup.state.isVisible) {
+            popup.hide()
+          }
+        },
+
         mouseleave(_view, e) {
           // Do not hide open popup on mouseleave.
           if (locked) {
--- a/packages/extension-drag-handle/src/drag-handle-plugin.ts
+++ b/packages/extension-drag-handle/src/drag-handle-plugin.ts
@@ -1,13 +1,16 @@
-import { Editor } from '@tiptap/core'
+import type { Editor } from '@tiptap/core'
 import { isChangeOrigin } from '@tiptap/extension-collaboration'
-import { Node } from '@tiptap/pm/model'
+import type { Node } from '@tiptap/pm/model'
+import type { EditorState, Transaction } from '@tiptap/pm/state'
+import { Plugin, PluginKey } from '@tiptap/pm/state'
+import type { EditorView } from '@tiptap/pm/view'
+import type { Instance, Props as TippyProps } from 'tippy.js'
+import tippy from 'tippy.js'
 import {
-  EditorState, Plugin, PluginKey,
-  Transaction,
-} from '@tiptap/pm/state'
-import { EditorView } from '@tiptap/pm/view'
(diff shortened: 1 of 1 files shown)


arnaugomez (approved):

.changeset/tall-eyes-wait.md
```diff
@@ -0,0 +1,5 @@
+---
+"@tiptap/extension-drag-handle": major
```
bdbch: Since you are pushing this to main right now - are you sure this is a major change?

.changeset/tall-eyes-wait.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+Dismissable popup on keydown
```
bdbch: Could this be a bit more descriptive? I think it would be helpful to know which popup this is about and which keys will dismiss the popup

packages/extension-drag-handle/src/drag-handle-plugin.ts
```diff
@@ -281,6 +304,15 @@ export const DragHandlePlugin = ({
 
     props: {
       handleDOMEvents: {
+        keydown() {
+          if (popup && popup.state.isVisible) {
```
bdbch: This will close the draghandle menu on any keypress right? What if the draghandle menu includes items that I may control via the keyboard arrowkeys? Or something like the notion like menu?

bdbch (changes requested):
I left a few comments.

On top of that, are you sure this needs to go to main? Since we want to go out with v3 we may not need this?

Aslam97:
> I left a few comments.
> 
> On top of that, are you sure this needs to go to main? Since we want to go out with v3 we may not need this?

@ThisDavidRichard Still want to keep the v2 version available. If users request it, Tiptap should provide it as a ZIP file or similar format.

packages/extension-drag-handle/src/drag-handle-plugin.ts
Aslam97: That wouldn't be the case since the focus has shifted from the editor. Also, I have added only if the is focused

.changeset/tall-eyes-wait.md
Aslam97: Change it from `major` to `minor`

.changeset/tall-eyes-wait.md
Aslam97: Updated the changeset to more descriptive "Dismissable popup on keydown, when the popup is visible and the editor has focus, typing should hide the popup." does that work for you?

bdbch (approved):

packages/extension-drag-handle/src/drag-handle-plugin.ts
```diff
@@ -281,6 +304,15 @@ export const DragHandlePlugin = ({
 
     props: {
       handleDOMEvents: {
+        keydown(view) {
+          if (popup && popup.state.isVisible && view.hasFocus()) {
+            popup.hide()
+            return true
```
bdbch, line 310: This is introducing a hefty problem with blocking other keyhandlers. Fix is already open - will be included in the next release

PR #6583, opened 2025-07-08
feat: enhance suggestion plugin with decoration content and empty state styling

Aslam97:
## Changes Overview
Enhanced the suggestion plugin with improved decoration handling and better support for empty query states. Added new configuration options for decoration content and empty state styling, while refactoring the import structure and callback signatures.

## Implementation Approach
- **Import optimization**: Reorganized imports to use type imports where appropriate and removed unused dependencies
- **Interface enhancement**: Extended `SuggestionOptions` to include `decorationContent` and `decorationEmptyClass` properties
- **Decoration system refactor**: 
  - Added support for custom decoration content with fallback to empty string
  - Implemented dynamic CSS class handling based on query state (empty vs populated)
  - Modified decoration creation to use an array of class names instead of a single class


packages/suggestion/src/suggestion.ts
```diff
@@ -68,6 +68,20 @@ export interface SuggestionOptions<I = any, TSelected = any> {
    */
   decorationClass?: string
 
+  /**
+   * The content of the decoration node.
+   * @default null
```
arnaugomez: The default is not null, it is an empty string

.changeset/clear-poets-occur.md
```diff
@@ -0,0 +1,5 @@
+---
+'@tiptap/suggestion': major
```
bdbch: Is this really a major change?

.changeset/clear-poets-occur.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+Support empty query class and content in decoration
```
bdbch:
```suggestion
Added support for empty query messages via decorations
```

packages/suggestion/src/suggestion.ts
```diff
@@ -68,6 +68,20 @@ export interface SuggestionOptions<I = any, TSelected = any> {
    */
   decorationClass?: string
 
+  /**
+   * The content of the decoration node.
+   * @default ''
+   * @example 'Type to search...'
+   */
+  decorationContent?: string
```
bdbch, line 76: See the other PR - maybe add to the JSDoc how to disable the decoration

bdbch (approved):

PR #6589, opened 2025-07-09
feat: add events for `mount` and `unmount`

nperez0111:
## Changes Overview
This adds both an `mount` and `unmount` event for tracking externally when the editor has been mounted or unmounted

## Testing Done
Trivial

## Checklist

### DIFF
--- a/.changeset/new-impalas-notice.md
+++ b/.changeset/new-impalas-notice.md
@@ -1,0 +1,5 @@
+---
+'@tiptap/core': minor
+---
+
+Add `mount` and `unmount` events to the `Editor` instance for tracking mounts and unmounts
--- a/packages/core/src/Editor.ts
+++ b/packages/core/src/Editor.ts
@@ -154,6 +154,7 @@ export class Editor extends EventEmitter<EditorEvents> {
       )
     }
     this.createView(el)
+    this.emit('mount', { editor: this })
 
     window.setTimeout(() => {
       if (this.isDestroyed) {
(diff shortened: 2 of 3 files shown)


arnaugomez (changes requested):
Looks good. Could you also add `onMount` and `onUnmount` options in the Editor constructor arguments, like this?

```ts
new Editor({onMount: () => console.log("mount")})
```

And to `useEditor` too.

Also, should `onUnmount` be fired after the editor unmounts, or before?

Make sure to clarify this in the documentation.

arnaugomez:
Could you add a few unit/integration tests? I believe it's valuable to add tests that check that the editor is mounted when `onMount` is called, and that the editor is unmounted when `onUnmount is called. This will help devs understand what these events do and prevent future changes that break their behaviour.

bdbch:
> And to useEditor too.

I think adding them to the Editor should be enough as the React/Vue editor's are just re-exports with added properties on top.

Cool idea, will take a look to merge it for the next minor - won't make it today.

nperez0111:
> Could you add a few unit/integration tests?

Eh, the other events don't have unit/integration tests for this, so I'm not really going to go out of my way for this

bdbch:
Agree - since we don't do it for other events we shouldn't enforce it for new events as well (otherwise we should also implement them for all existing events but that's something for another ticket 😉)

Taking another look.

bdbch:
@arnaugomez when you find time could you do another review? I don't think we need to enforce new tests when we're not consistent with them anyways.

bdbch:
@nperez0111 Could you rebase this to `develop` and reopen your PR? We're deprecating the next branch.

I will merge afterwards and bring it to main on the next minor release.

PR #6594, opened 2025-07-10
feat(link): add enableClickSelection option and Escape key support for link selection

Aslam97:
## Changes Overview

This PR introduces the `enableClickSelection` option to the `clickHandler` plugin, a feature that has been implemented in various projects. It provides more intuitive link interaction by allowing users to select the entire link on click. Additionally, it adds keyboard support for the `Escape` key to clear the link selection and move the cursor to the end of the link when `enableClickSelection` is enabled.

- Introduce `enableClickSelection` option to allow selecting the entire link on click.
- Add `Escape` key handler to clear link selection and move the cursor to the end when enabled.


.changeset/smooth-swans-exist.md
```diff
@@ -0,0 +1,5 @@
+---
+'@tiptap/extension-link': major
```
bdbch: Not a major change right? The default should be the same as before.

packages/extension-link/src/helpers/clickHandler.ts
```diff
@@ -42,6 +43,18 @@ export function clickHandler(options: ClickHandlerOptions): Plugin {
...
         const target = link?.target ?? attrs.target
 
+        if (options.enableClickSelection) {
+          const { doc, tr } = view.state
+          const range = getMarkRange(doc.resolve(pos), options.type)
+          if (range) {
+            const { from, to } = range
+            const $from = doc.resolve(from)
+            const $to = doc.resolve(to)
+            const transaction = tr.setSelection(new TextSelection($from, $to))
```
bdbch: I think you could use `extendMarkRange` instead to avoid handling the mark extension yourself

https://tiptap.dev/docs/editor/api/commands/nodes-and-marks/extend-mark-range

packages/extension-link/src/helpers/keydownHandler.ts
```diff
@@ -0,0 +1,24 @@
...
+    props: {
+      handleKeyDown: (view, event) => {
+        if (!view.editable) {
+          return false
+        }
+
+        const { selection } = view.state
+
+        if (event.key === 'Escape' && !selection.empty) {
+          view.dispatch(view.state.tr.setSelection(TextSelection.create(view.state.doc, selection.to)))
```
bdbch: Do we need to check if a link is selected here? A bit worried that this key handler will take precedence over other keyhandlers on escape

bdbch (changes requested):
small changes

bdbch (approved):

PR #6595, opened 2025-07-10
chore(drag-handle): add keyboard support dismissal popup

Aslam97:
## Changes Overview

This change adds keydown event handling logic to the drag handle plugin. Typing should hide the handle


.changeset/sweet-ravens-pay.md
```diff
@@ -0,0 +1,5 @@
+---
+'@tiptap/extension-drag-handle': major
```
bdbch: Is this a major change?

.changeset/sweet-ravens-pay.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+Typing should hide the drag handle
```
bdbch: I think the changelog message would be clearer as a statement of what the change does rather than what it should do—maybe rephrase it as "This changes..." instead?

```suggestion
Added logic to hide drag handle during typing to reduce visual clutter
```

packages/extension-drag-handle/src/drag-handle-plugin.ts
```diff
@@ -287,6 +287,18 @@ export const DragHandlePlugin = ({
 
       props: {
         handleDOMEvents: {
+          keydown() {
+            if (!element || locked) {
+              return false
+            }
+
+            hideHandle()
```
bdbch: This always hides the drag handle when any key is pressed - shouldn't this either be done when the content of the editor changes? Otherwise what happens to key handlers for enter, tab or arrow keys?

Aslam97: That wouldn't be the case since the focus has shifted from the editor. Also, I have added only if the is focused

packages/extension-drag-handle/src/drag-handle-plugin.ts
```diff
@@ -287,6 +287,22 @@ export const DragHandlePlugin = ({
...
         handleDOMEvents: {
+          keydown(view) {
+            if (!element || locked) {
+              return false
+            }
+
+            if (view.hasFocus()) {
+              hideHandle()
+              currentNode = null
+              currentNodePos = -1
```
arnaugomez, line 298: Why is currentNodePos `-1`?

If you find it appropriate, could you add a short comment that explains it?

Aslam97: Because it indicates that no node is selected, since valid positions start from 0

bdbch: I find this okay - usually `-1` is used in JS contexts to indicate no-index or not found (like with `Array.findIndex`)

packages/extension-drag-handle/src/drag-handle-plugin.ts
bdbch: In the worst case we can always revert this / change it

bdbch (approved):

PR #6613, opened 2025-07-11
fix problems with keydown events being blocked by drag handle plugin

bdbch:
## Changes Overview

This PR fixes a bug introduced by #6595 that blocks other key handlers from running

## Implementation Approach

I just always remove false from the keydown handler to allow other events to run through

## Checklist

### DIFF
--- a/packages/extension-drag-handle/src/drag-handle-plugin.ts
+++ b/packages/extension-drag-handle/src/drag-handle-plugin.ts
@@ -298,7 +298,7 @@ export const DragHandlePlugin = ({
               currentNodePos = -1
               onNodeChange?.({ editor, node: null, pos: -1 })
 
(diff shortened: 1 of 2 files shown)


packages/extension-drag-handle/src/drag-handle-plugin.ts
```diff
@@ -298,7 +298,7 @@ export const DragHandlePlugin = ({
               currentNodePos = -1
               onNodeChange?.({ editor, node: null, pos: -1 })
 
-              return true
+              return false
```
arnaugomez, line 302: Maybe add a comment on why return `false`, so that it is not changed at any other point in the future and the error reappears.

arnaugomez (approved):

PR #6679, opened 2025-07-19
Feature/resizable node views

### LINKED ISSUE #333: Resize images on editor
I would like to be able to resize images inside the editor. That way, content would be more flexible. I didn't find a way to do this through `tiptap`.

Adding a resize option like this (gif bellow) would help so much!

**Describe alternatives you've considered**
I really don't know how to implement this. Should this be on the Image plugin or a totally separated plugin? If you guys give some ideas I could make a PR.

bdbch:
## Changes Overview

This PR is still WIP as I'm not done with the full implementation - however would like to have a demo online already.

It implements a new helper that helps creating resizable node views (for example for images, videos or more advanced usecases).

## Implementation Approach

We create node views that auto-create all necessary resize handlers, event listeners given a configuration. This can be then used to render these resizable nodes for images, videos or other usecases.

## Testing Done
N/A - will do later

## Verification Steps
N/A - will do later

## Additional Notes
N/A - will do later

## Checklist

## Related Issues
Fixes #333 

#5336


marufmax:
waiting for this to be merged

bdbch:
Hey everyone, just a quick update

I’m planning to get back to this next now that Markdown support has been implemented. The PR is almost done - I’m just checking and fixing the last few bugs and looking into the final API design to make sure everything feels consistent.

Thanks for your patience, almost there!

arnaugomez:
From the PR, I see you can add a resizable behaviour to an element with the `ResizableNodeview` utility.

Would this utility work for making a React/Vue node view resizable? Or does it only work for Vanilla JS node views?

demos/src/Nodes/Image/React/styles.scss
```diff
@@ -6,12 +6,5 @@
 
   img {
     display: block;
-    height: auto;
```
svenadlung, line 9: What was the reason to remove the old styling?

demos/src/Nodes/Image/Vue/index.vue
```diff
@@ -64,13 +64,62 @@ export default {
 
   img {
     display: block;
-    height: auto;
+  }
```
svenadlung, line 74: I think these are some leftovers that can be reverted?

packages/extension-image/src/image.ts
```diff
@@ -22,6 +23,20 @@ export interface ImageOptions {
    * @example { class: 'foo' }
    */
   HTMLAttributes: Record<string, any>
+
+  /**
+   * Controls if the image should be resizable and how the resize is configured.
+   * @default false
+   * @example { directions: { top: true, right: true, bottom: true, left: true, topLeft: true, topRight: true, bottomLeft: true, bottomRight: true }, minWidth: 100, minHeight: 100 }
+   */
+  resize:
```
svenadlung, line 32: For tables it's called `resizable`, but it's just a boolean value. I'm not sure whether we should go with `resizeOptions` here instead?

svenadlung (changes requested):

packages/extension-image/src/image.ts
svenadlung: Ah, it's used for options and enabling resizing at the same time, right? Hm.

bdbch:
> Would this utility work for making a React or Vue node view resizable, or is it limited to vanilla JS node views?

It works with React views as well. You just need to do a bit of manual setup to connect the DOM element from your React node view with the regular node view.

What currently does not work (and this is unrelated to React or Vue) is using `contentDOM` inside resizable nodes. Editable content within a node view causes it to re-render on every state update, which breaks the elements attached to the resizable node view.

I'm still looking into a proper solution, but for an initial version this limitation should be fine.

packages/extension-image/src/image.ts
bdbch: I changed it to make it more obvious by requiring an `enabled` option inside the `resize` options. I think `resizeOptions` is a bit verbose since we are already in the options of the extension. (`options.resize`).

bdbch: So now you'd need:

```
resize: {
  enabled: true
}
```

for it to work. It doesn't really change anything but just makes it more obvious to users that they're enabling something.

svenadlung (approved):

Issue #6440, opened 2025-06-12
[Bug]: Performance issues when calling handleSelectionUpdate in VueNodeViewRenderer


AndreyChelnokov:
### Affected Packages

vue-3

### Version(s)

2.11.3

### Bug Description

The performance of the editor significantly degrades when creating a large number of nodes (e.g., 30,000 paragraphs) with VueNodeViewRenderer, compared to creating the same number of nodes without it.

<img width="976" alt="Image" src="https://github.com/user-attachments/assets/4dd71f34-7eb2-4cd7-b080-2a31293adc65" />

### Browser Used

Chrome

### Expected Behavior

The performance of nodes does not depend on the creation method

### Dependency Updates

- [x] Yes, I've updated all my dependencies.

AndreyChelnokov:
@ColemanDunn Is there anything that can be done about this?

PR #6682, opened 2025-07-19
Update VueNodeViewRenderer

AndreyChelnokov:
## Changes Overview

Removed handleSelectionUpdate as it did nothing useful but constantly calculated the position

## Implementation Approach

-

## Testing Done

Yes, I did various tests to make sure nothing would break.

For testing, you can run the repository's development mode and ensure that the ProseMirror-selectednode class and the selected property are set to bypass the remote method.

## Verification Steps

-

## Additional Notes

The selectNode and deselectNode methods are called independently of the handleSelectionUpdate method. Furthermore, previously the handleSelectionUpdate method performed position calculations every time, but never called selectNode and deselectNode because the conditions for calling them were not met.

## Checklist

## Related Issues

https://github.com/ueberdosis/tiptap/issues/6440


bdbch (changes requested):
We would need a changeset for this to make sure it's published correctly.

Also can you verify that the selection state of a Vue node view is still correctly updated when the selection changes? On top - is this also valid for React? I'm not sure right now if we had the same logic for the vue-2 and react implementations.

bdbch:
Closing this for inactivity

PR #6691, opened 2025-07-20
fix: avoid flushSync call during <EditorContent /> lifecycle v3

romarybi:
Original v2 PR -  #6318 

## Changes Overview
Fixed the bug where rendering the EditorContent dynamically would throw a flushSync error.

## Implementation Approach
Changed ReactRenderer flushSync condition to use a new boolean isEditorContentInitialized instead of isInitialized. This change makes sure that flushSync is not called in the EditorContent component lifecycle. 

## Testing Done
All tests passed locally.

## Verification Steps
[Sandbox](https://codesandbox.io/p/sandbox/tqc8hx) that shows the error that is now fixed. Create the tiptap editor that is initialized with content that includes a node that uses a react component. Create a conditional boolean to render EditorContent. Toggle the boolean and you should see the error in the console.

## Additional Notes
Previously, isInitialized was used and should probably be removed now. The reason I kept it in is it might be a breaking change for users who rely on it in their codebase.

## Checklist
- [] I have added tests where applicable.

## Related Issues
[Issue](https://github.com/ueberdosis/tiptap/issues/6315)


WilliamIPark:
Any progress on getting this fix merged?

sylvainar:
Hey, we found a bug on our side and this patches fixes it. Is going to be merged? Thanks

jelling:
+1 for merge please

Temporary workaround:
- `git clone https://github.com/romarybi/tiptap.git`
- `npm i && npm run build` 
- copy the `packages/react/dist` folder into your project somewhere
- in your `package.json`:  "@tiptap/react": "file:./lib/tiptap-react-patched",

nperez0111:
If this were to pass the lint check, then a pkg.pr.new version would be made. Can you fix the lint @romarybi

jelling:
> [warn] src/Editor.ts
> [warn] Code style issues found in the above file. Run Prettier with --write to fix.

@nperez0111 I get the best practice of having the PR requester fix the PR. But it's been two months and I'm not certain if they're going to come back to do that. Perhaps an exception could be made and someone from TipTap could do run the auto-fix as it looks like a 1m thing to do.

moekify:
Any progress here, it would be great to get this merged as it's causing bugs for our users @nperez0111 @romarybi

nperez0111:
I don't maintain Tiptap anymore, so I can't contribute anything to this. Someone else can re-open the PR based off the diff, fix the lint and people would at least be able to run the pkg.pr.new version. I was just recommending it

ecjs:
I can confirm that this fix works, I fix it in my own project via a pnpm patch. We did experience this bug in our own application.

bdbch (changes requested):
Could you add a regression test for this before we merge it? The fix looks good and it seems to solve the flushSync warning, but this area has caused node-view timing/desync issues for us before. A test around conditional EditorContent mounting with a React node view would make this a lot safer to merge.

Issue #6716, opened 2025-07-24
BubbleMenu jumps around when text element size changes


50bbx:
### Affected Packages

core, react

### Version(s)

3.2.0

### Bug Description

Using the [demo](https://tiptap.dev/docs/editor/extensions/functionality/bubble-menu) in the BubbleMenu extension, I get this weird behavior where the BubbleMenu changes position when applying bold. This happens because the text overflows to a new line, changing the size of the text element the `BubbleMenu` is attached to.

https://github.com/user-attachments/assets/7eac1f2c-a68d-4f77-8037-6eb77493e8c1

Is there any way to change this behavior and make sure that the `BubbleMenu` stays in place and does not change its placement due to size changes in the element it is attached to?

### Browser Used

Chrome

### Code Example URL

https://tiptap.dev/docs/editor/extensions/functionality/bubble-menu

### Expected Behavior

The BubbleMenu stays still when the element size changes.

### Dependency Updates

- [x] Yes, I've updated all my dependencies.

blessingnjoku:
Hi @50bbx  , I’m interested in working on this. Can I take it on?

50bbx:
@blessingnjoku I am not part of the maintainers here, so I am not the right person to ask that question. If I were you, I don't see anything wrong in trying to address it!

PR #6726, opened 2025-07-26
Fix: BubbleMenu jumps when text element size changes

blessingnjoku:
### What Was Changed
- Applied `position: fixed` to the BubbleMenu container.
- Used `left: 50%` `top:-50 to keep it centered horizontally `.

### Result
The BubbleMenu now remains visually steady and centered regardless of text wrapping or content changes.

https://github.com/ueberdosis/tiptap/issues/6716

### DIFF
--- a/demos/src/Extensions/BubbleMenu/React/styles.scss
+++ b/demos/src/Extensions/BubbleMenu/React/styles.scss
@@ -51,7 +51,7 @@
   h6 {
     font-size: 1rem;
   }
-
+  
   /* Code and preformatted text styles */
   code {
     background-color: var(--purple-light);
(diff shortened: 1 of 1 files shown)


50bbx:
I've tested the demo and it bumps around, in a different way, now. Just vertically in this case. The way to go with this is probably requires a lot more work. Keeping the BubbleMenu perfectly still if the selected text wraps would require a `position: fixed` plus static `top` and `left` values. But if you take scrolling into account, you would need to set the element to `position: absolute`.

https://deploy-preview-6726--tiptap-embed.netlify.app/preview/Extensions/BubbleMenu


https://github.com/user-attachments/assets/3fe4572f-f6b0-4308-b507-becc70d27821

Issue #6727, opened 2025-07-26
[v3] Mathematics extension would crash TipTap in older WebKit environments


harrisonlo:
### Affected Packages

extension-mathematics

### Version(s)

3.0.7

### Bug Description

Older WebKit environment (WKWebView) throws `Invalid regular expression: invalid group specifier name` error due to the new regex for input rules introduced in https://github.com/ueberdosis/tiptap/pull/6508

I'm not sure if input rules are still needed anymore for mathematics in v3? They don't seem to apply in the [online demo](https://tiptap.dev/docs/editor/extensions/nodes/mathematics). If they're not needed, I suppose we can just remove those input rules (for both `BlockMath` and `InlineMath` extensions) to solve this problem.

### Browser Used

Safari

### Expected Behavior

Let's support TipTap in older environments too.

### Dependency Updates

- [x] Yes, I've updated all my dependencies.

bdbch:
We still need the input rules as you can still input math nodes via `$$` and `$$$` input rules.

I guess it's because of the lookbehind / lookahead regex syntax which is not supported in older webkit envs? How old are those environments?

harrisonlo:
@bdbch I see. Did the input rules work for you in the online demo? Maybe I wasn't doing it correctly? Update: oh ok it's double $ for inline now. I see that it works.

In that case, can we change the regex syntax to support older environments? I have an app that uses Tauri which uses the local webview on my old mac (WebKit version 17618.3.11.11.7, 17618).

harrisonlo:
@bdbch I made a PR (https://github.com/ueberdosis/tiptap/pull/6731) to update the regex for the inline math extension and the migration util. I believe the new migration util regex also fixes some matching issue. Please take a look. Thanks!

giaBaoJS:
Confirmed on current `main`: the input rule `find` at `packages/extension-mathematics/src/extensions/InlineMath.ts:223` is a regex literal containing a lookbehind, so WebKit < Safari 16.4 fails to parse the whole chunk rather than just that rule — it is the only lookbehind left in the package, and it still reaches the built `dist`. I have a fix ready that replaces the regex with an `InputRuleFinder` function, which avoids the preceding-character-swallowing regression that got #6731 reverted by #7481; could this be assigned to me?

PR #5916, opened 2024-12-05
fix(extension-code): support safari versions 16.4 and older

lino-levan:
## Changes Overview
Rewrite the extension-code regex to avoid lookbehind to fix errors in safari versions 16.4 and older.

## Testing Done
I tested the regex in regex101 to ensure it continues to work as expected.

## Verification Steps
I checked to ensure that this actually works in older safari versions.

## Checklist
- [x] I have created a [changeset](https://github.com/changesets/changesets) for this PR if necessary.
- [x] My changes do not break the library.
- [x] I have added tests where applicable.
- [x] I have followed the project guidelines.
- [x] I have fixed any lint issues.

## Related Issues
https://github.com/ueberdosis/tiptap/issues/5889

PR #6731, opened 2025-07-27
fix(extension-mathematics): compatibility issues and migration util regex

harrisonlo:
## Changes Overview

I have updated the regex for the `InlineMath` input rule and the migration util function to replace the negative lookbehind syntax, which is not supported in all browser environments. It would otherwise crash tiptap in those older browsers and webviews. i.e. webkit 17618.3.11.11.7

## Implementation Approach

In addition to improving compatibility, I believe this PR also fixes some regex match issue for the migration util, which is not supposed to match [non-math expressions](https://github.com/ueberdosis/tiptap/blob/072b110a93862ccf95219fc8ca45f6961bddd904/packages/extension-mathematics/src/utils.ts#L4-L15). i.e. `$100$` (see below test links)

### For inline math
The previous input rule regex: https://regexr.com/8g704
https://github.com/ueberdosis/tiptap/blob/072b110a93862ccf95219fc8ca45f6961bddd904/packages/extension-mathematics/src/extensions/InlineMath.ts#L187

The **proposed** input rule regex: https://regexr.com/8gbjd
```
/(^|[^$])(\$\$([^$\n]+?)\$\$)(?!\$)/
```` 

### For migration util

The previous regex: https://regexr.com/8g704
https://github.com/ueberdosis/tiptap/blob/072b110a93862ccf95219fc8ca45f6961bddd904/packages/extension-mathematics/src/utils.ts#L16

The **proposed** regex: https://regexr.com/8g70a
```
/\$(?!\d+\$)(.+?)\$(?!\d)/g
```

## Checklist

## Related Issues

https://github.com/ueberdosis/tiptap/issues/6727
https://github.com/ueberdosis/tiptap/pull/5916


bdbch:
For me the example doesnt work anymore.

https://deploy-preview-6731--tiptap-embed.netlify.app/src/extensions/mathematics/react/

When I try to use the regex now, the content between the regex match will not be added to the new node.

harrisonlo:
@bdbch I've updated the regex and the capturing group reference. Without negative lookbehind, it has to check the preceding character in order to prevent `$$$...$$$` from matching in `InlineMath`, so the first capturing group would be that character. I've updated the capturing group reference in the input rule accordingly. https://deploy-preview-6731--tiptap-embed.netlify.app/src/extensions/mathematics/react/ seems to work fine now.

PR #6734, opened 2025-07-28
feat: BubbleMenu appendTo option

mweidner037:
## Changes Overview

Add an `appendTo` option to extension-bubble-menu, modeled on [tippy's appendTo option](https://atomiks.github.io/tippyjs/v6/all-props/#appendto).

## Implementation Approach

Add the option following the example of other options, and use it in the one relevant line (show's appendChild call).

## Testing Done

Copied the built code into our app and verified that menu was appended to the given element, similar to the 2.x behavior with tippy's appendTo option. In our app, this fixes some z-index issues by portaling the menu into the page's root stacking context.

## Checklist

### DIFF
--- a/.changeset/polite-beans-juggle.md
+++ b/.changeset/polite-beans-juggle.md
@@ -1,0 +1,5 @@
+---
+'@tiptap/extension-bubble-menu': patch
+---
+
+Add appendTo option
--- a/packages/extension-bubble-menu/src/bubble-menu-plugin.ts
+++ b/packages/extension-bubble-menu/src/bubble-menu-plugin.ts
@@ -82,6 +82,16 @@ export interface BubbleMenuPluginProps {
     | null
 
   /**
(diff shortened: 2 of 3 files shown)


felixgabler:
This would be great to have! I also wanted to add that the vue-3 [`<BubbleMenu>`](https://github.com/ueberdosis/tiptap/blob/develop/packages/vue-3/src/menus/BubbleMenu.ts#L76) (and `FloatingMenu`) component uses Teleport to achieve this, but that does not seem to work, it still renders it right into the TipTap elements. If this is merged, we could probably also remove the Teleport usage there.

felixgabler:
@bdbch thanks for merging this! Will you also make the relevant changes to the `vue-3` package (+ probably the others, as well as `FloatingMenu`? I guess most people use the `BubbleMenu` via the framework packages.

PR #6765, opened 2025-08-04
fix(unique-id): look for provider in collaborationCaret extension

Aslam97:
## Changes Overview

Fixed UniqueID extension creating unwanted paragraph nodes on page refresh in collaborative editing setups by correcting the provider lookup location.

## Implementation Approach

The issue was that the UniqueID extension's onCreate method was looking for the collaboration provider in the collaboration extension, but it's actually located in the collaborationCaret extension. Changed the provider lookup from:
```typescript
// Before - looking in wrong extension
const collab = this.editor.extensionManager.extensions.find(ext => ext.name === 'collaboration')
const provider = collab?.options?.provider
```

To:

```typescript
// After - looking in correct extension  
const collaborationCaret = this.editor.extensionManager.extensions.find(ext => ext.name === 'collaborationCaret')
const provider = collaborationCaret?.options?.provider
```

When the provider wasn't found, the extension would immediately call createIds() instead of waiting for the 'synced' event, causing it to process an unsynced document state and create unwanted nodes.

## Checklist


bdbch (changes requested):
Shouldn't we check for both since those unwanted paragraphs popped up in all kind of collaborative setups? I'm pretty sure the old fix looking for the collaboration extension was also valid.

Aslam97:
@bdbch Updated the PR. now it support should support both `collaboration` and `collaborationCaret`

PR #6776, opened 2025-08-05
Fix react bubble menu append to feature

phwebi:
## Changes Overview

Pass the `appendTo` prop from the react `BubbleMenu` to the bubble menu plugin instance, so that the `appendTo` prop will work.

## Testing Done

Tested in an application using tiptap that the updated `BubbleMenu` component respects the `appendTo` prop.

## Verification Steps

Confirm that the `appendTo` prop in the react `BubbleMenu` does nothing before this fix. After applying the fix, the bubble menu will be appended to the specified element.

## Checklist


bdbch (changes requested):
Good idea - however I would:

1. Add the same exposed prop to the Floating Menu if necessary
2. Add the same exposed props to the Vue menus
3. Update the changeset file

rahalm22:
It is also worth noting that since `appendTo` only accepts `HTMLElement` which can be limiting if the target container isn't yet in the DOM when the BubbleMenu mounts. My suggestion is to make `appendTo` to accept a function `() => HTMLElement` so that it will be possible to attach the BubbleMenu to containers that are mounted asynchronously.

Issue #6777, opened 2025-08-06
Can't unmount vue3 editor in test without causing an error


eriknygren:
### Affected Packages

core, vue

### Version(s)

3.0.9

### Bug Description

So in my vitest run of my tiptap editor, either with jsdom or happy-dom I get this when trying to unmount my tiptap editor component

```
TypeError: _a.remove is not a function
 ❯ Editor.unmount ../../node_modules/@tiptap/core/src/Editor.ts:185:15
 ❯ Editor.destroy ../../node_modules/@tiptap/core/src/Editor.ts:713:10
 ❯ ../../node_modules/@tiptap/vue-3/src/useEditor.ts:20:19
 ❯ ../../node_modules/@vue/runtime-core/dist/runtime-core.cjs.js:2818:40
 ❯ callWithErrorHandling ../../node_modules/@vue/runtime-core/dist/runtime-core.cjs.js:200:19
 ❯ callWithAsyncErrorHandling ../../node_modules/@vue/runtime-core/dist/runtime-core.cjs.js:207:17
 ❯ Array.hook.__weh.hook.__weh ../../node_modules/@vue/runtime-core/dist/runtime-core.cjs.js:2798:19
 ❯ Object.invokeArrayFns ../../node_modules/@vue/shared/dist/shared.cjs.js:90:11
     88| const invokeArrayFns = (fns, ...arg) => {
     89|   for (let i = 0; i < fns.length; i++) {
     90|     fns[i](...arg);
       |           ^
     91|   }
     92| };
 ❯ unmountComponent ../../node_modules/@vue/runtime-core/dist/runtime-core.cjs.js:5929:14
 ❯ unmount ../../node_modules/@vue/runtime-core/dist/runtime-core.cjs.js:5819:7
```

I haven't been able to isolate it in a repo as I can't recreate it without my extensions. However I don't think they are doing anything out of the ordinary.

But it seems to be pointing quite specificall to [this line](https://github.com/ueberdosis/tiptap/blob/224cb2c77a7921e62b18f4b510bb314b582926df/packages/core/src/Editor.ts#L185)

Which seems [to point to this specific PR](https://github.com/ueberdosis/tiptap/pull/6182)

Previously I was on version `2.9.1` where this was not an issue, which I believe was before this change.

I know this is not lots of info to go on, but maybe it's obvious to those who made this change? so CCing @nperez0111 and @bdbch for that reason.

Thanks in advance, any advice is highly appreciated 🙏 

[...]

PR #6779, opened 2025-08-06
fix(core): handle missing DOM remove method in test environments (#6777)

praneeth622:
## Changes Overview

Fixed the `Editor.unmount()` method to handle missing `remove()` method on DOM elements in test environments (jsdom/happy-dom). The issue was causing `TypeError: _a.remove is not a function` when unmounting Vue 3 editors in Vitest tests.

## Implementation Approach

Replaced the simple `this.css?.remove()` call with a robust fallback mechanism:

1. **Primary method**: Use `element.remove()` if available (modern browsers)
2. **Fallback method**: Use `parentNode.removeChild()` for environments missing `remove()` (jsdom/happy-dom)
3. **Error handling**: Wrap in try-catch to handle any unexpected DOM errors gracefully
4. **Graceful degradation**: Skip removal if neither method is available

The solution maintains backward compatibility while providing resilience against various test environment limitations.

## Testing Done

- ✅ Tested with mock DOM environments simulating jsdom/happy-dom (no `remove()` method)
- ✅ Tested with standard browser environments (has `remove()` method)
- ✅ Tested edge cases: orphaned elements, null elements, throwing DOM methods
- ✅ Verified Vue 3 `useEditor` hook scenario works correctly
- ✅ Performance tested: solution is 43% faster than original due to optimized checks
- ✅ All existing functionality preserved

## Verification Steps

1. **Reproduce original issue**:
   ```javascript
   // In a Vitest test with jsdom/happy-dom
   const editor = new Editor({ content: 'Hello world', extensions: [] })
   editor.css = document.createElement('style')
   delete editor.css.remove // Simulate missing method
   editor.unmount() // Should not throw

1. Test Vue 3 integration:
```js
// In Vue 3 component test
const { unmount } = mount(EditorComponent)
unmount() // Should not throw TypeError
```
2. Verify normal browser behavior:
```js 
const editor = new Editor({ content: 'Hello world', extensions: [] })
editor.unmount() // Should work as before
```

## Additional Notes

- Root cause: Test environments (jsdom/happy-dom) don't fully implement all DOM APIs, specifically missing Element.prototype.remove()
- Backward compatibility: Zero breaking changes - existing code continues to work
- Error resilience: Added console.warn for debugging unexpected DOM errors
- Performance: Improved performance due to optimized conditional checks
- Best practices: Follows defensive programming principles for cross-environment compatibility

## Checklist

- [x] I have created a [changeset](https://github.com/changesets/changesets) for this PR if necessary.
[...]

PR #6781, opened 2025-08-07
Fix unmounting vue-3 editor in test environments

eriknygren:
## Changes Overview

This adds a safe accessor to ensure that the `css` variable has a `remove` method before trying to invoke it

This is because of this error when unmounting the tiptap editor in vue-test-utils + vitest

```
TypeError: _a.remove is not a function
 ❯ Editor.unmount ../../node_modules/@tiptap/core/src/Editor.ts:185:15
```

I'm not sure of the exact cause, is `css` being set to an object by something that's not the right
type.. or is it lacking the right polyfills in vitest with `happy-dom` and `jsdom`. All I know is I'm
getting the above error.

So it's being set to something, that does not have the remove method. So updating this line to make
sure it actually exists before calling it.

## Implementation Approach

n/a, small safety accessor

## Testing Done

None, change only adds another accessor safety check

## Verification Steps

tests passing

## Checklist

## Related Issues

https://github.com/ueberdosis/tiptap/issues/6777

### DIFF
--- a/packages/core/src/Editor.ts
+++ b/packages/core/src/Editor.ts
@@ -182,7 +182,9 @@ export class Editor extends EventEmitter<EditorEvents> {
     }
     this.editorView = null
     this.isInitialized = false
-    this.css?.remove()
+    if (this.css && typeof this.css.remove === 'function') {
+      this.css.remove()
+    }
     this.css = null
   }
 


bdbch:
Seems to be a duplicate of https://github.com/ueberdosis/tiptap/pull/6779

PR #6783, opened 2025-08-07
fix(react): pass appendTo to BubbleMenuPlugin

adam187:
## Changes Overview

Pass appendTo props to BubbleMenuPlugin so it would be possible to render menu in different container.

## Checklist


bdbch (changes requested):
Good idea - however I would:

1. Add the same exposed prop to the Floating Menu if necessary
2. Add the same exposed props to the Vue menus
3. Add a changeset (patch) that describes this change via `npx changeset`

bdbch:
I'll close this in favor for https://github.com/ueberdosis/tiptap/pull/6776 as this seems to be a duplicate.

PR #6821, opened 2025-08-15
WIP: Markdown Support

bdbch:


dixalex:
Really waiting for this to switch from https://github.com/aguingand/tiptap-markdown :) Thank you!

Doriiaan:
> Really waiting for this to switch from https://github.com/aguingand/tiptap-markdown :) Thank you!

Same ahah :)

bdbch:
I'm currently finalizing the markdown renderer first - after that I'll look more into parsing from MD to Tiptap JSON.

```ts
const Blockquote = Node.create({
  // your typical setup

  markdown: {
    render: (node, h) => {
      if (!node.content) {
        return ''
      }

      const lines: string[] = []

      node.content.forEach(child => {
        const lineContent = h.renderChildren(child)
        const withPrefix = lineContent
          .split('\n')
          .map(line => `> ${line}`)
          .join('\n')
        lines.push(withPrefix)
      })

      const linesWithSeparators = lines.flatMap(line => [line, '> '])

      return linesWithSeparators.slice(0, -1).join('\n')
    },
  },
})
```

This is what the current markdown rendering function for the blockquote node looks like - so it's pretty flexible in defining how to render things with Tiptap extensions.

In general this render function will have access to:

- **Node**: The node currently being rendered as Tiptap JSON (including content and attributes)
- **Helpers**: A set of helper functions
  - **renderChildren**: The render function to render child content of the current node
  - **indent**: Used to indent content based off it's level, primarily used for lists or other indented items
  - **wrapInBlock**: A helper function that can be used to wrap a set of lines in a block
- **Context**: The context of the current rendering step
  - **Level**: The current nesting level (useful for indentation)
  - **Index**: The index of the current item based off the parent item
  - **parentType**: The parent node type (for example used in lists to determine if a list item is inside a bullet list or numbered list)
  - **meta**: A free-form meta object where custom data can be passed into the renderer

The render function will expect a string output.

Would welcome feedback.

nperez0111:
The only thing I'd question here is whether the markdown serialization is somewhat overlapping with the text serialization, like shouldn't the text serialization be re-used for markdown instead of a new separate property?

The static renderer should also be updated to pull from these new fields if you decide to keep them. I somewhat purposefully tried to make the markdown serialization outside of the core module so that only the people who need them have to pay the cost & weight of them. For serialization, it is relatively trivial, but for parsing, if you are parsing from a string or HTML or JSON makes a huge difference in bundle size, that I didn't want to ship in the core since not everyone will use it.

bdbch:
> The only thing I'd question here is whether the markdown serialization is somewhat overlapping with the text serialization, like shouldn't the text serialization be re-used for markdown instead of a new separate property?

True, I could look into reusing the existing text serialization if possible.

> The static renderer should also be updated to pull from these new fields if you decide to keep them.

Absolutely, makes sense to have one source of truth when it comes to serializing out markdown content here.

> I somewhat purposefully tried to make the markdown serialization outside of the core module so that only the people who need them have to pay the cost & weight of them. For serialization, it is relatively trivial, but for parsing, if you are parsing from a string or HTML or JSON makes a huge difference in bundle size, that I didn't want to ship in the core since not everyone will use it.

Also understand that point and thought about it how to not bloat the bundle. My problem is that this should live directly on the editor (as in `editor.getMarkdown()` or `editor.commands.setContent('# Hello world', { type: 'markdown' })` without the need to pass through an external parsing/serializer.

Not 100% fixed on my decisions here, I think the MarkdownManager could be extracted into an external package and used here. I'll get back to your thoughts after I'm done registering custom lexer definitions for custom nodes.

bdbch:
Also one thing that makes this a bit more complicated is that markdown configuration should/must happen in the extension similar to how we define how to parse and serialize HTML. I could make abstract typings for Marked.js types when it comes to parsing but it would be nicer if it was a one-fit solution

nperez0111:
> Also understand that point and thought about it how to not bloat the bundle. My problem is that this should live directly on the editor (as in editor.getMarkdown() or editor.commands.setContent('# Hello world', { type: 'markdown' }) without the need to pass through an external parsing/serializer.

Yea, I'm with you on `editor.getMarkdown()`, but `editor.commands.setContent(markdown)` harder since it requires a markdown parser involved. I don't think it is too much to ask for `editor.commands.setMarkdown(markdown)` & have a dedicated markdown extension, since that would slim the bundle significantly for those who don't need it.

> Also one thing that makes this a bit more complicated is that markdown configuration should/must happen in the extension similar to how we define how to parse and serialize HTML. I could make abstract typings for Marked.js types when it comes to parsing but it would be nicer if it was a one-fit solution

Yea, I agree this is difficult to design for. marked is a good choice of lib though, seems like it is "only" 40kb. But Tiptap/core is already a bit heavy when you consider all of it's deps to be 272kb, so I just wanna be considerate of it.

image

image

bdbch:
Absolutely worth to discuss. One idea I just had was to extend Tiptap's `Extendable` typings from a `@tiptap/markdown` package to allow for all the markdown configuration stuff if needed (or just ship is by default and don't do anything with it). I wonder if it would be possible to extend default commands like `setContent` or `insertContent` from an external package to allow for markdown insertions as well. Same with the `content` property on the editor instance.

Would love to have it extendable but still feel deeply integrated if actually installed and imported.

In general I'd be a fan if all of this logic could be used on the server without even initializing an Editor first but just passing through a set of Extensions.

bdbch:
I'm currently trying to migrate out the Markdown Manager from core to `@tiptap/markdown` so you need to install it manually.

Works as far as I can see, from the Markdown extension I'm overriding the editors default functions like setContent or the default commands and extend what is required to make this work.

I'll pursue this solution so we can have a less bloated core.

nperez0111:
> Works as far as I can see, from the Markdown extension I'm overriding the editors default functions like setContent or the default commands and extend what is required to make this work.

Very cool that it works 💪 

Appreciate the consideration going into this, Tiptap is used in a ton of places, so being light is good default goal

bdbch:
@nperez0111 happy to update that I found a good way to overwrite Tiptaps base functionality as soon as the Markdown extension is installed / used.

I had to export commands + their types and extend the Command Option types to make it work but now everything is optional and just exists in an external package (`@tiptap/markdown`) so if this is not installed, no types (except base types for the Extendable interface) are included in the core.

image

Stuff like this now works and is initially parsed from Markdown if `contentAsMarkdown` is `true`:

image

And running content related commands like this works as well:

image

nperez0111:
Glad to hear that it worked for you @bdbch, one suggestion, I'd probably have the option be `format: 'markdown'` instead of as a flag, because if you pass a json object it is unclear whether it will be interpreted differently or not, so it is best to be explicit (similar to using tagged unions as React props instead of boolean flags)

packages/core/src/index.ts
```diff
@@ -1,4 +1,5 @@
 export * from './CommandManager.js'
+export * from './commands/index.js'
```
nperez0111: May be worth namespacing this since it will be a number of commands that people might think they can use without the editor instance:

```suggestion
export * as commands from './commands/index.js'
```

Also reduces the long list of exports you'd see on the package

bdbch:
Quick update - I thought about the API on how people will define their markdown syntax for custom content not already in the GFM syntax.

We will expose a few markdown helpers that will allow users to quickly create the parser, serializer and tokenizer with opinionated syntax so there's no hassle with Regex.

It's still possible to do and the tokenizer logic is very flexible but also hard to learn.

## Block Nodes

To create a block node markdown spec, you can just simply call:

```js
markdown: createBlockMarkdownSpec({
  nodeName: 'myCustomNode',
  name: 'custom',
  content: 'block',
}),

// results in a supported syntax:
//
// :::custom {attributeA="hello" attributeB="world"}
//
// Custom content inside this node
//
// :::
```

## Atom Block Nodes

To create a atom block node markdown spec, you can just simply call:

```js
markdown: createAtomBlockMarkdownSpec({
  blockName: 'youtube',
}),

// results in a supported syntax:
//
// :::youtube {src="..." start="0" width="800" height="600"}
```

## Inline nodes

To create a inline node markdown spec, you can just simply call:

```js
// example from the markdown extension

markdown: createInlineMarkdownSpec({
  nodeName: 'mention',
  name: '@', // to make the syntax shorter
  selfClosing: true,
  allowedAttributes: ['id', 'label'],
}),

// results in a supported syntax:
//
// [@ id="john-doe" label="John Doe"]
// [tag type="bug"]Broken link[/tag]
```

This way most users shouldn't have to worry about creating Regex rules or creating custom rules just to have their nodes available in markdown output/input.

bdbch:
Next steps:

- Solve some smaller typing issues
- Check if commands work fine with Markdown input
- Follow feedback from @nperez0111 regarding `asMarkdown` / `type: 'markdown'`
- Write round-trip tests & smaller atomic unit tests for the markdown functionality
- Write first draft of docs

KaranJayakumar:
Hey @bdbch, this looks great!
Is there any ETA for when markdown support? We're attempting to integrate TipTap into our systems since the tiptap-markdown package doesn't currently work with Tiptap 3 so just curious if we should bother rolling our own version :)

- Tiptap User (Not associated with tiptap)

bdbch:
> Is there any ETA for when markdown support?

I'm pretty close to finalizing this, so probably end of this week or next week.

bdbch:
I created a first draft of the markdown docs here:
https://github.com/ueberdosis/tiptap-docs/pull/435

aguingand:
As the creator of [tiptap-markdown](https://github.com/aguingand/tiptap-markdown), I’m *hyped* to see this in core!  
I originally developed the package to add Markdown support to our Laravel CMS ([Sharp](https://github.com/code16/sharp)). With the rise of AI, it gained a lot of traction.  
One of the main challenges, however, was its limited compatibility with streamed Markdown content.

All in all, I’m really happy to have been part of this open-source adventure!

bdbch:
I finally figured out where to disable forced authentication for Vercel deployments. If you want to check the documentation for this PR already - feel free to do so here and send me your feedback:

https://tiptap-docs-git-docs-markdown-support-tiptap.vercel.app/docs/editor/markdown

packages/markdown/src/utils.ts
```diff
@@ -0,0 +1,146 @@
+import type { Content } from '@tiptap/core'
+import type { Fragment, Node } from '@tiptap/pm/model'
+
+import type { ContentType } from './types'
+
+export function wrapInMarkdownBlock(prefix: string, content: string) {
```
arnaugomez, line 12: Nitpick: add documentation comment explaining what the function does

bdbch:
> I see you add the markdown definitions to each and every extension. Doesn't this unnecessarily increment bundle size when you use Tiptap, but not with Markdown? Consider including the tool definitions of the official Tiptap extensions in a separate configuration object that is independent of the extension spec of the official extensions

Yes but this is wanted. I think the bundle size shouldn't be hit too hard by having markdown definitions in the base extensions.

Otherwise we will have additional steps when trying to get all extensions to run with markdown, like manually extending each extension with a markdown ready one.

We already excluded MarkedJS from the core bundle to keep the bundle size low.

I'd love to use vitest for tests but don't have time now to implement vitest and setup our Prosemirror setup with it. Will do when I have time. Cypress is really slow for tests like that.

tyn1998:
@bdbch Thank you for your markdown extension PR, really looking forward to its debut!

tyn1998:
I'm not sure if this is the appropriate place to ask the following question, but please allow me to do so here:

I found that [markdown conversion](https://tiptap.dev/docs/conversion/import-export/markdown/editor-extensions) cannot correctly handle markdown table syntax:

Payload:

image

The api response:

image

Is it expected that the markdown table is converted into a node of `type` text?

PR #6843, opened 2025-08-19
fix: improve types for rendered components props for ReactRenderer

michaellopez:
## Changes Overview

This PR improves the TypeScript types for the ReactRenderer to provide better type safety and autocompletion for the rendered component props.

Previously, the rendered component props were typed as `Record<string, any>`, which allowed for common errors like typos or incorrect prop value types to go unnoticed by the compiler. This change ensures that any props passed to the `ReactRenderer` are correctly validated against the rendered component's actual prop types, leading to a more robust developer experience.

## Implementation Approach

The implementation introduces a generic type parameter `P` to the `ReactRenderer` class and `ReactRendererOptions` interface.

This generic `P`, representing the rendered component's props, is used to strictly type the props object in the constructor.

The `updateProps` method is now typed to accept `Partial<Omit<P, 'ref'>>`, ensuring type-safe partial updates and preventing the internally-managed `ref` prop from being overwritten.

The constructor logic is updated to handle cases where props might be `undefined`, mostly to avoid type casting in the constructor declaration.

## Testing Done

As this is a type-only change, no runtime behavior is altered. All existing unit and integration tests continue to pass. The primary validation for this change is through TypeScript's static analysis, which now correctly identifies type errors at compile time when using the ReactRenderer. 

## Verification Steps

A verification test was done in a consuming project locally by installing the `@tiptap/react` package via `pnpm add file:<path to clone>/packages/react`. The improved types worked as expected when autocompleting and inspecting them.

## Checklist

~- [ ] I have added tests where applicable.~

### DIFF
--- a/.changeset/new-rabbits-hunt.md
+++ b/.changeset/new-rabbits-hunt.md
@@ -1,0 +1,5 @@
+---
+'@tiptap/react': patch
+---
+
+Improve typings for rendered component props in ReactRenderer.
--- a/packages/react/src/ReactRenderer.tsx
+++ b/packages/react/src/ReactRenderer.tsx
@@ -98,7 +98,7 @@ function isReact19Plus(): boolean {
   return false
 }
 
-export interface ReactRendererOptions {
+export interface ReactRendererOptions<P extends Record<string, any> = object> {
   /**
    * The editor instance.
    * @type {Editor}
@@ -107,10 +107,10 @@ export interface ReactRendererOptions {
 
   /**
    * The props for the component.
-   * @type {Record<string, any>}
+   * @type {P}
    * @default {}
    */
-  props?: Record<string, any>
+  props?: P
 
   /**
    * The tag name of the element.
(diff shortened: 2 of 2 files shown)


bdbch (changes requested):
Thanks for working on this. The general direction here makes sense, but I don't think the `ref` part is quite right yet.

`ReactRenderer` still allows a caller-provided `props.ref` at runtime and only injects its own ref when one isn't present. With this change, the public types now disallow `ref` in `props` and `updateProps`, which means the typings are narrower than the actual runtime behavior. That also seems risky with React 19, where `ref` is a normal prop for function components. In practice this looks like it could reject valid consumer code without actually preventing `ref` from being passed at runtime.

I think we should either keep `ref` in the public types, or change the runtime behavior as well if the intention is to fully reserve it for internal use. Right now it feels like we're in an in-between state.

I'd also like to see some type-level regression coverage here. Since this is a typings change, the main risk is breaking valid component shapes without noticing. A small type test covering plain function components, `forwardRef` components, and a case where the props already include `ref` would make this much safer to merge.

bdbch:
Closing this for now - feel free to re-open if you want to continue working on this.

PR #6857, opened 2025-08-22
fix(drag-handle-vue-2): fix vue2 warnign about objects as default

max-nextcloud:
## Changes Overview

* Vue2 requires factory functions for defaults for objects and arrays.
* `class` may not be used as prop.

## Implementation Approach

Use a factory function.

## Testing Done

Checked that it prevents the warning by building text with it:

> Invalid default value for prop "PluginKey": Props with type Object/Array must use a factory function to return the default value.

## Verification Steps

Actually testing is cumbersome. Vue 3 seems to have dropped the warning. [I'm still checking](https://github.com/vuejs/docs/pull/3282) if that actually also means it's not an issue anymore.
Checking the code and checking [the docs]( https://v2.vuejs.org/v2/guide/components-props#Prop-Validation) is probably best (search for 'factory function')

## Additional Notes

grafik

I'm still looking into fixing the class reserved name warning.

## Checklist

- [x] I have added tests where applicable (not applicable).
- [x] I have fixed any lint issues (lint is happy).

## Related Issues

Initially came up in https://github.com/nextcloud/text/pull/7537#issuecomment-3197240037

### DIFF
--- a/.changeset/moody-trees-rescue.md
+++ b/.changeset/moody-trees-rescue.md
@@ -1,0 +1,5 @@
+---
+'@tiptap/extension-drag-handle-vue-2': patch
+---
+
+Use factory function for object default value as required by vue 2.
--- a/.changeset/serious-tools-hear.md
+++ b/.changeset/serious-tools-hear.md
@@ -1,0 +1,5 @@
+---
+"@tiptap/extension-drag-handle-vue-2": patch
+---
+
+Fixed a bug that would cause Vue 2 to throw errors in console because Vue 2 expects factory functions for prop defaults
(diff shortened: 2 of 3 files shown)


max-nextcloud:
I tried to access `class` through `this.$attrs` but turns out it's not included there: https://v2.vuejs.org/v2/api/#vm-attrs

The only ways for fixing the other error I see are:
* Always using a `class="drag-handle"` on the div. More classes can be added via attribute fallbacks - but there would be no way to replace that class anymore. So behavior would change.
* Checking for other classes and removing the `drag-handle` class to preserve current behavior.
* Changing the prop name - also breaking the api.

mejo-:
@bdbch thanks for merging the PR. Would you consider to backport it to the `v2` branch? So far we're still stuck with Tiptap v2 in the Nextcloud Text editor.

bdbch:
Released as `2.26.4`

PR #6841, opened 2025-08-19
Support custom container via `appendTo` for `BubbleMenu`/`FloatingMenu` across React and Vue

felixgabler:
## Changes Overview
- Add `appendTo` support to `@tiptap/extension-floating-menu`.
- Pass `appendTo` through in React, Vue 2, and Vue 3 for both `BubbleMenu` and `FloatingMenu`.
- Add brief comments in Vue menus clarifying Teleport’s role vs plugin-controlled placement.
- No breaking changes; default behavior remains appending to the editor’s parent.

## Implementation Approach
- Extend `FloatingMenuPluginProps` with optional `appendTo?: HTMLElement`.
- Store and use `appendTo` in `show()` to append the element to `appendTo ?? view.dom.parentElement`.
- Thread `appendTo` prop through framework adapters:
  - React `FloatingMenu` and `BubbleMenu`
  - Vue 2 `FloatingMenu` and `BubbleMenu`
  - Vue 3 `FloatingMenu` and `BubbleMenu`
- Keep Teleport in Vue components; element is removed on mount and re-parented by the plugin. Comments clarify this.

## Testing Done
Manually tested in Vue 3 project

## Verification Steps
- In each framework, render both menus with and without `appendTo`:
  - Set a parent with `overflow: hidden` and verify menus are clipped without `appendTo` and not clipped with it.
  - Example (Vue 3):
    ```ts
    <BubbleMenu :editor="editor" :appendTo="document.body">...</BubbleMenu>
    <FloatingMenu :editor="editor" :appendTo="document.body">...</FloatingMenu>
    ```
  - Example (React):
    ```tsx
    <BubbleMenu editor={editor} appendTo={document.body}>...</BubbleMenu>
    <FloatingMenu editor={editor} appendTo={document.body}>...</FloatingMenu>
    ```
- Interact: make a text selection (BubbleMenu) and place caret in empty paragraph (FloatingMenu). Confirm correct positioning and visibility toggles.
- Resize window; ensure menus reposition correctly.
- Blur/focus editor and click inside menus; ensure menus stay visible appropriately.

## Additional Notes
- Teleport is only used to instantiate the Vue subtree; the plugin owns DOM placement. `appendTo` is the authoritative container.
- No API changes for existing users unless they opt into `appendTo`.

### Checklist
- [x] I have created a [changeset](https://github.com/changesets/changesets) for this PR if necessary.
- [x] My changes do not break the library.
- ~[ ] I have added tests where applicable.~
[...]

PR #6877, opened 2025-08-26
feat(BubbleMenu):add custom positioning support

donovan-fournier:
## Changes Overview

The BubbleMenu position is implied by the current selection (either text / node / cell selection).
In the V2, tippy allowed to set a [ `getreferenceclientrect`](https://atomiks.github.io/tippyjs/v6/all-props/#getreferenceclientrect) that allows deep customisation regarding the position.

For example, in a container / items scenario, we are using a BubbleMenu to display a global menu for the container when the selection is inside any item.
Right now, the menu is always positioned relative to the selection, but we want to be able to position to a parent node

## Implementation Approach

2 approaches were possible:
* Making the BubbleMenu more specific and provide node types info to define the position. Update the updatePosition method to compute that node position (if the selection is inside), instead of the selection.
    * This is very specific, and there could be a functional collision between shouldShow and the computation of the position
* Provide a generic method, similar to tippy, which provides the bounds of the expected element.
    * I chose this approach due to its 'similarity' with V2 and the generic behavior (It accepts more use case)

## Testing Done

Tested in the Demo (both Vue and React) 

## Verification Steps

Tested in the Demo (both Vue and React) 

## Additional Notes

We are using BubbleMenu for this use case, but due to the similarities with the FloatingMenu, I can add the same update to it  as well.

This PR may conflict with #6799 or #6841

## Checklist


bdbch:
Could you create a PR for ueberdosis/tiptap-docs describing your changes on the related packages? Otherwise our docs will drift out of context.

donovan-fournier:
@bdbch Here is the PR for the documentation.
https://github.com/ueberdosis/tiptap-docs/pull/387

Please let me know if there's any other change to do

bdbch (approved):

PR #6910, opened 2025-09-03
Allow key handling for suggestions when ESC is pressed

### LINKED ISSUE #6898: suggestion render setting onKeyDown not working on Esc
### Affected Packages

core mention suggestion

### Version(s)

3.3.0

### Bug Description

if keydown ESC, suggestion will not call onKeyDown because it intercepted inside and straight turn to call onExit().
it not a problem, but onExit doesn't pass the event, it cause i can't call event.preventDefault() or event.stopPropagation() if i need. 

this case is common when editor in sied menu or popup, i can't accurate intercept ESC KeydownEvent just when suggestion active.

https://github.com/ueberdosis/tiptap/blob/ba8d13fa83e17689b03a1fdee131fb870b0e4586/packages/suggestion/src/suggestion.ts#L519-L544

### Browser Used

Chrome

### Code Example URL

https://codesandbox.io/p/devbox/cocky-wind-rxwghd

### Expected Behavior

Can call event.stopPropagation() or event.preventDefault()  when in suggestion active and press ESC

CodeSandbox templates step:

1. open Drawer
2. type '@' to  trigger mention menu open.
3. press ESC
4. Can not call event.stopPropagation()/event.preventDefault()  , so Drawer Component get ESC KeydownEvent and colse self. (but i just want to close mention menu not the drawer.)

Expected Behavior:  can stopPropagation and Drawer Component not get ESC KeydownEvent.

### Dependency Updates

bdbch:
## Changes Overview

This pull request improves the flexibility of Escape key handling in the TipTap suggestion plugin, allowing consumers to intercept and manage Escape key events before the plugin auto-exits. This enables better integration with custom UI scenarios, such as editors embedded in modals or drawers.

## Implementation Approach

**Escape key handling improvements:**

* The suggestion plugin now invokes the consumer's `render().onKeyDown` handler for Escape key events before auto-exiting; if the handler returns `true`, the plugin assumes the consumer handled the event and does not auto-exit, letting the consumer control propagation and suggestion lifecycle.

## Checklist

## Related Issues

- Fixes #6898

### DIFF
--- a/.changeset/allow-escape-keydown-handling.md
+++ b/.changeset/allow-escape-keydown-handling.md
@@ -1,0 +1,9 @@
+---
+"@tiptap/suggestion": patch
+---
+
+Allow consumers to handle the Escape key via `render().onKeyDown` before the suggestion plugin auto-exits.
+
+Previously the suggestion plugin intercepted Escape internally and immediately called `onExit`, preventing `render().onKeyDown` from receiving the event and stopping propagation. Now `render().onKeyDown` is invoked first for Escape; if it returns `true` the plugin assumes the consumer handled the event (so they can call `event.preventDefault()` / `event.stopPropagation()` and optionally call `exitSuggestion(view)` themselves). If it returns `false` (or is absent), the plugin will continue to call `onExit` and close the suggestion as before.
+
+This change enables scenarios where the editor is inside a modal/drawer and the consumer needs to prevent the outer UI from reacting to Escape while still controlling the suggestion's lifecycle.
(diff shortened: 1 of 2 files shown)


PatrickLai007:
I believe this change does not address the root cause of the issue. Based on my investigation, the Escape event is being handled instead of being passed down to the suggestion extension. @bdbch, could you please review this?

bdbch:
@PatrickLai007 I'm currently limited in time to look into it. If you want to you could open a PR that passes through the event.

PatrickLai007:
> @PatrickLai007 I'm currently limited in time to look into it. If you want to you could open a PR that passes through the event.

@bdbch, I regret to inform you that the alert was erroneous. The issue was identified as originating from my Chrome extension. Upon testing in incognito mode, the functionality was confirmed to be as intended.

PR #6916, opened 2025-09-05
fix(link): respect openOnClick: false configuration

### LINKED ISSUE #6865: [Link] openOnClick option is ignored
### Affected Packages

extension-link

### Version(s)

3.3.0

### Bug Description

The `openOnClick: false` option seems to be ignored. When I click on a link in the editor to edit the text, it opens the URL in a new tab.

```ts
  const editor = useEditor({
    content: options.initialContent || '<p></p>',
    extensions: [
      StarterKit,
      Div,
      TextStyle,
      Superscript,
      TextAlign.configure({ types: ['heading', 'paragraph'] }),
      Link.configure({ openOnClick: false }),
      AllowStyleAttributes,
    ],
  // ...
```

### Browser Used

Chrome

### Code Example URL

_No response_

### Expected Behavior

I would expect clicking a link to not open it.

_No response_

### Dependency Updates

praneeth622:
Fixes #6865 where openOnClick: false was being ignored. The issue was that when openOnClick was false, no click handler plugin was added, allowing browser's default anchor behavior to take over.

Now the click handler plugin is always added and controls behavior through preventDefault() and conditional window.open() based on the openOnClick setting.

## Changes Overview
Modified the Link extension to always register the click handler plugin regardless of the `openOnClick` setting. The plugin now internally handles the `openOnClick` logic by always preventing default browser behavior and conditionally opening links based on the configuration.

## Implementation Approach
- **Root cause**: When `openOnClick: false`, no click handler plugin was registered, so browser default anchor behavior prevailed
- **Solution**: Always add the click handler plugin but control link opening behavior internally
- **Modified files**: `clickHandler.ts` to accept and handle `openOnClick` parameter, `link.ts` to always pass the setting to the plugin
- **Key logic**: `event.preventDefault()` always called, `window.open()` called conditionally based on `openOnClick` value

## Testing Done
- Manual testing in React and Vue demos confirms links no longer open when `openOnClick: false`
- Verified all three `openOnClick` states work correctly: `true`, `false`, `'whenNotEditable'`
- Build passes successfully with no TypeScript or linting errors
- Created Cypress test suite covering all openOnClick scenarios

## Verification Steps
1. Checkout the branch and run `npm run build`
2. Start dev server with `npm run dev`
3. Navigate to React Link demo at `http://localhost:3000/src/Marks/Link/React/`
4. Observe demo uses `openOnClick: false` configuration
5. Click on links in editor content - they should NOT open in new tabs
6. Verify you can edit link text instead of navigating away

## Additional Notes
- Maintains full backward compatibility with existing configurations
- No performance impact as click handler was already conditionally added
- Properly handles deprecated `'whenNotEditable'` option
- Solution prevents default behavior for all link clicks, then conditionally opens based on setting

## Checklist

## Related Issues
Fixes #6865


packages/extension-link/src/helpers/clickHandler.ts
```diff
@@ -8,6 +8,7 @@ type ClickHandlerOptions = {
   editor: Editor
   openOnClick?: boolean
   enableClickSelection?: boolean
+  openOnClick?: boolean | 'whenNotEditable'
```
arnaugomez, line 11: the `openOnClick` property appears duplicated

arnaugomez (commented):
It appears that the issue that this PR solves is already fixed.

https://github.com/ueberdosis/tiptap/issues/6865#issuecomment-3258763930

The issue was caused by the fact that the Link extension is added by the StarterKit, so the `openOnClick` option has to be passed there.

arnaugomez:
@praneeth622 the issue that this PR tries to fix has already been fixed. So can you do the following:

- If you think this PR is still necessary and that there is a bug in the Link extension, open a new issue describing the problem that this PR fixes, and how to reproduce the bug.
- Otherwise, you can close this PR.

Thank you anyway for your contribution to Tiptap and for trying to fix the issue.

PR #6917, opened 2025-09-05
feat(react): introduce `<Tiptap />` component for easier integration

bdbch:
## Changes Overview

This pull request introduces a new, optional React integration for Tiptap, providing a declarative `<Tiptap />` component that simplifies editor setup in React applications. This is an additive change, offering a more ergonomic, component-first API while keeping the existing imperative API fully supported for this major version. The update also includes minor robustness improvements in the core command logic.

With this new approach, we can later deprecate the current ways of setting up an editor that can be confusing and allows multiple ways that can lead to implementation errors.

Example:

```tsx
import { Tiptap, useEditor, useTiptap } from '@tiptap/react'

function MyEditor() {
  const editor = useEditor({ extensions: [StarterKit], content: '<h1>Hello from Tiptap</h1>' })

  return (
    <Tiptap instance={editor}>
      <Tiptap.Content />
      <Tiptap.BubbleMenu>My Bubble Menu</Tiptap.BubbleMenu>
      <Tiptap.FloatingMenu>My Floating Menu</Tiptap.FloatingMenu>
      <MenuBar /> {/* MenuBar can use the new `useTiptap` hook to read the editor instance from context */}
    </Tiptap>
  )
}

function MenuBar() {
  const { editor } = useTiptap()

  return (
    <div>
      <button onClick={() => editor.chain().focus().toggleBold().run()}>Bold</button>
      <button onClick={() => editor.chain().focus().toggleItalic().run()}>Italic</button>
      <button onClick={() => editor.chain().focus().toggleStrike().run()}>Strike</button>
    </div>
  )
}
```

As you can see, this approach automatically provides a **Context** for all sub-components, that components can hook in.

- `<Tiptap />` - The wrapper that provides context
- `<Tiptap.Loading />` - A loading wrapper that can be used to render a loading spinner / message while the editor is initializing, setting up. Future plan is to wrap this together with collaborative features to make sure editors are only rendered when everything is initialized
- `<Tiptap.Content />` - The replacement for `<EditorContent editor={editor} />`
- `<Tiptap.BubbleMenu />` - The replacement for `<BubbleMenu editor={editor} />`
- `<Tiptap.FloatingMenu />` - The replacement for `<FloatingMenu editor={editor} />`
- `useTiptap()` - A hook that can be used in sub-components to retrieve the Tiptap editor instance from children
  - `const { editor } = useTiptap()`
- `useTiptapState(selectorFn)` - A replacement for `useEditorState` which can be used to directly retrieve editor state from the current editor
## Implementation Approach

**React integration and API improvements:**

* Added a new `Tiptap` component and related context/hooks (`TiptapContext`, `useTiptap`, `useTiptapState`) to `packages/react/src/Tiptap.tsx`, enabling declarative, prop-driven editor setup and composition in React apps. This includes subcomponents for content, loading state, and menus.
* Exported the new `Tiptap` component and related utilities from the package entry point (`packages/react/src/index.ts`).
* Added a changeset documenting the new React integration, migration guidance, and deprecation plans for the legacy setup.

**Core robustness:**

* Improved error handling in the `focus` command (`packages/core/src/commands/focus.ts`) to gracefully handle cases where the editor view is not yet mounted, avoiding runtime errors.

## Additional Notes

- We will need to upgrade all demos later down the line

## Checklist

### DIFF
--- a/demos/src/Examples/Default/React/MenuBar.tsx
+++ b/demos/src/Examples/Default/React/MenuBar.tsx
@@ -1,0 +1,128 @@
+import { useTiptap, useTiptapState } from '@tiptap/react'
+import React from 'react'
+
+import { menuBarStateSelector } from './menuBarState.ts'
+
+/**
+ * Menu bar component that uses useTiptap and useTiptapState hooks
+ * to access the editor from context.
+ */
+export function MenuBar() {
+  const { editor, isReady } = useTiptap()
+  const editorState = useTiptapState(menuBarStateSelector)
+
+  if (!isReady || !editor) {
+    return null
+  }
+
+  return (
+    <div className="control-group">
+      <div className="button-group">
+        <button
+          onClick={() => editor.chain().focus().toggleBold().run()}
+          disabled={!editorState.canBold}
+          className={editorState.isBold ? 'is-active' : ''}
+        >
+          Bold
+        </button>
+        <button
+          onClick={() => editor.chain().focus().toggleItalic().run()}
+          disabled={!editorState.canItalic}
+          className={editorState.isItalic ? 'is-active' : ''}
+        >
+          Italic
+        </button>
+        <button
+          onClick={() => editor.chain().focus().toggleStrike().run()}
+          disabled={!editorState.canStrike}
+          className={editorState.isStrike ? 'is-active' : ''}
+        >
+          Strike
+        </button>
+        <button
+          onClick={() => editor.chain().focus().toggleCode().run()}
+          disabled={!editorState.canCode}
+          className={editorState.isCode ? 'is-active' : ''}
+        >
+          Code
+        </button>
+        <button onClick={() => editor.chain().focus().unsetAllMarks().run()}>Clear marks</button>
+        <button onClick={() => editor.chain().focus().clearNodes().run()}>Clear nodes</button>
+        <button
+          onClick={() => editor.chain().focus().setParagraph().run()}
+          className={editorState.isParagraph ? 'is-active' : ''}
+        >
+          Paragraph
+        </button>
+        <button
+          onClick={() => editor.chain().focus().toggleHeading({ level: 1 }).run()}
+          className={editorState.isHeading1 ? 'is-active' : ''}
+        >
+          H1
+        </button>
+        <button
+          onClick={() => editor.chain().focus().toggleHeading({ level: 2 }).run()}
+          className={editorState.isHeading2 ? 'is-active' : ''}
+        >
+          H2
+        </button>
+        <button
+          onClick={() => editor.chain().focus().toggleHeading({ level: 3 }).run()}
+          className={editorState.isHeading3 ? 'is-active' : ''}
+        >
+          H3
+        </button>
+        <button
+          onClick={() => editor.chain().focus().toggleHeading({ level: 4 }).run()}
+          className={editorState.isHeading4 ? 'is-active' : ''}
+        >
+          H4
+        </button>
+        <button
+          onClick={() => editor.chain().focus().toggleHeading({ level: 5 }).run()}
+          className={editorState.isHeading5 ? 'is-active' : ''}
+        >
+          H5
+        </button>
+        <button
+          onClick={() => editor.chain().focus().toggleHeading({ level: 6 }).run()}
+          className={editorState.isHeading6 ? 'is-active' : ''}
+        >
+          H6
+        </button>
+        <button
+          onClick={() => editor.chain().focus().toggleBulletList().run()}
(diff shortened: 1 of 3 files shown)


packages/react/src/Tiptap.tsx
```diff
@@ -0,0 +1,109 @@
...
+import { createContext, useContext, useEffect, useState } from 'react'
+
+import type { Editor, EditorContentProps, EditorStateSnapshot } from './index.js'
+import { EditorContent, useEditorState } from './index.js'
+import { type BubbleMenuProps, BubbleMenu } from './menus/BubbleMenu.js'
+import { type FloatingMenuProps, FloatingMenu } from './menus/FloatingMenu.js'
+
+export type TiptapContextType = {
+  editor: Editor
+  isReady: boolean
```
arnaugomez, line 20: Could you add a documentation comment explaining what `isReady` means? Like, for example, "true when the editor has been created".

Also not sure if the `isReady` is necessary. It could be argued that, if the user wants to listen to when the editor has been created, they could just register an event listener. I think many developers will not use the `isReady` property.

packages/core/src/commands/focus.ts
```diff
@@ -60,8 +60,14 @@ export const focus: RawCommands['focus'] =
...
-    if ((view.hasFocus() && position === null) || position === false) {
-      return true
+    try {
+      if ((view.hasFocus() && position === null) || position === false) {
+        return true
+      }
+    } catch {
+      // if view.hasFocus fails (view not mounted yet)
+      // we will return false because there's nothing to focus
+      return false
```
arnaugomez, line 79: Is this change related to the new Tiptap component?

arnaugomez (commented):
I love this approach, it could make my code simpler.

I'd like to ask whether the `isCreated` property is necessary. Another doubt I have is whether the `Tiptap` component could be called `TiptapProvider` to let the React developer know that it is a context provider

arnaugomez (commented):
I think there is some overlap between this feature and this one:

https://tiptap.dev/docs/editor/getting-started/install/react#using-the-editorcontext

bdbch:
> I think there is some overlap between this feature and this one:
> 
> https://tiptap.dev/docs/editor/getting-started/install/react#using-the-editorcontext

Yes - this will also become obsolete. It was the first draft at implementing React Context but since you can do:

1. Manual creation of `useEditor` without any context
2. Manual creation of `useEditor` and pass it to `<EditorContext.Provider>`
3. Automatic creation by using `<EditorProvider content="" extensions={[]} />`

I wanted to streamline it and make it more straight forward. So with 4.0.0 I'd get rid of at least 2 and 3.

Users should still be able to just create instances via `useEditor` and pass it to the EditorContent component BUT the `<Tiptap />` approach should become the new React standard for Tiptap.

packages/core/src/commands/focus.ts
bdbch: Yep - I think it's also a bug that wasn't discovered yet. Basically when you create an instance and render it out straight away - any command or `can()` check that checks for focus will throw an hard error because `dom.hasFocus` and just be checked when the DOM element for the view is mounted.

packages/react/src/Tiptap.tsx
bdbch: I will add JSDocs, I think I forgot them while working on it.

I would keep the isReady because in the future I'd like to use it for collaboration as well (so isReady would be true when the editor is initialized & the collaborative connection is ready if used).

bdbch:
> Another doubt I have is whether the Tiptap component could be called TiptapProvider to let the React developer know that it is a context provider

I understand your thought but I think lots of other libraries already do something similar (just checking Radix or FloatingUI for example - you just use `DropdownMenu`, not `DropdownMenuProvider`).

We could allow both - `<Tiptap>` or `<Tiptap.Provider>` as an additional mapping but that could also confuse people.

packages/react/src/Tiptap.tsx
```diff
@@ -0,0 +1,109 @@
...
+}
+
+export const TiptapContext = createContext<TiptapContextType>({
+  editor: null as unknown as Editor, // dummy editor instance
+  isReady: false,
+})
```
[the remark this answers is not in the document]
bdbch: Not sure about this. We can't create a real editor here as we don't have any extensions which would cause an error because editors always need a **doc** node type.

Allowing null here would mean that users ALWAYS have to do checks before, while our editor instance should always exist - this is only a problem if the context would be used without any editor instance being passed to it.

bdbch:
@arnaugomez updated the PR & added documentation. Would like to move this forward.

bdbch:
Thanks 🙏

PR #6920, opened 2025-09-05
Fix 6919 insert content at elsewhere

### LINKED ISSUE #6919: [BUG]: insertContentAt incorrectly modifies the insert position
### Affected Packages

core

### Version(s)

3.3.0

### Bug Description

The fix in #5651 seems to assume that the insert position is related to, or maybe the same as, the selection `from`. That is the case for `insertContent`, but `insertContentAt` can be used to insert content anywhere, not just at the current selection position. This assumption can lead to unexpected results when the selection happens to be at the beginning of a text block somewhere unrelated to the insert position. The fix also assumes that the current editor state is still useful when this command may be chained with other commands that have modified the document. I believe that the correct fix is to resolve the given position against the transaction's current document and then consider whether that position is at the beginning of a text block.

### Browser Used

Chrome

### Code Example URL

_No response_

### Expected Behavior

The content should be inserted at the position specified.

_No response_

### Dependency Updates

tilgovi:
When inserting content at a position other than the selection, consult the state of the resolved insert position when deciding whether or not to modify it in order to split the block at that position or not.

Fix #6919.

### DIFF
--- a/packages/core/src/commands/insertContentAt.ts
+++ b/packages/core/src/commands/insertContentAt.ts
@@ -74,7 +74,6 @@ export const insertContentAt: RawCommands['insertContentAt'] =
       }
 
       let content: Fragment | ProseMirrorNode
-      const { selection } = editor.state
 
       const emitContentError = (error: Error) => {
         editor.emit('contentError', {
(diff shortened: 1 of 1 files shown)


arnaugomez:
I close this PR because I've also [fixed it here](https://github.com/ueberdosis/tiptap/pull/6964). I've merged your changes into mine so that you get credit for your work.

tilgovi:
Thank you!

PR #6941, opened 2025-09-10
feat(character-count): added autotrim option

daanvangeloven:
## Changes Overview

Added an option in the CharacterCount extension to make autotrimming configurable.
This way you can set content past the limit programatically and manually trim it instead if it being cut off automatically.

## Implementation Approach

Added a boolean option to character count that disables the inital trimming.

## Verification Steps

- Open a text field and set content that exceeds the limit with autotrim disabled. You will now only be able to remove content from the editor until you get below the limit again.


bdbch:
Please submit a PR documenting this change over on https://github.com/ueberdosis/tiptap-docs

daanvangeloven:
@bdbch https://github.com/ueberdosis/tiptap-docs/pull/392

daanvangeloven:
Can this be merged?

arnaugomez:
In what situation can you set content past the limit when autoTrim is false? Only as the initial content of the document? Or can you also set it past the limit when creating a transaction programmatically by running a command?

arnaugomez:
I say it because it's not clear to me when the content will be trimmed when autoTrim is disabled. Will it be trimmed when I paste content, for example? So the property name might not be 100% descriptive of its behaviour because it is called "autoTrim:false" but when content is pasted it automatically trims it.

arnaugomez:
Also, apologies for the delay in reviewing.

bdbch:
I think the JSDoc helps a bit with clarity here:

> Sets whether the content will be automatically trimmed when programatically setting content over the limit.

The default still is `true` which means nothing changes for users, it's just meant as an escape hatch for users who may want to trim it manually for more specific use cases.

I'll see that I get this merged for the next release.

daanvangeloven:
We had a form with existing data. On some of the form fields users are able to set a max character limit but we don't want to automatically delete parts of the text when a new limit is set. 

This way users can set a limit lower than the current text length and then trim the text as needed until it is the correct length.

bdbch:
Added documentation for this new option here:
https://github.com/ueberdosis/tiptap-docs/pull/793

bdbch (dismissed):

Issue #4792, opened 2024-01-16
[Bug]: Removing spaces between tags


szymekjanaczek:
### Which packages did you experience the bug in?

core, react

### What Tiptap version are you using?

2.1.16

### What’s the bug you are facing?

Let's assume we write a text and format it like below:
```html
<p>Lorem <em>ipsum <strong>dolor</strong></em> <strong>sit</strong> amet.</p>
```
When saving this somewhere and then recovering and load into tiptap again, getting text in this format:
```html
<p>Lorem <em>ipsum <strong>dolor</strong></em><strong>sit</strong> amet.</p>
```
Without space between `</em>` and `<strong>`. So the words "dolor" and "sits" connects into one word "dolorsit".

### What browser are you using?

Chrome

### Code example

[codesandbox.io - React version](https://codesandbox.io/p/sandbox/charming-rumple-tslmsj?file=%2Fsrc%2FApp.js%3A15%2C15)
```javascript
const editor = useEditor({
  extensions: [StarterKit],
  editorProps: {
    attributes: {
      class: "Editor",
    },
  },
  content: `
    <p>Lorem <em>ipsum <strong>dolor</strong></em> <strong>sit</strong> amet.</p>
  `,
});

return (
  <div className="App">
    <EditorContent editor={editor} />
  </div>
);
```

### What did you expect to happen?

Reading the text as it has been previously written - with spaces between HTML tags.
So when I save
```html
<p>Lorem <em>ipsum <strong>dolor</strong></em> <strong>sit</strong> amet.</p>
```

And then set it as a content of the TipTap editor, I want to keep original spaces - without gluing words together.

### Anything to add? (optional)

<img width="450" alt="image" src="https://github.com/ueberdosis/tiptap/assets/74879123/c12e2bfd-b9dc-4b2c-8162-5cc77bce111b">

### Did you update your dependencies?

- [X] Yes, I’ve updated my dependencies to use the latest version of all packages.

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

andrewvasilchuk:
I'm getting the same problem. This is related to the changes made in #4767.

andrewvasilchuk:
@szymekjanaczek yes the problem is not  seen in `2.1.15`.

andrewvasilchuk:
Another example may be seen on the ["Mention" extension page](https://tiptap.dev/docs/editor/api/nodes/mention#usage).

```html
[...]

PR #6962, opened 2025-09-16
Add failing reproduction

reteps:
## Changes Overview

This PR serves as an issue report for #4792. I wanted to discuss implementation strategy before I implement this. Currently, whitespace is trimmed before the document is parsed by ProseMirror. This means that the `whitespace: pre` [option](https://tiptap.dev/docs/editor/extensions/custom-extensions/create-new/node#whitespace) isn't correctly respected.

I noticed this bug while implementing a custom node intended to function like a raw HTML node. The `removeWhitespaces(html)` call inside `elementFromString` prevents access to whitespace.

## Implementation Approach

There are a couple ways to solve this. The "best" way to solve this would be to invert the whitespace stripping process:

- Parse the document with ProseMirror
- Traverse the document tree, removing empty text nodes if `whitespace: pre` isn't set.

The alternative approach is to allow an "escape hatch" for the `useEditor` hook earlier in the process. The full call chain there looks like:

- `useEditor` (calls `EditorInstanceManager`)
- `packages/react/EditorInstanceManager.createEditor()`
- `Editor()`
- `Editor.createDoc`
- `createDocument`
- `createNodeFromContent`
- `elementFromString`

This entire call chain is private, which means that the only way to circumvent this behavior is to reimplement + extend from:

- `useEditor`, `EditorInstanceManager` and `Editor`.

This is a lot of code duplication to do this. Any option to override this in `createNodeFromContent` would make this much easier, such as:

- `CreateNodeFromContentOptions.removeWhitespaceNodes` (passed to `elementFromString` to strip whitespace or not)
- `CreateNodeFromContentOptions.elementFromString` (optional function override)

## Testing Done

[WIP] Will update

## Verification Steps

[WIP] Will update

## Additional Notes

N/A

## Checklist

## Related Issues

#4792

Let me know which approach is preferred (if any) and I will add it.

### DIFF
--- a/demos/src/Commands/InsertContent/React/index.spec.js
+++ b/demos/src/Commands/InsertContent/React/index.spec.js
@@ -45,6 +45,13 @@ context('/src/Commands/InsertContent/React/', () => {
     })
   })
 
+  it('should keep newlines in pre tag between nodes', () => {
+    cy.get('.tiptap').then(([{ editor }]) => {
+      editor.commands.insertContent('<pre><div>foo</div>\n<code>foo\nbar</code></pre>')
+      cy.get('.tiptap').should('contain.html', '<pre><div>foo</div>\n<code>foo\nbar</code></pre>')
+    })
+  })
+
   it('should keep newlines and tabs', () => {
     cy.get('.tiptap').then(([{ editor }]) => {
       editor.commands.insertContent('<p>Hello\n\tworld\n\t\thow\n\t\t\tnice.\ntest\tOK</p>')


demos/src/Commands/InsertContent/React/index.spec.js
```diff
@@ -45,6 +45,13 @@ context('/src/Commands/InsertContent/React/', () => {
     })
   })
 
+  it('should keep newlines in pre tag between nodes', () => {
+    cy.get('.tiptap').then(([{ editor }]) => {
+      editor.commands.insertContent('<pre><div>foo</div>\n<code>foo\nbar</code></pre>')
+      cy.get('.tiptap').should('contain.html', '<pre><div>foo</div>\n<code>foo\nbar</code></pre>')
+    })
```
reteps, line 52: This is a specific issue for our project, in which we have nodes like:

```
<my-custom-node {whitespace: `pre`}>
  <subnode>Item 1</subnode>
  <subnode>Item 2</subnode>
</my-custom-node>
```

The whitespace is getting collapsed, so we end up with a document like:

```
<my-custom-node><subnode>Item 1</subnode><subnode>Item 2</subnode>
</my-custom-node>
```

and a tree of

`my-custom-node`
- `subnode`
- `subnode`

We actually want a tree like:

`my-custom-node`
- `subnode-whitespace`
- `subnode`
- `subnode-whitespace`
- `subnode`

bdbch:
I'm unsure what to do with your PR but I can look into the bug you're describing. Should we rather re-open #4792 instead of going forward with issue-related discussions on a PR?

reteps:
That would be good. I opened this as a PR because I am willing to open a PR fixing this, if you suggest which of the two strategies you would prefer @bdbch !

reteps:
bump @bdbch

reteps:
bump @bdbch . Let me know if this is something I can contribute to.

reteps:
cc @arnaugomez

bdbch:
Sorry, I'm currently sick and can't work actively on Tiptap or react as fast as usually.

I reopened the issue. Lets discuss the problem there.

In general I'm having trouble with understanding your requirements. I guess you're using the default code block extension when you try to parse `<pre></pre>` with whitespaces inside?

bdbch:
I'm also closing this PR because it's not really making sense (description is saying it's fixing something while it's just adding a test and is more a bug-report).

I'd invite you to maybe explain your use case a little bit more in #4792.

reteps:
Thank you, I did that!!

PR #6967, opened 2025-09-16
fix: emit transactions for unmounted editors

nperez0111:
## Changes Overview

This changes the `editor.view` proxy to use the editor's existing `dispatchTransaction` handler rather than using the default ProseMirror one.

This makes it so that you can still listen for transactions in an unmounted editor.

## Implementation Approach

I've gone ahead & added a bunch of tests for the existing behavior to prove that it all works as intended

## Checklist


packages/react/src/menus/BubbleMenu.tsx
```diff
@@ -57,7 +57,6 @@ export const BubbleMenu = React.forwardRef<HTMLDivElement, BubbleMenuProps>(
         element: bubbleMenuElement,
         appendTo,
         pluginKey,
-        appendTo,
```
nperez0111: This was breaking the linter, so I just added it @bdbch

bdbch: Thanks!

PR #6972, opened 2025-09-17
fix: more accurate typing for `element` option

nperez0111:
## Changes Overview
This changes the `element` option on the `Editor` to more accurately reflect the different ways that ProseMirror can mount it's view. There are 3 options ([see docs here](https://prosemirror.net/docs/ref/#view.EditorView.constructor)):
 - passing an HTMLElement which appends to the element specified
 - passing a `{mount: HTMLElement}` which mounts to that element (A pretty useful thing to get rid of the wrapper div)
 - passing a function which takes an editor element and you can place it wherever you want in the document
 - passing null to not create an element

This now exposes the options & changes the react & vue bindings to pull the correct element from the `editor.view.dom` instead of relying on the option being a specific shape.

## Testing Done

All existing tests work, which means the mounting is done properly.

## Checklist

### DIFF
--- a/.changeset/hip-windows-retire.md
+++ b/.changeset/hip-windows-retire.md
@@ -1,0 +1,5 @@
+---
+'@tiptap/core': patch
+---
+
+Update the types of the element to create an editor instance from
--- a/packages/core/src/Editor.ts
+++ b/packages/core/src/Editor.ts
@@ -300,7 +300,7 @@ export class Editor extends EventEmitter<EditorEvents> {
           this.editorState = state
         },
         dispatch: (tr: Transaction): ReturnType<EditorView['dispatch']> => {
-          this.editorState = this.state.apply(tr)
+          this.dispatchTransaction(tr)
         },
 
         // Stub some commonly accessed properties to prevent errors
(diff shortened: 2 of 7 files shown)


bdbch (approved):

soren121:
Why is this not considered a breaking change?

bdbch:
In my tests this change didn't lead to breaking change, at least in it's intended change, maybe something unintended is caused by this PR which we should rather fix?

What problem are you running into that is different to the previous version?

soren121:
I access `editor.options.element` in a few places in my editor implementation. There are some instances where I want to access the editor container element instead of the ProseMirror view (`editor.view.dom`.) Updating to 3.6.0 caused a handful of errors in TypeScript.

I can change this on my end without too much hassle, but nonetheless I would not expect incompatible changes to public API types on a semver-minor update.

nperez0111:
Would disagree that just any external types changes should be considered breaking. Just because the options happened to be exposed does not mean that they are yours to rely on, they are the editor's options after-all?

soren121:
I'm not sure how I am to know what I can and cannot rely on then, when an interface is publicly exposed & documented.

At any rate, it's not my intention to argue about this. I will fix my code and consider `editor.options` as internal going forward.

bdbch:
In the future I'll take a closer look and think about the implications of a change a bit more before releasing. Thanks for bringing it up @soren121.

mejo-:
Turns out this PR introduced a bug, see https://github.com/ueberdosis/tiptap/issues/7282#issuecomment-3665907731

PR #6975, opened 2025-09-17
Unique ID utility

arnaugomez:
## Changes Overview

Creates a utility to add unique IDs to a document in the server

The utility is called `generateUniqueIds` and is exported from the `@tiptap/extension-unique-id` package.

It has the same functionality as the `UniqueID` extension, but without the need to create an `Editor` instance. This lets you add unique IDs to the document in the server.

It takes the following parameters:

- `doc`: The Tiptap JSON document to add unique IDs to.
- `extensions`: The extensions to use. Must include the `UniqueID` extension.

It returns the updated Tiptap JSON document, with the unique IDs added to the nodes.

## Checklist

### DIFF
--- a/packages/extension-unique-id/src/generate-unique-ids.ts
+++ b/packages/extension-unique-id/src/generate-unique-ids.ts
@@ -1,0 +1,71 @@
+import type { Extensions, JSONContent } from '@tiptap/core'
+import { findChildren, getSchema } from '@tiptap/core'
+import { Node } from '@tiptap/pm/model'
+import { EditorState } from '@tiptap/pm/state'
+
+import type { UniqueID } from './unique-id.js'
+
+/**
+ * Creates a new document with unique IDs added to the nodes. Does the same
+ * thing as the UniqueID extension, but without the need to create an `Editor`
+ * instance. This lets you add unique IDs to the document in the server.
+ *
+ * When you call it, include the `UniqueID` extension in the `extensions` array.
+ * The configuration from the `UniqueID` extension will be picked up
+ * automatically, including its configuration options like `types` and
+ * `attributeName`.
+ *
+ * @see `UniqueID` extension for more information.
+ *
+ * @throws {Error} If the `UniqueID` extension is not found in the extensions array.
+ *
+ * @example
+ * const doc = {
+ *   type: 'doc',
+ *   content: [
+ *     { type: 'paragraph', content: [{ type: 'text', text: 'Hello, world!' }] }
+ *   ]
+ * }
+ * const newDoc = addUniqueIds(doc, [StarterKit, UniqueID.configure({ types: ['paragraph', 'heading'] })])
+ * console.log(newDoc)
+ * // Result:
+ * // {
+ * //   type: 'doc',
+ * //   content: [
+ * //     { type: 'paragraph', content: [{ type: 'text', text: 'Hello, world!' }], id: '123' }
+ * //   ]
+ * // }
+ *
+ * @param doc - A Tiptap JSON document to add unique IDs to.
+ * @param extensions - The extensions to use. Must include the `UniqueID` extension.
+ * @returns The updated Tiptap JSON document, with the unique IDs added to the nodes.
+ */
+export function generateUniqueIds(doc: JSONContent, extensions: Extensions): JSONContent {
+  // Find the UniqueID extension in the extensions array. If it's not found, throw an error.
+  const uniqueIDExtension = extensions.find(ext => ext.name === 'uniqueID') as typeof UniqueID | undefined
+  if (!uniqueIDExtension) {
+    throw new Error('UniqueID extension not found in the extensions array')
+  }
+  const { types, attributeName, generateID } = uniqueIDExtension.options
+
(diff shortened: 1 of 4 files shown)


packages/extension-unique-id/src/generate-unique-ids.ts
nperez0111, line 1: Shouldn't this be a Tiptap command, and then you call the command on the server-side

alexvcasillas (dismissed):

packages/extension-unique-id/src/generate-unique-ids.ts
arnaugomez: The UniqueID extension automatically adds unique IDs to the document when the editor is created (see the `onCreate` method), so a command for that would not be necessary. 

However, this `generateUniqueIds` utility is meant to be run in the server without having to instantiate the Tiptap editor. Sort of like the HTML utility. I've created it because it helps a customer add unique ids to a document after it's generated in the server.

nperez0111: Right, but the editor _can_ run on the server, so by exposing it as a command rather than a one-off function (which could live on that customer's server), you would at least make it usable to both the server & client sides.
Especially, if they are going to do other data transformations which is likely.

e.g:

```js
const editor = new Editor({ element: null, content });

editor.chain().generateUniqueIds().somethingElse().run();
```

Having one-off functions is kind of antithetical to the architecture.

arnaugomez:
~~I'm waiting for feedback from a customer who is trying this code, then I'll merge it. I also need to add documentation before I merge.~~

Customer has tried the code and is happy with it, the PR can be merged.

packages/extension-unique-id/src/generate-unique-ids.ts
arnaugomez: I'm curious, what is your take on Tiptap's roadmap for reading/editing documents on the server? At the moment, it's true that the Editor can be instantiated on the server (provided you initialize it with the correct configuration). However, many extensions and capabilities are not supported outside the browser: collaboration, HTML parsing, etc.

What is your preferred approach moving forward? Adapt the Editor instance's capabilities so that they work on the server the same way they work on the client, or create specific lightweight utilities for manipulating content on the server?

Take into account that the former would take a big re-work of the Editor to make it work in different environments, while the latter is easier to implement and potentially less memory consuming, as it does not need to hold the Editor instance in-memory, and compute all side effects, events and transactions that happen in the client-side. Of course I'm biased here, I have my opinion, and I want to know if there are any disadvantages in the approach I'm taking to server-side editing, that I haven't thought out.

nperez0111: Good question!

My thoughts are that ideally the editor would work seamlessly on the server, because most of it actually is not dependent on client-side specific features except for HTML rendering & parsing. Transforming to html is handled by either the static-renderer (which is a simple mapping of a ProseMirror DomOutputSpec array to an html string, therefore _very lightweight_) or the existing `@tiptap/html/server` utility based on happy-dom. For HTML parsing, I'd say that there are two solutions A) don't do it, you should be using JSON B) also include something like happy-dom to stub any of that functionality on a server if you really must.

You make it sound like running the editor on the server is somehow heavyweight? I can assure you that if you use JSON as your input & output, that there would be no performance difference between the two. This is what happens when you do not mount the editor, it does not create an editor.view, and then at that point any manipulation you make is purely JavaScript objects (ProseMirror Nodes & Marks) [see here for reference](https://github.com/ueberdosis/tiptap/blob/1ea8906ba1927a7f219f46d4b2cb419782582843/packages/core/src/Editor.ts#L291-L329). ProseMirror was actually designed to be de-coupled for this reason, this is why setting up ProseMirror can be a bit tedious, but it is a powerful abstraction, all state changes (e.g. transactions, serialization & deserialization to/from JSON) can be done in pure JS, without access to a browser environment (e.g. document). And, what you gain here is the ability to express your transformation in terms of the familiar Tiptap API rather than some other bespoke API

You mention that collaboration is not supported outside of the browser, this just isn't the case? I mean hocuspocus & the collaboration server do just that on the server, if anything, it is trivial to implement on the server. Somewhere in Notion, you'll even find an example of me doing this.

To re-implement everything that the editor can do as a bunch of one-off functions you are doing 2 things: making more code to maintain( creating more surface area for bugs), and making it more difficult to work with Tiptap, by offering different APIs for each use case you happen upon along the way, rather than refining your current APIs.

So, all-in-all, you've presented a false dichotomy, there is little work required to make the editor work on the server. 

- If you need to convert to/from HTML, use the static renderer or `@tiptap/html/server` packages
- If you need collaboration, read & write to the ydoc directly (using y-prosemirror's built-ins). Here, I can admit some nicer abstraction of this as an extension would be nice to consumers, but it doesn't make it not possible.

arnaugomez: Thanks for your thoughts! Hope this can help us align our roadmap and figure out a way to make server-side content editing easier for developers.

alexvcasillas (approved):

PR #7011, opened 2025-09-25
fix: parse cell colwidth from nearest <colgroup> when missing cells

bdbch:
## Changes Overview

Parse a table cell's `colwidth` from the table's nearest `<colgroup>` when the `td` itself doesn't have a `colwidth` attribute. This preserves column width information when importing HTML that declares widths on `<colgroup>` / `<col>` instead of on each cell.

Files touched
- table-cell.ts — add fallback logic in the `colwidth` attribute parser.
- parse-colgroup-colwidth.md — new changeset (patch release for `@tiptap/extension-table`).

## Implementation Approach

- Keep original behavior (read `colwidth` attribute on the `<td>`).
- If the cell has no `colwidth`, find the enclosing `<table>`, query its `colgroup > col` elements, compute the cell index within its row, and read the corresponding `<col>` `width` attribute.
- Convert the `width` string to a number and return it as a single-item array (matching the existing `colwidth` shape), or `null` when not available.
- No public API changes; behaviour is additive and intended as a non-breaking bugfix.

## Testing Done

- Created a minimal HTML sample (see Verification Steps) with widths set only on `<colgroup>` and verified the parser code path returns the expected `colwidth` value.
- Verified a new changeset file was added at parse-colgroup-colwidth.md.

## Verification Steps

1. Create this sample HTML (or use your demo) and import/parse it with the editor:
   ```html
   <table>
     <colgroup>
       <col width="120">
       <col width="180">
     </colgroup>
     <tbody>
       <tr>
         <td>Cell A</td>
         <td>Cell B</td>
       </tr>
     </tbody>
   </table>
   ```
2. Start the demo app (if you use the repo demos):
   ```bash
   pnpm dev
   ```
3. Paste/import the HTML into a Tiptap instance that uses `@tiptap/extension-table` and inspect the node attributes for the first `tableCell` node. Expected:
   - The first `tableCell` should have attribute `colwidth: [120]`.
   - The second `tableCell` should have attribute `colwidth: [180]`.
4. Optionally add an automated test asserting that parsing the above HTML sets `colwidth` as expected.

## Additional Notes

- This is a patch (bugfix) — no breaking API changes.

## Checklist

- [x] I have followed the project guidelines (changeset added, small scope).

## Related Issues

- None linked here; this addresses lost column-width information when HTML uses `<colgroup>` instead of per-cell `colwidth`.

TL;DR: Fallback parsing added so `tableCell` reads column widths from the table's `<colgroup>` when a `td` lacks `colwidth`; changeset added for a patch release.

### DIFF
--- a/packages/extension-table/src/cell/table-cell.ts
+++ b/packages/extension-table/src/cell/table-cell.ts
@@ -40,6 +40,17 @@ export const TableCell = Node.create<TableCellOptions>({
           const colwidth = element.getAttribute('colwidth')
           const value = colwidth ? colwidth.split(',').map(width => parseInt(width, 10)) : null
 
+          // if there is no colwidth attribute on the cell, try to get it from the colgroup
+          if (!value) {
+            const cols = element.closest('table')?.querySelectorAll('colgroup > col')
+            const cellIndex = Array.from(element.parentElement?.children || []).indexOf(element)
+
+            if (cols && cols[cellIndex]) {
+              const colWidth = cols[cellIndex].getAttribute('width')
+              return colWidth ? [parseInt(colWidth, 10)] : null
+            }
+          }
+
           return value
         },
       },
--- a/.changeset/parse-colgroup-colwidth.md
+++ b/.changeset/parse-colgroup-colwidth.md
@@ -1,0 +1,9 @@
+---
+"@tiptap/extension-table": patch
+---
+
+Parse cell `colwidth` from nearest `<colgroup>` when missing on the cell
+
+When importing HTML, table column widths are often declared on a surrounding `<colgroup>` rather than on each `<td>`. Previously, `tableCell` only read the `colwidth` attribute from the cell itself and would lose width information in that case. The implementation now falls back to reading the corresponding `<col>`'s `width` from the table's `<colgroup>` using the cell's index.
+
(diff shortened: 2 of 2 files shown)


packages/extension-table/src/cell/table-cell.ts
```diff
@@ -40,6 +40,17 @@ export const TableCell = Node.create<TableCellOptions>({
           const colwidth = element.getAttribute('colwidth')
           const value = colwidth ? colwidth.split(',').map(width => parseInt(width, 10)) : null
 
+          // if there is no colwidth attribute on the cell, try to get it from the colgroup
+          if (!value) {
+            const cols = element.closest('table')?.querySelectorAll('colgroup > col')
+            const cellIndex = Array.from(element.parentElement?.children || []).indexOf(element)
```
arnaugomez, line 46: Can it be the case that the cellIndex is null or -1? (for example, when the array is []). In the next line, you access `cols[cellIndex]` so I wonder what would happen if it is null.

arnaugomez (changes requested):

packages/extension-table/src/cell/table-cell.ts
bdbch: fixed

PR #7024, opened 2025-09-28
fix: allow nullable types for configuration options

royce-mathew:
## Changes Overview

- Updated `CodeBlockOptions` interface to allow **nullable and undefined types** for configuration options.  
- This ensures flexibility when extending a `CodeBlock`, since default values exist and strict typing is not always necessary.  

## Implementation Approach

- Changed option types from strict primitives (`string`, `boolean`, `number`) to union types (`string | null | undefined`, etc.).  
- Preserved default values in comments to clarify intended behavior.  
- Ensured type safety while supporting optional overrides.  

## Testing Done

- Type-checked the updated interface in multiple scenarios:
  - With full config provided.  
  - With missing or `null` values.  
  - With only partial overrides.  
- Verified that defaults still apply when `null` or `undefined` are used.  

## Verification Steps

1. Extend `CodeBlockOptions` with partial or missing fields.  
2. Ensure code compiles without type errors.  
3. Confirm runtime defaults are applied correctly when fields are `null` or `undefined`.  

## Additional Notes

- This change improves **developer experience** by making the config more forgiving.  
- No breaking changes introduced.  

## Checklist

- [x] Nullable type support added for config options.  
- [x] Defaults preserved and documented.  
- [x] Verified no breaking changes.  
- [ ] Added tests (if necessary for config fallback logic).  
- [ ] Fixed any lint issues.  

## Related Issues

- Fix: allow nullable types for configuration options in `CodeBlockOptions`.

### DIFF
--- a/packages/extension-code-block/src/code-block.ts
+++ b/packages/extension-code-block/src/code-block.ts
@@ -6,17 +6,17 @@ export interface CodeBlockOptions {
    * Adds a prefix to language classes that are applied to code tags.
    * @default 'language-'
    */
-  languageClassPrefix: string
+  languageClassPrefix: string | null | undefined
   /**
    * Define whether the node should be exited on triple enter.
    * @default true
    */
-  exitOnTripleEnter: boolean
+  exitOnTripleEnter: boolean | null | undefined
   /**
    * Define whether the node should be exited on arrow down if there is no node after it.
    * @default true
    */
-  exitOnArrowDown: boolean
+  exitOnArrowDown: boolean  | null | undefined
   /**
    * The default language.
    * @default null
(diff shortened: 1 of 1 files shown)


packages/extension-code-block/src/code-block.ts
```diff
@@ -248,7 +253,7 @@ export const CodeBlock = Node.create<CodeBlockOptions>({
 
             const currentLine = lines[currentLineIndex]
             const leadingSpaces = currentLine.match(/^ */)?.[0] || ''
-            const spacesToRemove = Math.min(leadingSpaces.length, this.options.tabSize)
+            const spacesToRemove = Math.min(leadingSpaces.length, this.options.tabSize || 4)
```
bdbch: Just a nitpick - I don't like how we re-use the `4` hardcoded here.

Maybe it makes sense to move this out into a constant `const DEFAULT_TAB_SIZE = 4` and use it in `addOptions` and all other occurences.

You could also do something like this:

```
const { tabSize = DEFAULT_TAB_SIZE } = this.options
```

a few lines before you use the value and then just use `tabSize`

bdbch (changes requested):

Issue #6963, opened 2025-09-16
react BubbleMenu @tiptap/react/menus doesn't update on props change


TeemuKoivisto:
### Affected Packages

react

### Version(s)

3.4.3

### Bug Description

I am trying to pass `options` to floating-ui/dom, mainly for setting `scrollTarget`. I am retrieving the element in an useEffect hook and setting the options but after debugging for a while, I realized the options within `<BubbleMenu options={options}` is never updated, thus this listener `this.scrollTarget.addEventListener('scroll', this.resizeHandler)` is never triggered.

I am bybassing this bug by rendering the BubbleMenu conditionally with:

```tsx
  if (!options.scrollTarget) {
    return null
  }
  ...
```

### Browser Used

Chrome

### Code Example URL

https://codesandbox.io/p/sandbox/silly-fast-vclgjc

### Expected Behavior

Update BubbleMenu props automatically

### Dependency Updates

- [x] Yes, I've updated all my dependencies.

chase:
Ran into similar issues today.

From a glance, it seems to stem from the fact that options aren't updated after construction and the original options object isn't used, so any updates will only ever use the original due to this:

https://github.com/ueberdosis/tiptap/blob/develop/packages/extension-bubble-menu/src/bubble-menu-plugin.ts#L346-L349

I patched it so that the setMeta takes an object with a type and an optional `options` object that's `Object.assign`'d onto the internal options.

AdeboyeDN:
I will take a look at this and come up with a pr.

arnaugomez:
@AdeboyeDN your PR introduced a regression so I had to revert it ([More info on that here](https://github.com/ueberdosis/tiptap/pull/7028#issuecomment-3441663397)).

I re-open the issue.

I'm still in favour of providing an API that lets developers update the BubbleMenu. But be careful to not introduce breaking changes (and, if you do, merge your PR in the `next` branch). Also be mindful that, with the solution you provided, if a props is a function/object and it's not memoized, then the BubbleMenu plugin would re-register on every re-render of the React component.

bdbch:
I opened a PR that should fix this. Instead of reinitializing the whole plugin which leads to heavy regression the update now works via dispatching a transaction. while this will still be run when options are not memoized, it should not have heavy regressions like the previous approach as transactions that don't change the document and just pass metadata around are very fast.

PR #7028, opened 2025-09-29
fix(react BubbleMenu): BubbleMenu now reacts to prop changes

AdeboyeDN:
## Changes Overview

Fixed an issue where the React BubbleMenu didn’t pick up prop changes after initial mount. In particular, updates to options according to issue #6963

## Implementation Approach

The effect that registers the BubbleMenu plugin now includes all relevant configuration in its dependency list.

## Checklist

### DIFF
--- a/.changeset/fix-bubblemenu.md
+++ b/.changeset/fix-bubblemenu.md
@@ -1,0 +1,5 @@
+---
+"@tiptap/react": patch
+---
+
+Fix BubbleMenu so it now reacts to prop changes after initial mount
--- a/.changeset/fix-bubblemenu.md
+++ b/.changeset/fix-bubblemenu.md
@@ -2,4 +2,4 @@
 "@tiptap/react": patch
 ---
 
-Fix BubbleMenu so it now reacts to prop changes after initial mount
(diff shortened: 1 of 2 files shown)


.changeset/fix-bubblemenu.md
```diff
@@ -0,0 +1,5 @@
+---
+"@tiptap/react": patch
+---
+
+Fix BubbleMenu so it now reacts to prop changes after initial mount
```
bdbch:
```suggestion
Resolved an issue where the React BubbleMenu did not update when FloatingUI option props changed after initial mount. The BubbleMenu now correctly responds to updated option props.
```

bdbch (approved):

arnaugomez:
This is a breaking change and it caused a bug in one of our customers' code.

It's a breaking change because, if one of these options (appendTo, shouldShow, getReferencedVirtualElement, options) is not memoized, the bubble menu plugin is unregistered and registered again every time the component re-renders. This caused a lot of transactions and a bug where the editor selection was unset.

I'll have to rollback because this is a regression. I'm not against including the ability to re-load the options, but it should be made in a way that does not break the existing library.

PR #7060, opened 2025-10-09
Fix BubbleMenu and FloatingMenu's not updating/rendering correctly on first mount & race condition with Vue

bdbch:
## Changes Overview

- Fixes a bug where BubbleMenu / FloatingMenu sometimes mounted without correct positioning when `shouldShow` is `true` or when menus are conditionally rendered.
- Simplifies Vue menu components by using a direct DOM element that the plugin can re-parent instead of relying on the Teleport target string.
- Ensures menu position is updated immediately after being shown by calling `updatePosition()`.
- Makes `getShouldShow()` return a safe boolean for the BubbleMenu plugin.
- Adds changesets for a patch release.

## Implementation Approach

- Create a dedicated DOM element (`el = document.createElement('div')`) in the Vue component and use Teleport to render the menu content into that element.
- Remove the element from the DOM immediately so the plugin can re-parent it when showing the menu.
- Pass that direct element to the plugin so the plugin controls placement.
- After showing the menu (when `getShouldShow()` is true), call `updatePosition()` to compute and apply correct coordinates immediately.
- Return `shouldShow || false` from BubbleMenu's `getShouldShow()` to avoid nullable/undefined behavior.
- Use `nextTick()` in BubbleMenu registration to avoid a timing issue when Teleport is resolving children.

Why this helps
- Direct element + plugin re-parenting avoids race conditions where Vue/Teleport and the plugin disagree about where the DOM node lives.
- Calling `updatePosition()` immediately ensures correct placement when a menu is shown on creation.
- Returning a boolean avoids falsy/null edge-cases that prevented code paths from running.

## Testing Done

- Manual testing in the local dev demo:
  - Started demos (`pnpm dev`) and opened the Vue demo pages for BubbleMenu and FloatingMenu.
  - Verified menus are attached to the DOM and positioned correctly when:
    - The menu should be shown immediately on creation (shouldShow true).
    - The menu is conditionally rendered via parent state and toggled on/off.
  - Verified no console errors during mount/unmount and that plugins unregister on component unmount.
- Added changesets included for release notes.

## Verification Steps

1. Run the demo app:
   - pnpm dev
   - Open http://localhost:3000 and navigate to the Vue demo pages for FloatingMenu and BubbleMenu.
2. Reproduce scenarios:
   - Use examples where `shouldShow` returns true on mount. Confirm the menu appears and is positioned correctly.
   - Toggle conditional rendering / `shouldShow` and verify position updates and that the menu attaches to the DOM correctly.
3. Run repository checks:
   - pnpm lint
   - pnpm build
   - pnpm test (or pnpm test:open for cypress)
4. Inspect console for errors and confirm no orphaned DOM nodes remain after unmount.

## Additional Notes

- This is a non-breaking internal change; public API for the plugin/extension is unchanged.
- The change addresses two timing/race issues:
  - menu position not updated on creation when `shouldShow` is true (fixed by `updatePosition()`).
  - conditionally rendered BubbleMenu not attached correctly due to Teleport timing (fixed by using direct element + `nextTick()`).
- Changesets added for patch release: .changeset/*

## Checklist

- [x] My changes do not break the library. (please run CI)

## Related Issues

- Fixes internal positioning / rendering race conditions for BubbleMenu / FloatingMenu (no linked issue provided).

TL;DR: Use a direct DOM element + plugin-managed re-parenting and call updatePosition() after showing so floating/bubble menus mount with correct coordinates even when they are shown immediately on creation.

### DIFF
--- a/packages/vue-3/src/menus/BubbleMenu.ts
+++ b/packages/vue-3/src/menus/BubbleMenu.ts
@@ -95,6 +95,6 @@ export const BubbleMenu = defineComponent({
     })
 
     // Teleport only instantiates element + slot subtree; plugin controls final placement
-    return () => h(Teleport, { to: el }, h('div', attrs, slots.default?.()))
+    return () => h(Teleport, { to: el }, h('div', { ...attrs, ref: 'root' }, slots.default?.()))
   },
 })


felixgabler:
Unfortunately, this breaks the attributes inheritance so it is not possible anymore to specify things like the z-index, which is needed for positioning.

```html
<BubbleMenu :editor="editor" :append-to="body" class="z-50">
```

The old way of letting Vue fully own the element seemed better to me. To keep it simple, we could actually rather think about removing the Teleport. Why are you using that in the first place?

PR #7101, opened 2025-10-20
Remove unnecessary Teleport in menus and fix attribute forwarding

felixgabler:
## Changes Overview

A recent PR broke attribute forwarding for Vue 3 BubbleMenu and FloatingMenu components (see https://github.com/ueberdosis/tiptap/pull/7060#issuecomment-3422590191). Users can no longer set `class`, `style`, or other HTML attributes to control z-index and positioning. This PR fixes attribute forwarding by simplifying the implementation. Vue now owns the element directly via a `ref` again instead of using `Teleport` with a raw DOM element.

## Implementation Approach

- Removed the detached `document.createElement('div')` + `Teleport` approach
- Let Vue manage the element via `ref` and spread `attrs` on the root element
- Vue handles attribute reactivity automatically; the plugin still re-parents the element when showing

## Testing Done

- Manually tested in a local Bun + Vue 3 app with linked packages
- Verified `class` and `style` attributes are applied to the menu element
- Tested `appendTo` works correctly for fixing clipping issues
- Confirmed menus appear with correct z-index when set via `:style="{ zIndex: 9999 }"`
- No console errors on mount/unmount

## Verification Steps

1. Use BubbleMenu or FloatingMenu with attributes:
   ```vue
   <BubbleMenu :editor="editor" :append-to="document.body" :style="{ zIndex: 9999 }">
     <button>Bold</button>
   </BubbleMenu>
   ```
2. Wrap editor in a parent with `overflow: hidden`
3. Select text (BubbleMenu) or place cursor in empty paragraph (FloatingMenu)
4. Verify menu appears with the correct class and z-index applied
5. Check browser DevTools to confirm attributes are on the menu element

## Additional Notes

- No breaking changes; internal implementation detail only
- Simpler and more "Vue-ish" than the previous approach
- Attributes are now reactive if they change

## Checklist

- [ ] I have added tests where applicable (manual testing done).

## Related Issues

Fixes https://github.com/ueberdosis/tiptap/pull/7060#issuecomment-3422590191

### DIFF
--- a/.changeset/bubble-floating-menu-attrs.md
+++ b/.changeset/bubble-floating-menu-attrs.md
@@ -1,0 +1,5 @@
+---
+"@tiptap/vue-3": patch
+---
+
+Fix attribute forwarding for BubbleMenu and FloatingMenu Vue 3 components to allow setting z-index and other HTML attributes
--- a/packages/vue-3/src/menus/BubbleMenu.ts
+++ b/packages/vue-3/src/menus/BubbleMenu.ts
@@ -1,7 +1,7 @@
 import type { BubbleMenuPluginProps } from '@tiptap/extension-bubble-menu'
 import { BubbleMenuPlugin } from '@tiptap/extension-bubble-menu'
 import type { PropType } from 'vue'
-import { defineComponent, h, nextTick, onBeforeUnmount, onMounted, Teleport } from 'vue'
+import { defineComponent, h, nextTick, onBeforeUnmount, onMounted, ref } from 'vue'
 
 export const BubbleMenu = defineComponent({
   name: 'BubbleMenu',
@@ -51,7 +51,7 @@ export const BubbleMenu = defineComponent({
   },
 
   setup(props, { slots, attrs }) {
-    const el = document.createElement('div')
+    const root = ref<HTMLElement | null>(null)
 
     onMounted(() => {
       const {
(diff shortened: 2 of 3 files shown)


bdbch:
Before we merge this, can you verify that without the Portal the DOM element is inserted at the right position inside the DOM? (Not at the editor element but on the body so we don't have to fight with overflow/parent-child issues).

I didn't do the initial implementation & also am not that experienced with Vue anymore (at least after Vue 2) so I can just make guesses why the Portal was used here.

felixgabler:
> Before we merge this, can you verify that without the Portal the DOM element is inserted at the right position inside the DOM? (Not at the editor element but on the body so we don't have to fight with overflow/parent-child issues). 

Yes, when `:append-to="body"` is provided, the `BubbleMenuPlugin` appends it to the `body` and afterwards, FloatingUI manages it. I've tested it locally and it correctly appends to `body` like before but with the benefit that `attrs` are passed through again allowing for z-index changes, which are also critical for the overflow/parent-child issues.

image

I also did not see any other issues during my testing.

> I didn't do the initial implementation & also am not that experienced with Vue anymore (at least after Vue 2) so I can just make guesses why the Portal was used here.

As far as I understand it, the `Teleport` previously would attach the element to the body but the element itself would then immediately be removed from the DOM on mount and reattached wherever `appendTo` specified or at the editor's parent. That would leave and empty `div` in a `Teleport` in the DOM. Since this was done before my `appendTo` changes, I can only imagine that it was meant to combat the stacking context issues, but sadly not in the correct way.

bdbch (approved):

PR #7113, opened 2025-10-23
Fix/revert whitespace stripping

bdbch:
## Changes Overview

Removed the internal newline-only whitespace stripping from `elementFromString` in `@tiptap/core`. The function now returns the parsed DOM from `DOMParser` without recursively removing newline-only text nodes. This preserves newline-only whitespace nodes that were previously stripped.

## Implementation Approach

- Deleted the recursive `removeWhitespaces` helper and returned `new window.DOMParser().parseFromString(wrappedValue, 'text/html').body` directly from `elementFromString`.
- Kept a small changeset documenting the behavior change and added a short migration hint recommending `preserveWhitespace` when inserting content via editor commands.

## Testing Done

- Verified the change by inspecting the modified file and ensuring the helper was removed.
- Created a changeset remove-whitespace-elementfromstring.md explaining the behavior change and noting `editor.commands.insertContent(dom, { preserveWhitespace: true })` as a mitigation.

## Verification Steps

1. Build the packages (optional but recommended):
   - Run: pnpm build
2. Manual verification in Node environment:
   - In a browser-like environment (or demo), call:
     - `const dom = elementFromString('<div>\\n<p>text</p>\\n</div>')`
     - Inspect `dom.childNodes` — newline-only text nodes should now be present instead of removed.
3. Insert into an editor:
   - Use `editor.commands.insertContent(dom, { preserveWhitespace: true })` to preserve whitespace behavior when inserting parsed DOM into the editor.
4. Run or add tests that assert:
   - Previously failing input no longer produces unexpected extra newlines in the final editor output.
   - If consumers relied on the old stripping, the suggested `preserveWhitespace` flow restores the old outcome.

## Additional Notes

- The change may affect consumers who relied on the prior whitespace-stripping behavior. The changeset includes a short migration snippet showing how to strip whitespace nodes manually or use `preserveWhitespace` during `insertContent`.
- If this reintroduces previously fixed issues, we should either:
  - Revert the removal and rework the helper to only remove truly irrelevant whitespace, or
  - Provide a compatibility helper and document the change more prominently in CHANGELOG.md.

## Checklist

- [x] I have fixed any lint issues (no lint changes were required for this small edit).

## Related Issues

- See prior changelog notes referencing issue about new lines: (#4767) the changeset references similar previous fixes in the repository history.
- See also #4792 which describes a bug caused by the reverted code
- See #2720 which was the initial issue report for the attempted fix from #4767


arnaugomez (approved):

bdbch:
I thought about this PR again and I think we may have to postpone it. I'll close it for now because I think it will break content parsing for a lot of users.

I'm a bit annoyed by how it's setup now. We will probably fix this with a new major & write proper tests for all expected cases. but for now I'm just too worried this PR will break things for a lot of users.

PR #7139, opened 2025-10-29
fix(drag-handle): resolve parent when nodeAt(pos) is null for atom nodes with inline content

Aslam97:
## Changes Overview

Fix `findElementNextToCoords` so that when it maps a top-level DOM block to a ProseMirror position and `state.doc.nodeAt(pos)` returns `null` (for example, when the position falls inside an atom node that allows inline content), it falls back to the containing parent node instead. This ensures the drag handle targets the correct block element in those cases.

## Implementation Details

- In `findElementNextToCoords`, handle the case where `state.doc.nodeAt(pos)` is `null`.
- Resolve the position via `state.doc.resolve(pos)` and derive the parent node from the resolved position.
- Return that parent as `resultNode` and use `resolvedPos.start()` as the position for the block.

## Testing Done

Manually verified drag-handle behavior on **Notion-like** template

## Verification Steps

1. Create an atom node which allows inline content inside it.
2. Try dragging that block using the drag handle.
3. Confirm the behavior:
   - If the block contains regular content, dragging works as expected.
   - If the atom node is empty (no inline content), dragging failed to move the node — this fix ensures the correct parent block is targeted so dragging works in both cases.

## Additional Notes

https://github.com/user-attachments/assets/eb28f0b3-1255-41fa-9368-fde4efb96120

## Checklist

* [x] I have created a [[changeset](https://github.com/changesets/changesets)](https://github.com/changesets/changesets) for this PR if necessary.
* [x] My changes do not break the library.
* [x] I have added tests where applicable.
* [x] I have followed the project guidelines.
* [x] I have fixed any lint issues.

## Related Issues

#7328


arnaugomez:
Review the "Checklist" section of the PR description

Aslam97:
Reviewed the checklist

alexvcasillas (approved):

bdbch (approved):

PR #7162, opened 2025-11-03
fix(markdown): safely handle empty markdown content (fixes #7157)

### LINKED ISSUE #7157: Vue, Markdown - Empty string as content breaks EditorContent component
### Affected Packages

core, markdown

### Version(s)

3.10.1

### Bug Description

When using tiptap with markdown and set the initial content to an empty string, the EditorContent component doesn't work anymore. You cannot write into the field. Right now you have to set some initial text content to make the component work.

### Browser Used

Chrome

### Code Example URL

https://stackblitz.com/edit/vitejs-vite-nbmyhbrl?file=src%2Fcomponents%2FTiptapEditor.vue

### Expected Behavior

The EditorContent component works with markdown and an initial empty string as content.

In the `App.vue` you have the `editorContent` ref, that I use for model binding. The initial value is an empty string. When you set some content here, the `EditorContent` component works. Multiple empty characters dont work either, so I assume, it could have something to do with trimming content.

### Dependency Updates

likhith5697:
### Summary
Fixes a bug where initializing the Markdown extension with empty content (`''`) made the editor non-editable.

### Root Cause
The Markdown parser produced a document containing an empty text node, leading to `RangeError: Empty text nodes are not allowed` and preventing typing or focus.

### Fix
Added a guard in `MarkdownManager.parse()` to return a valid empty paragraph when markdown is empty or whitespace-only.

### Verification
Tested in `demos/src/Markdown/Full/Vue/EmptyContent/`  
– Editor loads with empty content  
– Typing works immediately  
– No console errors

### DIFF
--- a/.changeset/grumpy-forks-repeat.md
+++ b/.changeset/grumpy-forks-repeat.md
@@ -1,0 +1,5 @@
+---
+"@tiptap/markdown": patch
+---
+
+Fixes a bug where initializing the Markdown extension with empty content made the editor non-editable.


likhith5697:
Verified the deploy preview:
- Editor initializes with empty markdown (`''`)
- Typing works immediately
- Cursor and focus behave normally
- No console or schema errors

Ready for review :)

packages/markdown/src/index.ts
```diff
@@ -1,3 +1,7 @@
-export * from './Extension.js'
+import { Markdown } from './Extension.js'
+
+export { Markdown }
```
bdbch: Are those changes necessary?

packages/markdown/src/MarkdownManager.ts
```diff
@@ -267,21 +267,58 @@ export class MarkdownManager {
   /**
    * Parse markdown string into Tiptap JSON document using registered extension handlers.
    */
+  /**
```
bdbch: I think you duplicated the base JSDoc comment. Can you remove the old one?

packages/markdown/src/MarkdownManager.ts
```diff
@@ -267,21 +267,58 @@ export class MarkdownManager {
...
+   * Parse markdown string into Tiptap JSON document using registered extension handlers.
+   * Includes a fix for empty or invalid markdown content (#7157).
```
bdbch: Don't include this here. Not important for users.

packages/markdown/src/MarkdownManager.ts
```diff
@@ -267,21 +267,58 @@ export class MarkdownManager {
...
+   */
+  /**
+   * Parse markdown string into Tiptap JSON document using registered extension handlers.
+   * Includes a fix for empty or invalid markdown content (#7157).
```
bdbch: Same here. Also again duplicated JSDoc comments.

packages/markdown/src/MarkdownManager.ts
```diff
@@ -267,21 +267,58 @@ export class MarkdownManager {
...
+   */
   parse(markdown: string): JSONContent {
     if (!this.hasMarked()) {
       throw new Error('No marked instance available for parsing')
     }
 
-    // Use marked to tokenize the markdown
-    const tokens = this.markedInstance.lexer(markdown)
+    if (!markdown || markdown.trim() === '') {
+      return {
```
bdbch: Please use the `createDocument` utility so we follow the schema to create empty documents. Some users may not use the paragraph as the default content type.

packages/markdown/src/MarkdownManager.ts
```diff
@@ -267,21 +267,58 @@ export class MarkdownManager {
...
-      type: 'doc',
-      content,
+      return {
+        type: 'doc',
+        content,
+      }
+    } catch (err) {
+      console.warn('Markdown parse failed, using fallback empty document', err)
+      return {
+        type: 'doc',
```
bdbch: See above, same here.

likhith5697: > See above, same here.

Updated this block as well to use createDocument(content, schema).toJSON() for schema-safe document creation, matching the earlier fix.
Tested locally — works as expected with empty and parsed markdown.

likhith5697:
**Update Summary**

- Replaced manual doc creation with createDocument(content, schema).toJSON() for schema-safe initialization.
- Ensures compatibility with non-paragraph default nodes and prevents RangeError for empty markdown.
- Verified locally in Vue demo — editor loads, typing works, and no console errors.
- All requested fixes have been applied in the latest commit (a738509).

Thanks for the detailed review!

packages/markdown/src/MarkdownManager.ts
likhith5697: > I think you duplicated the base JSDoc comment. Can you remove the old one?

Done

packages/markdown/src/MarkdownManager.ts
likhith5697: > Don't include this here. Not important for users.

Done

packages/markdown/src/MarkdownManager.ts
likhith5697: > Same here. Also again duplicated JSDoc comments.

Done

packages/markdown/src/MarkdownManager.ts
likhith5697: > Please use the `createDocument` utility so we follow the schema to create empty documents. Some users may not use the paragraph as the default content type.

Done

packages/markdown/src/index.ts
likhith5697: > Are those changes necessary?

Thanks for catching that! I’ve reverted index.ts to only use direct exports — it now simply re-exports from Extension.js, MarkdownManager.js, and utils.js as before.

tushar-innovalabs:
@likhith5697 @bdbch Has there been any update on the release of this fix? We are facing the same issue and are stuck?

bdbch (dismissed):

bdbch:
Mh, seems like the unit tests are failing here because createDocument doesn't have all necessary extensions.

bdbch:
ping @likhith5697 unit tests are failing right now - could you take a look?

bdbch:
Closing this for now as it became stale.

PR #7165, opened 2025-11-04
feat: enhance table extension to respect user-defined styles for widths

bdbch:
## Changes Overview

Fixed a bug where the Table extension always set a fixed width (calculated as `cellMinWidth` × number of columns) instead of respecting user-provided `style` attributes, such as custom widths like `width: 100%`.

## Implementation Approach

- Added a `style` attribute to the Table node via `addAttributes()` to allow users to set custom styles.
- Modified `renderHTML()` in `table.ts` to check for user-provided styles and use them instead of calculated widths.
- Updated `TableView.ts` to apply user styles to the DOM element and prevent overriding in `updateColumns()` for resizable tables.

## Testing Done

- Verified that tables without custom styles still use calculated widths.
- Tested setting custom widths (e.g., `width: 100%`) via `editor.commands.updateAttributes('table', { style: 'width: 100%' })` and confirmed they are applied correctly.
- Ran existing tests to ensure no regressions.

## Verification Steps

1. Create a table in the editor.
2. Use `editor.commands.updateAttributes('table', { style: 'width: 100%' })` to set a custom width.
3. Check that the table renders with the specified width instead of the default calculated width.
4. Test with resizable tables to ensure the fix works for both rendering paths.

## Additional Notes

This resolves the issue where tables were always 100px wide, allowing users full control over table styling.

## Checklist

### DIFF
--- a/packages/extension-table/src/table/table.ts
+++ b/packages/extension-table/src/table/table.ts
@@ -273,15 +273,6 @@ export const Table = Node.create<TableOptions>({
 
   group: 'block',
 
-  addAttributes() {
-    return {
-      style: {
-        default: null,
-        parseHTML: element => element.getAttribute('style'),
-      },
-    }
-  },
-
   parseHTML() {
     return [{ tag: 'table' }]
   },


packages/extension-table/src/table/table.ts
```diff
@@ -273,17 +273,36 @@ export const Table = Node.create<TableOptions>({
 
   group: 'block',
 
+  addAttributes() {
+    return {
+      style: {
```
svenadlung: I think we should store a `width` attribute instead of the whole style object, WDYT?

packages/extension-table/src/table/TableView.ts
```diff
@@ -58,12 +58,17 @@ export function updateColumns(
...
+  // Check if user has set a width style on the table node
+  const hasUserWidth = node.attrs.style && typeof node.attrs.style === 'string' && node.attrs.style.includes('width')
+
+  if (!hasUserWidth) {
+    if (fixedWidth) {
+      table.style.width = `${totalWidth}px`
+      table.style.minWidth = ''
+    } else {
+      table.style.width = ''
+      table.style.minWidth = `${totalWidth}px`
```
svenadlung: Is it correct to ignore the min-width in that case? Even if the width is overridden.

packages/extension-table/src/table/table.ts
bdbch: I actually removed the addAttributes function because it's not really needed here. I just added it in case people want to pass through the style attribute directly - but it's not necessary for the logic check to run through.

PR #7171, opened 2025-11-05
Floating & Bubble Menu Show/Hide Events

mgreystone:
## Changes Overview

Adds "events" (transaction meta) for bot the bubble menu & floating menu extensions that will programmatically show & hide the respective menus.

## Implementation Approach

Followed the existing precedent set by the bubble menu's [`updatePosition`](https://tiptap.dev/docs/editor/extensions/functionality/bubble-menu?gad_source=1&gad_campaignid=22014820935&gbraid=0AAAAAqkAF26wCLp1PgeIh--dTtGG8bLjG&gclid=Cj0KCQiAiKzIBhCOARIsAKpKLAOFpNn7SUz03ZJkTypD3C7S73iYWT6ebtP_Z59HgV9CyVrMg1Fory0aAsnjEALw_wcB#force-update-the-position-of-the-bubble-menu) event. Added a transaction listener to the floating menu to listen for this new meta value.

## Testing Done

Ported these changes into my local application via `pnpm patch`. Tested programmatically showing & hiding the menus with React buttons.

```ts
<button
  type="button"
  onClick={() => editor.commands.setMeta('floatingMenu', 'show')}
>
  Show
</button>
```

## Additional Notes

Happy to add documentation upon approval.

## Checklist

### DIFF
--- a/.changeset/plenty-rice-shave.md
+++ b/.changeset/plenty-rice-shave.md
@@ -1,6 +1,6 @@
 ---
-'@tiptap/extension-floating-menu': minor
-'@tiptap/extension-bubble-menu': minor
+'@tiptap/extension-floating-menu': patch
+'@tiptap/extension-bubble-menu': patch
 ---
 
-Add "events" to programmatically show & hide the floating & bubble menus
+Add new transaction meta handlers to programmatically show & hide the floating & bubble menus via `transaction.setMeta('menuKey', 'show')` and `transaction.setMeta('menuKey', 'hide')`


bdbch:
Looks good. I think we may need to update the documentation so people know how to actually programmatically call this. Can you create a PR with a documentation for this here?

https://github.com/ueberdosis/tiptap-docs

mgreystone:
@bdbch Done https://github.com/ueberdosis/tiptap-docs/pull/581. Thanks so much.

mgreystone:
@bdbch 

> * When shouldShow is configured, it seems to take precedence over the meta event

Whoops. That seems fixable, though

> * When there are multiple menus, there's no way to distinguish which one should be shown

This is true. Follows the same pattern as the two `updatePosition` events, so i'm not sure why these would be treated differently? Maybe we could use a `pluginKey` to differentiate different menus, ~though i don't think plugin keys are supported in React or Vue?~

> In general I feel like this is the wrong way to do it. Maybe we can find a way that gives you access to the actuall BubbleMenu/FloatingMenu class via refs / getters so you can programmatically call `show()` or `hide()` on them directly without going through hoops.
> 
> For now I'm not happy with this solution. @mgreystone what are your thoughts?

I am but a visitor here; you all know better than I. But accessing a plugin instance directly seems odd to me. I don't think a plugin instance is event accessible via a PromseMirror view? Seems like more hoops to me, not less, but i might be missing something?

mgreystone:
> > * When there are multiple menus, there's no way to distinguish which one should be shown
> 
> This is true. Follows the same pattern as the two `updatePosition` events, so i'm not sure why these would be treated differently? Maybe we could use a `pluginKey` to differentiate different menus, ~though i don't think plugin keys are supported in React or Vue?~

I tried going this plugin key route. It is very clunky. Looks something like...


```ts
addCommands() {
  return {
    hideBubbleMenu: (pluginKey = 'bubbleMenu') =>
      ({ commands }) =>
        commands.setMeta(pluginKey, 'hide')
  }
}
```

The command (and all commands in general?) are a bit unpredictable if there are two instances of the same extension. It can't seem to fallback to `this.options.pluginKey`. Trying to support a `string` rather than a `PluginKey` also gets complicated, making sure that the command, plugin, & plugin view all share the same `PluginKey` instance.

I am not happy with it. But i don't have a better idea.

---

> In general I feel like this is the wrong way to do it. Maybe we can find a way that gives you access to the actuall BubbleMenu/FloatingMenu class via refs / getters so you can programmatically call `show()` or `hide()` on them directly without going through hoops.

I thought a lot about this, but i don't see a good way to get a `PluginView` instance from an `Extension` instance, much less a React or Vue component. They are abstracted away pretty well.

Maybe there could be something on the DOM that we could leverage--like how we had access to `element._tippy` in v2 (for better or worse). Though I worry if Floating UI were to be exposed directly, then the menu plugin's internal state will fall out of sync with the DOM.

---

Or if there was a good way for the plugin to optionally skip its [isSame](https://github.com/ueberdosis/tiptap/blob/3046a772dd4b40331435a0e350760d2dbb150e82/packages/extension-floating-menu/src/floating-menu-plugin.ts#L313) check when updating. That way, we could update `shouldShow` to return false based on some internal state, & then force an update. But right now, the plugin view only hides & shows if the selection or document changes.

mgreystone:
Afraid i'm going to abandon this.

bdbch:
Sorry, I didn't get your pings. Is this still relevant to you @mgreystone? I'd take a look again if you want.

mgreystone:
@bdbch This is still relevant to me, & i would be happy to keep working on this. I would need some [guidance how to proceed](https://github.com/ueberdosis/tiptap/pull/7171#issuecomment-3807174548). A clean option to expose show/hide functionality either with a command or a ref eludes me.

I would like to use the out-of-box bubble & floating menu extensions, if i can, but i have needed to abandon them in favor of a custom extension that meets my requirements.

bdbch:
Reopening then. Regarding the issue that you can't pick a specific menu via pluginKey - we have this PR that is almost ready to get merged:
https://github.com/ueberdosis/tiptap/pull/7512

I think this could be used to work around the limitation we have right now.

And yes, I think we should keep the plugin state in sync, otherwise we may run into weird side-effects. The best way then would be to add either commands that just send the meta transaction you proposed. That way we can trigger this on the plugin and separate by plugin key.

mgreystone:
@bdbch 

> Reopening then. Regarding the issue that you can't pick a specific menu via pluginKey - we have this PR that is almost ready to get merged: #7512
> 
> I think this could be used to work around the limitation we have right now.

That's good news. So i think the remaining limitation would be [commands](https://github.com/ueberdosis/tiptap/blob/253ca1c874d612e3629017f8a8009e76c5ffbf74/packages/extension-floating-menu/src/floating-menu.ts#L52). 

Something like this won't work if there are multiple menus. I believe it would only set meta for the last-registered extension.

```ts
  addCommands() {
    return {
      updateFloatingMenuPosition:
        () =>
        ({ tr, dispatch }) => {
          if (dispatch) {
            tr.setMeta(this.options.pluginKey, 'updatePosition')
          }
          return true
        },
    }
  },
```

Maybe something like this would work?

```ts
  addCommands() {
    return {
      updateFloatingMenuPosition:
        (pluginKey = this.options.pluginKey) =>
        ({ tr, dispatch }) => {
          if (dispatch) {
            tr.setMeta(pluginKey, 'updatePosition')
          }
          return true
        },
    }
  },
```

Or just not use commands at all & only use `setMeta`.

bdbch:
The second example would work and would be something I'd implement. That way users can just call this command with their desired plugin key and would be fine.

mgreystone:
@bdbch . Awesome. I am on it.

mgreystone:
@bdbch Done.

Based on my testing, `shouldShow` is ignored when `hide` or `show` is dispatched.

Also, commands aren't registered at all in React. So i did not add any new commands. The [existing command](https://github.com/ueberdosis/tiptap/blob/253ca1c874d612e3629017f8a8009e76c5ffbf74/packages/extension-floating-menu/src/floating-menu.ts#L52) is therefore problematic.

Will re-open & update the doc PR if you are happy.

bdbch:
> Also, commands aren't registered at all in React. So i did not add any new commands. The [existing command](https://github.com/ueberdosis/tiptap/blob/253ca1c874d612e3629017f8a8009e76c5ffbf74/packages/extension-floating-menu/src/floating-menu.ts#L52) is therefore problematic.

Yes, when using the React and Vue packages, you don't need to use the extension at all. I think the best solution there would be to

1. Just document it for React/Vue on the documentation
2. Have hooks that expose functions for an editor that will dispatch the meta transaction

mgreystone:
@bdbch 

Hmm.... I think i'd still vote for omitting commands & having to call `setMeta` directly. Different entrypoints for vanilla/react/vue seems like it would get confusing?

Or maybe a stand-alone function. Doesn't follow any other pattern i am aware of, but it would be consistent.


```ts
function hideBubbleMenu(editor: Editor, pluginKey = 'bubbleMenu'): void {
  const tr = editor.view.tr
  tr.setMeta(pluginKey, 'hide')
  editor.view.dispatch(tr)
}
```

mgreystone:
@bdbch One final check in. Can we move forward just with `setMeta` and without commands? Or do we need another direction here?

bdbch (approved):
Using transaction meta scoped by plugin key looks like the right direction here. It keeps the plugin state in sync, works with multiple menus, and avoids exposing internal plugin view instances.

mgreystone:
Thank you @bdbch 

Upon a rebase, it seems that https://github.com/ueberdosis/tiptap/pull/7611 has broken this using a default plugin key, as if you don't add a `pluginKey` prop to react, it will create a PM `PluginKey`, which the caller will not have. So for this to work with React, you _must_ specify a plugin key.

This affects the existing `updatePosition` functionality as well.

Documentation at https://github.com/ueberdosis/tiptap-docs/pull/716

bdbch:
@mgreystone I think that is the correct conclusion.

If a user wants to control a specific menu externally, that menu needs a stable, known `pluginKey`. After #7611, React/Vue generate a unique ProseMirror `PluginKey` when none is provided, which is the right behavior to avoid collisions between multiple menu instances. But it also means external code cannot reliably target that menu unless the key was explicitly passed in.

So I do not think we should assume a default menu here. That breaks down as soon as there is more than one instance, and it also means the caller may dispatch meta to a key that the actual menu plugin is not using.

I think the right contract is:

- if you want to control a specific menu externally via `setMeta`, you must provide `pluginKey`
- if you omit `pluginKey`, the menu still works normally, but it is not externally addressable

And yes, that applies to the existing `updatePosition` behavior as well, not just `show` / `hide`.

So from my side, I would keep the generated-key behavior from #7611 and update the docs to make the `pluginKey` requirement explicit for React/Vue users whenever they want programmatic control.

mgreystone:
@bdbch Sounds good. In that case, this is ready to go.

PR #7181, opened 2025-11-06
fix: prevent marks from continuing across hard breaks in Markdown

### LINKED ISSUE #7136: Markdown serializer reopens marks after hardBreak; following line keeps the mark
### Affected Packages

core, markdown, starter-kit, extension-hard-break

### Version(s)

3.9.0

### Bug Description

If a text node with a mark (bold, italic, strike, underline) is followed by a hardBreak, the markdown serializer closes the mark and immediately reopens it—even when the next text node doesn’t carry that mark. As a result the serialized markdown wraps the following line in the mark syntax (e.g. **line**), so when the document is reloaded that line regains the mark.

### Browser Used

Chrome

### Code Example URL

https://codesandbox.io/p/devbox/loving-https-7q87qr?workspaceId=ws_7DogZqLgKBuNY6TBGogudR

### Expected Behavior

Marks should only reopen when the inline node itself (or upcoming siblings) use that mark. In the example above the second line has no marks, so the serializer should emit:

**Line with mark:**
Line without mark.

Instead it currently outputs:

**Line with mark:**
**Line without mark.**

- The behavior lives in `MarkdownManager.renderNodesWithMarkBoundaries`.
- The non-text branch closes all active marks, renders the node (e.g. `hardBreak`), then reopens the marks unconditionally.
- Guarding the reopen so it only happens when the current inline node (or any upcoming sibling) actually uses the mark fixes the issue (had codex-high generate this fix for me)

Pseudo-code for the guard tested:

```ts
const nodeUsesMark = (subject, markType) => {
  if (!subject || typeof subject !== 'object') return false
  if (Array.isArray(subject.marks) && subject.marks.some(mark => mark.type === markType)) return true
  if (Array.isArray(subject.content)) {
    return subject.content.some(child => nodeUsesMark(child, markType))
  }
  return false
}

const upcoming = nodes.slice(i + 1)
const marksToReopen = new Map()

marksBeforeNode.forEach((mark, markType) => {
  if (nodeUsesMark(node, markType) || upcoming.some(n => nodeUsesMark(n, markType))) {
    marksToReopen.set(markType, mark)
  }
})

### Dependency Updates

bdbch:
## Changes Overview

Fixed hard break mark boundaries in Markdown serialization to prevent marks from continuing across line breaks.

## Implementation Approach

Modified the `renderNodesWithMarkBoundaries` method in `MarkdownManager.ts` to skip mark reopening after hardBreak nodes, ensuring that marks terminate at line breaks as expected in Markdown.

## Testing Done

- Added comprehensive Cypress integration tests covering hard break mark boundary scenarios
- Tests verify that marks don't continue after hard breaks and that marks can still be explicitly present on hard breaks when needed
- All existing tests continue to pass

## Verification Steps

1. Run the Cypress tests: `pnpm test:run --spec "**/conversion.spec.ts"`
2. Check that the hard break marks test cases pass
3. Verify that the fix prevents unwanted mark continuation in Markdown output

## Additional Notes

The fix is minimal and targeted, only affecting the mark boundary logic for hard break nodes. This ensures backward compatibility while fixing the specific issue.

## Checklist

## Related Issues

Fixes #7136

### DIFF
--- a/tests/cypress/integration/markdown/conversion-files/hard-break-marks.ts
+++ b/tests/cypress/integration/markdown/conversion-files/hard-break-marks.ts
@@ -1,9 +1,9 @@
 export const name = 'Hard Break with Marks'
 
-export const expectedInput = `**Speaker:**
+export const expectedInput = `**Speaker:**  
 John Doe.
 
-**Speaker:**
+**Speaker:**  
 **John Doe**`.trim()
 
 export const expectedOutput = {


bdbch:
I've separated the handling of soft breaks and hard breaks in Markdown parsing/serialization to align with the CommonMark spec:

- **Soft breaks**: Single newlines (`\n`) are now kept as `\n` characters in text nodes (no forced hard break conversion)
- **Hard breaks**: Two trailing spaces + newline (`  \n`) now correctly parse to `hardBreak` nodes and serialize back with the proper Markdown syntax

This ensures round-trip compatibility while respecting how Markdown actually works. Updated the tests to cover both cases separately, and fixed the `HardBreak` extension's `renderMarkdown` to output `  \n` instead of just `\n`.

Tests are now passing for both scenarios! 🎉

Issue #113, opened 2018-12-03
isActive.bold is not a function


PrimozRome:
I am trying to use tiptap in my Nuxt.js project but I am having problem integrating menu bar. I get this error "app.js:503 TypeError: isActive.bold is not a function".

I am following this basic example https://github.com/heyscrumpy/tiptap/blob/master/examples/Components/Routes/Basic/index.vue.

Here is my code

   ```
 <editor-menu-bar :editor="editor">
          <div
            slot-scope="{ commands, isActive }"
            class="html-eidor-menu">
            <button
              :class="{ 'is-active': isActive.bold() }"
              class="menubar__button"
              @click="commands.bold">
              <i class="fa fa-bold"/>
            </button>
          </div>
        </editor-menu-bar>

        <editor-content
          ref="input"
          :editor="editor"
          :editable="readonly || disabled"
          class="html-editor"
        />
```

isActive.bold() returns TypeError: isActive.bold is not a function

If I remove isActive from the template line then I have a problem with `@click="commands.bold"`. 

What am I doing wrong?

**Comment by philippkuehn** [comment]
It seems you forgot to import the bold extension?

**Comment by PrimozRome** [comment]
Hmm you are right I see in the example code now. 

I didn't know I need to do this. It's not mentioned in the docs for adding a menu bar.

**Comment by philippkuehn** [comment]
Yeah sorry for that. The docs are bad at the moment ☹️

**Comment by PrimozRome** [comment]
No worries, I cal live with that.

**Comment by PrimozRome** [comment]
@philippkuehn after that I am still getting error for commands `Invalid handler for event "click": got undefined`. I must be still missing something?

```
<template>
  <div>
        <editor-menu-bar :editor="editor">
          <div
            slot-scope="{ commands, isActive }"
            class="html-eidor-menu">
            <button
              :class="{ 'is-active': false }"
              class="menubar__button"
              @click="commands.bold">
              <i class="fa fa-bold"/>
            </button>
          </div>
        </editor-menu-bar>

        <editor-content
          ref="input"
          :editor="editor"
          :editable="readonly || disabled"
          class="html-editor"
        />
  </div>
</template>

<script>
import { Editor, EditorMenuBar, EditorContent } from 'tiptap'
import { Bold } from 'tiptap-extensions'

export default {
  name: 'HtmlInput',

  components: {
    EditorMenuBar,
    EditorContent
  },

  data() {
    return {
      editor: new Editor({
        extensions: [
          new Bold()
[...]

PR #7201, opened 2025-11-10
Create utility to update a position after a transaction

arnaugomez:
## Changes Overview

Creates a utility to compute the new position in the document after a transaction has been applied.

```ts
const {newPosition, newYRelativePosition} = editor.positionHelpers.getUpdatedPosition({
                  transaction,
                  position,
                  yRelativePosition,
})
```

Creates a demo to showcase this utility.

## Implementation Approach

ProseMirror has the following utility to get the updated position after a transaction:

```ts
const newPosition = transaction.mapping.map(position)
```

This utility works well in non-collaborative editors, but has a flaw: in collaborative editors, every Y.js transaction replaces the whole document.

To avoid this, Y.js transactions need to be handled differently. Y.js has a concept called "Y.js relative positions" that let you calculate the new position in the document after a Y.js transaction.

### API design

The approach is to create a `editor.positionHelpers` property which contains a list of methods that let developers map a position of a decoration without worrying about what type of transaction it is (a ProseMirror or a Y.js transaction).

`editor.positionHelpers` has 2 different implementations: the "pure ProseMirror" implementation and the "Collaboration" implementation. The "pure ProseMirror" implementation is located inside `@tiptap/core`.  The "Collaboration" implementation is located in the `@tiptap/collaboration` package and it is able to handle Y.js transactions. 

If the Editor has the Collaboration extension installed, the implementation is proxied to the "Collaboration" implementation, otherwise, it uses the "pure ProseMirror" implementation. This keeps the Editor code lean and simple so that it does not have any Collaboration logic: all the Y.js-related logic is kept in the Collaboration extension.

## Testing Done

Test the behaviour in the demo

Did not have time to add unit tests, would love to do it soon.

## Verification Steps

Try out the demo.

## Checklist

## Related Issues

Related Y.js issues:

[Edits by peers produce a prosemirror transaction that spans the entire document · Issue #113 · yjs/y-prosemirror](https://github.com/yjs/y-prosemirror/issues/113)

[Local decorations lost when remote client changes document · Issue #49 · yjs/y-prosemirror](https://github.com/yjs/y-prosemirror/issues/49)


arnaugomez:
I'm thinking of simplifying the naming of the methods:

`mapPositionFromTransaction` -> `mapPosition`

What do you think? Naming/API ideas welcome.

janthurau (approved):

.changeset/bright-walls-wash.md
```diff
@@ -0,0 +1,16 @@
+---
+'@tiptap/extension-collaboration': minor
+'@tiptap/core': minor
+'tiptap-demos': minor
```
bdbch, line 4: We don't really release the demos so we don't need that here.

packages/core/src/Editor.ts
```diff
@@ -773,4 +782,34 @@ export class Editor extends EventEmitter<EditorEvents> {
...
   }
+
+  /**
+   * Returns a set of utilities for working with positions and ranges. These
+   * utilities let you calculate the new position or range after applying a
+   * transaction.
+   *
+   * @returns The position helpers.
+   */
+  public get positionHelpers(): PositionHelpers {
```
bdbch, line 793: Does this make sense as a getter on the Editor class? I thought this would just be utility functions I important and pass my props through to map abs/rel positions. Why doe we need it here?

Also why do the positionHelpers exist on the collaboration state?

packages/extension-collaboration/src/helpers/getPositionHelpers.ts
```diff
@@ -0,0 +1,36 @@
+import type { Editor, PositionHelpers } from '@tiptap/core'
+
+import { mapPositionFromTransaction } from './mapPositionFromTransaction.js'
+import { mapRangeFromTransaction } from './mapRangeFromTransaction.js'
+import { getYAbsolutePosition, getYRelativePosition } from './yRelativePosition.js'
+import { getYAbsoluteRange, getYRelativeRange } from './yRelativeRange.js'
+
+export function getPositionHelpers(editor: Editor): PositionHelpers {
+  const { state } = editor
+  return {
```
bdbch, line 10: Why do we need to go through these helpers instead of calling those helpers directly from where I want to map? I'd rather keep it simple and expose simple helpers as utility functions.

Maybe you had a thought for this, but I think its confusing to go through collaboration storage <= utility functions <= editor class

packages/extension-collaboration/src/helpers/mapPositionFromTransaction.ts
```diff
@@ -0,0 +1,35 @@
...
+import { mapPositionFromTransaction as coreMapPositionFromTransaction } from '@tiptap/core'
+
+import { isChangeOrigin } from './isChangeOrigin.js'
+import { getYAbsolutePosition, getYRelativePosition } from './yRelativePosition.js'
+
+/**
+ * Returns the new position after applying a transaction. Handles both Y.js
+ * transactions and regular transactions.
+ */
+export function mapPositionFromTransaction(
```
bdbch, line 11: It's weird that the function itself is defined on the collaboration extension but the typing strictly named after this function lives in core.

packages/extension-collaboration/src/helpers/mapRangeFromTransaction.ts
```diff
@@ -0,0 +1,35 @@
+import type { MapRangeFromTransactionOptions, MapRangeFromTransactionResult } from '@tiptap/core'
+
+import { mapPositionFromTransaction } from './mapPositionFromTransaction.js'
+
+/**
+ * Returns the new Range after applying a transaction. Handles both Y.js
+ * transactions and regular transactions.
+ */
+export function mapRangeFromTransaction(options: MapRangeFromTransactionOptions): MapRangeFromTransactionResult {
```
bdbch, line 9: same here, if we do more of this the code gets really messy because you have more and more cross-dependencies/references between packages that could potentially not exist together.

packages/extension-collaboration/src/collaboration.ts
```diff
@@ -13,6 +16,11 @@ export interface CollaborationStorage {
    * Disabling collaboration will prevent any changes from being synced with other users.
    */
   isDisabled: boolean
+
+  /**
+   * Get helper methods for working with Y.js positions and ranges.
+   */
+  getPositionHelpers: (editor: Editor) => PositionHelpers
```
bdbch, line 23: As described above it feels weird that I need to go through the editor class, to the collabs extension storage just to call a mapping function. Is that really necessary?

bdbch (changes requested):
A few questions / things I'd like to clarify before we move forward.

arnaugomez:
Hi @bdbch . Thanks for your review. I totally agree that this could be a standalone utility function instead of a method in the Editor. That would really simplify the design.

However, I'm not sure where this utility would be imported from. Would we import it from `@tiptap/core` or from `@tiptap/collaboration`? I believe it should be imported from the Collaboration extension to avoid mixing concerns.

I suggest, to keep it simple, that there is a single API that is imported from `@tiptap/collaboration` and that handles both collaborative and non-collaborative transactions.
```ts
import {getUpdatedPosition} from "@tiptap/collaboration"
const {newPosition, newYRelativePosition} = getUpdatedPosition({
                  transaction,
                  position,
                  yRelativePosition,
})
```
Would this be a better API?

safwansamsudeen:
@arnaugomez this is such great work, thanks! Need it for my editor, so I'm waiting for it to get merged. Is there a timeline?

arnaugomez:
> @arnaugomez this is such great work, thanks! Need it for my editor, so I'm waiting for it to get merged. Is there a timeline?

We're working on it! Hopefully we release it this week

nperez0111:
@arnaugomez & @bdbch I'd recommend for you to look at what I did in BlockNote: https://github.com/TypeCellOS/BlockNote/blob/90665db51eea65ea4581f79b2bcbd7b8e605bb97/packages/core/src/api/positionMapping.ts

For Tiptap, you should definitely keep any Y.js implementations away from the core package. I think how this would work for Tiptap would be to have a `trackPosition`, in the editor core. Which, the collaboration extension can replace the implementation of at runtime `onCreate` (so it has an editor instance).

Happy to answer questions

PR #7207, opened 2025-11-11
Add RTL/LTR textDirection support

### LINKED ISSUE #3957: [Bug]: Why RTL is not working in tiptap editor?

core

2.0.0

 when I applied RTL and it should take RTL styles but its not. Otherwise, tiptap can't recognize which one "LTR"/"RTL"

Chrome

### Code example

_No response_

Hi there,
I think one of the most popular editors tiptap in react js. why missing RTL? Please add RTL support in tiptap editor, it would be great to get more users. `dir="auto"` is not a solution is a take us bluff. when I style something on RTL it's not applied because of the `'auto'` direction, it cannot catch 'RTL' or 'LTR'. so `dir='auto'` is not a good solution, either you make an extension for that.
OR
If you guys are already implemented/manage somehow to solve this problem (like: above problem) please let me know, and I will implement that.

_No response_

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

bdbch:
## Changes Overview

  This PR adds native text direction support for RTL (right-to-left) and bidirectional content in Tiptap. Users can now
  set a global text direction for the editor or apply direction attributes to specific nodes via commands.

Documentation: https://github.com/ueberdosis/tiptap-docs/pull/492

  ## Implementation Approach

  - Added a new `textDirection` option to `EditorOptions` that accepts `'ltr'`, `'rtl'`, `'auto'`, or `undefined`
  - Created a new `TextDirection` core extension that:
    - Adds a global `dir` attribute to all non-text nodes
    - Sets the `dir` attribute on the editor element itself via a ProseMirror plugin
    - Uses extension options to properly configure the default direction during schema creation
  - Implemented two new commands:
    - `setTextDirection(direction, position?)` - Sets text direction on nodes (uses current selection if no position
  provided)
    - `unsetTextDirection(position?)` - Removes text direction from nodes
  - The extension is enabled by default but can be disabled via `enableCoreExtensions` option

  ## Testing Done

  - Created a comprehensive React demo showcasing:
    - Global direction control (None/LTR/RTL/Auto)
    - Per-node direction commands
    - Examples with English, Arabic, Hebrew, and bidirectional text
    - Various node types (headings, paragraphs, lists, blockquotes)
  - Added Playwright test specs for the demo
  - Manually tested the demo to verify proper rendering and attribute application

  ## Verification Steps

  1. Check out this branch and run the demos: `npm run dev`
  2. Navigate to the TextDirection demo at `/Examples/TextDirection/React`
  3. Test global direction by clicking the direction buttons (None/LTR/RTL/Auto)
  4. Verify that the `dir` attribute appears on the editor element and all content nodes
  5. Select text and use the "Set Direction on Selection" buttons to apply direction to specific nodes
  6. Test with different content types (headings, lists, blockquotes) to ensure all nodes support the attribute

  ## Additional Notes

  This implementation follows the approach used by other editors like Lexical, which add `dir` attributes to individual
  nodes rather than relying solely on CSS or parent-level directives. This ensures consistent rendering across different
   contexts and supports proper bidirectional text handling for languages like Arabic and Hebrew mixed with English.

  ## Checklist

  ## Related Issues

  Fixes #3957

### DIFF
--- a/demos/src/Examples/TextDirection/React/index.spec.js
+++ b/demos/src/Examples/TextDirection/React/index.spec.js
@@ -1,61 +1,25 @@
-import { expect, test } from '@playwright/test'
-
-test.describe('TextDirection', () => {
-  test('should apply text direction attributes', async ({ page }) => {
-    await page.goto('http://localhost:3000/Examples/TextDirection/React')
-
-    await page.waitForSelector('.tiptap')
-
-    const firstParagraph = page.locator('.tiptap p').first()
-    const dirAttribute = await firstParagraph.getAttribute('dir')
-
-    expect(dirAttribute).toBe('auto')
+context('/src/Examples/TextDirection/React/', () => {
+  beforeEach(() => {
+    cy.visit('/src/Examples/TextDirection/React/')
   })
 
-  test('should change global direction', async ({ page }) => {
-    await page.goto('http://localhost:3000/Examples/TextDirection/React')
-
-    await page.waitForSelector('.tiptap')
-
-    await page.click('button:has-text("RTL")')
-
-    await page.waitForTimeout(100)
-
-    const firstParagraph = page.locator('.tiptap p').first()
-    const dirAttribute = await firstParagraph.getAttribute('dir')
-
-    expect(dirAttribute).toBe('rtl')
+  it('should apply text direction attributes', () => {
+    cy.get('.tiptap p').first().should('have.attr', 'dir', 'auto')
   })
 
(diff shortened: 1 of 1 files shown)


AlonMiz:
Hi thanks for that @bdbch works great.
one gap that still remains is that the placeholder should also conform to the rtl if it's in rtl language
`@tiptap/extension-placeholder`
want me to open an issue for that?

bdbch:
@AlonMiz sounds good - feel free to open an issue.

PR #7216, opened 2025-11-12
feat(core): allow custom resize handle rendering via `createCustomHandle` in `ResizableNodeView`

Aslam97:
## Changes Overview

1. **Added** an optional `createCustomHandle` callback to `ResizableNodeView`, allowing developers to define and style their own resize handles. When provided, the internal logic uses this callback instead of the default handle creation, bypassing the built-in `positionHandle` behavior. This gives developers full control over markup, styling, and positioning while maintaining backward compatibility.

2. **Removed** predefined styles from the `wrapper` element to better support dynamic alignment. The previous inline styles required developers to use `!important` overrides, which are no longer necessary.

3. **Added** an `update` event listener to the editor to dynamically attach or remove resize handles based on the editor’s editable state. This implementation tracks the last editable state to prevent unnecessary re-renders.

## Implementation Approach

Introduced a new optional method:

```ts
createCustomHandle?: (direction: ResizableNodeViewDirection, className: string) => HTMLElement
```

Hooked into the constructor:

```ts
if (options.options?.createCustomHandle) {
  this.createCustomHandle = options.options.createCustomHandle
}
```

Updated `attachHandles()` to: Use `createCustomHandle` when defined. Skip the default `positionHandle()` in that case.

## Checklist

- [ ] I have added tests where applicable

## Related Issues

#7218


packages/core/src/lib/ResizableNodeView.ts
```diff
@@ -567,8 +619,14 @@ export class ResizableNodeView {
    */
   private attachHandles(): void {
     this.directions.forEach(direction => {
-      const handle = this.createHandle(direction)
-      this.positionHandle(handle, direction)
+      const handle = this.createCustomHandle
+        ? this.createCustomHandle(direction, this.classNames.handle)
+        : this.createHandle(direction)
+
```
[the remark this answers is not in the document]
Aslam97: Hmm, not sure whether to include this or not — let’s wait for the maintainer’s though first.

packages/core/src/lib/ResizableNodeView.ts
```diff
@@ -204,6 +204,52 @@ export type ResizableNodeViewOptions = {
...
+     *
+     *   if (direction === 'left' || direction === 'right') {
+     *     handle.style.top = '0'
+     *     handle.style.bottom = '0'
+     *   }
+     *
+     *   return handle
+     * }
+     * ```
+     */
```
arnaugomez, line 257: Can you think of a more simple example?

packages/core/src/lib/ResizableNodeView.ts
```diff
@@ -204,6 +204,52 @@ export type ResizableNodeViewOptions = {
...
+    createCustomHandle?: (direction: ResizableNodeViewDirection, className: string) => HTMLElement
```
arnaugomez: Can you think of other configuration options that a developer might need, besides the `className`? For example: style or any other html property

Aslam97: Passing `className` to the handle isn’t strictly necessary—since developers can customize the handle in any way they prefer—I decided to include it so that those using the `className` option can easily reuse their existing styles. It’s basically the same `className` value that the developer defines in the configuration. For example:

```ts
className: {
  handle: 'tiptap-handle'
}
```

This allows developers to reuse the same styling within `createCustomHandle`, though they’re free to ignore it if they prefer their own approach.

packages/core/src/lib/ResizableNodeView.ts
Aslam97: The reason I’m using that example is because it aligns with the default behavior, so developers can use it as a reference for how to implement their own.

packages/core/src/lib/ResizableNodeView.ts
bdbch: Since this seems to be a factory function for a new HTMLElement, why not avoid passing through any unnecessary properties here and just give it a direction? I'll have access to the className anyway?

```js
createCustomHandle: (direction) => {
  const myHandle = document.createElement('div')
  myHandle.classList.add('resize-handle')
  myHandle.classList.add(`resize-handle--${direction}`)

  return myHandle
}
```

Aslam97: Good point, I’ve already removed `className` from the callback. The function now only receives the `direction`

Vico-CS:
Hi @bdbch,
Are changes still required on this PR or can it be merged?
I'm pushing a bit since it will solves [this issue](https://github.com/ueberdosis/tiptap/issues/7218)

bdbch:
Sorry, currently a bit slow with reviewing all the things coming in. Will come back to this today.

packages/core/src/lib/ResizableNodeView.ts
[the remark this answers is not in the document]
bdbch: I think it's a good idea for IDE's that don't pick up typings automatically. I'm just not sure about a throw. How about we just fall back to the default handle & use `console.warn` instead.

bdbch (approved):
Looks good. Just a small thing around the creation of the handle. See the comment below.

Aslam97:
Got it! I’ve switched to the explicit branching and added a safeguard so if a custom handle doesn’t return a proper HTMLElement, it falls back to the default and logs a warning.

Vico-CS:
Hi @bdbch,
One last approve needed and we should be good I believe 🙏

PR #7219, opened 2025-11-12
Make Tiptap JSON types more accurate

arnaugomez:
## Changes Overview

This PR refactors the `JSONContent` type and related JSON types in `@tiptap/core` to be more accurate and type-safe. The previous implementation used a loose type definition with optional fields and an index signature (`[key: string]: any`), which allowed invalid structures and provided poor type checking. The new implementation introduces distinct, well-documented interfaces that accurately represent the Tiptap JSON structure.

## Implementation Approach

The refactoring introduces several new type definitions:

1. **`JSONContent`** - Now an alias for `JSONNode` (previously a loose object type)
2. **`JSONFragment`** - A type representing a list of Tiptap JSON nodes: `JSONNode[]`
3. **`JSONNode`** - A unified interface that represents both text and non-text nodes:
   - Required `type: string` (e.g., "text", "doc", "paragraph", "heading", "image")
   - Optional `attrs?: Record<string, any>`
   - Optional `text?: string` - Only present for text nodes (where `type: "text"`)
   - Optional `marks?: JSONMark[]` - Applied to inline nodes
   - Optional `content?: JSONFragment` - Only present for non-text nodes
   
   The interface documentation clarifies that:
   - Text nodes have `type: "text"`, a `text` property, and never have `content`
   - Non-text nodes have other types, can have `content`, but never have `text`
4. **`JSONMark`** - Interface for marks with:
   - Required `type: string`
   - Optional `attrs?: Record<string, any>`

Key improvements:

- Unified node interface simplifies the type system while maintaining accuracy
- Removed the permissive index signature `[key: string]: any`
- Changed `attrs` from `Record<string, any>` to `Record<string, any>` for better type safety
- Added comprehensive JSDoc documentation explaining the distinction between text and non-text nodes
- Types now accurately reflect the actual structure used throughout the codebase

## Testing Done

- Existing tests in `tests/cypress/integration/core/` continue to pass, validating that the new types are compatible with existing usage patterns
- Tests verify JSON structures with text nodes (`type: 'text'` and `text` property) work correctly
- The types align with the JSON structures used in tests like `createNodeFromContent.spec.ts`, `generateJSON.spec.ts`, and `onContentError.spec.ts`

## Verification Steps

1. Review the type definitions in `packages/core/src/types.ts` (lines 444-503)
2. Verify that TypeScript compilation succeeds: `pnpm build`
3. Check that existing tests pass: `pnpm test`
4. Verify type checking in IDE when using `JSONContent` - text nodes should have `type: "text"` and `text` property
5. Test that the unified `JSONNode` interface correctly handles both text and non-text nodes

## Additional Notes

This is a **minor** version bump as it improves type accuracy but may require updates in codebases that were relying on the previous loose typing. The changes are backward compatible at runtime, but may surface type errors in TypeScript projects that were using invalid JSON structures.

The new types provide better:

- **Type safety**: Invalid structures are now caught at compile time
- **Developer experience**: Better autocomplete and IntelliSense
- **Documentation**: Clear JSDoc comments explain each type's purpose
- **Maintainability**: More explicit types make the codebase easier to understand

## Checklist

## Related Issues

https://github.com/ueberdosis/tiptap-docs/pull/495

### DIFF
--- a/.changeset/popular-rules-own.md
+++ b/.changeset/popular-rules-own.md
@@ -1,0 +1,5 @@
+---
+'@tiptap/core': minor
+---
+
+Make Tiptap JSON types more accurate
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -470,7 +470,7 @@ export interface JSONNode {
    * Optional attributes for the node. The structure depends on the node type.
    * Attribute values can be any JSON-serializable value.
    */
-  attrs?: Record<string, unknown>
+  attrs?: Record<string, any>
   /**
    * The text content of the node. Only present for text nodes.
    * Text nodes have the `type` property set to "text".
@@ -499,7 +499,7 @@ export interface JSONMark {
    * Optional attributes for the mark. The structure depends on the mark type.
    * Attribute values can be any JSON-serializable value.
    */
-  attrs?: Record<string, unknown>
+  attrs?: Record<string, any>
 }
 
 /**
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -452,13 +452,13 @@ export type JSONContent = JSONNode
 export type JSONFragment = JSONNode[]
 
 /**
- * Represents a non-text node in Tiptap JSON format. Nodes can contain other
+ * Represents a node in Tiptap JSON format. Nodes can contain other
  * nodes in their `content` property.
  *
  * There are two types of nodes:
- * - Text nodes: nodes with the type "text". They have a `text` property and
+ * - Text nodes: nodes with the `type` property set to "text". They have a `text` property and
  *   never have a `content` property. Text nodes are inline nodes.
- * - Non-text nodes: nodes with type other than "text". They can have a `content`
+ * - Non-text nodes: nodes with `type` other than "text". They can have a `content`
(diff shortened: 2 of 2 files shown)


.changeset/popular-rules-own.md
```diff
@@ -0,0 +1,5 @@
+---
+'@tiptap/core': minor
```
bdbch, line 2: I'm a bit worried to change such an important type in a minor version. I could imagine this change could break plenty of existing integrations / typings because people never had to add type guards or had to decide between JSON nodes or Marks.

JSONFragment also isn't really needed because there is no real fragment in just our JSON content structure. Fragments in Prosemirror usually have additional properties that normal nodes don't have.

Do you have any examples where this new typing would actually help? Because right now the current, simple typing already fulfilled pretty much all json usecases where you want to pass json as content since nodes will always follow nodes except you go into a mark.

bdbch (changes requested):
I'm really not a fan changing such an important type in a minor version. Not sure if people would need to add typeguards or their typings suddenly requiring changes.

What do you think? Is this really necessary to split types by nodes/marks? Since marks can only be in a nodes `marks` field our loose typing pretty much did everything we needed.

PR #7248, opened 2025-11-19
Utility to update a position after a transaction (approach 2)

arnaugomez:
## Changes Overview

Creates a utility to compute the new position in the document after a transaction has been applied.

```ts
const {newPosition,mapResult} = editor.positionHelpers.getUpdatedPosition({
                  transaction,
                  position,
})
```

## Implementation Approach

ProseMirror has the following utility to get the updated position after a transaction:

```ts
const newPosition = transaction.mapping.map(position)
```

This utility works well in non-collaborative editors, but has a flaw: in collaborative editors, every Y.js transaction replaces the whole document.

To avoid this, Y.js transactions need to be handled differently. Y.js has a concept called "Y.js relative positions" that let you calculate the new position in the document after a Y.js transaction.

This PR tries an alternative approach to map positions that consists of:

- Using a plugin to store the state of the Y.doc before and after the transaction
- Using the state of the Y.doc before the transaction to get the relative position
- Using the state of the Y.doc after the transaction to convert the relative position into an absolute position.

It exposes a default implementation for mapping positions that works in non-collaborative editors. This default implementation is overridden by the collaboration extension so that it can handle collaborative transactions too.

This PR also includes a demo that showcases this functionality.

## Checklist

## Related Issues

Related Tiptap issues/PRs:

https://github.com/ueberdosis/tiptap/pull/7201

Related Y.js issues:

[Edits by peers produce a prosemirror transaction that spans the entire document · Issue #113 · yjs/y-prosemirror](https://github.com/yjs/y-prosemirror/issues/113)

[Local decorations lost when remote client changes document · Issue #49 · yjs/y-prosemirror](https://github.com/yjs/y-prosemirror/issues/49)


packages/extension-collaboration/src/collaboration.ts
```diff
@@ -48,7 +52,7 @@ export interface CollaborationOptions {
    * @default 'default'
    * @example 'my-custom-field'
    */
-  field?: string
+  field: string
```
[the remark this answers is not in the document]
arnaugomez: This is a misunderstanding in Tiptap types. Even though the property is required, it is optional in the `Extension.configure()` method.

packages/core/src/Editor.ts
```diff
@@ -778,4 +784,26 @@ export class Editor extends EventEmitter<EditorEvents> {
   get $doc() {
     return this.$pos(0)
   }
+
+  positionHelpers: PositionHelpers = {
```
bdbch: Not a fan of the specific naming. If we put utility functions directly on the editor instance I'd rather give the interface & property name a more generic name. `editor.utils.getUpdatedPosition()` sounds more generic and could be reused in the future for other generic utils that live on the editor instance.

packages/extension-collaboration/src/collaboration.ts
```diff
@@ -102,6 +106,17 @@ export const Collaboration = Extension.create<CollaborationOptions, Collaboratio
     }
   },
 
+  onBeforeCreate() {
+    this.editor.positionHelpers = {
+      getUpdatedPosition: (position, transaction) => {
```
bdbch: I would spread ...this.editor.utils before here so if we have more generic helpers later, this wouldn't override the whole object.

bdbch (changes requested):
Looks good, however I don't like having a very specific "positionHelpers" property on the editor. Lets make this a generic `utils` property that can hold more utility functions.

packages/extension-collaboration/src/collaboration.ts
arnaugomez: fixed

packages/core/src/Editor.ts
arnaugomez: Renamed to `utils`

arnaugomez:
I applied the suggested changes @bdbch . Let me know if that works!

Should we mark that as an experimental API? Or is it OK to publish it as it is?

An experimental API would look like this:

```ts
const updatedPosition = editor.utils.experimental__getUpdatedPosition(position, transaction)
```

bdbch:
> I applied the suggested changes @bdbch . Let me know if that works!
> 
> Should we mark that as an experimental API? Or is it OK to publish it as it is?
> 
> An experimental API would look like this:
> 
> ```ts
> const updatedPosition = editor.utils.experimental__getUpdatedPosition(position, transaction)
> ```

I think we can sent it without the experimental prefix as long as it works, we could reuse that util later for the Decorations API so we don't even need to remove it again.

arnaugomez:
I have another doubt: 

In order for his API to work, the Collaboration extension has to store a snapshot of the YDoc on every transaction. 
According to Jan, the performance hit is not large. However, should we make the collaborative position mapping opt-in/opt-out? So that, if you don't activate this option, there is zero performance overhead. Not everybody is going to map positions in their editor.

I see it could be configured like this:
```ts
Collaboration.configure({specialPositionMapping: true})
```
(I have no idea how to call this configuration option)

Let me know what you think.

bdbch:
Maybe we can activate that option as soon as mapping is called once instead of having to opt-in? Not sure if that works or not.

bdbch (approved):

arnaugomez:
> Maybe we can activate that option as soon as mapping is called once instead of having to opt-in? Not sure if that works or not.

To map a position that way, the snapshot of the Y.doc already needs to be stored before the mapping is called. So it would not be possible to do that.

arnaugomez:
After thinking about it I've made collaborative position mapping optional. It can be enabled with this option:
```ts
Collaboration.configure({enablePositionMapping: true})
```
This way, developers that update do not suffer the performance overhead. This also gives us time to study and measure the actual performance overhead of enabling this option in real-time scenarios, and if the performance hit is very low, we can enable it by default in the future.

arnaugomez:
I close this because, after discussing with the team, we've decided to follow approach 3 (see PR #7288). We've decided to not go with this PR because it introduced extra overhead: you had to serialize and deserialize the Y.js document on every transaction. We did some profiling and performance testing and we found that approach 3 was more performant.

PR #7260, opened 2025-11-24
fix(markdown): handle unknown XML tags in markdown content

### LINKED ISSUE #7256: `setContent` in Markdown crashes if XML is present
### Affected Packages

markdown

### Version(s)

3.11.0

### Bug Description

When you attempt to `setContent` with text including XML such as <abc><abc/> in a tiptap markdown editor, such as this:

> `const content = "Test failing content <abc></abc>";`

...it'll fail with this error:

> Invalid content for node paragraph: <"Test failing content ", paragraph>

### Browser Used

Chrome

### Code Example URL

https://codesandbox.io/p/sandbox/ancient-dew-wychg6

### Expected Behavior

You should be able to write arbitrary XML in a tiptap editor. Ideas:
- In the old [community made tiptap markdown](https://github.com/aguingand/tiptap-markdown), it would automatically transform invalid XML into escaped characters. This could be a potential option, another idea could be just to sanitize and remove invalid XML when set.

Image

### Dependency Updates

nehaaprasad:
fix: #7256 

**Problem:**
When markdown contains XML tags that aren't recognized by any extension, `generateJSON` may create block nodes (like paragraphs) in inline contexts, causing ProseMirror to throw: "Invalid content for node paragraph".

**Solution:**
- Added detection for block nodes in inline HTML parsing contexts
- When block nodes are detected, fall back to converting the HTML to plain text
- This preserves the content while preventing the crash


ElPicador:
Any news on merging this fix? 🙏 

We have the same issue and would gladly have it merged, instead of relying on a monkey patch

kaihenthorniwane:
https://github.com/user-attachments/assets/57f0e429-4345-47e0-8dad-734ac2b52dc0

seems to work well in the preview!

kaihenthorniwane:
Not sure if I have the credentials to approve this though?

packages/markdown/src/MarkdownManager.ts
```diff
@@ -32,6 +32,22 @@ export class MarkdownManager {
   private nodeTypeRegistry: Map<string, MarkdownExtensionSpec[]>
   private indentStyle: 'space' | 'tab'
   private indentSize: number
+  private blockLikeNodeNames: Set<string> = new Set([
```
bdbch, line 35: Not sure if we should hard-code block nodes here. If you have access to the extensions, you should be able to get this information from the extension configuration I think (or in a last resort from the nodes themselves, this should **not** be hardcoded ever).

packages/markdown/src/MarkdownManager.ts
```diff
@@ -751,60 +767,68 @@ export class MarkdownManager {
...
    */
   private parseHTMLToken(token: MarkdownToken): JSONContent | JSONContent[] | null {
-    const html = token.text || token.raw || ''
+    const html = (token.raw ?? token.text ?? '').toString()
 
     if (!html.trim()) {
       return null
     }
-
-    // Check if we're in a server-side environment (no window object)
```
bdbch, line 774: Is there a reason why this was removed? I don't think this is related to XML tags at all?

packages/markdown/src/MarkdownManager.ts
```diff
@@ -751,60 +767,68 @@ export class MarkdownManager {
...
-      }
-    }
-
-    // Use generateJSON to parse the HTML using extensions' parseHTML rules
     try {
       const parsed = generateJSON(html, this.baseExtensions)
 
-      // If the result is a doc node, extract its content
       if (parsed.type === 'doc' && parsed.content) {
-        // For block-level HTML, return the content array
```
bdbch, line 802: I think keeping some of those comments is a good idea just to understand why something is returned early / for what purpose

bdbch (changes requested):

packages/markdown/src/MarkdownManager.ts
```diff
@@ -32,6 +32,22 @@ export class MarkdownManager {
...
+    'orderedList',
+    'listItem',
+    'taskList',
+    'taskItem',
+    'codeBlock',
+    'horizontalRule',
+    'table',
+    'tableRow',
+    'tableCell',
+    'tableHeader',
```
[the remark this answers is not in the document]
bdbch: See my comment above. I think we should not hard-code blocklevel nodes.

bdbch:
Any update here?

bdbch:
Closing this for now until the contributor picks it up again.

PR #7261, opened 2025-11-24
fix: trim text from end when character limit exceeded

### LINKED ISSUE #7254: Character count extension trims text from the beginning instead of the end when content exceeds limit
### Affected Packages

character-count

### Version(s)

*

### Bug Description

When adding the [character count extension](https://tiptap.dev/docs/editor/extensions/functionality/character-count) to an existing project where the editor content already exceeds the maximum allowed length, the plugin cuts the text as expected. However, the text is cut from left to right (beginning to end), so the starting text disappears instead of trimming from the end as expected.

### Browser Used

Chrome

### Code Example URL

https://codesandbox.io/p/devbox/twilight-leaf-kfmcqv

### Expected Behavior

**Current:** 
**…cutoff** you could build one of the biggest social networks on that idea.

**Expected:**
Let‘s make sure people can’t write more than 100 characters. I bet you could build one **…cutoff**

I think the issue needs to be fixed inside this if condition: https://github.com/ueberdosis/tiptap/blob/e4b95fab00f9ac9f286e82bb0d179043c1f135e7/packages/extensions/src/character-count/character-count.ts#L121

### Dependency Updates

nehaaprasad:
fix: #7254 

### Changes Overview

- Fixed character count extension to trim from the end instead of the beginning when initial content exceeds the limit.

### Implementation Approach

- Updated the trimming logic to find the position where the limit is reached and delete from that position to the end, preserving the beginning text.

### DIFF
--- a/packages/extensions/src/character-count/character-count.ts
+++ b/packages/extensions/src/character-count/character-count.ts
@@ -133,21 +133,17 @@ export const CharacterCount = Extension.create<CharacterCountOptions, CharacterC
                 } else {
                   const remainingChars = limit - charCount
                   let charIndex = 0
-                  let currentCharCount = 0
 
                   for (let i = 0; i < nodeText.length; i++) {
                     const substring = nodeText.slice(0, i + 1)
                     const substringCharCount = this.options.textCounter(substring)
-                    
                     if (substringCharCount > remainingChars) {
                       break
                     }
-                    
                     charIndex = i + 1
-                    currentCharCount = substringCharCount
                   }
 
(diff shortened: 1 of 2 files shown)


bdbch (changes requested):
See the copilot comments. There are a few functional and performance related problems with the PR (bad-scaling loops, positions being calculated off, etc.)

I would also like to have a test for this.

Feel free to add a vitest unit test in `packages/extensions/src/character-count/__test__` that tests those cases.

nehaaprasad:
@bdbch 
please check the updates!

PR #7262, opened 2025-11-24
fix(link): prevent crash when class is undefined in HTMLAttributes

### LINKED ISSUE #7214: "No value supplied for attribute class" error when passing undefined class to HTMLAttributes in Link
### Affected Packages

extension-link

### Version(s)

3.10.5

### Bug Description

When you pass an undefined class to `HTMLAttributes` when configuring a link, it causes the following error:
`No value supplied for attribute class`

eg.
```
Link.configure({
  HTMLAttributes: {
    class: undefined
  },
}),
```

### Browser Used

Chrome

### Code Example URL

https://codesandbox.io/p/sandbox/trusting-glitter-nyl9g3

### Expected Behavior

Should throw an error but not crash when passing an undefined class

_No response_

### Dependency Updates

nehaaprasad:
- fixed a crash that occurred when passing `undefined` as the `class` value (or other attributes) in `HTMLAttributes` when configuring the Link extension. The error "No value supplied for attribute class" was thrown because ProseMirror doesn't accept `undefined` values for HTML attributes.

### Implementation Approach

1. **Filter undefined values in `renderHTML()`**: Added filtering to remove `undefined` values from merged attributes before passing them to ProseMirror's renderer. This prevents the crash by ensuring only valid attribute values are rendered.

2. **Handle undefined defaults in `addAttributes()`**: Added nullish coalescing operator (`??`) to convert `undefined` values to `null` for `target`, `rel`, and `class` attribute defaults, ensuring consistent behavior.

## Related Issues

fixes : #7214


bdbch (changes requested):
See the comments from copilot. The code should not be duplicated.

bdbch (changes requested):
To come back to this.

**Code Duplication**

The filtering logic is duplicated in both branches of the conditional. Consider extracting it into a utility function you can reuse in both conditionals.

**Test Coverage**

Having a unit test for this behavior would be great. You can use vitest to add it. A simple "should not crash when HTMLAttributes contain undefined values" test should be enough.

nehaaprasad:
@bdbch 
please check the updates!

PR #7277, opened 2025-11-27
Fix TS imports for NodeNext

bdbch:


bdbch:
Those comments are wrong as the inherited tsconfig dictates the base url (as the root of the project).

Issue #7299, opened 2025-12-03
Text color does not affect underline


Clindbergh:
### Affected Packages

core, static-renderer

### Version(s)

3.12

### Bug Description

Adding a color to an underlined text does not change the color of the underline. This is because the underline is rendered outside.

See the failing test case (PR).

```
 FAIL  packages/static-renderer/__tests__/json-string.spec.ts > static render json to string (with prosemirror) > supports underline and spans
AssertionError: expected '<p><u><span style="color: #ED1F1F">Un…' to be '<p><span style="color: #ED1F1F"><u>Un…' // Object.is equality

Expected: "<p><span style="color: #ED1F1F"><u>Underlined</u></span></p>"
Received: "<p><u><span style="color: #ED1F1F">Underlined</span></u></p>"

 ❯ packages/static-renderer/__tests__/json-string.spec.ts:293:18
    291|     })
    292| 
    293|     expect(html).toBe('<p><span style="color: #ED1F1F"><u>Underlined</u></span></p>')
       |                  ^
    294|   })
    295| 

```

### Browser Used

Chrome

### Expected Behavior

The underline is rendered outside.

### Dependency Updates

- [x] Yes, I've updated all my dependencies.

Clindbergh:
This was already discussed in #4742 (@nperez0111).

Issue #4742, opened 2024-01-03
[Bug]: Changing text color does not affect underline


Daniel-Rauh:
### Which packages did you experience the bug in?

react, extension-underline, extension-color

### What Tiptap version are you using?

2.1.12

### What’s the bug you are facing?

When underlining text that had it's color changed or when changing the color of underlined text the underline is always black. There should be an option to make it the same color as the text.

### What browser are you using?

Chrome

### What did you expect to happen?

Expected the underline color to be the same as the text color

### Did you update your dependencies?

- [X] Yes, I’ve updated my dependencies to use the latest version of all packages.

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

MaximilianKresse:
I stumbled into the same issue today. To add on this - the problem is the order of the html tags:
![image](https://github.com/ueberdosis/tiptap/assets/545671/c8548d52-12bf-45fa-b2ac-d5f19c4a0b8f)

The underline tag should be inside span tag not around it.

laurenga:
This affects me too. I also noticed it happens with strikethrough.

ymc9:
You can check this comment here

https://github.com/ueberdosis/tiptap/discussions/4689#discussioncomment-10131672

nperez0111:
hm, maybe we could just change the priority of the extensions to have them apply in a different order? I forgot how the order is set

EskelCz:
@nperez0111 Changing the order of extensions did the trick for me, thanks!

nperez0111:
> @nperez0111 Changing the order of extensions did the trick for me, thanks!

What order did you set so we can resolve this in future versions?

EskelCz:
@nperez0111 maybe my config was bad to start with
I had this:
```
const extensions = [
      StarterKit,
      Underline,
      TextStyle,
      ...
```
change it to this:
```
const extensions = [
      StarterKit,
      TextStyle,
      Underline,
      ...
```

EskelCz:
@nperez0111 Just found out that strikethrough suffers with the same exact issue. Had to put TextStyle extension first.

nperez0111:
This is now resolved in v2.6.0 (I gave text style a higher priority)

PR #7300, opened 2025-12-03
:cop: Add failing test case for underline with text color in static r…

Clindbergh:
…enderer

## Changes Overview

Added a test case for #7299.

## Checklist

### DIFF
--- a/packages/extension-underline/src/underline.ts
+++ b/packages/extension-underline/src/underline.ts
@@ -38,6 +38,11 @@ declare module '@tiptap/core' {
 export const Underline = Mark.create<UnderlineOptions>({
   name: 'underline',
 
+  /**
+   * Increase the priority to ensure that u elements are rendered within spans, so their text colors are applied.
+   */
+  priority: 101,
+
   addOptions() {
     return {
       HTMLAttributes: {},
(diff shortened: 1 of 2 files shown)


Clindbergh:
The proposed solution here is similar to what was done in [#5457](https://github.com/ueberdosis/tiptap/pull/5457) (for #4742) by @nperez0111 for v2.x.x.

Clindbergh:
This may have to be tested in the frontend too. I tried the same fix on my existing version:

```typescript
  const underline = Underline;
  underline.config.priority = 101;
  return underline;
```

...which causes the editor to behave the same wrong way as the static renderer (underline v 3.4.4).

bdbch:
I'm not sure if we should merge a failing test only without a fix. Do you mind fixing it or should we take this PR/branch over?

Clindbergh:
> I'm not sure if we should merge a failing test only without a fix. Do you mind fixing it or should we take this PR/branch over?

I think my second commit _did_ fix the test. However there is no test that for the same behavior in the frontend/visual editor. I am afraid that it won't work there (see my previous comment).

Clindbergh:
> I'm not sure if we should merge a failing test only without a fix. Do you mind fixing it or should we take this PR/branch over?

@bdbch Of course you may take over the PR. Anyway, before I would be able to work on it I would be glad to receive some feedback on the linked issue #7299 which is what I am trying to show here. Can you quickly confirm that this is actually an issue and not just a misconfiguration on my side? For reproduction: Underline colored text and render it with the static renderer.

PR #7311, opened 2025-12-05
fix(extension-unique-id): check next node attribute before transaction modification

amarnath-memfold:
## Changes Overview

Fixed a bug in the `unique-id` extension where in series of changed nodes in a transaction, if in the entire document, you had just an empty node with an id followed by any node with an existing id, the second node's id was being regenerated. I have added a check to check if the node followed by an empty node has an existing id, in which case, this regeneration should not happen. 

## Implementation Approach

The bug was in the `appendTransaction` handler when processing empty nodes with a next node. Previously, the code would:
1. Modify the next node's attributes via `tr.setNodeMarkup()`
2. Then check if the next node already had an attribute value

This was incorrect because the check should happen before modifying the transaction. The fix:
- Moves the check to occur before `tr.setNodeMarkup()` is called
- Uses `tr.doc.nodeAt(nextNode.pos)` to check the current state in the transaction document
- Returns early if the next node already has a different ID, preventing incorrect ID assignment

This ensures we're checking the actual current state of the node in the transaction document before making any modifications.

## Testing Done

- Verified that linting passes (`pnpm run lint`)
- Code formatting verified with Prettier
- manually tested the scenario listed above

## Verification Steps

1. Create an editor instance with the `UniqueID` extension configured for specific node types
2. Create a scenario with an empty node(with a unique id) followed by a node that already has a unique ID
3. Verify that the empty node gets a new unique ID without affecting the next node's existing ID
4. Verify that if the next node already has a different ID, the code returns early and doesn't modify it

## Additional Notes

The fix is minimal and focused - it only reorders the check to happen before the transaction modification. This prevents race conditions where the node state could be checked after it's already been modified within the same transaction.

## Checklist


amarnath-memfold:
Hey @bdbch, requesting your review.

amarnath-memfold:
Hey @KevinHosey, @bdbch seems to be away. Would you be able to look into this PR?

bdbch:
I'm there just catching up on a few things over the last few days of the year. I'll try to review your PR soon

amarnath-memfold:
Thanks @bdbch !

Issue #7349, opened 2025-12-18
Marks (bold/italic/color) are not reset on Enter by default


romancavry:
### Affected Packages

core, react

### Version(s)

3.13.0

### Bug Description

In Tiptap, when pressing Enter inside a paragraph that has active marks (bold, italic, color, etc.), the marks are preserved in the newly created paragraph. There is no built-in option to automatically reset formatting on Enter, which can be surprising for users expecting a fresh paragraph to start with default styling.

### Browser Used

Chrome

### Code Example URL

https://tiptap.dev/docs/examples/basics/default-text-editor

### Expected Behavior

Steps to reproduce:

1. Type text with bold styles.
2. Press Enter
3. See preserved bold mark on new line.

https://github.com/user-attachments/assets/08337490-ed75-47a0-8436-d1505890b699

### Additional Context (Optional)

It would be helpful if Tiptap provided a built-in option like resetMarksOnEnter: true/false to handle this behavior natively.

### Dependency Updates

- [x] Yes, I've updated all my dependencies.

bdbch:
Hi @romancavry,

Thanks for reporting this! This is actually intended behavior. By default, Tiptap (like ProseMirror and most rich text editors) preserves active marks (bold, italic, etc.) when you split a block. This ensures that users typing a bold header or list can hit Enter and continue typing in the same style without interruption.

However, if you prefer to enforce a "clean slate" on every new line, you can achieve this globally by replacing the default Enter handler. You don't need to modify individual nodes; you can simply create a small extension to override the behavior editor-wide.

Here is how you can create a custom extension to handle this:

```typescript
import { Extension } from '@tiptap/core'

export const ResetMarksOnEnter = Extension.create({
  name: 'resetMarksOnEnter',

  addKeyboardShortcuts() {
    return {
      Enter: () => {
        // splitBlock({ keepMarks: false }) ensures marks are not carried over
        return this.editor.commands.splitBlock({ keepMarks: false })
      },
    }
  },
})

```

Then just add `ResetMarksOnEnter` to your editor's extension list.

I'm closing this issue for now as the default behavior works as designed, but the snippet above should solve your specific use case!

romancavry:
Thank you for your reply @bdbch!

I've implemented your ResetMarksOnEnter extension, but I'm encountering some compatibility issues with certain block types.
Issues observed:

[...]

PR #7356, opened 2025-12-19
feat: Add Optional Mark Reset on Enter

jordan-sussman:
## Changes Overview
Closes - https://github.com/ueberdosis/tiptap/issues/7349
Adds support to reset marks on `enter` as an optional boolean prop named `resetMarksOnEnter`.

## Implementation Approach
 Adds resetMarksOnEnter as a global editor option (similar to enableInputRules and enablePasteRules), implemented in the keymap extension's Enter handler. This boolean defaults to false. When true, marks are unset after splitting a block on enter. IE:

```
  const editor = new Editor({
    extensions: [StarterKit],
    resetMarksOnEnter: true,
  })
```

## Testing Done
1. Add a paragraph with text and active marks (bold, italic, etc.)
2. Press `enter` to create a new line and add text to the new line

## Verification Steps
If you've set `resetMarksOnEnter` to `true`, the marks should be unset, otherwise, when `false` or not provided, the marks continue on the new line

## Checklist

## Related Issues
https://github.com/ueberdosis/tiptap/issues/7349


bdbch:
Hey @jordan-sussman, thanks for working on this!  

After reviewing the implementation, I think we should take a different approach. Rather than adding `resetMarksOnEnter` to the Paragraph extension specifically, **I'd prefer to implement this as a global editor-level option**. 

Why a global option is better:

1. **Consistency**:  Users would expect this behavior everywhere (blockquotes, list items, etc.), not just in top-level paragraphs
2. **Simpler API**: One configuration point instead of per-extension
3. **Avoids conflicts**: The current approach adds an Enter handler at the Paragraph extension level which could interfere with lists, code blocks, etc. 
4. **More flexible**: Users could enable/disable the option globally

Would you be fine looking into this - or what are your thoughts about that?

jordan-sussman:
@bdbch thanks - that's a great recommendation! The change for that is up and ready for another look.

jordan-sussman:
Closing since linked issue has been

Issue #7347, opened 2025-12-17
Images cannot be selected when enableClickSelection of the Link extension is set to true


rktyt:
### Affected Packages

core, react, extension-link, extension-image

### Version(s)

3.13.0

### Bug Description

When using both the Image extension and the Link extension,
setting enableClickSelection to true on the Link extension makes it very difficult to select an image.
It requires clicking the image about five times in a row to select it.

### Browser Used

Chrome

### Code Example URL

https://codesandbox.io/p/sandbox/billowing-framework-y6kjk2

### Expected Behavior

The expected behavior is that an image can be selected with a single click.
This issue does not occur in v3.12.1, so the behavior should be the same as in v3.12.1.

### Dependency Updates

- [x] Yes, I've updated all my dependencies.

cdunn07:
Just ran into this same issue when upgrading to v3.14

Aslam97:
Thanks for reporting this. I’ll try to reproduce and investigate the interaction between `enableClickSelection` on the Link extension and the Image extension

alexvcasillas:
This was a regression in v3.13.0 and has already been fixed in **v3.16.0** (#7359), which restored the guard that lets clicks on non-link elements fall through to the editor's default selection. Upgrading `@tiptap/extension-link` (and the rest of your Tiptap packages) to v3.16.0 or later resolves it, so a single click selects the image again.

To make sure this exact path can't silently break again, I added a regression test covering `enableClickSelection: true` in #8110.

Closing as resolved. Please reopen if you still see it on the latest version. 🙏

PR #7359, opened 2025-12-22
fix(extension-link): prevent click handler from blocking non-link element selection

Aslam97:
## Changes Overview

Fixes an issue where clicking on images (and other non-link elements) required multiple clicks to properly select them.

## Implementation Approach

- Move link element detection to the beginning of the click handler
- Early return `false` when the clicked element is not a link

## Testing Done

Added unit tests

## Checklist

## Related Issues

#7347


arnaugomez (approved):

packages/extension-link/src/helpers/clickHandler.ts
```diff
@@ -23,6 +23,26 @@ export function clickHandler(options: ClickHandlerOptions): Plugin {
...
         }
 
+        let link: HTMLAnchorElement | null = null
+
+        if (event.target instanceof HTMLAnchorElement) {
+          link = event.target
+        } else {
+          let a = event.target as HTMLElement
+
+          while (a.nodeName !== 'DIV' && a.parentNode) {
```
bdbch: This would break if I have a custom node that renders a div, right?
Why is it doing a specific check on DIV here?

bdbch (commented):
Quick question from my side.

packages/extension-link/src/helpers/clickHandler.ts
bdbch: Or rather - what does this check accomplish where you'd need to specifically check for a wrapping div?

Aslam97: I have change it to use `closest` instead

packages/extension-link/src/helpers/clickHandler.ts
```diff
@@ -23,6 +23,18 @@ export function clickHandler(options: ClickHandlerOptions): Plugin {
           return false
         }
 
+        let link: HTMLAnchorElement | null = null
+
+        if (event.target instanceof HTMLAnchorElement) {
+          link = event.target
+        } else {
+          link = (event.target as HTMLElement).closest('a')
```
bdbch: Just to make sure to not break out of the prosemirror div, I'd limit this lookup to the `editor.view.dom` element. Otherwise this is fine for me

Aslam97: Added

bdbch (approved):

PR #7362, opened 2025-12-22
fix: race condition in ReactRenderer

YousefED:
## Changes Overview

There's a race condition that can occur in React StrictMode, that causes an already destroyed renderer to be re-added. This happens because `render` is called after `destroy`, and `render` then calls `setRenderer`.

## Implementation Approach

Keep a `destroyed` flag and don't call `render` for already destroyed components

## Testing Done

Tested manually

## Checklist

### DIFF
--- a/.changeset/thick-months-train.md
+++ b/.changeset/thick-months-train.md
@@ -2,4 +2,4 @@
 '@tiptap/react': patch
 ---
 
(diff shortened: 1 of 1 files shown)


YousefED:
@bdbch I see you self-assigned this. Let me know if there's anything you need from me to move this forward!

PR #7384, opened 2026-01-06
Add shouldShow option for suggestions to control visibility / trigger

bdbch:
## Changes Overview

This PR adds a new `shouldShow` callback to the Suggestion utility. This allows developers to filter when suggestions are displayed, specifically to prevent autocomplete popups (like mentions) from opening for remote users in collaborative environments.

## Implementation Approach

- Added a `shouldShow` property to SuggestionOptions.
- Modified the apply method in the suggestion plugin state to optionally invoke `shouldShow` when a match is found.
- Passed the current `transaction` to the callback via `props`, enabling the use of helpers like `isChangeOrigin`.

## Testing Done

- Added unit tests to verify that `shouldShow` correctly filters suggestion activation.
- Verified that the `transaction` object and editor instance are correctly passed to the callback.

## Verification Steps

Reviewers can verify this by configuring a suggestion plugin (e.g., via the Mention extension) with a `shouldShow` callback:

```js
Mention.configure({
  suggestion: {
    shouldShow: ({ props }) => {
      // Return false to prevent the suggestion from showing
      return isChangeOrigin(props.transaction)
    },
  },
})
```

## Additional Notes

Documentation has been updated in suggestion.mdx and mention.mdx to include a guide on using this feature in collaborative setups.

## Checklist

## Related Issues

Resolves the issue where suggestions pop open for remote users in collaborative documents.

### DIFF
--- a/packages/suggestion/src/suggestion.ts
+++ b/packages/suggestion/src/suggestion.ts
@@ -24,6 +24,7 @@ export interface SuggestionOptions<I = any, TSelected = any> {
    * @param props.text The current suggestion text.
    * @param props.transaction The current transaction.
    * @returns {boolean}
+   * @example ({ transaction }) => isChangeOrigin(transaction)
    */
   shouldShow?: (props: {
     editor: Editor


packages/suggestion/src/suggestion.ts
```diff
@@ -14,6 +14,16 @@ export interface SuggestionOptions<I = any, TSelected = any> {
...
 
+  /**
+   * A function that returns a boolean to indicate if the suggestion should be active.
+   * This is useful to prevent suggestions from opening for remote users in collaborative environments.
+   * @param props The props object.
+   * @param props.editor The editor instance.
+   * @param props.range The range of the suggestion.
+   * @returns {boolean}
+   */
+  shouldShow?: (props: { editor: Editor; range: Range; props: any }) => boolean
```
arnaugomez: Could the type of the `props` property be made more specific, instead of `any`? If not, could you add a description (a `TSDoc` comment) to the `props` property, so the developer knows what it is and has the context to know what data it contains?

Issue #7422, opened 2026-01-16
Floating element fails to hide on scroll container boundary


ikrong:
### Affected Packages

extension-bubble-menu, extension-floating-menu

### Version(s)

3.15.3

### Bug Description

The hide middleware of Floating UI returns promise-resolved data to mark the visibility state of the floating element. However, the extension-bubble-menu and extension-floating-menu packages do not handle this resolved data at all — this seems to be an oversight or bug.

<img width="801" height="613" alt="Image" src="https://github.com/user-attachments/assets/32ae2621-8e3a-4795-91a1-e97aab25db45" />

<img width="1759" height="829" alt="Image" src="https://github.com/user-attachments/assets/eb360033-1c2d-49b4-9d68-d02a2a787e16" />

<img width="981" height="586" alt="Image" src="https://github.com/user-attachments/assets/feec6801-d979-4d59-b9fa-e1b8db27b0b7" />

### Browser Used

Chrome

### Code Example URL

https://codesandbox.io/p/sandbox/tiptap-floating-j93d2g

### Expected Behavior

I created a demo that integrates the hide middleware of Floating UI. Here are the reproduction steps and observed issues:

1. Select some text within the editor
2. Scroll the editor’s scrollbar all the way to the bottom

The floating element should disappear as soon as it touches the edge of the scroll container.

Instead of hiding at the scroll container’s edge, the floating element remains visible and only disappears when it reaches the edge of the window viewport.

### Dependency Updates

- [x] Yes, I've updated all my dependencies.

nehaaprasad:
Hello! @ikrong 
please check the PR

ikrong:
Hello @naaa760,
Thanks for fixing this issue! Floating UI’s hide middleware comes with both the referenceHidden and escaped strategies, and each of these strategies resolves to slightly different states. It would be better to add additional condition checks here.

[click here to view floating-ui hide middleware](https://github.com/floating-ui/floating-ui/blob/e7d3f00df9cdb037bb116f098e3e4695a11d8bfd/packages/core/src/middleware/hide.ts#L32)

<img width="830" height="927" alt="Image" src="https://github.com/user-attachments/assets/9b44f3fd-afd5-48e5-a50d-730318b19844" />

bdbch:
I created a PR that checks both if the reference is hidden or the reference element escaped.

PR #7423, opened 2026-01-16
fix: handle hide middleware visibility in bubble-menu and floating-menu

nehaaprasad:
### Changes Overview

- fixed bubble-menu and floating-menu not hiding at scroll container boundaries. The hide middleware data was collected but not applied to control visibility.

fix: #7422

### Implementation Approach

- Added handling for `middlewareData.hide.referenceHidden` in the `updatePosition()` method to set visibility based on scroll container boundaries.

### Testing Done

- Tested with hide middleware enabled. Floating elements now hide when reaching scroll container edges, not just viewport edges.

### Verification Steps

1. Enable hide middleware: `options={{ hide: {} }}`
2. Select text and scroll the editor container
3. Verify the floating element hides at container boundaries


bdbch:
Closed in favor for https://github.com/ueberdosis/tiptap/pull/7444

PR #7424, opened 2026-01-16
fix(react): handle undefined options.current in useEditor callbacks

### LINKED ISSUE #7417: useEditor `onUpdate` callback not called when `options.current` is `undefined` (with Collaboration extension)
### Affected Packages

react

### Version(s)

3.15.3

### Bug Description

- TipTap version: 3.15.3
- `editor.options.current` is `undefined`
- The `onUpdate` wrapper silently fails (no error, just doesn't call the callback)
- Only happens with Collaboration extension enabled
- Workaround: Use `editor.on('update', handler)` in a `useEffect`

### Browser Used

Chrome

### Code Example URL

_No response_

### Expected Behavior

The `options.current` should be set by TipTap's `useEditor` hook, but it's `undefined`.

_No response_

### Dependency Updates

nehaaprasad:
### Changes Overview

Fixed `onUpdate` and other callbacks not being called when `options.current` is `undefined`, especially with the Collaboration extension.

fix: #7417

## Implementation Approach

Added `wrapOptionsWithCallbacks` helper that safely wraps all callbacks using optional chaining (`this.options.current?.onUpdate?.(...args)`). Applied in `createEditor()`, `onRender()`, and `scheduleDestroy()` to ensure callbacks always check `this.options.current` at call time.

## Testing Done

Verified `onUpdate` works when `options.current` is defined or `undefined`, and tested with Collaboration extension.

## Verification Steps

1. Use `useEditor` with `onUpdate` and Collaboration extension
2. Verify `onUpdate` is called on content changes
3. Check console for errors

## Additional Notes

- Prevents silent callback failures when `options.current` is temporarily `undefined`.

## Checklist

## Related Issues

Fixes #7417


sahiee-dev:
# Deep Dive Analysis: Root Cause Investigation

Thanks for working on this fix, @naaa760! Here's my analysis of the underlying issue.

> **Note:** I’m not proposing or submitting code changes here.  
> The code snippets below are only to illustrate the existing control flow and lifecycle timing based on reading the current implementation. This is purely a root-cause analysis and discussion.


## Why `options.current` Should Never Be Undefined

In `useEditor.ts`:

```typescript
export function useEditor(options: UseEditorOptions = {}, deps: DependencyList = []): Editor | null {
  const mostRecentOptions = useRef(options)  // initialized with options object
  mostRecentOptions.current = options        // updated on every render
  const [instanceManager] = useState(() => new EditorInstanceManager(mostRecentOptions))
}
```

The `options` ref is initialized with `useRef(options)` where `options` defaults to `{}`, so `mostRecentOptions.current` should always be defined.

## How Can `.current` Become Undefined?

### 1. React StrictMode Race Condition
- `EditorInstanceManager` constructor calls `scheduleDestroy()` immediately (line 76)
- This sets a 1ms timeout accessing `this.options.current`
- During StrictMode, React unmounts/remounts quickly—the timeout may fire between mount cycles

### 2. Async Event Timing with Yjs/Provider
- The Collaboration extension's `ySyncPlugin` fires events asynchronously during sync
- These can trigger `onUpdate` after React begins unmounting but before editor destruction
- The component's ref may be in an inconsistent state at this moment

### 3. Provider-Specific Behavior
- Third-party providers (like Liveblocks) may trigger sync events at unpredictable lifecycle moments

## Why This PR Is Valuable

Despite not solving the root cause, this PR is correct because:

1. **Defensive pattern**: Callbacks reference `this.options` via closure. If `.current` is undefined when called, optional chaining is the right response
2. **Root cause may be unsolvable**: React's StrictMode timing and external provider behavior are outside TipTap's control

## Suggestions to Improve This PR

### 1. Add Dev-Mode Warning

```typescript
private wrapOptionsWithCallbacks(options: Partial<EditorOptions>): Partial<EditorOptions> {
  return {
    ...options,
    onUpdate: (...args) => {
      if (!this.options.current) {
        if (isDev) {
          console.warn(
            '[TipTap] onUpdate callback skipped: options.current is undefined. ' +
            'This may occur during React StrictMode cleanup or with async collaboration providers. ' +
            'Consider using editor.on("update", handler) in a useEffect as a workaround.'
          )
        }
        return
      }
      return this.options.current.onUpdate?.(...args)
    },
    // ... same pattern for other callbacks
  }
}
```

### 2. Document the Workaround

The `editor.on('update', handler)` pattern should be documented as recommended for Collaboration extensions:

```typescript
useEffect(() => {
  if (!editor) return
  
  const handler = ({ editor, transaction }) => {
    // This bypasses options.current entirely
    console.log('Content updated')
  }
  
  editor.on('update', handler)
  return () => editor.off('update', handler)
}, [editor])
```

## Questions for Maintainers

1. Is the 1ms timeout in `scheduleDestroy()` specifically for React 18 StrictMode? Would increasing it help?
2. Should the Collaboration extension documentation recommend `editor.on()` instead of hook-based callbacks?
3. Would it make sense to track this as a separate "root cause investigation" issue?

---

**Conclusion**: The optional chaining approach is the right defensive measure. Adding dev-mode warnings would improve it by helping developers understand why callbacks aren't firing.

packages/react/src/useEditor.ts
```diff
@@ -127,30 +127,47 @@ class EditorInstanceManager {
...
+      onUpdate: (...args: Parameters<NonNullable<EditorOptions['onUpdate']>>) =>
+        this.options.current?.onUpdate?.(...args),
+      onContentError: (...args: Parameters<NonNullable<EditorOptions['onContentError']>>) =>
+        this.options.current?.onContentError?.(...args),
+      onDrop: (...args: Parameters<NonNullable<EditorOptions['onDrop']>>) =>
+        this.options.current?.onDrop?.(...args),
+      onPaste: (...args: Parameters<NonNullable<EditorOptions['onPaste']>>) =>
+        this.options.current?.onPaste?.(...args),
+      onDelete: (...args: Parameters<NonNullable<EditorOptions['onDelete']>>) =>
+        this.options.current?.onDelete?.(...args),
```
arnaugomez: Nitpick: I believe the parameters type (`...args`) can be inferred and does not need to be declared explicitly here, is it correct?

arnaugomez (commented):
I see that before this PR the callbacks were added to the options only when the Editor was created, now they are re-added when the Editor is updated (when the component is re-rendered). I like this approach.

However, I'm not sure if this fixes the issue (See [this comment](https://github.com/ueberdosis/tiptap/issues/7417#issuecomment-3764592426)). I think we should first investigate why `options.current` becomes `undefined` in the first place.

nehaaprasad:
@arnaugomez 
sir i fixed the root cause by normalizing options in the hook with options ?? {} before assigning to the ref, so the ref is never undefined and callbacks (including onUpdate) always run

PR #7431, opened 2026-01-20
fix(extension-image): support showCaption and figcaption in NodeView …

### LINKED ISSUE #7425: Content loss when typing in React NodeView with conditionally rendered NodeViewContent
### Affected Packages

@tiptap/react, @tiptap/extension-image, @tiptap/core

### Version(s)

3.15.3

### Bug Description

When a custom node extension has `content: "inline*"` and the `NodeViewContent` component is conditionally rendered in the React NodeView, users can still position their cursor inside the node's content area. However, any text typed in this state is not properly stored in the document.

  **Steps to reproduce:**

  1. Create a custom Image extension with `content: "inline*"` and a `showCaption` attribute
  2. In the React NodeView, conditionally render `NodeViewContent` only when `showCaption` is true:

  ```tsx
  // Extension
  export const Image = TiptapImage.extend({
    content: "inline*",
    addAttributes() {
      return {
        showCaption: { default: false },
      }
    },
  })

  // NodeView
  function ImageNodeView({ node, editor }) {
    const showCaption = node.attrs.showCaption
    const hasContent = node.content.size > 0
    const shouldShowCaption = showCaption || hasContent

    return (
      <NodeViewWrapper>
        <img src={node.attrs.src} />
        {editor?.isEditable && shouldShowCaption && (
          <NodeViewContent as="div" />
        )}
      </NodeViewWrapper>
    )
  }

  3. Insert the node with showCaption: false
  4. Use arrow keys or click to position cursor next to the image (cursor enters the node's content area)
  5. Type some text - it appears visually on screen
  6. Check editor.getJSON() - the typed text is NOT in the document (content.size is 0)

### Browser Used

Chrome

### Code Example URL

_No response_

### Expected Behavior

 One of the following:

  1. When NodeViewContent is not rendered, ProseMirror should prevent the cursor from entering the node's content area
  2. Text typed should be stored correctly regardless of whether NodeViewContent is rendered
  3. TipTap should provide a warning when a NodeView's schema has content but NodeViewContent is not in the DOM

 This issue is particularly problematic for image nodes with optional captions where:
  - Captions are allowed but hidden by default
  - The caption area should only appear when the user explicitly enables it

  The current behavior creates a confusing UX where users can type text that silently disappears when the document is saved or reloaded.

  Workaround attempted:
  - Always rendering NodeViewContent fixes the content storage issue, but breaks cursor navigation (cursor gets "trapped" inside the image node)
  - Adding arrow key handlers to escape the node helps, but mouse click behavior is still problematic

### Dependency Updates

unseoJang:
First of all, apologies for my previous attempt which only updated the demo files and did not address the root cause in the actual extension code. That was a mistake on my part.

This PR is directly connected to https://github.com/ueberdosis/tiptap/issues/7425#issuecomment-3767091574 and provides a real fix in the extension-image package to prevent text loss (fix #7425).

## Changes Overview

- Added the `showCaption` attribute to the image node schema.
- Updated the NodeView logic to support a `<figure>` and `<figcaption>` structure when `showCaption` is true.
- Ensured that text typed before/after an image node is preserved when the image is deleted, addressing the original issue.

## Implementation Approach

- Modified `addAttributes` in `extension-image` to include `showCaption`.
- Changed the NodeView to render a `<figure>` wrapper and a `<figcaption>` (with `data-node-view-content`) when `showCaption` is true.
- Kept the resizable NodeView logic for images with resizing enabled, and only applied the new structure for non-resizable images.

## Testing Done

- Manually tested in the demo: inserted an image with `showCaption: false`, typed text before/after the image, deleted the image, and confirmed the text remains in the document.
- Also tested with `showCaption: true` to ensure figcaption content is editable and preserved.

## Verification Steps

1. Insert an image node with or without `showCaption`.
2. Place the cursor before or after the image and type some text.
3. Delete the image node.
4. Verify that the text remains in the document and is not lost.

## Additional Notes

- This change affects the core extension, not just the demo, so all users benefit from the fix.
- A changeset will be included to document this user-facing change.

## Checklist

## Related Issues

- https://github.com/ueberdosis/tiptap/issues/7425


unseoJang:
Added showCaption as a node attribute (default false) and enabled the image node to hold inline content (content: 'inline*') so caption text can be stored inside the node.

Updated HTML parsing/serialization to support a <figure><figcaption/></figure> structure:

parseHTML() now recognizes <figure> and reads attributes from the nested , while treating <figcaption> as the node’s content container via contentElement: 'figcaption'.

renderHTML() now preserves caption content during export: it wraps in <figure> not only when showCaption is true, but also when the node actually has caption content (node.content.size > 0), using a ProseMirror “content hole” (0) in <figcaption> so the stored caption text is emitted to HTML instead of being dropped.

Reworked the NodeView for the non-resizable case to avoid the #7425 class of content-loss bugs by keeping a stable contentDOM:

The NodeView always creates a <figcaption> and always returns it as contentDOM; showCaption only toggles visibility (display) rather than conditionally removing contentDOM. This avoids “conditional contentDOM/NodeViewContent” patterns that can cause typed text to appear visually but not be reflected in the document state.

Prevented accidental/invalid attributes on :

showCaption is filtered out so it is not written as an HTML attribute on the  element.

Restored safe defaults via addOptions() (inline, allowBase64, HTMLAttributes, resize) to ensure option reads like this.options.HTMLAttributes are always defined.

Kept the existing ResizableNodeView path intact when resize.enabled === true, limiting the caption NodeView structure changes to the non-resizable configuration.

 @NoFearInMyHeart

Issue #7064, opened 2025-10-10
TypeError: Cannot read properties of null (reading 'domFromPos')


MathiasWP:
### Affected Packages

@tiptap/extension-bubble-menu

### Version(s)

3.4.3

### Bug Description

I am not sure how this has happened, but it was triggered in production and captured with Sentry: https://kvist.sentry.io/share/issue/914d2f155f934654ae262a161f6a8daa/

So it seems like it's a bug that is possible to trigger.

Sorry for lack of good information.

### Browser Used

Edge

### Expected Behavior

Error should not throw

### Dependency Updates

- [x] Yes, I've updated all my dependencies.

bdbch:
Seems that somehow your start/end selection position ended up at a position outside of the document. While this probably happens very rarely we could add a safeguard there.

metznerml:
in my application I have minimum screen width and display the editor as long as the screen doesn't get scmaller than x px, when I render my error Message and the editor get's destructed I can reproduce this by using the responsive design mode in my dev tools.

timheerwagen:
I encountered this bug in production when using the bubble menu, too. However, I cannot reproduce it reliably.

kalda341:
We are also encountering this issue, and it's very hard to reproduce. All the cases I've seen it occur have been within nested editors, but I'm not totally sure that's related.
I'm just about to upgrade - can anyone confirm whether this is fixed in [v3.6.2](https://github.com/ueberdosis/tiptap/releases/tag/v3.6.2)?

volarname:
the issue is still there on v3.15.3, we have same issue after migrating to v3, this was not on latest v2 tiptap

volarname:
I just created a simple [PR](https://github.com/ueberdosis/tiptap/pull/7433) to fix the issue

bdbch:
Merged, I will close this for now - if this is not fixed with the next version, please ping me to reopen this.

kfitzpatrick:
@bdbch Regretfully, it appears still to be occurring after updating to 3.22.4

Our stacktrace doesn't appear to have too many new details in it, but here it is:
```
Cannot read properties of null (reading 'domFromPos')
TypeError: Cannot read properties of null (reading 'domFromPos')
    at coordsAtPos (webpack-internal:///./node_modules/prosemirror-view/dist/index.js:541:47)
    at EditorView.coordsAtPos (webpack-internal:///./node_modules/prosemirror-view/dist/index.js:5627:16)
[...]

PR #7433, opened 2026-01-20
fix: bubble menu TypeError: Cannot read properties of null (reading 'domFromPos')

volarname:
## Changes Overview

Just simple safe guard to prevent `TypeError: Cannot read properties of null (reading 'domFromPos')` error to appear. This can appear when editor is about to destroy and bubble menu is still calculating the position - the editor view doesnt exist anymore. By early return, it will prevent to call posToDOMRect where the error is happening. This check will not break anything and has 0 performance impact.

## Checklist

## Related Issues
[TypeError: Cannot read properties of null (reading 'domFromPos')](https://github.com/ueberdosis/tiptap/issues/7064)

### DIFF
--- a/.changeset/curly-lizards-hang.md
+++ b/.changeset/curly-lizards-hang.md
@@ -1,0 +1,5 @@
+---
+"@tiptap/extension-bubble-menu": patch
+---
+
+Added a safeguard to avoid `TypeError: Cannot read properties of null (reading 'domFromPos')` being thrown when the editor was being destroyed


packages/extension-bubble-menu/src/bubble-menu-plugin.ts
```diff
@@ -269,6 +269,10 @@ export class BubbleMenuView implements PluginView {
       return referencedVirtualElement
     }
 
+    if (!this.view || !this.view.dom || !this.view.dom.parentNode) {
```
EriksonBahr:
```suggestion
    if (!this.view?.dom?.parentNode) {
```

PR #7439, opened 2026-01-23
Fix/firefox drag handle caret visibility

bdbch:
## Changes Overview

Fixed a Firefox-specific bug where the text caret becomes invisible after dragging and dropping a node using the drag handle extension.

## Implementation Approach

1. Added a new `isFirefox()` utility function to `@tiptap/core` following the existing pattern of browser detection utilities (`isSafari`, `isAndroid`, etc.)
2. In the drag handle plugin, added a `drop` event listener that applies a workaround for Firefox by toggling the `contentEditable` attribute after a drop operation, which forces Firefox to re-render the caret

This is a well-documented Firefox bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1327834

## Testing Done

Manual testing in Firefox by dragging and dropping nodes using the drag handle and verifying the caret remains visible.

## Verification Steps

1. Open the editor in Firefox
2. Use the drag handle to drag a node (e.g., a paragraph)
3. Drop it at a different position
4. Click in the editor to place the caret
5. Verify the caret is visible (without this fix, it would be invisible)

## Additional Notes

The workaround uses `requestAnimationFrame` to ensure the drop operation has completed before toggling `contentEditable`. This approach is commonly used by other editors (CKEditor, TinyMCE) to work around this Firefox limitation.

## Checklist

## Related Issues

- https://bugzilla.mozilla.org/show_bug.cgi?id=1327834
- https://github.com/ProseMirror/prosemirror/issues/583


packages/extension-drag-handle/src/drag-handle-plugin.ts
```diff
@@ -151,15 +151,34 @@ export const DragHandlePlugin = ({
...
+    // This workaround forces Firefox to re-render the caret by toggling contentEditable.
+    // See: https://bugzilla.mozilla.org/show_bug.cgi?id=1327834
+    if (isFirefox()) {
+      const editorElement = editor.view.dom
+
+      // Use requestAnimationFrame to ensure the drop operation has completed
+      requestAnimationFrame(() => {
+        if (editorElement.isContentEditable) {
+          editorElement.contentEditable = 'false'
+          editorElement.contentEditable = 'true'
```
arnaugomez, line 165: You set the property twice

PR #7457, opened 2026-01-26
fix(core): respect node inline setting in ResizableNodeView

People-Sea:
## Changes Overview

ResizableNodeView forces display: flex, ignoring the node's inline setting. Fixed by using inline-flex for inline nodes and flex for block nodes.

## Implementation Approach

One line change:
```js
element.style.display = this.node.type.isInline ? 'inline-flex' : 'flex'
```

## Testing Done

 - Added unit test in `packages/core/__tests__/resizableNodeView.spec.ts`                                                                                                         
  - Verifies inline nodes get `display: inline-flex`
  - Verifies block nodes get `display: flex`                                                                                                                                       
  - Lint and prettier passed.
 

## Verification Steps

1. Configure Image extension with inline: true + resize: { enabled: true }
2. Check that the resize container has display: inline-flex
3. Image should render inline with text

## Checklist

### DIFF
--- a/packages/core/__tests__/resizableNodeView.spec.ts
+++ b/packages/core/__tests__/resizableNodeView.spec.ts
@@ -1,0 +1,114 @@
+import { Editor, Node, ResizableNodeView } from '@tiptap/core'
+import Document from '@tiptap/extension-document'
+import Paragraph from '@tiptap/extension-paragraph'
+import Text from '@tiptap/extension-text'
+import { afterEach, describe, expect, it } from 'vitest'
+
+const InlineImage = Node.create({
+  name: 'inlineImage',
+  group: 'inline',
+  inline: true,
+  atom: true,
+
+  addAttributes() {
+    return {
+      src: { default: null },
+      width: { default: null },
+      height: { default: null },
+    }
+  },
+
+  parseHTML() {
+    return [{ tag: 'img[data-type="inline"]' }]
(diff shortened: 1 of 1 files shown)


people-sea:
@arnaugomez 

Thanks for the review! This is a bugfix - enabling `resize` was causing `inline` to have no effect. This PR fixes that, so I don't think doc changes are needed. But happy to update if you think otherwise!

packages/core/src/lib/ResizableNodeView.ts
```diff
@@ -548,7 +548,7 @@ export class ResizableNodeView {
     const element = document.createElement('div')
     element.dataset.resizeContainer = ''
     element.dataset.node = this.node.type.name
-    element.style.display = 'flex'
+    element.style.display = this.node.type.isInline ? 'inline-flex' : 'flex'
```
bdbch, line 551: Can we add a regression test for this behavior before merging? This is the only changed logic in the PR, and right now there’s nothing asserting that an inline node using `ResizableNodeView` keeps an inline container (`display: inline-flex`) instead of falling back to block-level `flex`. A focused test around an inline resizable image would make this fix much safer.

people-sea: Added test for inline/block display. Please re-review when you get a chance. Thanks!

PR #7481, opened 2026-01-31
Prevent inline math input rule from replacing previous character

n-siddarth:
## Changes Overview

Changed the InlineMath extension's input rule to prevent it from consuming the preceding character to a `$$latex$$` expression. 

For example, the string `Hello $$\sigma$$` would be transformed into → `Helloσ` (with the space deleted).

## Implementation Approach

We can utilize a fixed-width negative lookbehind that validates the preceding character is not a $-sign. This mimics the negative lookahead that's used at the end of the regular expression.

## Testing Done

Change is small enough to not require tests.

## Verification Steps

Build the editor. Utilize the mathematics extension and verify that `$$somelatex$$` does not replace the preceding character.

## Checklist

- [x] I have created a [changeset](https://github.com/changesets/changesets) for this PR if necessary
- [x] I have added tests where applicable: `(N/A here)`

### DIFF
--- a/.changeset/yellow-bags-pump.md
+++ b/.changeset/yellow-bags-pump.md
@@ -1,7 +1,0 @@
----
-'@tiptap/extension-mathematics': minor
----
-
-Prevent inline math input rule from replacing previous character
-
-- Changed input rule to utilize negative lookbehind rather than regular character match to ensure preceding character is not consumed in final match. This ensures the range's `from` position is correctly at the start of the double `$` signs.
--- a/packages/extension-mathematics/src/extensions/InlineMath.ts
+++ b/packages/extension-mathematics/src/extensions/InlineMath.ts
@@ -220,7 +220,7 @@ export const InlineMath = Node.create<InlineMathOptions>({
   addInputRules() {
     return [
       new InputRule({
-        find: /(?<!$)(\$\$([^$\n]+?)\$\$)(?!\$)/,
+        find: /(?<!\$)(\$\$([^$\n]+?)\$\$)(?!\$)/,
         handler: ({ state, range, match }) => {
           const latex = match[2]
           const { tr } = state
--- a/.changeset/fix-inline-mathematics-input-rule.md
+++ b/.changeset/fix-inline-mathematics-input-rule.md
@@ -1,0 +1,5 @@
+---
+'@tiptap/extension-mathematics': patch
+---
+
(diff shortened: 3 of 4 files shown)


bdbch:
Is there a related issue for this?

bdbch:
One thing that worries me is a negative lookup. While Safari catched up and supports them since 2023 I wonder if we should keep this information on the docs. Not sure though.

n-siddarth:
> Is there a related issue for this?

Sorry, I did not create one due to the simplicity of the change. Would you like me to create a corresponding issue?

n-siddarth:
> One thing that worries me is a negative lookup. While Safari catched up and supports them since 2023 I wonder if we should keep this information on the docs. Not sure though.

Hmm, good point. We of course do not need to use negative lookbehinds. We could capture the previous character and insert content that re-adds the character in a generic `InputRule`. Is this approach better?

packages/extension-mathematics/src/extensions/InlineMath.ts
```diff
@@ -220,9 +220,9 @@ export const InlineMath = Node.create<InlineMathOptions>({
   addInputRules() {
     return [
       new InputRule({
-        find: /(^|[^$])(\$\$([^$\n]+?)\$\$)(?!\$)/,
+        find: /(?<!$)(\$\$([^$\n]+?)\$\$)(?!\$)/,
```
bdbch: `(?<!$)` does not mean 'previous character is not a literal `$`' here - the `$` is still being parsed as the end-of-input anchor. Because of that, this now matches inside inputs like `$$$x$$` and `$$$$x$$`, which the old regex correctly rejected. This likely needs to be `(?<!\$)` (or another equivalent assertion) so we only stop consuming the preceding character without allowing matches after another dollar sign. Please also add a regression test for both `Hello $$x$$` preserving the space and triple-dollar cases staying unmatched.

.changeset/yellow-bags-pump.md
```diff
@@ -0,0 +1,7 @@
+---
+'@tiptap/extension-mathematics': minor
```
bdbch: This reads like a bug fix with no intended API change, so the changeset should probably be `patch` rather than `minor` unless the mathematics package intentionally treats fixes as minor releases.

n-siddarth:
@bdbch Updated the PR to include the requested changes

PR #7500, opened 2026-02-04
feat: Added `showOnMouseUp` option to bubble menu

Aslam97:
## Changes Overview

Adds a new `showOnMouseUp` option to the bubble menu that defers showing the menu until the mouse button is released, similar to Notion's text selection behavior.

## Implementation Approach

Added a new optional `showOnMouseUp` prop (default: `false`) to the bubble menu that defers menu display until mouseup occurs. 

Key implementation details:
- Track mouse state via `isMouseDown` flag
- Store pending updates while mouse is down instead of immediately showing the menu
- Process pending updates on mouseup to display the menu

## Testing Done

- Verified menu appears only after mouseup when `showOnMouseUp: true`

## Verification Steps

1. Enable the `showOnMouseUp`: `<BubbleMenu showOnMouseUp={true} />`
2. Click and drag to select text - menu should NOT appear during selection
3. Release mouse button - menu should appear at selection

## Checklist

### DIFF
--- a/.changeset/yummy-queens-rush.md
+++ b/.changeset/yummy-queens-rush.md
@@ -1,0 +1,5 @@
+---
+'@tiptap/extension-bubble-menu': minor
+---
+
+Added `showOnMouseUp` option to bubble menu. This option allows the bubble menu to be displayed when the user releases the mouse button after a text selection, enhancing the user experience during text editing.
--- a/packages/extension-bubble-menu/src/bubble-menu-plugin.ts
+++ b/packages/extension-bubble-menu/src/bubble-menu-plugin.ts
@@ -101,6 +101,17 @@ export interface BubbleMenuPluginProps {
   getReferencedVirtualElement?: () => VirtualElement | null
 
   /**
+   * When enabled, the bubble menu will only appear after the mouse button is released
+   * (on mouseup), similar to Notion's behavior. This prevents the menu from appearing
+   * and following the cursor while the user is still making a selection.
+   *
+   * When disabled (default), the menu appears immediately as the selection changes.
+   * @type {boolean}
+   * @default false
+   */
+  showOnMouseUp?: boolean
+
+  /**
    * The options for the bubble menu. Those are passed to Floating UI and include options for the placement, offset, flip, shift, arrow, size, autoPlacement,
    * hide, and inline middlewares.
    * @default {}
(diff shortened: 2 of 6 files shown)


arnaugomez (approved):

.changeset/yummy-queens-rush.md
```diff
@@ -0,0 +1,5 @@
+---
+'@tiptap/extension-bubble-menu': minor
+---
+
+Added `showOnMouseUp` option to bubble menu. This option allows the bubble menu to be displayed when the user releases the mouse button after a text selection, enhancing the user experience during text editing.
```
bdbch, line 5: Is this conflicting with the Bubble Menu's `shouldShow` option? How do they work together?

bdbch (commented):
Not sure if I should approve or reject. Quick question:

- This is meant that while I select something, the Menu only opens after the selection is finished and my mouse is let go right? Should this also work with keyboard selections? In that case - maybe it's better to not handle mouse/keyboard events but build this logic around transactions that update the selection?
- Is this in conflict with `shouldShow`

overall the code looks good to me, just want to make sure we build this the right way as I don't know where exactly the requirement came from.

.changeset/yummy-queens-rush.md
Aslam97: It shouldn’t conflict with `shouldShow`. The menu is shown only after the user releases the mouse (`mouseup`), and then it goes through `shouldShow` for validation.

Aslam97: It does not handle keyboard selection, it only apply to mouse selection

Aslam97: "maybe it's better to not handle mouse/keyboard events but build this logic around transactions that update the selection?" I am not sure about this one, since what user want is they dont want to see the floating before releasing the mouse (similiar to Notion)

bdbch: Yeah that would work - what we would need to do is onTransaction, check if the transaction changes the selection object - and if it does, we'd need to start a timeout / debounce that only resolves when no other transaction in X time occured again - and just then we "activate" the bubble menu so it can be shown. That would mean that it would work for all kind of interactions that change the selection.

Aslam97: I see what you mean. The current showOnMouseUp works by deferring updates while the mouse is down, I can refactor it so the menu only shows on a commit selection transaction instead.

bdbch: Sorry Aslam, coming back to this. I'm not sure if I'd want this functionality on top of `shouldShow`. Don't you think this behavior is more an integration topic and should be handled by the dev implementing it?

`shouldShow` should be able to handle this just fine if the user tracks the mousedown state somewhere & checks for it in shouldShow.

Aslam97: I have send them how they can achieve the same behaviour without adding this implementation to the core package. closing this PR

Issue #7508, opened 2026-02-06
[@tiptap/extension-drag-handle-react]: Position 3 out of range Error when nested set to false


siamahnaf:
### Affected Packages

extension-drag-handle, extension-drag-handle-react

### Version(s)

3.19.0

### Bug Description

When editor is empty and the setting is set to false, and usually I am trying to drag the placeholder it showing error. But when set to nested true, no error coming-

See video

https://github.com/user-attachments/assets/a9a15096-9756-40db-8ad7-65fcd396bfe3

Code

```
<DragHandle editor={editor} nested={false}>
    {/* Content */}
</DragHandle>
```

Details Screenshots of error-

<img width="991" height="759" alt="Image" src="https://github.com/user-attachments/assets/e8b1ab71-ef15-4f0b-867c-aa645b209c4d" />

### Browser Used

Chrome

### Expected Behavior

I tried all ways, but the error is always coming-

Dependencies-

```
"dependencies": {
    "@floating-ui/react": "^0.27.17",
    "@siamf/react-color-pick": "^1.5.4",
    "@tiptap/extension-blockquote": "^3.19.0",
    "@tiptap/extension-bold": "^3.19.0",
    "@tiptap/extension-collaboration": "^3.19.0",
    "@tiptap/extension-document": "^3.19.0",
    "@tiptap/extension-drag-handle": "^3.19.0",
    "@tiptap/extension-drag-handle-react": "^3.19.0",
    "@tiptap/extension-emoji": "^3.19.0",
    "@tiptap/extension-hard-break": "3.19.0",
    "@tiptap/extension-heading": "^3.19.0",
    "@tiptap/extension-horizontal-rule": "^3.19.0",
    "@tiptap/extension-image": "^3.19.0",
    "@tiptap/extension-italic": "^3.19.0",
    "@tiptap/extension-link": "^3.19.0",
    "@tiptap/extension-list": "^3.19.0",
    "@tiptap/extension-node-range": "^3.19.0",
    "@tiptap/extension-paragraph": "^3.19.0",
    "@tiptap/extension-strike": "^3.19.0",
    "@tiptap/extension-subscript": "^3.19.0",
    "@tiptap/extension-superscript": "^3.19.0",
    "@tiptap/extension-table": "^3.19.0",
    "@tiptap/extension-text": "^3.19.0",
    "@tiptap/extension-text-align": "^3.19.0",
    "@tiptap/extension-text-style": "^3.19.0",
[...]

PR #7509, opened 2026-02-07
Fix: Position 3 out of range Error [@tiptap/extension-drag-handle]

siamahnaf:
## Changes Overview

Current version has following code on  `extension-drag-handle/src/helpers/dragHandler.ts`-
```javascript
const $from = doc.resolve(result.pos)
const $to = doc.resolve(result.pos + result.resultNode.nodeSize)
```

And I changed to
```javascript
const $from = doc.resolve(result.pos)
const $to = doc.resolve(result.pos + 1)
```

## Implementation Approach

I do following steps-
- Very first tried to reproduce the error on tiptap repository code, I find out the error still happening on there also
Screenshot 2026-02-07 at 1 28 17 PM

-  Then search previous version code where `nested` options was not introduced like `v3.16.0`. And on that previous version the line was `const $to = doc.resolve(result.pos + 1)`.
- Then I change only `const $to = doc.resolve(result.pos + result.resultNode.nodeSize)` to `const $to = doc.resolve(result.pos + 1)`
- I checked and it get solved

## Testing Done

Testing done on `tiptap` repository demos.

## Verification Steps

Just the review the changes, its one line code change only.

## Checklist

## Related Issues

https://github.com/ueberdosis/tiptap/issues/7508

### DIFF
--- a/packages/extension-drag-handle/src/helpers/dragHandler.ts
+++ b/packages/extension-drag-handle/src/helpers/dragHandler.ts
@@ -49,8 +49,9 @@ function getDragHandleRanges(
   }
 
   // For non-nested mode, use depth 0 to select the outermost block
+  const offset = result.resultNode.isText ? 0 : -1
   const $from = doc.resolve(result.pos)
-  const $to = doc.resolve(result.pos + 1)
+  const $to = doc.resolve(result.pos + result.resultNode.nodeSize + offset)
 
   return getSelectionRanges($from, $to, 0)
 }


packages/extension-drag-handle/src/helpers/dragHandler.ts
```diff
@@ -50,7 +50,7 @@ function getDragHandleRanges(
 
   // For non-nested mode, use depth 0 to select the outermost block
   const $from = doc.resolve(result.pos)
-  const $to = doc.resolve(result.pos + result.resultNode.nodeSize)
+  const $to = doc.resolve(result.pos + 1)
```
bdbch: This is just happening on empty documents, right? Maybe we can do a better fix here that still selects the whole node range BUT limits it to the document size / actual node size.

Give me a minute to checkout your PR and test it out myself.

bdbch: Okay, I'd rather do this:

```
  // For non-nested mode, use depth 0 to select the outermost block
  const offset = result.resultNode.isText ? 0 : -1
  const $from = doc.resolve(result.pos)
  const $to = doc.resolve(result.pos + result.resultNode.nodeSize + offset)
```

The reason why this is happening is because usually, the position usually points into the text node which has a valid length, but as soon as there's no text node, the position will point into the wrapping node (the empty paragraph) which in turn then will add the closing boundary on top leading to a range outside of the document range.

PR #7512, opened 2026-02-10
fix: use instance-specific pluginKey as transaction meta key in BubbleMenu/FloatingMenu to prevent cross-contamination

Re-cool:
# Changes Overview

When multiple `BubbleMenu`/`FloatingMenu` instances use different `pluginKey` values, all instances respond to any instance's option updates because the transaction meta key is hardcoded as `"bubbleMenu"`/`"floatingMenu"` instead of using the provided `pluginKey`.

## Implementation Approach

- **`BubbleMenuView`/`FloatingMenuView`**: Accept `pluginKey` in constructor and use it in `getMeta()` instead of hardcoded string
- **React `BubbleMenu`/`FloatingMenu`**: Use `pluginKey` prop in `setMeta()` instead of hardcoded string
- **`FloatingMenu` extension**: Use `this.options.pluginKey` in `updateFloatingMenuPosition` command

Default values remain `"bubbleMenu"`/`"floatingMenu"`, maintaining full backward compatibility.

## Testing Done

- Added 11 new test cases across `extension-bubble-menu` and `extension-floating-menu` packages
- Tests cover `pluginKey` isolation with both string and `PluginKey` instance types, cross-contamination prevention between instances, command meta key scoping, and backward compatibility with default `pluginKey`
- Tests verified to fail before the fix and pass after (TDD approach)

## Verification Steps

1. Create two `BubbleMenu` or `FloatingMenu` instances with different `pluginKey` values and different `options` (e.g., different `placement`)
2. Update one instance's options dynamically
3. Verify the other instance's options remain unchanged

## Checklist

### DIFF
--- a/.changeset/few-rice-end.md
+++ b/.changeset/few-rice-end.md
@@ -1,0 +1,5 @@
+---
+'@tiptap/core': patch
+---
+
+Fix checking if mark is active and toggling off marks when part of the selection does not allow the mark (e.g. a code block)
--- a/.changeset/rich-squids-fold.md
+++ b/.changeset/rich-squids-fold.md
@@ -1,0 +1,5 @@
+---
+'@tiptap/extension-code-block-lowlight': patch
+---
+
+Fixed a runtime error when initializing `CodeBlockLowlight` by switching the `CodeBlock` import to a named export. This prevents `extend is not a function` errors caused by ESM/CJS interop issues.
--- a/.changeset/seven-beers-search.md
+++ b/.changeset/seven-beers-search.md
@@ -1,0 +1,5 @@
+---
+"@tiptap/extension-invisible-characters": patch
+---
+
+Added missing storage typings
--- a/.changeset/shy-bugs-raise.md
+++ b/.changeset/shy-bugs-raise.md
@@ -1,0 +1,5 @@
+---
+"@tiptap/extension-drag-handle": patch
+---
+
+Fix drag position resolving outside the document when dragging an empty text node at the end of the document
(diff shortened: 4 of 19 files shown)


Re-cool:
This builds on the work in #7445 by scoping the transaction meta key to each instance's `pluginKey`, so that multiple instances with different `pluginKey` values can be updated independently.

packages/extension-floating-menu/src/floating-menu-plugin.ts
```diff
@@ -358,10 +362,10 @@ export class FloatingMenuView {
   /**
    * Handles the transaction event to update the position of the floating menu.
    * This allows external code to trigger a position update via:
-   * `editor.view.dispatch(editor.state.tr.setMeta('floatingMenu', 'updatePosition'))`
+   * `editor.view.dispatch(editor.state.tr.setMeta(pluginKey, 'updatePosition'))`
```
bdbch, line 365: I think we should mention the default pluginKey in the JSDoc so people know what the default is.

packages/extension-bubble-menu/src/bubble-menu-plugin.ts
```diff
@@ -567,7 +571,7 @@ export class BubbleMenuView implements PluginView {
   }
 
   transactionHandler = ({ transaction: tr }: { transaction: Transaction }) => {
```
bdbch, line 579: Do you think we should replicate the JSDoc from the FloatingMenu here?

PR #7549, opened 2026-03-01
fix: handle flushSync error when remounting useEditor in React 19

### LINKED ISSUE #7543: [Bug]: flushSync error when remounting useEditor with existing ID and ReactNodeViewRenderer in React 19
### Affected Packages

react, extension-document, extension-text

### Version(s)

3.14.0, 3.20.0

### Bug Description

React 19 appears to have a [bug](https://github.com/facebook/react/issues/29585) in which some components in an array will remount if the array is reordered, even if every element is given a unique key tied to the data it represents. When the list is re-ordered, and we use the same id in the dependency array for useEditor, we get a flushSync error on any component which moved to be later in the list that uses a custom extension with a ReactNodeViewRenderer.

React 18 Sandbox: https://codesandbox.io/p/sandbox/tiptap-flushsync-reorder-18-w7ttcg
React 19 Sandbox: https://codesandbox.io/p/sandbox/tiptap-flushsync-reorder-19-x47nq7

This is likely related to any or all of #6315 #7438 #6386 

### Browser Used

Firefox

### Code Example URL

https://codesandbox.io/p/sandbox/tiptap-flushsync-reorder-19-x47nq7

### Expected Behavior

We do not get a flush sync error if useEditor remounts with the same dependency array

* The version where I first encountered this is 3.14.0
* React core team is aware of the issue on their end

### Dependency Updates

bxff:
Closes #7543

## What this PR does

Wraps the `flushSync` call in `ReactRenderer`'s constructor with a try-catch. When `flushSync` throws (because React is already in a render/commit phase), it falls back to `queueMicrotask` — the same path already used for non-initialized editors.

- **Normal case**: `flushSync` succeeds → synchronous ProseMirror/React rendering (no behavior change)
- **Edge case** (React 19 remount during commit): `flushSync` throws → graceful fallback to async render

## Why

React 19 may remount keyed components when a list is reordered, even with stable keys. When this happens during `useEditor`'s effect cycle, `ReactRenderer` gets constructed during React's commit phase, and `flushSync` throws a hard error.

## Reproduction

Reporter's React 19 sandbox: https://codesandbox.io/p/sandbox/tiptap-flushsync-reorder-19-x47nq7

React 18 comparison (works fine): https://codesandbox.io/p/sandbox/tiptap-flushsync-reorder-18-w7ttcg


bdbch (changes requested):
Thanks for the PR and for digging into the React 19 remount case. I'm going to reject this as-is because the new try/catch is a bit too broad: it doesn't just catch the flushSync lifecycle error, it will also swallow real errors thrown by this.render() and then retry them in a microtask. That changes the current failure mode, makes debugging harder, and could hide unrelated regressions.

I think this needs to be narrowed so we only fall back for the specific flushSync-during-render/commit failure, while rethrowing everything else. A focused test for that fallback path would also help a lot.

bdbch (approved):

bdbch (changes requested):
My bad, didn't mean to approve - see above.

bxff:
Fwiw I did have a look and this PR is wrong. I did redo it but I wasn't confident enough to make another PR. Closing for now. Thanks for the time, I really appreciate it. :)

PR #7559, opened 2026-03-04
Add `types: 'all'` support to UniqueID extension

arnaugomez:
## Summary
- add support for `types: 'all'` in `@tiptap/extension-unique-id`
- resolve `'all'` to all node types except `doc` and `text`
- apply this resolution consistently in runtime ID assignment and `generateUniqueIds`
- include a changeset for `@tiptap/extension-unique-id`

## Why
The extension currently defaults to `types: []`, which means no IDs are added unless users explicitly list node types. This change adds a shorthand to target all document nodes (excluding `doc` and `text`) without manually maintaining a list.

## Related PRs
- Docs update: https://github.com/ueberdosis/tiptap-docs/pull/648

## Validation
- `corepack pnpm --filter @tiptap/extension-unique-id lint`

### DIFF
--- a/packages/extension-unique-id/src/generate-unique-ids.ts
+++ b/packages/extension-unique-id/src/generate-unique-ids.ts
@@ -46,10 +46,13 @@ export function generateUniqueIds(doc: JSONContent, extensions: Extensions): JSO
   if (!uniqueIDExtension) {
     throw new Error('UniqueID extension not found in the extensions array')
   }
-  const { types, attributeName, generateID } = uniqueIDExtension.options
-
   // Convert the JSON content to a ProseMirror node
   const schema = getSchema([...extensions.filter(ext => ext.name !== 'uniqueID'), uniqueIDExtension])
+  const { types: configuredTypes, attributeName, generateID } = uniqueIDExtension.options
+  const types =
+    configuredTypes === 'all'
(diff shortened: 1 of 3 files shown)


packages/extension-unique-id/src/generate-unique-ids.ts
```diff
@@ -46,10 +46,13 @@ export function generateUniqueIds(doc: JSONContent, extensions: Extensions): JSO
...
     throw new Error('UniqueID extension not found in the extensions array')
   }
-  const { types, attributeName, generateID } = uniqueIDExtension.options
-
   // Convert the JSON content to a ProseMirror node
   const schema = getSchema([...extensions.filter(ext => ext.name !== 'uniqueID'), uniqueIDExtension])
+  const { types: configuredTypes, attributeName, generateID } = uniqueIDExtension.options
+  const types =
+    configuredTypes === 'all'
+      ? Object.keys(schema.nodes).filter(type => type !== 'doc' && type !== 'text')
```
Aslam97, line 54: Shouldnt the unique ID only be applied to top-level nodes?

bdbch: No I don't think so, they should be applied to all available non-text nodes that are not `doc` so we also need to go deeper in depth.

bdbch:
@arnaugomez merged, please make sure to move forward your docs PR after the next release is out.

PR #7566, opened 2026-03-05
fix(nodeview): ensure NodeView re-renders on position changes without…

bdbch:
**Changes Overview**

- Fixed NodeView re-rendering so sibling NodeViews update when their document position changes (e.g., when another child is moved). Edited React, Vue 3 and Vue 2 renderers and updated the DragHandle demo.

**Implementation Approach**

- Added per-NodeView `currentPos` tracking and an `editor.on('update')` listener that:
  - rAF-batches position checks,
  - compares `getPos()` to `currentPos`,
  - when changed, forces a re-render by calling `renderer.updateProps({ getPos: () => this.getPos() })` (fresh function ref to bypass memo/shallow-equality).
- Kept the existing `update()` early-return optimization but updated it to also update `currentPos` so `handleEditorUpdate` is cheap when ProseMirror already handled the node.
- Demo: show `pos` and give each `Recommendation` node a unique `data-id` and start with five nodes to make verification easy.
- No public API changes; custom `update` option path unchanged.

**Testing Done**

- Manual verification in the demo:
  - React now contains five recommendation nodes and each view shows `pos:`.
  - Dragging any node updates its own `pos` and the sibling nodes’ displayed `pos` immediately (validated interactively).
- Quick static checks: TypeScript/compile error scan on edited files (no new file errors reported).
- Added a .changeset entry for the three affected packages.

**Verification Steps**

- Start demos:
```bash
pnpm dev
# open http://localhost:3000 -> Extensions -> DragHandleWithNodeViews
```
- Confirm five recommendation blocks are present (IDs 1–5).
- Drag block 1 to a different position and observe:
  - The dragged block’s `pos:` updates immediately.
  - All sibling blocks that changed positions update their `pos:` immediately (no need to edit/focus them).
- Run CI-style checks locally:
```bash
pnpm lint
pnpm build
pnpm test:run   # optional; runs existing Cypress tests
```

**Additional Notes**

- Performance: position checks are rAF-batched and early-return quickly when nothing changed, matching the existing `selectionUpdate` pattern. For documents with very large numbers of NodeViews this is O(N) per frame in the worst case; batching keeps it practical for normal use.
- Non-breaking: no API signature changes. Users with custom `update` callbacks are unaffected (the custom `update` path runs

### DIFF
--- a/.changeset/fix-nodeview-position-updates.md
+++ b/.changeset/fix-nodeview-position-updates.md
@@ -1,4 +1,5 @@
 ---
+'@tiptap/core': patch
 '@tiptap/react': patch
 '@tiptap/vue-3': patch
 '@tiptap/vue-2': patch


packages/core/src/utilities/nodeViewPositionRegistry.ts
```diff
@@ -0,0 +1,70 @@
...
+ * Per-editor registry for centralized NodeView position-change checks.
+ * A single editor.on('update') listener + rAF is shared across all NodeViews
+ * for a given editor, keeping overhead bounded regardless of NodeView count.
+ *
+ * This is consumed by React, Vue 3, and Vue 2 NodeView renderers.
+ */
+interface PositionUpdateRegistry {
+  callbacks: Set<() => void>
+  rafId: number | null
+  handler: () => void
```
arnaugomez, line 13: Could you add TSDoc comments to these properties? Especially rAF

.changeset/fix-nodeview-position-updates.md
```diff
@@ -0,0 +1,7 @@
+---
+'@tiptap/react': patch
+'@tiptap/vue-3': patch
+'@tiptap/vue-2': patch
+---
```
arnaugomez, line 6: Core package also has changes

PR #6833, opened 2025-08-17
feat(suggestion): add safe API to exit suggestions

bdbch:
## Changes Overview

This pull request introduces a safe and reliable API for exiting suggestions in `@tiptap/suggestion`, along with improvements to user interaction and visual feedback. The changes ensure that suggestions can be closed programmatically or via user actions (Escape key, clicking outside), and that decorations are removed deterministically without document edits. Additionally, new styles enhance the visibility of suggestion ranges in demos.

## Implementation Approach

**Suggestion plugin API and behavior improvements:**

* Added a safe API for exiting suggestions by dispatching a metadata-only transaction with `{ exit: true }`, which reliably deactivates the suggestion plugin and removes inline decorations. Programmatic exit is now available via the exported `exitSuggestion(view, pluginKey?)` helper. [[1]](diffhunk://#diff-08a3639c233904dd897a79ccc92227a3b367ea0b42fd6c996e5087361c23482aR1-R10) [[2]](diffhunk://#diff-960594a3a1cb2fc968abde71ae4ea81b830d4bc5cbb36d592d419bd9c006da97L1-R7) [[3]](diffhunk://#diff-c16484042622aaa7ff275b7931a38149dfd69125ed9737cce7ffa397d3a63005R536-R545)
* Escape key now triggers `renderer.onExit` and dispatches the exit meta, ensuring suggestions close immediately without document edits. [[1]](diffhunk://#diff-08a3639c233904dd897a79ccc92227a3b367ea0b42fd6c996e5087361c23482aR1-R10) [[2]](diffhunk://#diff-c16484042622aaa7ff275b7931a38149dfd69125ed9737cce7ffa397d3a63005L397-R504)
* Clicking outside the editor closes active suggestions by installing a global click handler when suggestions are active. [[1]](diffhunk://#diff-08a3639c233904dd897a79ccc92227a3b367ea0b42fd6c996e5087361c23482aR1-R10) [[2]](diffhunk://#diff-c16484042622aaa7ff275b7931a38149dfd69125ed9737cce7ffa397d3a63005R207-R259) [[3]](diffhunk://#diff-c16484042622aaa7ff275b7931a38149dfd69125ed9737cce7ffa397d3a63005R333-R344)

**Visual and styling improvements:**

* Added new inline decoration styles to visually highlight the active suggestion range in demos, including handling for empty decorations and consistent appearance with mention nodes.

**Internal code refactoring:**

[...]

PR #7570, opened 2026-03-06
fix(suggestion): keep dismissed state after dismissal

bdbch:
## Changes Overview

After dismissing a suggestion via Escape, the suggestion menu would reappear as soon as the user typed another character in the same word. This makes dismissal feel broken — one Escape press should suppress the suggestion until the user clearly starts a
 new input context.

## Implementation Approach

Added a `dismissedFrom: number | null` field to the suggestion plugin's internal state. When the user dismisses via Escape (dispatching `{ exit: true }` metadata), the trigger character's position (`range.from`) is stored in `dismissedFrom`.

On every subsequent transaction, before re-activating a match, the plugin checks:
- **Same word?** — if `match.range.from === dismissedFrom`, the cursor is still in the dismissed word → stay suppressed
- **Whitespace/newline inserted?** — if the transaction inserted any whitespace, the user has moved on → clear `dismissedFrom`
- **Different trigger position?** — if the match starts at a different position, it's a different word → clear `dismissedFrom`
- **No match at all?** — cursor has left any trigger context entirely → clear `dismissedFrom`

The key design decision: `dismissedFrom` intentionally survives the `!next.active` cleanup block at the bottom of `apply()`, so it persists across inactive transactions.

## Testing Done

Added four new unit tests in `suggestion.test.ts` covering the full matrix:

1. Typing more characters in the same word after dismissal → suggestion stays hidden
2. Inserting a space after dismissal then typing a new `@` → suggestion reopens
3. Backspacing past the trigger char, then retyping it → suggestion reopens
4. Typing a new `@` at a different position after dismissal → suggestion reopens

## Verification Steps

1. Open an editor with the mention extension
2. Type `@foo` — suggestion menu opens
3. Press **Escape** — menu closes
4. Type more characters (`bar`) — menu should **not** reappear (was broken before)
5. Press **Space**, then type `@` — menu should reappear ✓
6. Repeat steps 2–3, then click somewhere else in the document and type `@` — menu should reappear ✓

## Additional Notes

No public API changes. The `dismissedFrom` field lives entirely in internal plugin state. The existing `allow`, `shouldShow`, and `exitSuggestion` API are unaffected.

## Checklist

## Related Issues

Follows up on #6833 (Escape key to dismiss suggestions).

### DIFF
--- a/packages/suggestion/src/__tests__/suggestion.test.ts
+++ b/packages/suggestion/src/__tests__/suggestion.test.ts
@@ -128,7 +128,13 @@ describe('suggestion integration', () => {
 
 describe('suggestion dismissal', () => {
   /** Builds a minimal editor with a single @-mention suggestion and returns helpers. */
-  function setup() {
+  function setup(
+    options: {
+      allowSpaces?: boolean
+      allowToIncludeChar?: boolean
+      shouldResetDismissed?: Parameters<typeof Suggestion>[0]['shouldResetDismissed']
+    } = {},
+  ) {
     const onStart = vi.fn()
     const onUpdate = vi.fn()
     const onExit = vi.fn()
@@ -140,7 +146,10 @@ describe('suggestion dismissal', () => {
           Suggestion({
             editor: this.editor,
             char: '@',
+            allowSpaces: options.allowSpaces,
+            allowToIncludeChar: options.allowToIncludeChar,
             items: () => [],
+            shouldResetDismissed: options.shouldResetDismissed,
             render: () => ({ onStart, onUpdate, onExit }),
           }),
         ]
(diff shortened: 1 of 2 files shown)


oBusk:
Will this implementation mean that adding/removing any character before the trigger. And then editing after the trigger again will activate it again? Maybe an acceptable compromise.

Are you also testing with `allowSpaces: true`?

bdbch:
This implementation is dismissing the mention state until you leave the mention via a whitespace or newline character or you move the selection out of the suggestion range. That means, if you insert characters right before your mention and go back in, you'll have the mention state again which I think is fair.

I think having support for allowSpaces makes it tricky to restore the suggestion state again except maybe using a newline.

If you want I could add an escape hatch to give you more control over when you restore the mention state again, but then you'd need to handle the logic yourself if the default isn't fitting your needs.

oBusk:
I was considering how to solve it with allowSpaces, and there was some mentions of tracking the triggering position to identify "already closed", but the fact that something that could cause that trigger to move would reset and begins showing suggestions again made it slightly flimsy. I was considering if there could be some decoration on the triggering char to mark it as "consumed", but there might be other flimsyness with that.

Since we want `allowSpaces` for our usecase, I think the most reliable we can do is to try to trigger suggestions based on keydown rather than text content, but I understand that would mean we can't really use `tiptap/suggestion` to accomplish that.

This PR will definely make escape work more like you expect, especially without `allowSpaces`, so that's great! Could you clarify what control the escapehatch would give me? Are you saying I call some function to "reactivate" mention state manually?

gethari:
Hi @bdbch any updates on this please ? Are we moving forward with this fix, considering `allowSpaces` will work smoothly after this ?

bdbch:
Pushed a small follow-up:  now maps with the transaction, and I added a newline regression test too. Should cover the review concern a bit better now.

bdbch:
@oBusk pushed an update here: Escape now clears the decoration properly, `allowSpaces` dismissal is handled, and there is a small `shouldResetDismissed` escape hatch for custom reset logic too.

bdbch:
Added documentation for the escape hatch function here:
https://github.com/ueberdosis/tiptap-docs/pull/710

oBusk:
> @oBusk pushed an update here: Escape now clears the decoration properly, `allowSpaces` dismissal is handled, and there is a small `shouldResetDismissed` escape hatch for custom reset logic too.

@bdbch sounds good! I will see if i time to try it out in it's current state tmrw

PR #7596, opened 2026-03-14
fix(markdown): prevent crossing marks when serializing overlapping marks

### LINKED ISSUE #7590: [Markdown]: overlapping bold and italic text doesn't output correct markdown
### Affected Packages

@tiptap/markdown

### Version(s)

3.20.1

### Bug Description

Steps to reproduce:
1. Toggle bold on
2. Type "123"
3. Toggle italic on
4. Type "456"
5. Toggle bold off
6. Type "789"
7. Toggle italic off

The overlapping styles render find in the editor, but the markdown it outputs looks like this

```md
**123*456**789*
```

### Browser Used

Chrome

### Code Example URL

https://github.com/christianjuth/tiptap-markdown-bug/blob/main/src/App.jsx

### Expected Behavior

It should output:

```
**123*456***_789_
```

Confirming this is valid to GitHub markdown:
**123*456***_789_

_No response_

### Dependency Updates

bhaveshrode:
Fixes #7590

When bold and italic marks overlap, the markdown serializer could produce
crossing marks such as:

**123*456**789*

which results in invalid HTML nesting:

<b>123<i>456</b>789</i>

This change ensures marks are closed in reverse order so they are properly nested.

Example output after fix:

**123*456***789*

which corresponds to:

<b>123<i>456</i></b><i>789</i>

A test was added to cover overlapping mark serialization.

### DIFF
--- a/packages/markdown/__tests__/markdown-overlapping-marks.test.ts
+++ b/packages/markdown/__tests__/markdown-overlapping-marks.test.ts
@@ -1,0 +1,45 @@
+import { Editor } from '@tiptap/core'
+import { Markdown } from '@tiptap/markdown'
+import StarterKit from '@tiptap/starter-kit'
+
+describe('Markdown serialization – overlapping marks', () => {
+  it('serializes overlapping bold and italic without crossing marks', () => {
+    const editor = new Editor({
+      extensions: [StarterKit, Markdown],
+      content: {
+        type: 'doc',
+        content: [
+          {
+            type: 'paragraph',
+            content: [
+              {
+                type: 'text',
+                text: '123',
+                marks: [{ type: 'bold' }],
+              },
+              {
+                type: 'text',
+                text: '456',
+                marks: [{ type: 'bold' }, { type: 'italic' }],
+              },
+              {
+                type: 'text',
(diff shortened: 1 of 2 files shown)


bdbch:
Your PR does not have new tests and also breaks pre-existing unit tests. Could you double verify that your changes don't break existing logic?

packages/markdown/__tests__/markdown-overlapping-marks.test.ts
```diff
@@ -0,0 +1,45 @@
+import { Editor } from '@tiptap/core'
+import { Markdown } from '@tiptap/markdown'
+import StarterKit from '@tiptap/starter-kit'
+
+describe('Markdown serialization – overlapping marks', () => {
```
bdbch, line 5: Please fix this test.

bdbch (changes requested):

bdbch:
Actually, I fixed this in https://github.com/ueberdosis/tiptap/issues/7590 - which also includes further fixes and improvements on top where using the same delimitters would confuse the parser by using html tags as optional delimitters.

Thanks for your PR though!

PR #7600, opened 2026-03-15
fix(collaboration): move content validation from beforeTransaction to filterTransaction

### LINKED ISSUE #6927: Content still gets removed after disabling collaboration on content errors
### Affected Packages

core, extension-collaboration

### Version(s)

3.0.0

## Problem

When collaboration is enabled and invalid content is loaded (i.e. unknown node types), those invalid nodes get removed even if [the suggested setup to deal with schema errors is implemented](https://tiptap.dev/docs/editor/core-concepts/schema#invalid-schema-handling).

According to the [code here](https://github.com/ueberdosis/tiptap/blob/develop/packages/extension-collaboration/src/collaboration.ts#L231), returning `false` on a `beforeTransaction` event handler should `prevent the transaction from being applied` but that's not what's happening.

Looking at the yjs code that dispatches the `beforeTransaction` event, it doesnt seem to care about the returned value: https://github.com/yjs/yjs/blob/main/src/utils/Transaction.js#L467

### Browser Used

Chrome

### Code Example URL

_No response_

### Expected Behavior

Content should not be wiped out if the doc recommendations are implemented

_No response_

### Dependency Updates

dearlordylord:
Fixes #6927

## Problem

`enableContentCheck` + `onContentError` don't work for collaboration:

1. `beforeTransaction` return value is [ignored by Yjs](https://github.com/yjs/yjs/blob/main/src/utils/Transaction.js#L467) - `return false` is a no-op
2. `beforeTransaction` reads pre-transaction state via `yXmlFragmentToProsemirrorJSON(fragment)` - the first corrupt transaction is invisible
3. `filterTransaction` always returns `true` - the [TODO at line 260](https://github.com/ueberdosis/tiptap/blob/main/packages/extension-collaboration/src/collaboration.ts#L260) confirms this was known-incomplete

## Fix

Remove `beforeTransaction`. Move validation into `filterTransaction`:
- `transaction.doc.check()` validates post-transaction state
- `return false` is respected by ProseMirror - transaction is rejected
- `isChangeOrigin(transaction)` targets only Yjs sync transactions
- Non-Yjs transactions and awareness updates (`!docChanged`) pass through unvalidated
- On failure: sets `isDisabled`, emits `contentError` with `disableCollaboration` callback


packages/extension-collaboration/src/collaboration.ts
```diff
@@ -217,49 +218,38 @@ export const Collaboration = Extension.create<CollaborationOptions, Collaboratio
...
+              return false
+            }
+            if (!transaction.docChanged) {
+              return true
+            }
+            try {
+              transaction.doc.check()
+              return true
+            } catch (error) {
+              this.storage.isDisabled = true
```
bdbch, line 242: Is this correct? Asking because you also do this again in the disableCollaboration callback.

packages/extension-collaboration/src/collaboration.ts
```diff
@@ -217,49 +218,38 @@ export const Collaboration = Extension.create<CollaborationOptions, Collaboratio
...
+              this.editor.emit('contentError', {
+                error: error as Error,
+                editor: this.editor,
+                disableCollaboration: () => {
+                  fragment.doc?.destroy()
+                  this.storage.isDisabled = true
```
bdbch: see above

packages/extension-collaboration/src/collaboration.ts
```diff
@@ -217,49 +218,38 @@ export const Collaboration = Extension.create<CollaborationOptions, Collaboratio
...
+            try {
+              transaction.doc.check()
+              return true
+            } catch (error) {
+              this.storage.isDisabled = true
+              this.editor.emit('contentError', {
+                error: error as Error,
+                editor: this.editor,
+                disableCollaboration: () => {
+                  fragment.doc?.destroy()
```
bdbch, line 247: Should we also call this outside of the disableCollaboration? What happens, when the disableCollaboration cb is never called?

packages/extension-collaboration/src/collaboration.ts
dearlordylord: the second one is redundant, my oversight

packages/extension-collaboration/src/collaboration.ts
dearlordylord: (commented the redundant one / removed)

dearlordylord:
I'll get back on 3rd comment soon. Tackled the 1 and 2 comments now because that case was more obvious

packages/extension-collaboration/src/collaboration.ts
dearlordylord: indeed, in this flow, isDisabled blocks future transactions, but the resource stays alive if disableCollaboration is never called - was already present in the older design so I didn't change this behaviour

ProseMirror is protected either way since filterTransaction returns false (that was not in the main branch design!)

what I thought is that giving control over when exactly to clean up would provide users more flexibility, but looking at the flow again I see that the user would have to recreate the Y.Doc either way

so it seems that the correct behaviour would be to call `fragment.doc?.destroy()` just before `this.editor.emit('contentError',`

HOWEVER, that means disableCollaboration becomes effectively a noop (the doc is already destroyed by that point) (alternatively, removing it would break public api surface) - are we ok with disableCollaboration being noop?

bdbch: All good for me to just keep it how it was. I was just wondering it would be needed to be moved outside of the cb function to make sure the fragment is destroyed correctly.

dearlordylord:
Added a changeset
(I elected to force push to keep commits clean, but probably should've just left it for squash...)

PR #7611, opened 2026-03-16
fix(menus): forward HTML props and auto-generate unique menu keys

### LINKED ISSUE #6551: BubbleMenu component doesn't give access to the props of the bubble menu,

bdbch:
## Changes Overview

- forward BubbleMenu and FloatingMenu HTML props to the actual menu container in React and Vue 2
- auto-generate stable per-instance menu plugin keys when `pluginKey` is omitted, so multiple menus can coexist without collisions
- add React regression tests for DOM prop forwarding and duplicate-menu key isolation

## Implementation Approach

- in React, keep a stable menu element, sync HTML attributes/listeners onto that real element, and generate a stable `PluginKey` per component instance when no `pluginKey` is provided
- in Vue 2, forward attrs/listeners/class/style to the rendered root element and generate a fallback plugin key during component setup
- in Vue 3, preserve reactive attr forwarding on the root element and switch omitted `pluginKey` handling to stable per-instance `PluginKey` generation

## Testing Done

- ran `pnpm vitest run packages/react/src/menus/BubbleMenu.spec.ts packages/react/src/menus/FloatingMenu.spec.ts packages/extension-bubble-menu/__tests__/bubble-menu-plugin.spec.ts packages/extension-floating-menu/__tests__/floating-menu-plugin.spec.ts`
- ran `pnpm turbo run build --filter=@tiptap/react --filter=@tiptap/vue-2 --filter=@tiptap/vue-3`
- verified the React BubbleMenu demo locally in Chrome by selecting text, toggling bold, and confirming the text and list bubble menus continue to work without explicit `pluginKey` props

## Verification Steps

- open the React BubbleMenu demo and mount two bubble menus without explicit `pluginKey` props
- select text, click `Bold`, and confirm the inline bubble menu stays available after the command runs
- select a list item and confirm the list-specific bubble menu still appears
- render BubbleMenu/FloatingMenu with `className`, `style`, `data-*`, and event handlers and confirm those bind to the positioned menu container itself

## Additional Notes

- this PR includes two changesets: one for menu HTML prop forwarding and one for automatic menu plugin-key generation
- explicit `pluginKey` behavior remains unchanged; only the omitted-key path now gets isolated per-instance keys

## Checklist

## Related Issues

- Closes #6551

### DIFF
--- a/packages/react/src/menus/useMenuElementProps.ts
+++ b/packages/react/src/menus/useMenuElementProps.ts
@@ -1,5 +1,7 @@
 import type { CSSProperties, HTMLAttributes } from 'react'
-import { useLayoutEffect, useRef } from 'react'
+import { useEffect, useLayoutEffect, useRef } from 'react'
+
+const useIsomorphicLayoutEffect = typeof window !== 'undefined' ? useLayoutEffect : useEffect
 
 type MenuElementProps = HTMLAttributes<HTMLDivElement>
 type MenuSyntheticEvent = Event & {
@@ -333,7 +335,7 @@ export function useMenuElementProps(element: HTMLDivElement, props: MenuElementP
   const previousPropsRef = useRef<MenuElementProps>({})
   const listenersRef = useRef<EventListenerEntry[]>([])
 
-  useLayoutEffect(() => {
+  useIsomorphicLayoutEffect(() => {
     const previousProps = previousPropsRef.current
 
     syncClassName(element, previousProps.className, props.className)


bdbch:
Addressed the Copilot review feedback in `ba6749e7c`:

- tightened React attribute forwarding to known-safe DOM/global attributes plus `aria-*` / `data-*`
- wrapped forwarded event handlers with a lightweight SyntheticEvent-like object (`nativeEvent`, `persist`, `isDefaultPrevented`, `isPropagationStopped`, `currentTarget`) and added explicit event-name mappings for focus, blur, double-click, mouse-enter, and mouse-leave
- made the Vue 2 fallback `pluginKey` lazy so the immediate `editor` watcher can never register with an undefined key
- added a React regression assertion to cover the event wrapper and filtered React-only props

Re-ran:
- `pnpm vitest run packages/react/src/menus/BubbleMenu.spec.ts packages/react/src/menus/FloatingMenu.spec.ts packages/extension-bubble-menu/__tests__/bubble-menu-plugin.spec.ts packages/extension-floating-menu/__tests__/floating-menu-plugin.spec.ts`
- `pnpm turbo run build --filter=@tiptap/react --filter=@tiptap/vue-2 --filter=@tiptap/vue-3`

PR #7626, opened 2026-03-19
fix: prevent cursor jump during IME composition in colored text

### LINKED ISSUE #7621: The cursor will jump when typing text before the end of colored text using an input method
### Affected Packages

extension-text-style

### Version(s)

3.20.3

### Bug Description

1. Select any piece of text in the editor and apply color to it (make sure the text at the end of the paragraph is colored).
2. Insert the cursor before the last character in the paragraph.
3. Insert a character using the input method.
4. The cursor ends up at the end of the paragraph.

https://github.com/user-attachments/assets/3b5d513f-eede-49cd-a552-75596e998cab

Move the cursor to the end of the paragraph.

### Browser Used

Chrome

### Code Example URL

https://tiptap.dev/docs/ui-components/components/color-text-button

### Expected Behavior

The cursor is expected to remain before the last character, not after it.

In fact, this issue does not occur in Firefox or Safari. Even on Tiptap’s official website, the problem exists on some pages but not others.

**Affected pages (with the issue):**
- https://tiptap.dev/docs/ui-components/components/color-text-button
- https://templates.tiptap.dev/Fjt9rjmh8h

**Unaffected pages (no issue):**
- https://tiptap.dev/docs/editor/extensions/functionality/color
- The Notion-like editor at https://tiptap.dev/

Based on my debugging, the code execution flow when inputting characters via an input method is as follows:

1. In `prosemirror-view/domchange.ts` (lines 108–122): `parseBetween` and `findDiff` compare the current ProseMirror state with the DOM-parsed result.
2. In `prosemirror-model/diff.ts` (line 37): `findDiffEnd` calls `sameMarkup` and returns early when marks do not match.
3. Since `injectExtensionAttributesToParseRule` in `@tiptap/core` skips `null`/`undefined` values but retains empty strings (`""`), it causes an inconsistency between the ProseMirror state (`null`) and the DOM parse result (`""`).
4. In `prosemirror-view/domchange.ts` (lines 233–237): When `findDiff` returns `{start:1, endA:3, endB:3}` (where `3` represents the positional coordinate of the cursor at the end of the line), the condition `sel.head == tr.mapping.map(chTo) - 1` evaluates to `true`. This prevents `parse.sel` from being applied, leaving the cursor stuck at the incorrect position.

Based on point 3 above, I believe the logic of `injectExtensionAttributesToParseRule` is correct. Instead, I suggest modifying the default values of the `parseHTML` method in the following files under the `packages/extension-text-style` directory to either `null` or `undefined`:
- `src/background-color/background-color.ts`
- `src/color/color.ts`
- `src/font-family/font-family.ts`
- `src/font-size/font-size.ts`
- `src/line-height/line-height.ts`

Additionally, this issue also occurs if the version of `prosemirror-model` in `demos/package.json` is upgraded to [v1.25.2](https://github.com/ProseMirror/prosemirror-model/releases/tag/1.25.2).

ProseMirror introduced a change in version 1.25.2 (commit [e73b74c](https://github.com/ProseMirror/prosemirror-model/commit/e73b74ca17cc2e1ca8d7a6a608310278502e9370), July 2, 2025):
- **v1.24.1**: `dom.setAttribute("style", "color: #F98181")` → `getAttribute('style')` returns `"color: #F98181"` → `parseHTML` receives the original hex value → consistent with the ProseMirror state.
- **v1.25.2**: `dom.style.cssText = "color: #F98181"` → Chrome normalizes the value → `getAttribute('style')` returns `"color: rgb(249, 129, 129);"` → `parseHTML` receives the RGB value → inconsistent with the hex value in the ProseMirror state.

This change was implemented to bypass CSP (Content Security Policy) restrictions (Issue [#1522](https://github.com/ProseMirror/prosemirror/issues/1522)), but it introduced an unintended side effect of color value normalization. I hope this issue can also be addressed and fixed.

### Dependency Updates

semimikoh:
When typing with IME (Korean/Chinese/Japanese) before the last character
  of colored text, the cursor would jump to the end of the paragraph.

  Root cause: ProseMirror's DOMSerializer sets styles via style.cssText,
  which causes Chrome to normalize hex colors (#FF0000) to rgb format
  (rgb(255, 0, 0)). During IME composition, this mismatch between the
  stored mark attributes and re-parsed DOM values causes findDiff to
  detect the entire mark region as changed, triggering a full replacement
  that misplaces the cursor.

  Fix: Add a custom markView to TextStyle that uses setAttribute instead
  of cssText, preserving the original color format. Also filter empty
  strings in parseHTML and injectExtensionAttributesToParseRule to prevent
  additional attribute mismatches.

  Closes #7621
 ## Changes Overview

  Fix cursor jumping to end of paragraph when typing with IME (Korean/Chinese/Japanese) before the last character of text styled with
  color/textStyle marks.

  - Added `addMarkView` to TextStyle extension that uses `setAttribute` instead of `style.cssText`
  - Added empty string filtering in `parseHTML` for color, backgroundColor, fontFamily, fontSize, lineHeight
  - Added empty string check in `injectExtensionAttributesToParseRule`

  ## Implementation Approach

  **Root cause:** ProseMirror's `DOMSerializer` sets styles via `dom.style.cssText`, which causes Chrome to normalize hex colors (`#FF0000`) to
  rgb (`rgb(255, 0, 0)`). During IME composition, `MarkViewDesc` becomes `CONTENT_DIRTY` and falls back to re-parsing the DOM — reading the
  normalized rgb value instead of the original hex. This mismatch makes `findDiff` detect the entire mark region as changed, triggering a
  full-range replacement where Chrome's composition guard blocks correct cursor positioning.

  **Fix:** A custom `addMarkView` creates the span using `span.setAttribute('style', ...)` which preserves the original color format, preventing
   the hex→rgb normalization.

  **Alternatives considered:**
  - Normalize colors to rgb at storage time — loses user's original format, wider change scope
  - Patch ProseMirror's DOMSerializer upstream — maintainer is conservative about changes
  - Set `inclusive: false` on textStyle mark — causes side effects (new text won't inherit style at mark boundary)

  ## Testing Done

  - Manually tested IME input (Korean) before last character of colored text — cursor stays in place
  - Manually tested IME input in middle of colored text — works normally
  - Manually tested non-IME input with color/font-family/font-size — works as before
  - All 684 unit tests pass

  ## Verification Steps

  1. Open demo with TextStyle/Color extensions
  2. Select text and apply a color
  3. Place cursor before the last character of the colored text
  4. Type with IME (Korean/Chinese/Japanese input method)
  5. **Before fix:** cursor jumps to end of paragraph after composition
  6. **After fix:** cursor stays at the correct position

  ## Additional Notes

  Debug logging confirmed the issue — during IME composition, `findDiff` returns `{start: 410, endA: 429, endB: 429}` (endA==endB means
  mark-only change), replacing the entire colored range instead of just the inserted text. This only happens in Chrome due to `cssText`
  normalization.

  Open to feedback if there's a better approach!

  ## Checklist

  ## Related Issues

  Closes #7621


packages/core/src/helpers/injectExtensionAttributesToParseRule.ts
```diff
@@ -31,7 +31,7 @@ export function injectExtensionAttributesToParseRule(
           ? item.attribute.parseHTML(node)
           : fromString(node.getAttribute(item.name))
 
-        if (value === null || value === undefined) {
+        if (value === null || value === undefined || value === '') {
```
snovey: I don't think it's necessary to modify this.

semimikoh: @snovey  You're right, reverted. This change was unnecessary for the fix.
https://github.com/ueberdosis/tiptap/pull/7626/changes/e4da8efc850812907edb568f74770de5beb98a8c

packages/extension-text-style/src/text-style/index.ts
```diff
@@ -123,6 +123,26 @@ export const TextStyle = Mark.create<TextStyleOptions>({
     return ['span', mergeAttributes(this.options.HTMLAttributes, HTMLAttributes), 0]
   },
 
+  addMarkView() {
```
bdbch: Introducing a `MarkView` here feels too heavy for this bug. The issue is about keeping color attributes stable during IME composition, but a mark view changes the live rendering path for every `textStyle` mark, not just color handling. That increases the surface area for regressions and also makes the editor DOM behavior diverge from the normal `renderHTML()` path.

packages/extension-text-style/src/text-style/index.ts
```diff
@@ -123,6 +123,26 @@ export const TextStyle = Mark.create<TextStyleOptions>({
...
+    return ({ HTMLAttributes }) => {
+      const span = document.createElement('span')
+
+      // Use setAttribute instead of style.cssText to preserve original
+      // color format (e.g. hex #FF0000) and prevent browser normalization
+      // to rgb() which causes mark attribute mismatches during IME composition.
+      Object.entries(HTMLAttributes).forEach(([attr, value]) => {
+        if (value != null && value !== false) {
+          span.setAttribute(attr, String(value))
```
bdbch: This brings back the CSP problem ProseMirror fixed in `e73b74c`. Using `setAttribute('style', ...)` can be blocked in stricter CSP environments, which is why upstream switched to `style.cssText` in the first place. Fixing the IME issue by reintroducing that behavior would trade one bug for a security/compatibility regression.

bdbch (changes requested):
Thanks for digging into the IME cursor jump. The diagnosis is useful, but I don't think we should merge this approach as-is. The new `textStyle` mark view changes the rendering path for every `textStyle` mark, and the implementation relies on `setAttribute('style', ...)`, which reintroduces the CSP/security compatibility issue ProseMirror explicitly fixed upstream. I left line comments with the main concerns.

packages/extension-text-style/src/text-style/index.ts
bdbch: The safer direction here seems to be stabilizing color values instead of introducing a `MarkView`. In particular, we should canonicalize color values at the `Color` / `BackgroundColor` extension boundary so the stored mark attrs and the reparsed DOM use the same representation during IME composition. That keeps the normal rendering path intact, avoids reintroducing the CSP issue from `setAttribute('style', ...)`, and narrows the fix to the actual source of the mismatch rather than changing how every `textStyle` mark is rendered.

semimikoh: @bdbch 

Thanks for the review! I've replaced the MarkView approach with color normalization at the Color/BackgroundColor extension boundary. Now
   both parseHTML and setColor/setBackgroundColor run through a shared normalizeColor() utility that canonicalizes values to rgb() via the 
  browser's CSS parser, so stored attrs and reparsed DOM always match during IME composition.

bdbch (changes requested):
Thanks for reworking this away from the MarkView approach. Normalizing at the Color / BackgroundColor extension boundary is much closer to the right shape and addresses the CSP concern from the earlier version.

I still don’t think we can merge it as-is, though, because the fix currently only stabilizes color values when they enter through parseHTML() or setColor() / setBackgroundColor(). Tiptap content can also come from JSON / Yjs / persisted document state, and those paths can still contain existing hex values like #ff0000 in mark attrs. In that case we still render the mark from the stored hex value, Chrome reparses the DOM as rgb(...) during IME composition, and we end up with the same attr mismatch again.

I think the fix needs to make color attrs stable regardless of how the document was created. A better direction would be to canonicalize the textStyle color attrs before they are compared/rendered in the editor state as well, so JSON-loaded content and command-created content converge on the same representation. For example, that could mean normalizing these attrs when content is imported into the editor / parsed into the ProseMirror doc, or otherwise ensuring the mark attrs are canonical even when they originate from JSON rather than HTML.

Also, the current changeset still includes @tiptap/core, but this version of the PR doesn’t change anything in core anymore, so that release entry should be removed unless core changes come back.

Please also add a regression test that starts from JSON content with a hex color attr, since that’s the path the current implementation still misses.

semimikoh:
@bdbch  Thanks for the feedback! I've addressed all three points:

  1. JSON / Yjs / persisted content normalization — Added onCreate and onTransaction hooks to both Color and BackgroundColor extensions. onCreate normalizes
  hex values in the initial document (e.g. JSON-loaded content), and onTransaction catches any subsequent external updates (e.g. Yjs). Also added
  normalizeColor() in renderHTML as a safety net so the DOM output is always canonical regardless of stored attr format.
  2. Removed @tiptap/core from changeset — Only @tiptap/extension-text-style remains since core is no longer touched.
  3. Added regression tests — Integration tests that create an editor with JSON content containing hex color attrs and verify they get normalized to rgb().
  Also covers the external update (Yjs-like) path.

packages/extension-text-style/src/utilities/normalize-color.ts
```diff
@@ -0,0 +1,19 @@
+/**
+ * Normalize a CSS color value to a canonical form using the browser's
+ * built-in CSS parser.  The browser always returns colors in `rgb(…)` or
+ * `rgba(…)` format, which gives us a stable representation that won't
+ * change across re-parses (e.g. during IME composition).
+ *
+ * Falls back to the original value when `document` is unavailable (SSR).
+ */
+export function normalizeColor(color: string): string {
```
snovey, line 22: Every call to the `normalizeColor` method creates a `<span>` tag. If we have to use this approach, would it be better to implement it as a singleton tied to the editor's lifecycle?

packages/extension-text-style/src/utilities/normalize-color-attrs.ts
```diff
@@ -0,0 +1,66 @@
...
+ * when at least one attr was rewritten.
+ *
+ * Returns `true` if a normalizing transaction was dispatched.
+ */
+export function normalizeDocColorAttrs(editor: Editor, attrName: 'color' | 'backgroundColor'): boolean {
+  const { state } = editor
+  let hasChanges = false
+  const tr = state.tr
+
+  state.doc.descendants((node, pos) => {
```
snovey: `onTransaction` fires very frequently, and the `state.doc.descendants` method is quite expensive. For the current implementation, this solution is too heavy-handed.

snovey:
Following the current approach, I believe the minimal complete solution consists of two parts:
1. Change all fallback return values of `parseHTML` under the `extension-text-style` directory to `null` (i.e., exclude the `addMarkView` related code in the initial commit).
2. Maintain a singleton to retrieve the browser-computed color values (i.e., refactor the `normalizeColor` method into the singleton pattern).

However, I consider the second part of the solution rather tricky and dependent on the browser environment.

I haven’t come up with a better alternative either, which is why I only filed a bug report instead of submitting a PR.

packages/extension-text-style/src/utilities/normalize-color.ts
semimikoh: Refactored to reuse a singleton <span> element and added a Map cache so repeated lookups skip DOM access entirely.

packages/extension-text-style/src/utilities/normalize-color-attrs.ts
semimikoh: Removed the onTransaction hook. Colors are normalized at parseHTML, renderHTML, and command boundaries. Initial content is
  handled once in onCreate.

bdbch (changes requested):
Hey, thanks for the updates! The approach of normalizing via `parseHTML`/`renderHTML` and handling the initial load in `onCreate` is definitely a step in the right direction.

However, we still can't merge this as-is because handling programmatic or collaborative (e.g. Yjs) updates is missing, and scanning the entire document on every transaction would cause severe performance issues on larger documents.

Instead, could you handle this by creating a ProseMirror plugin with an `appendTransaction` hook (using `addProseMirrorPlugins`)? 

This is the standard ProseMirror way to enforce document constraints. With `appendTransaction`, you can:
1. Check if the document actually changed (`transaction.docChanged`).
2. Use the transaction's step map (`transaction.mapping`) to iterate **only** over the specific ranges of the document that were modified.
3. If you find un-normalized hex colors in those specific ranges, you can return a new transaction that fixes them.

This ensures we don't tank performance by checking the whole document on every keystroke, and it will safely intercept inbound Yjs syncs.

Let me know if you need any pointers on how to utilize `appendTransaction`!

semimikoh:
@bdbch  ---
  Thanks for the great suggestion! I've refactored the approach to use a ProseMirror plugin with appendTransaction via
  addProseMirrorPlugins.

  What changed:
  - Replaced the onCreate full-document scan with a plugin that has two hooks:
    - view() — normalizes initial JSON/HTML content on editor mount
    - appendTransaction — uses findDiffStart/findDiffEnd to compute the changed range and only walks those nodes
  - Removed the now-unused normalizeDocColorAttrs utility
  - Added a test case for programmatic/collaborative updates being normalized via appendTransaction

  This ensures Yjs syncs and programmatic mark updates are intercepted without scanning the entire document on every
  transaction.

bdbch:
Sorry @semimikoh - was this on purpose? I wasn't able to get back to your PR yet because my inbox is filled up again.

semimikoh:
@bdbch Sorry, not on purpose! I deleted my fork while cleaning up and didn't realize it would close the PR. The appendTransaction version is still intact — I'll re-fork and open a new PR and link it here. Thanks for following up!

semimikoh:
@bdbch Re-opened as #7899 (same commits, fresh fork). The `appendTransaction` plugin refactor you suggested is included. Thanks again for following up!

Issue #3872, opened 2023-03-23
getMarkRange should be handled differently for link


hasakilol:
### What’s the bug you are facing?

If two links are ajacent with no white space between, getMarkRange will return the range containing both the links, but not one.

### Which browser was this experienced in? Are any special extensions installed?

Chrome

### How can we reproduce the bug on our side?

I will provide with a demo later. Please don't delete my post.

### What did you expect to happen?

Range with only one link is returned.

### Did you update your dependencies?

- [X] Yes, I’ve updated my dependencies to use the latest version of all packages.

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

hasakilol:
```
let startIndex = start.index
  let startPos = $pos.start() + start.offset
  let endIndex = startIndex + 1
  let endPos = startPos + start.node.nodeSize

  findMarkInSet([...start.node.marks], type, attributes) **// this code seems useless???**

  while (startIndex > 0 && mark.isInSet($pos.parent.child(startIndex - 1).marks)) {
    startIndex -= 1
    startPos -= $pos.parent.child(startIndex).nodeSize
  }

  while (
    endIndex < $pos.parent.childCount
    && isMarkInSet([...$pos.parent.child(endIndex).marks], type, attributes) 
    **// in git history, this was mark.isInSet($pos.parent.child(endIndex).marks). Why did the change happen?**
  ) {
    endPos += $pos.parent.child(endIndex).nodeSize
    endIndex += 1
  }
```

nperez0111:
This is released with 2.10.0

PR #7628, opened 2026-03-19
fix(core): extendMarkRange match only the current mark of a type

mweidner037:
## Changes Overview

Currently, If two links are adjacent with no white space between, extendMarkRange will select the range containing both the links. This PR changes it to only select the link containing the original selection.

An analogous change was already made for the underlying getMarkRange function (https://github.com/ueberdosis/tiptap/pull/5826). However, extendMarkRange's default of `attributes = {}` was overriding that change.

## Implementation Approach

Remove the `attributes = {}` default, so that omitted/undefined attributes pass through to getMarkRange. getMarkRange then applies its own default, which is to use the attributes of the original mark.

## Testing Done

See changes to extendMarkRange tests.

## Verification Steps

In the Marks/Link preview, create two adjacent links with different hrefs. Then, place your cursor in one of the links, click Set Link, and update it to a 3rd href.

Before this PR: both adjacent links acquire the 3rd href.

After this PR: only the link containing the cursor acquires the 3rd href.

## Additional Notes

You can recover the original behavior by passing explicitly passing `{}` as the attributes parameter.

## Checklist

## Related Issues

Analogous getMarkRange issue: https://github.com/ueberdosis/tiptap/issues/3872

### DIFF
--- a/.changeset/mean-islands-change.md
+++ b/.changeset/mean-islands-change.md
@@ -2,4 +2,4 @@
 '@tiptap/core': patch
 ---
 
-Remove extendMarkRange attributes default, to avoid extending across adjacent links with different hrefs
+extendMarkRange defaults to using the attributes of the first mark of the given type, instead of `attributes = {}`. In particular, `extendMarkRange('link')` no longer extends to adjacent links with different hrefs; restore the previous behavior with `extendMarkRange('link', {})`.


packages/core/__tests__/extendMarkRange.spec.ts
```diff
@@ -1,5 +1,5 @@
 // eslint-disable-next-line
```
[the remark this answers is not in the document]
mweidner037: This was probably here to suppress warnings for the unused getDebugJSON import, which my IDE auto-removed. Unless there is a pressing need to keep that unused import (for a commented debug line), I will keep it removed and also remove this eslint suppression.

packages/core/src/commands/extendMarkRange.ts
```diff
@@ -31,7 +31,7 @@ declare module '@tiptap/core' {
 }
 
 export const extendMarkRange: RawCommands['extendMarkRange'] =
-  (typeOrName, attributes = {}) =>
+  (typeOrName, attributes) =>
   ({ tr, state, dispatch }) => {
```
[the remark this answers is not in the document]
mweidner037: I've updated the `attributes` typedoc to use the same wording as getMarkRange.

packages/core/src/commands/extendMarkRange.ts
```diff
@@ -31,7 +32,7 @@ declare module '@tiptap/core' {
 }
 
 export const extendMarkRange: RawCommands['extendMarkRange'] =
```
bdbch, line 34: Passing undefined through here fixes the adjacent-link case, but it also exposes a broader issue in getMarkRange: when attributes are omitted it defaults to start.node.marks[0]?.attrs, which is the first mark on the text node, not necessarily the requested mark type. On text with multiple marks, extendMarkRange(type) can now fail depending on mark order. I think we need to either derive the default attrs from the matching mark of type in getMarkRange, or keep this command from delegating to that default until the helper is fixed.

packages/core/__tests__/extendMarkRange.spec.ts
```diff
@@ -286,4 +273,134 @@ describe('extendMarkRange', () => {
 
     editor.destroy()
   })
```
bdbch, line 276: The new tests cover the adjacent-link behavior well, but I think we also need a regression case for a text node carrying multiple marks, for example link plus bold or two marks where the target mark is not first in the mark set. Without that, this change can look green while extendMarkRange(type) still regresses for omitted attributes on multi-mark text.

bdbch:
One additional concern here: I think this is effectively a public API behavior change, not just an internal implementation tweak.

Today, omitting `attributes` in `extendMarkRange(type)` behaves like passing an empty object, so adjacent marks of the same type are treated as one range even if their attrs differ. With this change, omitting `attributes` starts meaning "match only the current mark". That is a better default for links, but it is still a semantic change for existing consumers.

Because of that, I think we should be careful about how we merge and ship this so we do not create migration surprises:

- if we keep this behavior change, the changelog/changeset should explicitly call out that omitted `attributes` and `{}` now behave differently
- we should document the migration path clearly: pass `extendMarkRange(type, {})` to preserve the previous behavior
- if maintainers feel this is too risky for a patch release, another option is to preserve the old `extendMarkRange` default for now and introduce the link-specific fix in a more targeted way, or defer the default-semantics change to the next major

I do think the new behavior is more intuitive for the link case; I just want us to be explicit about whether we are shipping a bug fix or intentionally changing the command contract for existing users.

packages/core/src/commands/extendMarkRange.ts
mweidner037: Good catch. I've added a fix for getMarkRange, and a test that fails without that fix.

packages/core/__tests__/extendMarkRange.spec.ts
mweidner037: Added!

mweidner037:
> One additional concern here: I think this is effectively a public API behavior change, not just an internal implementation tweak.
> 
> Today, omitting `attributes` in `extendMarkRange(type)` behaves like passing an empty object, so adjacent marks of the same type are treated as one range even if their attrs differ. With this change, omitting `attributes` starts meaning "match only the current mark". That is a better default for links, but it is still a semantic change for existing consumers.
> 
> Because of that, I think we should be careful about how we merge and ship this so we do not create migration surprises:
> 
>     * if we keep this behavior change, the changelog/changeset should explicitly call out that omitted `attributes` and `{}` now behave differently
> 
>     * we should document the migration path clearly: pass `extendMarkRange(type, {})` to preserve the previous behavior
> 
>     * if maintainers feel this is too risky for a patch release, another option is to preserve the old `extendMarkRange` default for now and introduce the link-specific fix in a more targeted way, or defer the default-semantics change to the next major
> 
> 
> I do think the new behavior is more intuitive for the link case; I just want us to be explicit about whether we are shipping a bug fix or intentionally changing the command contract for existing users.

I agree that this is technically a breaking change, since it changes the behavior of existing code. However it would be nice to release it sooner than Tiptap v4 :)

I did a quick check for uses of extendMarkRange in the wild, by searching GitHub globally for ".extendMarkRange". Within the first five pages of code results, essentially all used the command for its "standard" purpose: `extendMarkRange('link')`, followed by either `setLink` or `unsetLink`. I would guess that these uses expect the updated behavior (don't expand to adjacent links) and will be happy with this change.

I've also updated the changeset to clarify the change and how to preserve the old behavior.

PR #7629, opened 2026-03-20
fix(drag-handle): support RTL direction for drag ghost image

wbcoder0:
## Changes Overview

Fix drag ghost (`setDragImage`) for RTL editors and mixed-direction blocks: the temporary wrapper now mirrors the **dragged block’s** text direction, and the drag hotspot is aligned to the **pointer** using the wrapper’s bounds (not a fixed corner).

## Implementation Approach

1. **Direction on the ghost wrapper**  
   Resolve the DOM node at the drag range start (`view.domAtPos`), normalize text nodes to their `parentElement`, then read `getComputedStyle(...).direction` (falling back to `view.dom`). Set `wrapper`’s `dir` so the clone matches the block (including RTL page + LTR paragraph cases).

2. **Drag image hotspot**  
   `DataTransfer.setDragImage` uses the off-screen `wrapper` as the image. After cloning and `append`, measure `wrapper.getBoundingClientRect()` and set the hotspot x to `clientX - wrapperRect.left`, clamped to `[0, wrapperRect.width]`, so the ghost tracks the cursor in both LTR and RTL.

## Testing Done

- Manual: drag handle in an RTL-configured editor; mixed Arabic/Latin paragraphs where block `direction` differs from `view.dom`.
- Confirmed ghost text direction and pointer alignment during drag.

## Verification Steps

1. Set editor / root to `dir="rtl"` (or CSS `direction: rtl`).
2. Drag a block via the drag handle; confirm the ghost renders with correct direction and stays under/near the cursor.
3. Optional: `dir="auto"` or LTR block inside RTL editor — ghost should follow the **block** direction.

## Additional Notes

- Supersedes an earlier approach that only toggled `dir` on the editor root and used `offsetWidth` as the hotspot, which mis-handled mixed-direction content and mis-aligned the ghost.

### DIFF
--- a/packages/extension-drag-handle/src/helpers/getDraggedBlockDir.ts
+++ b/packages/extension-drag-handle/src/helpers/getDraggedBlockDir.ts
@@ -21,6 +21,8 @@ export function getDraggedBlockDir(view: EditorView, pos: number): string {
       draggedDom = child
     } else if (node instanceof Element) {
       draggedDom = node
+    } else if (node.nodeType === Node.TEXT_NODE && node.parentElement) {
+      draggedDom = node.parentElement
     }
   }
 
--- a/packages/extension-drag-handle/__tests__/getDraggedBlockDir.spec.ts
+++ b/packages/extension-drag-handle/__tests__/getDraggedBlockDir.spec.ts
@@ -143,6 +143,29 @@ describe('getDraggedBlockDir', () => {
 
       expect(getDraggedBlockDir(view, 0)).toBe('rtl')
     })
+
+    it('should normalize a Text node from domAtPos to its parentElement instead of falling back to editor root', () => {
+      const editorRoot = document.createElement('div')
+      const rtlParagraph = document.createElement('p')
+      const textNode = document.createTextNode('مرحبا')
+
+      rtlParagraph.appendChild(textNode)
+
+      const dirMap = new Map<Element, string>([
+        [editorRoot, 'ltr'],
+        [rtlParagraph, 'rtl'],
+      ])
+
+      mockDirection(dirMap)
+
+      const view = createMockView({
+        dom: editorRoot,
+        nodeDOM: vi.fn(() => null),
+        domAtPos: vi.fn(() => ({ node: textNode, offset: 3 })),
+      })
+
+      expect(getDraggedBlockDir(view, 0)).toBe('rtl')
+    })
   })
 
   describe('ultimate fallback to editor root', () => {


packages/extension-drag-handle/src/helpers/dragHandler.ts
```diff
@@ -89,6 +89,17 @@ export function dragHandler(
 
   const { tr } = view.state
   const wrapper = document.createElement('div')
+
+  const domNode = view.domAtPos(ranges[0].$from.pos).node
```
bdbch: I don't think `domAtPos(...).node` is a stable way to identify the dragged block here. At block boundaries ProseMirror can return the parent container plus an offset, so this can still read the editor root/list wrapper direction instead of the actual block being dragged. That means mixed RTL/LTR cases may still produce the wrong ghost direction. Could we derive the direction from the dragged DOM node itself (for example via `view.nodeDOM(from)` or by resolving the child at the returned offset) before setting `wrapper.dir`?

packages/extension-drag-handle/PULL_REQUEST_BODY.md
```diff
@@ -0,0 +1,28 @@
+<!-- Copy the sections below into your GitHub PR description when opening/updating the MR. -->
```
bdbch: I think this is a left-over from your agent session. Please remove this.

bdbch (changes requested):
Thanks for tackling the RTL drag-ghost issue. I think the hotspot adjustment is moving in the right direction, but I don't think the current direction lookup is reliable enough to merge yet. `view.domAtPos(...).node` can resolve to a container at block boundaries instead of the dragged block itself, so mixed-direction content can still inherit the wrong `dir`. This also needs automated regression coverage for the new drag-image behavior instead of relying only on manual verification.

packages/extension-drag-handle/src/helpers/dragHandler.ts
wbcoder0: Thanks for the catch! You're right — `domAtPos(...).node` can return the parent container at block boundaries, making it unreliable for reading the dragged block's direction.

I've refactored the direction detection into a standalone helper (`getDraggedBlockDir`) that:

1. **Primarily** uses `view.nodeDOM(pos)`, which maps a position directly to the block-level DOM element — no parent-container ambiguity. This is also consistent with how the same file already uses `view.nodeDOM` at the cloning step.
2. **Falls back** to `view.domAtPos(pos)` but now properly resolves `node.childNodes[offset]` instead of using `node` directly.
3. **Ultimate fallback** to `view.dom`'s direction, defaulting to `'ltr'`.

Also added 8 unit tests in `getDraggedBlockDir.spec.ts` covering:
- `nodeDOM` returning an RTL / LTR element directly
- `domAtPos` fallback: parent container with child-at-offset resolution
- Mixed RTL/LTR regression: ensures direction is read from the dragged block, not the parent wrapper
- Ultimate fallback to editor root direction / default `'ltr'`

All 111 tests pass (including the 8 new ones).

packages/extension-drag-handle/PULL_REQUEST_BODY.md
wbcoder0: Removed — thanks for flagging.

packages/extension-drag-handle/src/helpers/getDraggedBlockDir.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  if (nodeDom instanceof Element) {
+    draggedDom = nodeDom
+  } else {
+    const { node, offset } = view.domAtPos(pos)
+    const child = node.childNodes[offset]
+
+    if (child instanceof Element) {
+      draggedDom = child
+    } else if (node instanceof Element) {
```
bdbch, line 22: This still misses the `domAtPos()`-returns-a-`Text`-node case. If `nodeDOM(pos)` is null and `domAtPos(pos)` lands inside a text node within an RTL paragraph, `childNodes[offset]` is undefined and `node instanceof Element` is false, so we fall back to `view.dom` and can still inherit the editor root direction instead of the dragged block's direction. Could we normalize `Text` nodes to `node.parentElement` here before falling back to the editor root, and add a regression test for that shape?

wbcoder0: Good catch — that was indeed a gap. Fixed by adding a `Text`-node normalization branch in `getDraggedBlockDir`:

```ts
} else if (node.nodeType === Node.TEXT_NODE && node.parentElement) {
  draggedDom = node.parentElement
}
```

So when `nodeDOM(pos)` is null and `domAtPos(pos)` lands inside a Text node (e.g. `{ node: Text("مرحبا"), offset: 3 }` within an RTL `<p>`), we now walk up to `node.parentElement` and read its direction instead of falling through to `view.dom`.

Added a regression test for this exact shape: LTR editor root + RTL paragraph containing an Arabic Text node — asserts that direction resolves to `'rtl'`, not the editor root's `'ltr'`.

All 112 tests pass.

PR #7652, opened 2026-03-27
fix(core): When dragging between editors, only delete the source content if dropEffect is 'move'

mweidner037:
## Changes Overview

The browser Drag&Drop API allows specifying a [dropEffect](https://developer.mozilla.org/en-US/docs/Web/API/DataTransfer/dropEffect). Usually, this is "move", but ProseMirror lets you change it to "copy" by holding down a modifier key when you initiate dragging ([dragstart line](https://github.com/ProseMirror/prosemirror-view/blob/17af0f4b25933933f644a655ad6bb9a0249d7cb5/src/input.ts#L707), [dragMoves def](https://github.com/ProseMirror/prosemirror-view/blob/17af0f4b25933933f644a655ad6bb9a0249d7cb5/src/input.ts#L675)).

Tiptap's PasteRule drop handler correctly implements the "move" behavior when dragging between editors. However, it does not actually check that the intended dropEffect is "move". This PR adds a check.

## Implementation Approach

Before removing the source content, check if `dropEvent.dataTransfer?.dropEffect === 'move'`.

## Testing Done

My own project involves two editors side-by-side. Before the change, holding Option (on Mac) while initiating dragging from editor to another would show the Drag-Copy cursor (a green plus sign in Chrome), but dropping would actually move the content.

After copying the change to Tiptap's PasteRule.ts into the same project, repeating the action successfully copies the content - leaving the original editor unchanged.

Dragging without holding Option still moves the content, as expected.

## Checklist

### DIFF
--- a/.changeset/small-icons-double.md
+++ b/.changeset/small-icons-double.md
@@ -1,0 +1,5 @@
+---
+'@tiptap/core': patch
+---
+
+When dragging between editors, only delete the source content if dropEffect is 'move'


mweidner037:
Note that respecting the dropEffect in this way increases the impact of https://github.com/ProseMirror/prosemirror/issues/1208: if you drag onto the non-contentDOM portion of a NodeView, the browser chooses the "copy" effect, even if the user initiated with the "move" action (no modifier keys). (Tested in Chrome and Firefox.)

This already causes an unexpected Drag-Copy cursor to appear, though the operation is still "move" because Tiptap ignores the dropEffect. But after the current PR, it will cause the operation to actually copy - consistent with the cursor but not with user expectations.

If desired, I could rework this to re-check dragMoves instead of trusting the dropEffect, using the same logic [as ProseMirror](https://github.com/ProseMirror/prosemirror-view/blob/17af0f4b25933933f644a655ad6bb9a0249d7cb5/src/input.ts#L740).

mweidner037:
Closing and I will create a separate PR that copies ProseMirror's behavior, instead of copying the browser's behavior.

PR #7654, opened 2026-03-27
Add aria-label support to details toggle button

arnaugomez:
## Summary
- add an `a11y.toggleButtonLabel` option to `@tiptap/extension-details`
- apply a default `aria-label` to the generated toggle button and keep it in sync with the open state
- add unit coverage for the default label and a custom override, and include a changeset

## Testing
- `pnpm test:unit -- packages/extension-details/__tests__/details.spec.ts`
- `pnpm lint:fix`

### DIFF
--- a/.changeset/flat-taxis-promise.md
+++ b/.changeset/flat-taxis-promise.md
@@ -1,0 +1,5 @@
+---
+'@tiptap/extension-details': patch
+---
+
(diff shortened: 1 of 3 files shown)


bdbch:
@arnaugomez can you. also update the docs for the extension to reflect the new option?

PR #7659, opened 2026-03-27
fix: isolate markdown lexer helpers during tokenization

bdbch:
## Changes Overview

Fixes a markdown parsing bug where custom tokenizer helper methods could bind to a retained lexer instead of Marked's active lexer. This caused ordered list parsing and other re-entrant tokenizer flows to break inline tokenization in subsequent paragraphs.

## Implementation Approach

Removed the persistent lexer stored in `MarkdownManager` and switched helper creation to parse-local lexers. Custom markdown tokenizer callbacks now use Marked's active `this.lexer`, so recursive `inlineTokens()` and `blockTokens()` calls stay within the active parse session instead of leaking into shared lexer state.

## Testing Done

Ran:
- `pnpm vitest run packages/markdown/__tests__/manager.spec.ts`
- `pnpm vitest run packages/markdown/__tests__/conversion.spec.ts`
- `pnpm --filter @tiptap/markdown build`

Added regression coverage for:
- ordered list followed by an inline-marked paragraph
- parity with an isolated `new Marked()` instance
- a custom tokenizer that calls `inlineTokens()` and no longer corrupts later inline parsing

## Verification Steps

1. Parse markdown with an ordered list followed by a paragraph containing inline marks such as `**bold**`.
2. Confirm the paragraph after the list still receives inline tokens and renders as marked content.
3. Parse the same markdown through both the default markdown manager and one configured with an isolated `new Marked()` instance.
4. Confirm both outputs match.

## Additional Notes

This keeps the fix contained to `@tiptap/markdown` and does not change `@tiptap/extension-list` behavior or the public API.

## Checklist

## Related Issues

- Customer-reported markdown lexer state bug affecting ordered-list parsing and subsequent inline tokenization.

### DIFF
--- a/packages/markdown/src/MarkdownManager.ts
+++ b/packages/markdown/src/MarkdownManager.ts
@@ -94,7 +94,7 @@ export class MarkdownManager {
    * `markdownName`, `parseMarkdown`, `renderMarkdown` and `priority` from the
    * extension config (using the same resolution used across the codebase).
    */
-  registerExtension(extension: AnyExtension, _recreateLexer?: boolean): void {
+  registerExtension(extension: AnyExtension): void {
     // Keep track of all extensions for HTML parsing
     this.extensions.push(extension)
 


packages/markdown/src/MarkdownManager.ts
```diff
@@ -143,13 +141,24 @@ export class MarkdownManager {
...
+  private createTokenizerHelpers(lexer: Lexer): MarkdownLexerConfiguration {
+    return {
+      inlineTokens: (src: string) => lexer.inlineTokens(src),
+      blockTokens: (src: string) => lexer.blockTokens(src),
     }
   }
 
+  private tokenizeInline(src: string): MarkdownToken[] {
+    return this.createLexer().inlineTokens(src) as MarkdownToken[]
+  }
```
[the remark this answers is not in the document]
bdbch: Followed up on this now. `MarkdownManager.parse()` creates a parse-scoped lexer and keeps it available during token-to-JSON conversion, so post-lex inline tokenization paths reuse that lexer instead of allocating a fresh one per call. This keeps the active-tokenizer fix intact while removing the hot-path allocation Copilot called out. Verified with the markdown manager tests, conversion tests, and package build.

packages/markdown/src/MarkdownManager.ts
```diff
@@ -289,16 +288,26 @@ export class MarkdownManager {
...
+    this.activeParseLexer = parseLexer
 
-    // Return a document node containing the parsed content
-    return {
-      type: 'doc',
-      content,
+    try {
+      // Use a parse-scoped lexer so follow-up inline tokenization can reuse
+      // the same configured lexer state without sharing it across parses.
+      const tokens = parseLexer.lex(markdown) as MarkdownToken[]
```
arnaugomez, line 299: Why is this `as` operator necessary?

packages/markdown/src/MarkdownManager.ts
```diff
@@ -95,7 +94,7 @@ export class MarkdownManager {
    * `markdownName`, `parseMarkdown`, `renderMarkdown` and `priority` from the
    * extension config (using the same resolution used across the codebase).
    */
-  registerExtension(extension: AnyExtension, recreateLexer: boolean = true): void {
+  registerExtension(extension: AnyExtension, _recreateLexer?: boolean): void {
```
arnaugomez: If _recreateLexer is not used and it's the last parameter, maybe you can remove it.

packages/markdown/src/MarkdownManager.ts
```diff
@@ -143,13 +142,24 @@ export class MarkdownManager {
...
+
+  private createTokenizerHelpers(lexer: Lexer): MarkdownLexerConfiguration {
+    return {
+      inlineTokens: (src: string) => lexer.inlineTokens(src),
+      blockTokens: (src: string) => lexer.blockTokens(src),
     }
   }
 
+  private tokenizeInline(src: string): MarkdownToken[] {
+    return (this.activeParseLexer ?? this.createLexer()).inlineTokens(src) as MarkdownToken[]
```
arnaugomez, line 160: Why is this as operator necessary?

packages/markdown/src/MarkdownManager.ts
bdbch: Marked’s lexer returns Marked token types (Token[] / TokensList), while MarkdownManager works with our MarkdownToken abstraction. Those shapes overlap at runtime, but our MarkdownToken type allows additional fields used by Tiptap extensions/tokenizers, so TypeScript won’t treat Marked’s return type as assignable automatically. The cast is just telling TS that these lexer results are safe to use as MarkdownToken[] in our pipeline.

packages/markdown/src/MarkdownManager.ts
bdbch: See above

PR #7663, opened 2026-03-27
fix(core) Don't stop dragover/dragenter events in NodeViews, to prevent spurious drag-copy cursors

mweidner037:
## Changes Overview

When you drag some content over a NodeView but outside of its contentDOM, NodeView.stopEvent stops the corresponding dragover event. This sometimes causes the drag cursor to become the drag-copy cursor (a green plus sign), visually suggest that the drag will copy the source content, even though it will actually move.

You can see this reliably in the GuideNodeViews/DragHandle demo in Firefox: if you start dragging and place your cursor just to the right of another drag cursor (in its right margin), the drag-copy cursor appears. I've also seen this behavior in my own app in Chrome, though it does not seem to occur in that demo.

The PR changes NodeView's stopEvent to allow dragover and dragenter events, falling back on [ProseMirror's behavior](https://github.com/ProseMirror/prosemirror-view/blob/5bcfa0ebd4cff7f13c936bcde6d39b4d7df22b75/src/input.ts#L717), which is to preventDefault (signaling that [a drop is allowed here](https://developer.mozilla.org/en-US/docs/Web/API/HTML_Drag_and_Drop_API/Drag_operations#dragging_over_elements_and_specifying_drop_targets)).

With this change, I no longer can produce the unintended drag-copy cursor, either in the demo + Firefox or in my app + Firefox/Chrome.

## Implementation Approach

Add dragenter/dragover events to the section under `// these events are handled by prosemirror`.

## Checklist

### DIFF
--- a/packages/core/src/NodeView.ts
+++ b/packages/core/src/NodeView.ts
@@ -173,7 +173,7 @@ export class NodeView<
     }
 
     const isDragEvent = event.type.startsWith('drag')
-    const isDragOverEnterEvent = event.type === 'dragenter' || event.type === 'dragover'
+    const isDragOverEnterEvent = event.type === 'dragover' || event.type === 'dragenter'
     const isDropEvent = event.type === 'drop'
     const isInput = ['INPUT', 'BUTTON', 'SELECT', 'TEXTAREA'].includes(target.tagName) || target.isContentEditable
 


packages/core/src/NodeView.ts
```diff
@@ -237,7 +238,15 @@ export class NodeView<
     }
 
     // these events are handled by prosemirror
-    if (isDragging || isDropEvent || isCopyEvent || isPasteEvent || isCutEvent || (isClickEvent && isSelectable)) {
+    if (
+      isDragging ||
+      isDragOverEnterEvent ||
```
bdbch, line 243: Can we add a regression test for this path before merging? This fix is currently verified manually, but `stopEvent` is subtle enough that a future refactor could easily reintroduce the drag cursor regression without anyone noticing. A focused test around `dragover`/`dragenter` on NodeView DOM outside `contentDOM` would make the intent here much safer to maintain.

mweidner037: Would such a test go in tests/cypress? I'm not sure that we can test this in vitest given its reliance on the browser DOM.

bdbch: I would say this could be an E2E Cypress test in that case.

mweidner037: I spent some time on a Cyprus test for the GuideNodeViews/DragHandle demo, but I could not get it to fail before this PR and pass afterwards. I don't think it is worth too much effort given how small this change is.

PR #7670, opened 2026-03-29
fix(react): prevent crash when ReactNodeViewRenderer returns early

### LINKED ISSUE #7641: `ReactNodeViewRenderer` crashes when nodes exist in initial document and renderHTML returns text
### Affected Packages

core, react

### Version(s)

latest

### Bug Description

  When a custom atom node meets these conditions
- `atom: true`
- Uses `ReactNodeViewRenderer` for its node view
- Has a `renderHTML` that returns a string (text node, not an element)
- Exists in the initial document at parse time (via `parseHTML` rules or content)

ProseMirror crashes with this error

```
TypeError: dom.hasAttribute is not a function during EditorView construction.
```

[Codesandbox Repro](https://codesandbox.io/p/sandbox/interesting-ully-826d7l?file=%2Fpackage.json%3A9%2C6-9%2C19)

It's not clear to me if this is a TipTap or a ProseMirror issue, but I was able to workaround it. The error fires when the component mounts, and when it unmounts.

-  **Mount** - There's an empty object getting returned when the editor is unavailable, which happens here: https://github.com/ueberdosis/tiptap/blob/18f9afeecf9d19afb60efcee35de93df4cb97afb/packages/react/src/ReactNodeViewRenderer.tsx#L382-L382
- **Unmount** - When it unmounts, again there's no view for it to use: https://github.com/ueberdosis/tiptap/blob/18f9afeecf9d19afb60efcee35de93df4cb97afb/packages/react/src/EditorContent.tsx#L168-L170

Looking at it from the ProseMirror side, maybe this is an issue with `NodeViewDesc.create` not handling the case where `dom` could be a text node.

Here's how I worked around the errors for now.

### Before

```ts
addNodeView() {
  return ReactNodeViewRenderer(TemplateTagNodeView);
}
```

### After
```ts
// Wrap `ReactNodeViewRenderer` to fix TypeError on mount
addNodeView() {
    const reactRenderer = ReactNodeViewRenderer(TemplateTagNodeView);
    return (props, ...rest) => {
      const spec = reactRenderer(props, ...rest);
      if (!spec.dom) {
        const placeholder = document.createElement('span');
        return { dom: placeholder };
      }
      return spec;
    };
  },

addProseMirrorPlugins() {
  return [
  // Keep a placeholder element to fix TypeError on unmount
    new Plugin({
      props: {
        nodeViews: {
          [this.name]: (node: ProseMirrorNode) => {
            const dom = document.createElement('span');
            dom.textContent = node.attrs.value;
            return { dom };
          },
        },
      },
    }),
  ];
},
```

### Browser Used

Chrome

### Code Example URL

https://codesandbox.io/p/sandbox/interesting-ully-826d7l?file=%2Fpackage.json%3A9%2C6-9%2C19

### Expected Behavior

No type error should be thrown.

_No response_

### Dependency Updates

cyphercodes:
## Problem

When a custom atom node uses ReactNodeViewRenderer and exists in the initial document at parse time, the renderer could return an empty object if `contentComponent` wasn't available yet. This caused ProseMirror to crash with:

```
TypeError: dom.hasAttribute is not a function
```

This happens because ProseMirror expects a valid NodeView with a `dom` property, but received an empty object.

## Solution

Return a minimal valid NodeView with a placeholder DOM element instead of an empty object when `contentComponent` is not available. The placeholder NodeView implements all required methods with safe defaults.

Fixes #7641


package.json
```diff
@@ -81,7 +81,7 @@
     "ts-loader": "9.3.1",
     "tsup": "^8.5.1",
     "turbo": "2.3.3",
-    "typescript": "^5.7.3",
```
bdbch, line 84: I don't mind updating this, but I think we should remove it from this PR and keep it clean and focused on the specific fix, not generic dependency upgrades

bdbch (dismissed):

cyphercodes:
Resolved the merge conflict with current `main` and kept the PR focused on the React node view fix.

Verification:
- `pnpm install --frozen-lockfile`
- `pnpm exec vitest run packages/react/src/ReactNodeViewRenderer.spec.ts`
- `pnpm exec oxlint packages/react/src/ReactNodeViewRenderer.tsx packages/react/src/ReactNodeViewRenderer.spec.ts`
- `pnpm exec oxfmt --check packages/react/src/ReactNodeViewRenderer.tsx packages/react/src/ReactNodeViewRenderer.spec.ts`
- `(cd packages/react && pnpm build)`
- `git diff --check origin/main...HEAD`

cyphercodes:
Added focused fallback lifecycle coverage for the React node view placeholder case.

Verification:
- `pnpm exec vitest run packages/react/src/ReactNodeViewRenderer.spec.ts`
- `pnpm exec oxlint packages/react/src/ReactNodeViewRenderer.tsx packages/react/src/ReactNodeViewRenderer.spec.ts`
- `pnpm exec oxfmt --check packages/react/src/ReactNodeViewRenderer.tsx packages/react/src/ReactNodeViewRenderer.spec.ts`
- `(cd packages/react && pnpm build)`
- `git diff --check origin/main...HEAD`
- `git diff --check`

Commit: 07c0e55a92f278f66860fd212d410afc3eb48274

.changeset/fix-react-node-view-renderer.md
```diff
@@ -0,0 +1,12 @@
...
+---
+
+Fix ReactNodeViewRenderer crash when contentComponent is not available
+
+When a custom atom node uses ReactNodeViewRenderer and exists in the initial document,
+the renderer could return an empty object if contentComponent wasn't available yet.
+This caused ProseMirror to crash with "TypeError: dom.hasAttribute is not a function"
+because it expected a valid NodeView with a dom property.
+
+The fix returns a minimal valid NodeView with a placeholder DOM element instead of an empty object.
```
bdbch:
```suggestion
Fix ReactNodeViewRenderer crash when contentComponent is not available
```

bdbch (approved):

PR #7683, opened 2026-04-01
Fix/7562 (core): toggle list correctly when document selection is AllSelection

### LINKED ISSUE #7562: Unable to toggle list item when all content is selected (Ctrl + A)
### Affected Packages

ListItem

### Version(s)

3.20

### Bug Description

The issue is reproducible in the official example:
https://tiptap.dev/docs/editor/extensions/nodes/list-item

Steps to reproduce:
- Clear all editor content.
- Create a list item with some text.
- Press Ctrl + A to select all content.
- Try to toggle the list item.

Actual result:
It is impossible to toggle the list item when all content is selected.

### Browser Used

Chrome

### Code Example URL

_No response_

### Expected Behavior

The content should no longer be a list after toggling (i.e., it should be converted back to a normal paragraph).

_No response_

### Dependency Updates

okorie2:
## Changes Overview

Fix an issue where toggling lists fails when the entire document is selected (Ctrl/Cmd + A). When ProseMirror creates an AllSelection, the selection spans the entire document (0..doc.content.size). In this scenario, findParentNode cannot detect the surrounding list because the selection sits at the document root, causing toggleList to incorrectly fall back to the wrap branch instead of lifting or converting the existing list.

This change detects the special case where the document consists of a single top-level list and treats it as the active list context.

## Implementation Approach

The fix introduces handling for AllSelection when the document contains a single top-level list node.

When this condition is detected:

- The list is treated as the current list context even though findParentNode cannot detect it.
- The selection is normalized to an inner TextSelection within the list before executing liftListItem, since the underlying ProseMirror command expects the selection to be inside a list item.

This allows the existing list toggling logic to work without modifying ProseMirror's internal commands.

## Testing Done

Added regression tests covering the following scenarios when the entire document is selected:

- toggling off an ordered list
- toggling off a bullet list
- converting bullet → ordered list

## Verification Steps

Reviewers can verify the behavior using the official example:

- Clear the editor content.
- Create a list item with some text.
- Press Ctrl/Cmd + A to select the entire document.
- Toggle the same list type → the list should be removed.
- Toggle a different list type → the list should convert correctly.

This behavior previously failed when the selection was AllSelection.

## Additional Notes

Demo of fix

https://github.com/user-attachments/assets/2c29759c-46d5-4466-8857-b96322e5625f

## Checklist

## Related Issues

Closes #7562

### DIFF
--- a/packages/core/src/commands/toggleList.ts
+++ b/packages/core/src/commands/toggleList.ts
@@ -167,11 +167,7 @@ export const toggleList: RawCommands['toggleList'] =
       }
 
       // change list type
-      if (
-        isList(currentList.node.type.name, extensions) &&
-        listType.validContent(currentList.node.content) &&
-        dispatch
-      ) {
+      if (isList(currentList.node.type.name, extensions) && listType.validContent(currentList.node.content)) {
         return chain()
           .command(() => {
             tr.setNodeMarkup(currentList.pos, listType)
--- a/packages/extension-list/__tests__/taskItem.spec.ts
+++ b/packages/extension-list/__tests__/taskItem.spec.ts
@@ -489,4 +489,61 @@ describe('TaskItem', () => {
 
     expect(editor.can().toggleBulletList()).toBe(true)
   })
+  it('reports ordered list toggle as available when the whole bullet list document is selected', () => {
+    editor = new Editor({
+      extensions: [Document, Paragraph, Text, ListItem, OrderedList, BulletList],
+      content: {
+        type: 'doc',
+        content: [
+          {
+            type: 'bulletList',
+            content: [
+              {
+                type: 'listItem',
+                content: [
+                  {
+                    type: 'paragraph',
+                    content: [{ type: 'text', text: 'Test item' }],
+                  },
+                ],
+              },
+            ],
+          },
+        ],
+      },
+    })
+
+    editor.commands.selectAll()
+
+    expect(editor.can().toggleOrderedList()).toBe(true)
+  })
+
+  it('reports bullet list toggle as available when the whole ordered list document is selected', () => {
+    editor = new Editor({
+      extensions: [Document, Paragraph, Text, ListItem, OrderedList, BulletList],
+      content: {
+        type: 'doc',
+        content: [
+          {
+            type: 'orderedList',
+            content: [
+              {
+                type: 'listItem',
+                content: [
+                  {
+                    type: 'paragraph',
+                    content: [{ type: 'text', text: 'Test item' }],
+                  },
+                ],
+              },
+            ],
+          },
+        ],
+      },
+    })
+
+    editor.commands.selectAll()
+
+    expect(editor.can().toggleBulletList()).toBe(true)
+  })
 })


packages/core/src/commands/toggleList.ts
```diff
@@ -96,17 +112,69 @@ export const toggleList: RawCommands['toggleList'] =
...
+    const currentList = parentList ?? allSelectionList
+
+    const isInsideExistingList = !!parentList && range.depth >= 1 && range.depth - parentList.depth <= 1
+
+    const hasWholeDocSelectedList = !!allSelectionList
+    if ((isInsideExistingList || hasWholeDocSelectedList) && currentList) {
       // remove list
-      if (parentList.node.type === listType) {
+      if (currentList.node.type === listType) {
+        if (isAllSelection && hasWholeDocSelectedList && dispatch) {
```
bdbch: This new branch only runs when `dispatch` is available, so the fix does not carry over to the dry-run path used by `editor.can()`. In `can()` calls, this still falls through to `commands.liftListItem(itemType)` with the original root-level `AllSelection`, which means `can().toggleBulletList()` / `can().toggleOrderedList()` can continue to report `false` for the exact case this PR is fixing. Please normalize the selection for the `can()` path as well and add a regression test for that behavior.

okorie2: Updated

okorie2:
> Thanks for working on this. The runtime fix looks directionally right, but the new AllSelection path only normalizes the selection when dispatch is available. Please also handle the dry-run path used by editor.can() so toggleList reports the command as available when the whole document is a single top-level list. Right now integrations that drive toolbar state from can().toggleBulletList() / can().toggleOrderedList() can still show the action as disabled even though execution succeeds.

Thanks for the feedback. I have made updates and included additional tests as well.

packages/core/src/commands/toggleList.ts
```diff
@@ -96,17 +112,69 @@ export const toggleList: RawCommands['toggleList'] =
...
+            .liftListItem(itemType)
+            .run()
+        }
+
         return commands.liftListItem(itemType)
       }
 
       // change list type
-      if (isList(parentList.node.type.name, extensions) && listType.validContent(parentList.node.content) && dispatch) {
+      if (
```
bdbch: This still leaves a `.can()` gap for list-type conversion. The same-list removal path now normalizes `AllSelection` without requiring `dispatch`, but the conversion branch is still gated on `dispatch` here. That means selecting a whole-document bullet list and checking `editor.can().toggleOrderedList()` (or the inverse) can still report `false` even though execution succeeds. Please extend the dry-run handling to the conversion case as well and add a regression test for bullet -> ordered and ordered -> bullet availability under whole-document selection.

okorie2: > This still leaves a `.can()` gap for list-type conversion. The same-list removal path now normalizes `AllSelection` without requiring `dispatch`, but the conversion branch is still gated on `dispatch` here. That means selecting a whole-document bullet list and checking `editor.can().toggleOrderedList()` (or the inverse) can still report `false` even though execution succeeds. Please extend the dry-run handling to the conversion case as well and add a regression test for bullet -> ordered and ordered -> bullet availability under whole-document selection.

updated

okorie2:
> The `.can()` support is improved, but one part is still missing: changing one list type into another under whole-document `AllSelection`. The conversion branch in `toggleList` still requires `dispatch`, so command availability can remain incorrect for `editor.can().toggleOrderedList()` / `editor.can().toggleBulletList()` when the whole document is a single top-level list of the other type. Please handle that dry-run path too and cover it with a regression test.

Updated as well.

PR #7689, opened 2026-04-02
chore: do not publish src to npm

gameroman:
## Changes Overview

Stop publishing `src/` to npm

## Additional Notes

The `src/` directory should not be published to npm. It increases install size for no reason

## Checklist


bdbch:
I’m not sure yet if we actually want to remove the published source files. I get the package-size argument, but having `src/` available in the package is also pretty useful for debugging, inspecting internals, and even LLM/editor workflows that look directly in `node_modules`. This probably wouldn’t break most users, but it does remove some real ergonomics for advanced ones. Do we have a strong reason to optimize this away, or evidence that the size win is worth that tradeoff?

gameroman:
> having `src/` available in the package is also pretty useful for debugging

Source maps, i.e. `.js.map` files that are already there, provide everything needed for debugging

> inspecting internals

If someone actually needs to look at the source files, it is better to look at github and docs

> This probably wouldn’t break most users, but it does remove some real ergonomics for advanced ones

Most users don't need source files in their dependencies and it is rare that someone looks at `node_modules` instead of github or docs

> Do we have a strong reason to optimize this away, or evidence that the size win is worth that tradeoff?

For [`@tiptap/core`](https://npmx.dev/package-code/@tiptap/core/v/3.22.0) for example, this would decrease the install size by 25-30%

bdbch:
I get where you’re coming from, but I still think the tradeoff is worth calling out clearly.

1. The saved size is not noticeable in any bundled app, because `src/` is never what gets consumed there. This is only an npm package/install footprint win.
2. The main benefit of shipping `src/` is that you can open the exact installed version and read the real code immediately, without going through GitHub, finding the matching tag/version, and jumping around there.
3. I don’t think we know yet how meaningful this win is for users in practice. For `@tiptap/core`, the current unpacked package is about 2.33 MB, and `src/` is about 426 KB of that. So it is a real reduction, but it is not a bundle/runtime reduction, and sourcemaps are already a much larger part of the package than `src/` is.

So I’m not against it outright, I just think we should frame it accurately: smaller installed package size, but less inspectability in `node_modules`.

I will discuss this internally on what we want to achieve here, I don't have a strong opinion because it comes down to what our end users prefer / like. (Smaller installed package size vs. local inspectability)

gameroman:
> smaller installed package size, but less inspectability in `node_modules`

Yes, that is correct 👍

PR #7690, opened 2026-04-02
fix(markdown): preserve adjacent marks with different attrs

### LINKED ISSUE #7682: Adjacent marks of same type with different attributes are merged in output of getMarkdown
### Affected Packages

markdown

### Version(s)

3.20.4

### Bug Description

If 2 or more marks of the same type but with different attributes are adjacent to each other, they are merged into one.

Example:
The following HTML 
```html
<p>
<a target="_blank" href="https://example.com">example</a><a target="_blank" href="https://github.com">github</a><a target="_blank" href="https://tiptap.dev">tiptap</a>.
</p>
```

Is converted to the following markdown when calling `editor.getMarkdown()`:
```md
[examplegithubtiptap](https://tiptap.dev).
```

This is also the case for custom marks.

### Browser Used

Firefox

### Code Example URL

https://codesandbox.io/p/sandbox/jovial-lalande-r3539x

### Expected Behavior

I expect that only marks with matching attributes are merged.
The markdown should thus look more like this:

```md
[example](https://example.com)[github](https://github.com)[tiptap](https://tiptap.dev).
```

_No response_

### Dependency Updates

bdbch:
## Changes Overview

Fix markdown serialization so adjacent marks of the same type are only merged when their attributes also match.

## Implementation Approach

Updated markdown mark-boundary handling to compare both mark type and mark attributes when deciding whether a mark continues into the next text node. Added a regression test for adjacent link marks with different `href` values.

## Testing Done

- `pnpm vitest run packages/markdown/__tests__/manager.spec.ts`
- `pnpm vitest run packages/markdown/__tests__`
- `pnpm run test:unit`

## Verification Steps

1. Create a document with adjacent link marks that use different URLs.
2. Call `editor.getMarkdown()`.
3. Confirm the output keeps them as separate markdown links instead of merging them into one link targeting the last URL.

## Additional Notes

A changeset is included for `@tiptap/markdown`.

## Checklist

## Related Issues

Closes #7682

### DIFF
--- a/packages/markdown/src/MarkdownManager.ts
+++ b/packages/markdown/src/MarkdownManager.ts
@@ -16,6 +16,7 @@ import {
   flattenExtensions,
   generateJSON,
   getExtensionField,
+  objectIncludes,
 } from '@tiptap/core'
 import { type Lexer, type Token, type TokenizerExtension, type TokenizerThis, marked } from 'marked'
 
@@ -1124,7 +1125,13 @@ export class MarkdownManager {
             ...activeMarksClosingHere, // outer (were active before) — close last
           ]
         } else {
-          marksToCloseAtEnd = findMarksToCloseAtEnd(activeMarks, currentMarks, nextNode, this.markSetsEqual.bind(this))
+          marksToCloseAtEnd = findMarksToCloseAtEnd(
+            activeMarks,
+            currentMarks,
+            nextNode,
+            this.markSetsEqual.bind(this),
+            this.marksEqual.bind(this),
+          )
         }
 
         // Extract trailing whitespace before closing marks to prevent invalid markdown like "**text **"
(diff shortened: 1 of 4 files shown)


packages/markdown/src/MarkdownManager.ts
```diff
@@ -1283,6 +1290,24 @@ export class MarkdownManager {
...
+
+    const attrs1 = mark1?.attrs ?? {}
+    const attrs2 = mark2?.attrs ?? {}
+
+    return (
+      Object.keys(attrs1).length === Object.keys(attrs2).length &&
+      objectIncludes(attrs1, attrs2) &&
+      objectIncludes(attrs2, attrs1)
+    )
+  }
```
arnaugomez, line 1309: Isn't there a ProseMirror built-in method to check if two marks and mark attributes are equal?

bdbch: Checked this and `mark.eq()` isn’t applicable here.

In `MarkdownManager` we’re comparing serialized JSON marks from `JSONContent` (`{ type, attrs }`), not ProseMirror `Mark` instances, so there’s no reliable `eq()` method to call at this layer. I tightened the types to reflect that by replacing the local `any` mark params/maps with `NonNullable<JSONContent['marks']>[number]` instead.

PR #7692, opened 2026-04-03
fix(core): markPasteRule incorrectly removes storedMarks when mark is at end of pasted text (since v3.3.0, #6862)

Nantris:
## Summary

When pasting content where a mark extends to the end of the text (e.g. `test **bold**`), `markPasteRule` unconditionally calls `tr.removeStoredMark(config.type)`, which sets `storedMarks` to an empty array `[]`. This overrides the mark's `inclusive: true` behavior, causing the cursor to land **outside** the mark instead of inside it.

**Expected:** Pasting `test **bold**` → cursor is inside the bold mark: `test <strong>bold|</strong>`
**Actual:** Pasting `test **bold**` → cursor is outside the bold mark: `test <strong>bold</strong>|`

## Root cause

The `removeStoredMark` call in `markPasteRule` exists to prevent marks from bleeding into subsequently typed text when a mark appears mid-content (e.g. pasting `**bold** not bold` should not make "not bold" inherit the bold mark). This is correct behavior.

However, when the mark is at the **end** of the pasted content, there is no following content to protect. In this case, `removeStoredMark` sets `storedMarks = []` (explicitly "no marks"), which overrides ProseMirror's position-based mark resolution. With `storedMarks = null`, ProseMirror would check the cursor position, see it's at the end of an `inclusive: true` mark, and correctly include that mark.

This was surfaced by the per-extension paste rule plugin change in #6862 (v3.3.0), which altered how `appendTransaction` chains propagate `storedMarks` between independently-running paste rule plugins.

## Fix

Only call `removeStoredMark` when the regex match is **not** at the end of the text being processed. When it is at the end, skip the call and let the mark's `inclusive` setting determine cursor behavior naturally.

## Proposed test plan (not included)

- Paste `**bold**` → cursor should be inside bold mark, typing continues in bold
- Paste `**bold** not bold` → "not bold" should not be bold, typing after paste should not be bold
- Paste `test **bold**` in a list context → cursor inside bold mark
- Paste `__underline__` at end of text → same fix applies to underscore variant
- Paste `*italic*` at end of text → same fix applies via italic's markPasteRule

### DIFF
--- a/.changeset/grumpy-bugs-pull.md
+++ b/.changeset/grumpy-bugs-pull.md
@@ -1,0 +1,5 @@
+---
+"@tiptap/core": patch
+---
+
+Fix incorrect selection placement when pasting at the end of a marked text node, ensuring inclusive marks are respected


Nantris:
@bdbch this is causing some failing tests on our end after upgrading. Any comments or concerns would be appreciated and otherwise I would love to see this merged.

To clarify the issue, if you past `this is **bold**` then storedMarks gets cleared and the `bold` mark loses its intended inclusive value if so specified by the schema. I hope that's clear but if not happy to explain.

PR #7696, opened 2026-04-03
fix: preserve twitch src on HTML round-trip

### LINKED ISSUE #7563: [Bug] Twitch Extension nodes converted to HTML don't convert back properly if loaded into the editor again
### Affected Packages

@tiptap/extension-twitch

### Version(s)

3.20.0

### Bug Description

The issue occurs when the editor's content, after being read as HTML, is fed back into itself. For some reason, it strips the unique parts of the links out from the iframe elements, leaving generic player.twitch.tv URLs for Channel and Video links (which turn into "Invalid Twitch URL"). For Clips links, the unique clip ID is stripped, resulting in a "clip not found" entry. While it never turns into "Invalid Twitch URL", this is also wrong.

### Browser Used

Chrome

### Code Example URL

https://codesandbox.io/p/devbox/boring-star-xr4mhf?workspaceId=ws_7RxxXybA9sS3v4RjCuas2r

### Expected Behavior

I expect the HTML to be converted back into PM nodes in Tiptap. This is how the YouTube extension functions.

I assume this has to do with how Twitch's URLs have to be converted from the "normal" format to those weird player.twitch.tv links - the conversion probably isn't working right in the other direction. The YouTube extension doesn't need to do this, I think, so it doesn't have this issue.

In the example I provided, the issue is present. Just click on the big button at the top to trigger the issue repeatedly. I'm not sure how to get around codesandbox.io blocking Twitch content from loading, but that doesn't seem to matter since the underlying div/iframe structure is correct and the rest shouldn't matter (plus this code works fine locally and has the same bug).

### Dependency Updates

bdbch:
## Changes Overview

- Fixes a bug where Twitch embeds lost their original video, clip, or channel URL when content was converted to HTML and then reloaded into the editor.

## Implementation Approach

- Added a reverse-embed URL parser in `packages/extension-twitch/src/utils.ts` and wired `parseHTML` attribute parsing for the Twitch node in `packages/extension-twitch/src/twitch.ts`.
- Added a round-trip regression test to `packages/extension-twitch/__tests__/twitch.spec.ts`.

## Testing Done

- Ran the Twitch unit tests locally:

```
pnpm exec vitest run packages/extension-twitch/__tests__/twitch.spec.ts
```

- Built the Twitch package:

```
cd packages/extension-twitch && pnpm build
```

## Verification Steps

- Run the Twitch unit test locally:

```
pnpm exec vitest run packages/extension-twitch/__tests__/twitch.spec.ts
```

- Optionally build the package and inspect dist:

```
cd packages/extension-twitch && pnpm build
```

## Additional Notes

- A changeset is included at `.changeset/green-stingrays-shop.md` with a short user-facing note. No public API changes were made.

## Checklist

- [x] My changes do not break the library (local unit test passed for the Twitch spec).
- [x] I have followed the project guidelines (run `pnpm lint` if needed).

## Related Issues

- closes #7563

### DIFF
--- a/packages/extension-twitch/src/twitch.ts
+++ b/packages/extension-twitch/src/twitch.ts
@@ -7,6 +7,16 @@ import {
   TWITCH_REGEX_GLOBAL,
 } from './utils.js'
 
+const getParsedDimension = (value: string | null) => {
+  if (!value) {
+    return null
+  }
+
+  const parsedValue = Number.parseInt(value, 10)
+
+  return Number.isNaN(parsedValue) ? null : parsedValue
+}
+
 const getParsedTwitchAttributes = (element: HTMLElement) => {
   const src = element.getAttribute('src')
 
@@ -26,7 +36,9 @@ const getParsedTwitchAttributes = (element: HTMLElement) => {
     }
   }
 
-  return null
+  return {
+    src,
+  }
 }
 
 export interface TwitchOptions {
@@ -185,19 +197,11 @@ export const Twitch = Node.create<TwitchOptions>({
       },
       width: {
         default: this.options.width,
-        parseHTML: element => {
-          const width = (element as HTMLElement).getAttribute('width')
-
-          return width ? Number.parseInt(width, 10) : null
-        },
+        parseHTML: element => getParsedDimension((element as HTMLElement).getAttribute('width')),
       },
       height: {
         default: this.options.height,
-        parseHTML: element => {
-          const height = (element as HTMLElement).getAttribute('height')
-
-          return height ? Number.parseInt(height, 10) : null
-        },
+        parseHTML: element => getParsedDimension((element as HTMLElement).getAttribute('height')),
       },
       autoplay: {
         default: this.options.autoplay,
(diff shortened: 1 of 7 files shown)


packages/extension-twitch/src/twitch.ts
```diff
@@ -154,21 +193,27 @@ export const Twitch = Node.create<TwitchOptions>({
     return {
       src: {
         default: null,
+        parseHTML: element => getParsedTwitchAttributes(element as HTMLElement)?.src,
```
arnaugomez: Is the casting (`as` operator) always valid here? Or could there be a situation where the element is not an HTML element?

arnaugomez (dismissed):

arnaugomez:
Should we do the same in the YouTube extension?

bdbch:
Yep, good call. I pushed a follow-up that does the same round-trip fix for YouTube as well, and updated the changeset too.

packages/extension-twitch/src/twitch.ts
bdbch: Yep, this cast was redundant here. `parseHTML` already gets an `HTMLElement`, so I removed the `as` usage in the follow-up commit.

arnaugomez (approved):

PR #7705, opened 2026-04-04
perf(extensions): avoid full-document scan for Placeholder when showOnlyCurrent is true

HMarzban:
### Summary

Optimizes `@tiptap/extensions/placeholder` so that when **`showOnlyCurrent` is `true`** (the default), placeholder decorations are built from the **selection anchor** only (**O(depth)**) instead of walking the **entire document** with `doc.descendants` on every update (**O(n)**).

When **`showOnlyCurrent` is `false`**, behavior is unchanged in intent: we still perform a **full-document scan** so placeholders can appear on **every** empty textblock.

### Motivation

Large documents spend unnecessary work in the placeholder plugin on each transaction because the previous implementation always traversed the full node tree, even though the default configuration only needs the **current** empty block at the caret.

### What changed

- **`showOnlyCurrent: true`**: resolve `anchor`, walk relevant depths (`includeChildren` still respected via `maxDepth`), only **textblock** nodes, same emptiness rules as before (`isNodeEmpty`).
- **`showOnlyCurrent: false`**: keep a **`doc.descendants`** pass for all empty textblocks (unavoidable for “show everywhere” semantics).
- **Unchanged API**: `PlaceholderOptions`, `dataAttribute` / `preparePlaceholderAttribute`, `PluginKey('placeholder')`, and existing defaults.

### Performance

- **Typical editing (default options)**: cost scales with **nesting depth**, not document size.
- **“All empty blocks” mode**: still **O(n)** — by design.

### Testing / QA

Suggested manual checks:

- [ ] Empty paragraph at caret — placeholder shows as before.
- [ ] Nested blocks (e.g. list → item → paragraph) — placeholder at caret.
- [ ] `showOnlyCurrent: false` — placeholders on multiple empty blocks across the doc.
- [ ] Read-only / `showOnlyWhenEditable` — matches prior behavior.
- [ ] Custom `dataAttribute` — attribute still applied correctly.

### Risk / compatibility

- **No intentional breaking changes**; public options and decoration shape are preserved.
- Please run the **extensions** package tests / CI in this repo.

---

**Related file:** `packages/extensions/src/placeholder/placeholder.ts`

### DIFF
--- a/packages/extensions/src/placeholder/placeholder.ts
+++ b/packages/extensions/src/placeholder/placeholder.ts
@@ -57,9 +57,7 @@ export interface PlaceholderOptions {
    * You can use a function to return a dynamic placeholder or a string.
    * @default 'Write something …'
    */
-  placeholder:
-    | ((props: { editor: Editor; node: ProseMirrorNode; pos: number; hasAnchor: boolean }) => string)
-    | string
+  placeholder: ((props: { editor: Editor; node: ProseMirrorNode; pos: number; hasAnchor: boolean }) => string) | string
 
   /**
    * **Checks if the placeholder should be only shown when the editor is editable.**
@@ -144,9 +142,7 @@ function collectPlaceholdersFullScan(
       return options.includeChildren
     }
 
-    decorations.push(
-      createPlaceholderDecoration(editor, options, node, pos, hasAnchor, dataAttributeKey),
-    )
+    decorations.push(createPlaceholderDecoration(editor, options, node, pos, hasAnchor, dataAttributeKey))
 
     return options.includeChildren
   })
@@ -186,9 +182,7 @@ function collectPlaceholdersShowOnlyCurrent(
     const isEmpty = !node.isLeaf && isNodeEmpty(node)
 
     if (hasAnchor && isEmpty) {
-      decorations.push(
-        createPlaceholderDecoration(editor, options, node, pos, hasAnchor, dataAttributeKey),
-      )
+      decorations.push(createPlaceholderDecoration(editor, options, node, pos, hasAnchor, dataAttributeKey))
     }
   }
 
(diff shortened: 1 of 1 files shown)


packages/extensions/src/placeholder/placeholder.ts
```diff
@@ -89,6 +87,108 @@ export interface PlaceholderOptions {
...
+
+    if (!node.type.isTextblock) {
+      continue
+    }
+
+    const hasAnchor = nodeContainsAnchor(anchor, pos, node)
+    const isEmpty = !node.isLeaf && isNodeEmpty(node)
+
+    if (hasAnchor && isEmpty) {
+      decorations.push(createPlaceholderDecoration(editor, options, node, pos, hasAnchor, dataAttributeKey))
```
arnaugomez, line 185: Should you return early here? If I understand it correctly, the function only returns one decoration for the current node.

arnaugomez (changes requested):

packages/extensions/src/placeholder/placeholder.ts
HMarzban: We shouldn’t return early inside the loop.

The loop is over depths (1 … maxDepth). In principle more than one empty textblock on the resolved path could match (unusual schemas / nested textblocks). Stopping after the first push could drop a valid decoration, so we finish the loop.

On the API side: we collect a Decoration[] (length 0 or more) and pass it to DecorationSet.create(doc, decorations), which is what ProseMirror expects — an array of Decorations; an empty array is valid when there’s nothing to show.

Separately, the props.decorations hook may return null when the plugin is inactive (!active); that’s valid for “no decorations” and is separate from the Decoration[] passed into DecorationSet.create.

References: [DecorationSet](https://prosemirror.net/docs/ref/#view.DecorationSet%5Ecreate), [EditorProps.decorations](https://prosemirror.net/docs/ref/#view.EditorProps.decorations).

Issue #7706, opened 2026-04-05
remove unnecessary `xmlns="http://www.w3.org/1999/xhtml"` from generateHTML output


kcoet:
### Affected Packages

core, react

### Version(s)

3.22.2

### Description of the Bug

The `generateHTML` function generates html tags with the attribute `xmlns="http://www.w3.org/1999/xhtml"` I think this is not really necessary.

This adds to the HTML size, I think it will impact SEO and page load times.

### Browser Used

Chrome

### Expected Behavior

The HTML tag generated by `generateHTML` must not have the attribute `xmlns="http://www.w3.org/1999/xhtml"`

### Dependency Updates

- [x] Yes, I've updated all my dependencies.

PR #7709, opened 2026-04-05
Remove unnecessary xmlns attribute from generateHTML output

arnaugomez:
## Bug

`generateHTML` (both browser and server versions in `@tiptap/html`) produces HTML with `xmlns="http://www.w3.org/1999/xhtml"` on every top-level element. For example:

```html
<p xmlns="http://www.w3.org/1999/xhtml">Hello</p>
```

This happens because the serialization uses `XMLSerializer.serializeToString()`, which is designed for XML/XHTML output and always includes namespace declarations. The attribute is unnecessary for HTML documents, adds to the HTML size, and can negatively impact SEO and page load times.

## Fix

Replaced `XMLSerializer.serializeToString()` with a wrapper `div` + `innerHTML` approach in both `getHTMLFromFragment` implementations (browser and server). This is the same technique already used in the `options.document` code path and in the `@tiptap/core` version of the function.

The output now produces clean HTML without namespace attributes:

```html
<p>Hello</p>
```

## Changed files

- `packages/html/src/getHTMLFromFragment.ts` — browser version
- `packages/html/src/server/getHTMLFromFragment.ts` — server version
- `packages/html/__tests__/generateHTML.spec.ts` — updated test expectations
- `packages/html/__tests__/server-with-jsdom.spec.ts` — updated test expectations

Fixes https://github.com/ueberdosis/tiptap/issues/7706

### DIFF
--- a/demos/src/GuideContent/GenerateHTML/React/index.spec.js
+++ b/demos/src/GuideContent/GenerateHTML/React/index.spec.js
@@ -6,6 +6,6 @@ context('/src/GuideContent/GenerateHTML/React/', () => {
   it('should render the content as an HTML string', () => {
     cy.get('pre code').should('exist')
 
-    cy.get('pre code').should('contain', '<p xmlns="http://www.w3.org/1999/xhtml">Example <strong>Text</strong></p>')
+    cy.get('pre code').should('contain', '<p>Example <strong>Text</strong></p>')
   })
 })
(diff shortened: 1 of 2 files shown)


arnaugomez:
The failing test is a flaky test that already existed before this change:

```
  /src/Examples/Images/React/
    ✓ finds image elements inside editor (232ms)
    1) allows removing images
    ✓ allows images to be added via URL (2164ms)


  2 passing (33s)
  1 failing

  1) /src/Examples/Images/React/
       allows removing images:
     AssertionError: Timed out retrying after 30000ms: Expected to find element: `.tiptap img.ProseMirror-selectednode`, but never found it.
      at Context.eval (webpack:///../demos/src/Examples/Images/React/index.spec.js:14:51)
```

[This PR](https://github.com/ueberdosis/tiptap/pull/7710) fixes this bug

PR #7725, opened 2026-04-08
refactor: remove redundant id attribute from table of contents node m…

### LINKED ISSUE #7721: Table of Content plugin overriding `id` attr on nodes
### Affected Packages

core, extension-table-of-contents

### Version(s)

latest

### Bug Description

I have a custom unique id plugin that sets `id` attrs on nodes but when I use table of contents plugin with that it overrides the custom unique id.

### Browser Used

Chrome

### Code Example URL

_No response_

### Expected Behavior

It should not override `id` attr. There is not need to do so in table of contents

https://github.com/ueberdosis/tiptap/blob/ac994c5daab02f603d61eb759a90c2a1c46dfc03/packages/extension-table-of-contents/src/plugin.ts#L49-L53

### Dependency Updates

emrullahtastan:
…arkup

## Changes Overview

This PR fixes an issue where the TableOfContents extension was incorrectly overriding the id attribute on nodes (headings) whenever a data-toc-id was assigned. This caused conflicts with other extensions (like unique-id) or manual ID management systems.

## Implementation Approach

The `TableOfContents` extension was explicitly setting the id attribute to match the `data-toc-id`during both initial creation (`onCreate`) and document updates (`appendTransaction`).

I have:

- Removed the `id` assignment from the `setNodeMarkup` calls in `packages/extension-table-of-contents/src/plugin.ts`.
- Removed the `id` assignment from the `setNodeMarkup` calls in `packages/extension-table-of-contents/src/tableOfContents.ts`.

By removing these assignments, the extension now only manages the `data-toc-id` attribute, leaving the standard `id` attribute to be managed by other extensions or the user.

## Testing Done

- Verified using the local TOC demo.
- Confirmed that headings no longer receive a redundant id attribute automatically upon editor load or update.
- Verified that the Table of Contents sidebar still functions correctly, as it relies on data-toc-id for state tracking and link generation.
- Confirmed that there are no console errors or regressions in TOC indexing.

## Verification Steps

1. Open the Table of Contents demo page.
2. Inspect any heading in the DOM; verify it has a `data-toc-id` but **no** id attribute (unless provided by another extension).
3. Type in the editor to trigger a transaction; verify that no id is added to the headings.
4. If testing with a unique ID extension, verify that its IDs are now preserved and NOT overwritten by TOC.

## Additional Notes

This makes the TableOfContents extension a "good citizen" by not interfering with global attributes it doesn't strictly need to own.

## Checklist

- [x] I have added tests where applicable. (Manual verification performed).

## Related Issues

Fixes #7721

### DIFF
--- a/.changeset/chilly-eels-grin.md
+++ b/.changeset/chilly-eels-grin.md
@@ -1,0 +1,5 @@
+---
+'@tiptap/extension-table-of-contents': patch
+---
+
+fix(extension-table-of-contents): stop overriding the id attribute on nodes
--- a/packages/extension-table-of-contents/src/plugin.ts
+++ b/packages/extension-table-of-contents/src/plugin.ts
@@ -49,7 +49,6 @@ export const TableOfContentsPlugin = ({
             tr.setNodeMarkup(pos, undefined, {
               ...node.attrs,
               'data-toc-id': id,
-              id,
             })
 
             modified = true
--- a/packages/extension-table-of-contents/src/tableOfContents.ts
+++ b/packages/extension-table-of-contents/src/tableOfContents.ts
@@ -296,7 +296,6 @@ export const TableOfContents = Extension.create<TableOfContentsOptions, TableOfC
         tr.setNodeMarkup(pos, undefined, {
           ...node.attrs,
           'data-toc-id': id,
-          id,
         })
       }
 


arnaugomez (changes requested):
The problem goes beyond the fact that the table of contents overrides the ID. It also defines an `id` attribute in the extension file (`packages/extension-table-of-contents/src/tableOfContents.ts`). So I wonder if that attribute is necessary at all. If we do not set the value of the `id` property, it does not make sense to define it in the extension file (`packages/extension-table-of-contents/src/tableOfContents.ts`)

PR #7742, opened 2026-04-14
fix(collaboration): onContentError not called when enableContentCheck is true

yuichi0301:
## Changes Overview

Fix `onContentError` not being called even when `enableContentCheck` is set to `true`.

## Implementation Approach

Changed `return false` to `return true` in the `filterTransaction` catch block. Since Yjs transactions cannot be cancelled, returning `false` caused a state mismatch between ProseMirror and Yjs, which prevented `onContentError` from functioning properly.

## Testing Done

Updated the expected value in the existing test (`filterInvalidContent.spec.ts`) and confirmed all tests pass.

## Verification Steps

1. Create a Collaboration editor with `enableContentCheck: true` and an `onContentError` handler
2. Send invalid content via Yjs
3. Confirm that `onContentError` is called

## Additional Notes

Regression introduced in commit 556b0a3, which moved content validation from `beforeTransaction` to `filterTransaction`.

## Checklist

## Related Issues

556b0a31239b2d51569b125034d47b5678af6561


yuichi0301:
Friendly ping on this 👋

Without this fix, `onContentError` is effectively a no-op when `enableContentCheck: true` — users can configure the callback but it never fires on invalid collaborative content, so there's no way to recover.

@dearlordylord — since you moved the validation into `filterTransaction` in #7600, your eyes would be especially valuable here. I tried to keep the original intent intact (rejecting invalid Yjs content) while making sure `onContentError` still fires. Let me know if I missed any nuance from the original design.

@bdbch — would appreciate a maintainer look when you have a moment. Copilot suggestions are addressed and tests + changeset are included.

Happy to rebase or iterate. Thanks!

bdbch (approved):
thanks @yuichi0301 - looks good to me

yuichi0301:
Thanks for the review and merge, @bdbch! I appreciate it.

Issue #7729, opened 2026-04-08
Feature Request: Allow returning undefined/false from parseMarkdown


bdbch:
### Discussed in https://github.com/ueberdosis/tiptap/discussions/7571

<div type='discussions-op-text'>

<sup>Originally posted by **n-siddarth** March  7, 2026</sup>
### Description

It would be nice if users had the option of returning `undefined` or `false` from `parseMarkdown` implementations indicating that the parsing was not successful and that the tokens should be skipped for this handler. This would allow multiple extensions to serialize to the same markdown string but parse differently based on certain conditions.

### Use Case

For example, let's say files and websites are treated differently in the editor. Both should serialize to standard markdown links. For example, files can be serialized to `[filename](/path/to/file)` while normal links can be represented as standard `[display](http://example-url.com)`. However, when it comes to parsing, the two extensions should be able to conditionally parse the tokens depending on the structure of the `href` token passed to the `parseMarkdown` function.

### Type

New Tiptap API</div>

okorie2:
I opened a PR that updates MarkdownParseResult to allow false | undefined and adds tests demonstrating handler fallthrough.

While investigating, I noticed that top-level parsing already supports multiple handlers via getHandlersForToken(...), but inline parsing still uses getHandlerForToken(...), which only selects the first handler.

```

  /** Get the first handler for a token type (for backwards compatibility). */
  private getHandlerForToken(type: string): MarkdownExtensionSpec | undefined {
    // First try the markdown token registry (for parsing)
    const markdownHandlers = this.getHandlersForToken(type)
    if (markdownHandlers.length > 0) {
      return markdownHandlers[0]
    }

    // Then try the node type registry (for rendering)
    const nodeTypeHandlers = this.getHandlersForNodeType(type)

    return nodeTypeHandlers.length > 0 ? nodeTypeHandlers[0] : undefined
  }

```

If desired, I can also adapt inline parsing to use the same fallthrough mechanism so that multiple handlers can participate there as well.

PR #7752, opened 2026-04-18
feat(markdown): enhance parseMarkdown handler to support fallthrough b…

okorie2:
## Changes Overview

This PR updates the MarkdownParseResult type to allow false and undefined as valid return values from parseMarkdown handlers.

Returning false or undefined signals that the handler declines the current token, allowing the parser to continue trying other registered handlers for the same token type.

Although the runtime parser already supports this behavior (via normalizeParseResult()), the TypeScript type definition did not previously allow these values.

## Implementation Approach

The runtime parser already treats falsy parse results as a declined token:

```
if (!result) {
  return null
}

```
Since false and undefined normalize to null, the parser continues trying other registered handlers.

This PR formalizes that existing behavior by:

- Updating the MarkdownParseResult type to include false | undefined.
- Updating the documentation comment explaining the return shape of parseMarkdown.

No runtime logic changes were required.

## Testing Done

Added tests verifying handler fallthrough behavior:

- A handler returning false allows the next registered handler to parse the token.
- A handler returning undefined allows the next registered handler to parse the token.
- Parsing stops at the first handler that returns a valid result.

## Verification Steps

Reviewers can verify the change by:

- Creating two extensions with the same markdownTokenName.
- Making the first handler return false or undefined.
- Confirming the second handler successfully parses the token.

## Additional Notes

While investigating this issue, I noticed that top-level parsing already supports multiple handlers via getHandlersForToken(...), but inline parsing still uses getHandlerForToken(...), which selects only the first registered handler.

I left a comment in the issue asking whether inline parsing should also support handler fallthrough to make behavior consistent across parsing paths.

Happy to extend the change if desired.

## Checklist

## Related Issues

#7729


packages/core/src/types.ts
```diff
@@ -939,8 +939,17 @@ export default MarkdownHelpers
...
+ * the current token. In that case the parser will continue trying other
+ * registered handlers for the same token type.
  */
-export type MarkdownParseResult = JSONContent | JSONContent[] | { mark: string; content: JSONContent[]; attrs?: any }
+export type MarkdownParseResult =
+  | JSONContent
+  | JSONContent[]
+  | { mark: string; content: JSONContent[]; attrs?: any }
+  | false
+  | undefined
```
arnaugomez, line 970:
```suggestion
export type MarkdownParseResult =
  | JSONContent
  | JSONContent[]
  | { mark: string; content: JSONContent[]; attrs?: any }
  | false
  | undefined
  | null
```

Perhaps `null` could be included in the type?

arnaugomez (changes requested):

packages/core/src/types.ts
okorie2: Sure, will include it!

arnaugomez (approved):
@bdbch you might give it a last review

CHANGELOG.md
```diff
@@ -320,9 +320,10 @@
 
 - 032f8f1: Fix dependencies installation after packages updates producing peer dependency resolution conflicts
 - 032f8f1: Treat non-indented continuation lines following an ordered list marker as part of the same list item.
-  
+
```
bdbch, line 2049: please revert those changes to the CHANGELOG.md

turbo.json
```diff
@@ -3,7 +3,14 @@
   "ui": "tui",
   "tasks": {
     "build": {
-      "inputs": ["src/**", "tsup.config.ts", "tsconfig*.json", "../../tsconfig.json", "../../tsconfig.build.json", "package.json"],
+      "inputs": [
```
bdbch, line 6: Please revert those changes, not related to your fix I think.

bdbch (changes requested):

okorie2:
Resolved conflicts locally but unable to commit because of the following error in code that was previously merged
`tiptap/packages/vue-3/__tests__/useEditor.spec.ts
  28:26  error  React Hook "useEditor" is called in function "setup" that is neither a React function component nor a custom React Hook function. React component names must start with an uppercase letter. React Hook names must start with the word "use"  react-hooks/rules-of-hooks`
  
  Please advise @bdbch

bdbch:
> Resolved conflicts locally but unable to commit because of the following error in code that was previously merged `tiptap/packages/vue-3/__tests__/useEditor.spec.ts 28:26 error React Hook "useEditor" is called in function "setup" that is neither a React function component nor a custom React Hook function. React component names must start with an uppercase letter. React Hook names must start with the word "use" react-hooks/rules-of-hooks`
> 
> Please advise @bdbch

When did you encounter this error, while committing?

okorie2:
> > Resolved conflicts locally but unable to commit because of the following error in code that was previously merged `tiptap/packages/vue-3/__tests__/useEditor.spec.ts 28:26 error React Hook "useEditor" is called in function "setup" that is neither a React function component nor a custom React Hook function. React component names must start with an uppercase letter. React Hook names must start with the word "use" react-hooks/rules-of-hooks`
> > Please advise @bdbch
> 
> When did you encounter this error, while committing?

Yes, while committing.

bdbch:
I think this seems to be a local problem. https://github.com/ueberdosis/tiptap/blob/main/packages/vue-3/src/useEditor.ts is a Vue hook so it should be fine. I also didn't see the error on my end.

demos/src/Marks/Link/React/index.jsx
```diff
@@ -5,83 +5,22 @@ import Document from '@tiptap/extension-document'
 import Link from '@tiptap/extension-link'
 import Paragraph from '@tiptap/extension-paragraph'
 import Text from '@tiptap/extension-text'
-import { EditorContent, useEditor, useEditorState } from '@tiptap/react'
+import { EditorContent, MarkViewContent, ReactMarkViewRenderer, useEditor, useEditorState } from '@tiptap/react'
 import React, { useCallback } from 'react'
 
 export default () => {
+  const CustomLink = Link.extend({
```
bdbch: Why do we have this Link here with a MarkView? If you need it for whatever reason - I think we should keep the configuration from line 20.

Otherwise this breaks our Link tests.

bdbch (changes requested):

CHANGELOG.md
```diff
@@ -4897,4 +4897,4 @@
 - build(deps): bump actions/cache from 3.2.6 to 3.3.1 by @dependabot in https://github.com/ueberdosis/tiptap/pull/3846
 - docs: prepare for stable release by @svenadlung in https://github.com/ueberdosis/tiptap/pull/3892
 - New Issue & Discussion Templates by @bdbch in https://github.com/ueberdosis/tiptap/pull/3907
-- feat(core): add editor to this context in schema functions by @bdbch in https://github.com/ueberdosis/tiptap/pull/3909
+- feat(core): add editor to this context in schema functions by @bdbch in https://github.com/ueberdosis/tiptap/pull/3909
```
bdbch: can you revert the whole CHANGELOG.md? It don't know what kind of changes happened here.

bdbch (changes requested):

bdbch:
After the CHANGELOG is cleaned this looks good for me to merge.

okorie2:
> After the CHANGELOG is cleaned this looks good for me to merge.

Just letting you now I made the requested change @bdbch

bdbch:
@coderabbitai full review

PR #7761, opened 2026-04-21
Default immediatelyRender to false in SSR environments

arnaugomez:
## Summary

When using `useEditor` in an SSR environment (e.g. Next.js), omitting the `immediatelyRender` option would throw an error in development and silently return `null` in production. This was a very common pitfall — especially when AI coding agents set up the editor without explicitly passing `immediatelyRender: false`, causing the entire page to crash.

## What changed

- `immediatelyRender` now defaults to `true`, but is automatically set to `false` when SSR is detected (server-side or Next.js environment)
- In development, a `console.warn` is logged instead of throwing an error
- The logic is simplified from multiple branches with different throw/return paths into a single, clear flow

## Why

The previous behavior was too strict — throwing in dev and silently degrading in prod made debugging harder and created a sharp edge for new users and automated tooling. A warning is sufficient to inform developers while keeping the app functional.

### DIFF
--- a/.changeset/react-ssr-immediately-render.md
+++ b/.changeset/react-ssr-immediately-render.md
@@ -1,0 +1,7 @@
+---
+'@tiptap/react': patch
+---
+
+Default `immediatelyRender` to `false` in SSR environments instead of throwing an error
+
+Previously, omitting `immediatelyRender` in an SSR environment (e.g. Next.js) would throw an error in development and silently return `null` in production. This was a common source of crashes, especially when AI-generated code set up the editor without explicitly passing `immediatelyRender: false`. The hook now defaults `immediatelyRender` to `true`, but automatically sets it to `false` when SSR is detected, logging a warning in development instead of throwing.
--- a/packages/react/src/useEditor.ts
+++ b/packages/react/src/useEditor.ts
@@ -93,38 +93,16 @@ class EditorInstanceManager {
   }
 
   private getInitialEditor() {
-    if (this.options.current.immediatelyRender === undefined) {
-      if (isSSR || isNext) {
-        if (isDev) {
-          /**
-           * Throw an error in development, to make sure the developer is aware that tiptap cannot be SSR'd
-           * and that they need to set `immediatelyRender` to `false` to avoid hydration mismatches.
-           */
-          throw new Error(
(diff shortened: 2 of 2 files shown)


.changeset/react-ssr-immediately-render.md
```diff
@@ -0,0 +1,7 @@
+---
+'@tiptap/react': patch
```
nperez0111, line 2: Definitely not a patch, is a change in runtime behavior no?

arnaugomez: I made it a patch because it does not break anything. The behaviour in production remains the same.

bdbch (approved):

.changeset/react-ssr-immediately-render.md
zack-maxima: This is breaking... 

I have a nextjs app that passes `immediatelyRender: true` because I don't use nextjs with server-side rendering. I don't want to handle editor being null for no reason, the component only renders on the client in my case.

The previous behaviour would only throw if `isSSR` or `isNext && immediatelyRender === undefined` which allowed me to pass `immediatelyRender: true`.

This new behaviour is forcing `immediatelyRender` to be set to false with no way of overriding because it's now checking for `isSSR || isNext`, so if you're running nextjs the `immediatelyRender` param is completely ignored. It's always going to be false with no way to set it to true.

I think the proper behaviour would be to automatically set `immediatelyRender: false` only if `isSSR` is true.

zack-maxima:
I opened a follow-up PR to fix this: ueberdosis/tiptap#7842. It restores the ability to pass an explicit `immediatelyRender: true` in client-side Next.js apps while keeping the safer default for the unset case.

.changeset/react-ssr-immediately-render.md
zack-maxima: PR for suggested fix: https://github.com/ueberdosis/tiptap/pull/7842

arnaugomez: Thanks for finding this issue. I've approved your PR

nperez0111: Ah, yea I meant to push back on this further, but didn't have the energy to write it up. Glad to see it fixed though!

I spent _a lot_ of time on those conditionals, so I knew it wasn't going to be a straightforward refactor 😂

Issue #2819, opened 2022-05-23
Nodeview component not recieving the 'selected' prop in React.


mattiastofte:
### What’s the bug you are facing?

I've used the documentation as described [here](https://tiptap.dev/guide/node-views/react#selected). The problem is that I get all the other props, just not the 'selected' prop which I've defined as a boolean value in typescript. Why is this? I would like this prop so I could switch how the component is rendered based on if you're editing or not.

Note: I'm using React with NextJS.

The code for the component that's added into my extension with addNodeView():
```typescript
import React from "react";
import {
  NodeViewWrapper,
  NodeViewContent,
  Editor,
} from "@tiptap/react";
import styles from "./LatexBlock.module.css";

var Latex = require("react-latex");

const LatexBlockComponent = ({
  node,
  editor,
  selected,
}: {
  node: Node;
  editor: Editor;
  selected: boolean;
}) => {
  return (
    <NodeViewWrapper className={styles.block}>
      <NodeViewContent className={styles.content} />
      <Latex>{node.textContent}</Latex>
    </NodeViewWrapper>
  );
};

export default LatexBlockComponent;
```

### Which browser was this experienced in? Are any special extensions installed?

This was experienced in the Google Chrome browser. I've only used official extensions such as code, codeblock, etc.

### How can we reproduce the bug on our side?

index.ts
```typescript
import { LatexBlock } from './LatexBlock'

export * from './LatexBlock'

export default LatexBlock
```

LatexBlock.ts
```typescript
import { Node, mergeAttributes } from '@tiptap/core'
import { ReactNodeViewRenderer } from '@tiptap/react'
import LatexBlockComponent from './LatexBlockComponent'

declare module '@tiptap/core' {
  interface Commands<ReturnType> {
    LatexBlock: {
      /**
       * Set a code block
       */
      setLatexBlock: (attributes?: { language: string }) => ReturnType,
      /**
       * Toggle a code block
       */
      toggleLatexBlock: (attributes?: { language: string }) => ReturnType,
    }
  }
}

export const LatexBlock = Node.create({
  name: 'LatexBlock',

  group: 'block',

  content: 'inline*',

  parseHTML() {
    return [
      {
        tag: 'LatexBlock',
      },
    ]
  },

  renderHTML({ HTMLAttributes }) {
    return ['LatexBlock', mergeAttributes(HTMLAttributes), 0]
  },

  addNodeView() {
    return ReactNodeViewRenderer(LatexBlockComponent)
[...]

PR #7763, opened 2026-04-22
feat(core): add selectedOnTextSelection option to node views

### LINKED ISSUE #3040: Node view `selected` prop doesn't get updated when node is selected

I'm trying to use the node view `selected` prop (https://tiptap.dev/guide/node-views/vue#selected) but it doesn't get updated when the node is selected.

Firefox 103.0 on Ubuntu 22.04.

No special extensions, just `@tiptap/starter-kit` + a custom node view to show the issue.

1. Open the CodeSandbox provided below
2. Click anywhere on the bold text (which is the custom node)
3. Text between parenthesis should change to "selected" but it's not the case

https://codesandbox.io/s/tiptap-node-view-selected-prop-issue-yneccj

Node view `selected` prop value should be `true` if selection is at node position (as stated in the docs).

This is the first time I'm using this prop so I don't know if it already worked and if it did, when it stopped working.

It's probably the same issue than https://github.com/ueberdosis/tiptap/issues/2819.

### Are you sponsoring us?

- [X] Yes, I’m a sponsor. 💖

ozdemircibaris:
## Changes Overview

Adds a new `selectedOnTextSelection` option to `NodeViewRendererOptions`. When enabled, a node view's `selected` prop becomes `true` while a `TextSelection` is fully contained within the node's range — not only when a `NodeSelection` wraps the node. Default behavior is unchanged (opt-in).

## Implementation Approach

Extracted the selection-matching logic into a shared `isNodeViewSelected` helper in `@tiptap/core` and replaced the inline `handleSelectionUpdate` checks in the Vue 3, Vue 2, and React renderers so all three share the same source of truth.

## Testing Done

Added `packages/core/__tests__/isNodeViewSelected.spec.ts` covering six cases: NodeSelection, TextSelection engulfing the node, cursor inside with the option disabled and enabled, selection outside the node, and partial overlap across the node boundary. All 793 existing tests still pass.

## Verification Steps

1. Create a block node with `content: 'text*'` and `addNodeView: () => VueNodeViewRenderer(Component, { selectedOnTextSelection: true })`.
2. Place the cursor inside the node's text — the `selected` prop should become `true`.
3. Move the cursor to another node — `selected` should become `false`.
4. Verify default behavior is unchanged when the option is omitted or set to `false`.

## Related Issues

Fixes #3040
Related to #2819


alexvcasillas (dismissed):
Looks good to me 👍🏻

bdbch (approved):

PR #7788, opened 2026-05-01
fix(extension-mathematics): honor empty-string latex in updateBlockMath

iamgio:
## Changes Overview

Fix `editor.commands.updateBlockMath({ latex: '' })` silently ignoring the empty string. Previously, passing `latex: ''` left the node's `latex` attribute unchanged; it now correctly sets the attribute to `''`, matching `updateInlineMath`.

**Use case:** a custom editor popover where clearing the textarea should clear the rendered node while the user is mid-edit.

## Implementation Approach

The command body used a falsy fallback:

```ts
tr.setNodeMarkup(pos, this.type, {
  ...node.attrs,
  latex: latex || node.attrs.latex,
})
```

`'' || node.attrs.latex` evaluates to the old value, so an explicit empty-string update is dropped. `updateInlineMath` does not have this bug — it uses `{ ...node.attrs, latex }` directly.

The fix:

- Replace `latex || node.attrs.latex` with `latex ?? node.attrs.latex` so only `undefined` falls back. Empty strings now flow through. The two commands behave consistently.
- Relax the command type signature from `{ latex: string; pos?: number }` to `{ latex?: string; pos?: number }` to match `updateInlineMath`, since the command already supports being called without a `latex` value.

## Testing Done

Added tests in `packages/extension-mathematics/__tests__/blockMath.spec.ts`:

- `updateBlockMath({ latex: '' })` on a node with `latex: 'x^2'` → attrs updated to `latex: ''`.
- `updateBlockMath({})` (no latex passed) on a node with `latex: 'x^2'` → attrs unchanged.
- The block-math command's empty-string and missing-latex behaviour matches `updateInlineMath`.

`pnpm --filter @tiptap/extension-mathematics run lint` is clean.

## Verification Steps

1. `pnpm install`
2. `pnpm exec vitest run packages/extension-mathematics`
3. In a demo editor with `BlockMath` registered, insert a block math node (`editor.commands.insertBlockMath({ latex: 'x^2' })`), then run `editor.commands.updateBlockMath({ latex: '' })` and inspect `editor.getJSON()` — the node's `latex` attr should be `''`.

## Additional Notes

None.

## Checklist

### DIFF
--- a/packages/extension-mathematics/src/extensions/BlockMath.ts
+++ b/packages/extension-mathematics/src/extensions/BlockMath.ts
@@ -55,7 +55,7 @@ declare module '@tiptap/core' {
        * @param options - Options for updating block math.
        * @returns ReturnType
        */
-      updateBlockMath: (options?: { latex?: string; pos?: number }) => ReturnType
+      updateBlockMath: (options?: { latex: string; pos?: number }) => ReturnType
     }
   }
 }
--- a/packages/extension-mathematics/src/extensions/BlockMath.ts
+++ b/packages/extension-mathematics/src/extensions/BlockMath.ts
@@ -55,7 +55,7 @@ declare module '@tiptap/core' {
        * @param options - Options for updating block math.
        * @returns ReturnType
        */
-      updateBlockMath: (options?: { latex: string; pos?: number }) => ReturnType
+      updateBlockMath: (options?: { latex?: string; pos?: number }) => ReturnType
     }
   }
 }
@@ -157,7 +157,7 @@ export const BlockMath = Node.create<BlockMathOptions>({
 
           tr.setNodeMarkup(pos, this.type, {
             ...node.attrs,
-            latex: latex || node.attrs.latex,
+            latex: latex ?? node.attrs.latex,
           })
 
           return true
--- a/.changeset/fix-update-block-math-empty-latex.md
+++ b/.changeset/fix-update-block-math-empty-latex.md
@@ -1,0 +1,5 @@
+---
+"@tiptap/extension-mathematics": patch
+---
+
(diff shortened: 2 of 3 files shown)


arnaugomez (dismissed):
Good catch, thank you very much.

I've reviewed and this does not seem to introduce any regressions.

LGTM

arnaugomez:
@iamgio can you solve the merge conflicts?

packages/extension-mathematics/src/extensions/BlockMath.ts
```diff
@@ -55,7 +55,7 @@ declare module '@tiptap/core' {
        * @param options - Options for updating block math.
        * @returns ReturnType
        */
-      updateBlockMath: (options?: { latex: string; pos?: number }) => ReturnType
+      updateBlockMath: (options?: { latex?: string; pos?: number }) => ReturnType
```
arnaugomez: Sorry for bringing this again. This change modifies the public API of the extension. I think this could be avoided.

I'm not sure it makes sense to allow `undefined` as a possible value of `latex`. If I understand correctly, it would be equivalent to not calling `updateBlockMath` at all.

Couldn't it be possible to keep it like this:

```suggestion
      updateBlockMath: (options?: { latex: string; pos?: number }) => ReturnType
```

And update the changeset accordingly.

arnaugomez (changes requested):

packages/extension-mathematics/src/extensions/BlockMath.ts
iamgio: Makes sense - updated

arnaugomez (approved):

arnaugomez:
Thank you Gio!!

PR #7799, opened 2026-05-05
fix: preserve raw inline style values to avoid &quot; encoding

### LINKED ISSUE #7016: Style attribute values are encoded with `&quot;` in `getHTML()` output (Tiptap / ProseMirror 1.25.2)
### Affected Packages

Core

### Version(s)

2.12.0

### Bug Description

After a clean install of my application that using the **Tiptap** editor, I noticed that certain inline style attributes are being HTML-encoded when retrieving content using `getHTML()`.

Specifically, I observed that when setting a `font-family` with multiple words (e.g., `Times New Roman`, `Red Hat Display`, `Open Sans`), the resulting HTML wraps the font value in `&quot;` entities, such as:

`<span style="font-family: &quot;Open Sans&quot;">Some text</span>`

This issue appears to have started after updating to **prosemirror-model version 1.25.2**, and seems related to this [commit](https://github.com/ProseMirror/prosemirror/issues/1522), which introduced changes to improve style injection safety.

While this has so far been observed with `font-family`, it’s likely that other style-related attributes with quoted values may also be affected.

## Actual Behavior

The `getHTML()` method returns style attributes with encoded quotes, like:

`<span style="font-family: &quot;Open Sans&quot;;">Some text</span>`

## Scope of Impact

Although currently observed with `font-family`, this behavior likely affects **any style property** that requires quoted values, such as:

- `font-family`
- `background-image` (e.g., URLs)
- `content` (in pseudo-elements)
- `font` shorthand values, etc.

### Browser Used

Chrome

### Code Example URL

_No response_

### Expected Behavior

The output of `getHTML()` should preserve the intended formatting of style attribute values without converting quotes to HTML entities. For example:

- `<span style='font-family: "Open Sans;"'>Some text</span>`
- or `<span style="font-family: 'Open Sans;'">Some text</span>`
- or `<span style="font-family: Open Sans;">Some text</span>` // The current rendering before the prosemirror-model upgrade

This ensures the styling remains valid and usable in both rendering and post-processing contexts.

This issue is being encountered within **Tiptap**, which uses ProseMirror under the hood. Since `getHTML()` is the primary way Tiptap outputs editor content, this change in ProseMirror's `prosemirror-model` is affecting the output format unexpectedly.

## Question

- Is this change expected behavior due to ProseMirror's style sanitization updates?
- Is there a recommended workaround or Tiptap-specific solution to avoid or reverse the `&quot;` encoding?

Any insight or recommendations would be appreciated. Thank you!

### Dependency Updates

ozdemircibaris:
## Changes Overview

Fixes `&quot;` HTML entity encoding in `getHTML()` output for inline style attributes parsed by the FontFamily, FontSize, LineHeight, and Highlight extensions.

## Implementation Approach

CSSOM accessors (e.g. `element.style.fontFamily`) canonicalize multi-word values to a quoted form, which `innerHTML` serialization then encodes as `&quot;`. Updated `parseHTML` in those extensions to parse the raw `style` attribute directly (last-declaration-wins), matching the existing pattern in `Color` and `BackgroundColor`.

## Testing Done

- Added 15 unit tests across `font-family.spec.ts`, `font-size.spec.ts`, `line-height.spec.ts` covering multi-word/quoted/unquoted values, multiple declarations, CSS variables, and case-insensitivity.
- Verified `pnpm test:unit` (837 tests pass), `pnpm lint` (71 packages), and DTS build all pass.
- Reproduced the original issue in the FontFamily React demo and confirmed `getHTML()` no longer emits `&quot;` for unquoted/single-quoted multi-word fonts.

## Verification Steps

1. `pnpm install && pnpm dev`
2. Open the FontFamily extension demo.
3. In the browser console, run `editor.getHTML()`.
4. Confirm the output for multi-word fonts like `Comic Sans MS, Comic Sans` or `'Exo 2'` preserves the original formatting without `&quot;` entities.

## Additional Notes

The issue speculated this started with `prosemirror-model 1.25.2`, but the bug reproduces on 1.24.1 too — root cause is the per-extension CSSOM-based `parseHTML`, unrelated to the linked PM CSP change.

## Checklist

## Related Issues

Closes #7016


arnaugomez (commented):
I know it's different package but still it's strange to me that there is so much repeated code. Maybe we could add a utility to tiptap-core? Presumably, there will be other style marks in the future. Maybe some of them custom extensions created by customers. And they'll encounter the same problem.

arnaugomez (dismissed):

ozdemircibaris:
> I know it's different package but still it's strange to me that there is so much repeated code. Maybe we could add a utility to tiptap-core? Presumably, there will be other style marks in the future. Maybe some of them custom extensions created by customers. And they'll encounter the same problem.

Great point, agreed, the duplication had already crept up to 4 places (`Color` and `BackgroundColor` had the same inline pattern). Pushed an update that extracts `getStyleProperty` to `@tiptap/core` and migrates all six affected extensions (`Color`, `BackgroundColor`, `FontFamily`, `FontSize`, `LineHeight`, `Highlight`) to use it. Custom extensions can now reuse the helper for the same issue.

arnaugomez (changes requested):
What if the customer is using an older version of Tiptap? Then `getStyleProperty` import would not be available.

ozdemircibaris:
> What if the customer is using an older version of Tiptap? Then `getStyleProperty` import would not be available.

As far as I checked, all `@tiptap/*` packages release together (changeset `"fixed"` config) and peer deps resolve to exact pins on publish (`@tiptap/extension-text-style@3.22.5` → `"@tiptap/core": "3.22.5"`). So npm/pnpm should prevent the mismatch: customer can't install a newer extension against an older core.

Only edge case I can think of is third-party custom extensions importing `getStyleProperty` with a loose peer-dep range, but that's standard API-addition responsibility on the extension author.

Happy to add a runtime fallback (`typeof getStyleProperty === 'function'` + CSSOM) if you'd prefer extra safety, but it would weaken the fix on older cores. Leaning towards trusting the package manager, but let me know if I'm missing a scenario you have in mind @arnaugomez

arnaugomez:
> > What if the customer is using an older version of Tiptap? Then `getStyleProperty` import would not be available.
> 
> As far as I checked, all `@tiptap/*` packages release together (changeset `"fixed"` config) and peer deps resolve to exact pins on publish (`@tiptap/extension-text-style@3.22.5` → `"@tiptap/core": "3.22.5"`). So npm/pnpm should prevent the mismatch: customer can't install a newer extension against an older core.
> 
> Only edge case I can think of is third-party custom extensions importing `getStyleProperty` with a loose peer-dep range, but that's standard API-addition responsibility on the extension author.
> 
> Happy to add a runtime fallback (`typeof getStyleProperty === 'function'` + CSSOM) if you'd prefer extra safety, but it would weaken the fix on older cores. Leaning towards trusting the package manager, but let me know if I'm missing a scenario you have in mind @arnaugomez

Ok. It's the first time I come across this issue where we introduce an API for the first time and another package depends on it. @bdbch what do you think we should do in situations like this? Maybe it has happened to you before

bdbch:
I think it's reasonable to expect our users to upgrade their package versions correctly. I think right now the required package versions / peer deps are also setup in a way where users would receive a warning by their package manager if they have drifting versions.

arnaugomez (approved):

arnaugomez:
Ok, then, approved. Thanks @bdbch

PR #7374, opened 2025-12-29
Make tabindex configurable

crisp-tweakers:
## Changes Overview

Makes the tabindex for the Tabindex core extension configurable, in order to accommodate specific requirements wrt focus order when used in forms.

## Implementation Approach

An approach was already outlined in issue https://github.com/ueberdosis/tiptap/issues/4632 using the coreExtensionOptions configuration option for passing a custom value to the extension. So basically when initializing the editor it can be configured like follows:

```
new Editor({
    coreExtensionOptions: {
        tabindex: {
            value: '4',
        },
    },
})
```

## Testing Done

Tested locally, with and without specifying a value through the configuration option. When no configuration is specified the behaviour remains unchanged (the extension will add a `tabindex="0"` to the editor).

## Verification Steps

When specifying a value through the configuration options the tabindex on the editor instance in the DOM will reflect this value.

## Additional Notes

As there is no existing documentation for the Tabindex core extension I have not updated any documentation. This feature probably has a very niche use-case, and I am not sure what the best way is to document this.

## Checklist

- [ ] I have created a [changeset](https://github.com/changesets/changesets) for this PR if necessary.
- [x] My changes do not break the library.
- [ ] I have added tests where applicable.
- [ ] I have followed the project guidelines.
- [ ] I have fixed any lint issues.

## Related Issues

https://github.com/ueberdosis/tiptap/issues/4632

PR #7804, opened 2026-05-07
Make tabindex configurable

### LINKED ISSUE #4632: [Bug]: Cannot set tabIndex on the React EditorContent

core

2.1.12

`tabIndex` cannot be set on the editor, but we require this to allow focusing the editor even when it's read-only.

Chrome

### Code example

_No response_

Focusing editor `div` should be possible regardless of editability, but current API prevents it via all methods.

_No response_

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

arnaugomez:
This PR makes the `tabindex` attribute of the editor configurable via `coreExtensionOptions`.

It builds on the work started in #7374 and addresses the review feedback from that PR.

A `TabindexOptions` type is exported for type safety. The `coreExtensionOptions` interface now includes the `tabindex` property, and JSDoc documentation was added to the extension.

The logic now supports setting a custom tabindex on non-editable editors (for example, `tabindex="-1"` to allow programmatic focus), which fully resolves #4632.

Tests were added to cover default behavior, custom values on editable and non-editable editors, and undefined handling.

### DIFF
--- a/.changeset/many-houses-occur.md
+++ b/.changeset/many-houses-occur.md
@@ -1,0 +1,5 @@
+---
+"@tiptap/core": minor
+---
+
+Add configurable tabindex option to allow customizing the tabindex attribute on the editor element via coreExtensionOptions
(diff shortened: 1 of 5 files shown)


bdbch (approved):

bdbch:
We should make sure to add documentation for this as well.

PR #7807, opened 2026-05-07
Fix DragHandle React unmount cleanup

### LINKED ISSUE #7806: [Bug]: DragHandle React wrapper can throw removeChild NotFoundError on unmount
### Affected Packages

extension-drag-handle-react, extension-drag-handle

### Version(s)

3.20.4

### Bug Description

`@tiptap/extension-drag-handle-react` can throw a DOM `NotFoundError` when the `DragHandle` component unmounts after the underlying drag handle plugin has reparented its element.

The error is:

```txt
Uncaught NotFoundError: Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node.
```

The React wrapper currently renders the drag handle element directly:

```tsx
return (
  <div ref={setElement}>
    {children}
  </div>
)
```

Then it passes that same React-owned DOM element to `DragHandlePlugin`.

Inside `DragHandlePlugin`, the element is moved into an imperatively-created wrapper:

```ts
const wrapper = document.createElement('div')
wrapper.appendChild(element)
editor.view.dom.parentElement?.appendChild(wrapper)
```

and the plugin later removes that wrapper on destroy:

```ts
removeNode(wrapper)
```

This means the same DOM node is owned by React’s host tree, but is also reparented/removed by the ProseMirror plugin. When the React component unmounts, React may try to remove the `<div>` from its original parent, but it is no longer a child of that parent, so the browser throws `NotFoundError`.

This is easier to trigger when an editor switches between editable/read-only modes or when a view swap conditionally mounts/unmounts `DragHandle`.

### Browser Used

Chrome

### Code Example URL

_No response_

### Expected Behavior

Unmounting `DragHandle` should not throw, even if the drag handle plugin has moved the handle element into its own wrapper.

The React adapter should keep DOM ownership clear: if `DragHandlePlugin` reparents/removes the handle container, React should not also own that same container as a normal host node.

One possible implementation would be for `@tiptap/extension-drag-handle-react` to create the plugin-managed element imperatively and render `children` into it with `createPortal`, so:

- Tiptap/ProseMirror owns the drag handle container and can move/remove it.
- React owns only the portal children rendered inside that container.
- Toggling editor modes or conditionally unmounting `DragHandle` does not produce a `removeChild` `NotFoundError`.

_No response_

### Dependency Updates

Tintef:
## Summary

Fixes #7806.

This updates `@tiptap/extension-drag-handle-react` so `DragHandle` no longer renders the plugin-managed handle element as a normal React host node.

Instead, the handle element is created imperatively and React renders `children` into it with `createPortal`. This keeps DOM ownership clear: the drag handle plugin can move/remove the handle container, while React owns only the portal children.

The cleanup also avoids unregistering the plugin after the editor has already been destroyed.

## Checks

- `corepack pnpm --filter @tiptap/extension-drag-handle-react lint`
- `corepack pnpm --filter @tiptap/extension-drag-handle-react build`

Note: these commands passed locally with a Node engine warning because this shell is using Node 20 while current `main` requires Node 24.

### DIFF
--- a/.changeset/fix-drag-handle-react-unmount.md
+++ b/.changeset/fix-drag-handle-react-unmount.md
@@ -1,0 +1,7 @@
+---
+"@tiptap/extension-drag-handle-react": patch
+---
+
+Fix `DragHandle` unmounts by rendering children into the plugin-managed drag handle element with a React portal.
+
+This avoids React trying to remove a host node after the drag handle plugin has moved it into its own wrapper.
--- a/packages/extension-drag-handle-react/src/DragHandle.tsx
+++ b/packages/extension-drag-handle-react/src/DragHandle.tsx
@@ -7,9 +7,9 @@ import {
   normalizeNestedOptions,
 } from '@tiptap/extension-drag-handle'
 import type { Node } from '@tiptap/pm/model'
-import type { Plugin } from '@tiptap/pm/state'
 import type { Editor } from '@tiptap/react'
-import { type ReactNode, useEffect, useMemo, useRef, useState } from 'react'
+import { type ReactNode, useEffect, useMemo, useState } from 'react'
+import { createPortal } from 'react-dom'
 
 type Optional<T, K extends keyof T> = Pick<Partial<T>, K> & Omit<T, K>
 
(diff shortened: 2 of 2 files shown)


arnaugomez (approved):

arnaugomez:
Thanks for looking into this. I love React portals and I think they're a more idiomatic React solution. I'll ask @bdbch to review, too

bdbch (approved):
Looks good to me, thanks for your PR! I initially had some thoughts about creating the element via useState - specifically when React renders in strict mode or concurrent rendering which could lead to unattached DOM nodes but I think most browsers should be fine and just GC it.

I'll let the tests run through, afterwards will merge.

PR #7813, opened 2026-05-10
fix React NodeView batching and selection sync

### LINKED ISSUE #7811: [Bug]: Maximum update depth exceeded permanently breaks editor after large React-NodeView dispatch
### Affected versions
`@tiptap/react@3.22.3` and `@3.23.1` (latest). React 19, Vite + Vanilla JS or Next.js. All file/line references below are pinned to `v3.23.1`.

### Repro
https://github.com/szilard-dobai/tiptap-flushsync-repro
CodeSandbox: https://codesandbox.io/s/github/szilard-dobai/tiptap-flushsync-repro

Three buttons demonstrate that the trigger doesn't matter — `view.dispatch` from a React event handler, `view.dispatch` via `setTimeout`, and a synthetic `ClipboardEvent` all reproduce identically. Threshold is ~50.

### Trigger conditions
Two ingredients are both required:

1. A single transaction creates >50 React-backed NodeView instances. (Realistic for paste / `setContent` / programmatic dispatch.)
2. The NodeView's React component fires `setState` from a mount-time `useEffect` — i.e. anything that lazy-initializes a library, measures the DOM, hydrates state from a ref, etc. _Trivial_ NodeViews that render purely from props don't reproduce, but most real-world NodeViews do. (We originally hit this with an `equation` NodeView that lazy-loads a MathML editor.)

### Persistence
Once a single dispatch trips the limit, _every subsequent dispatch_ throws — even one inserting only ~40 atoms. Page reload required. React's `nestedUpdateCount` only resets when a commit drains all pending work; the editor's churn from `useSyncExternalStore` re-renders never lets that happen.

### Why
1. Paste / dispatch creates one transaction containing N>50 React-backed NodeView nodes.
2. `EditorView.updateStateInner` walks the doc and constructs a `ReactNodeView` per new node.
3. Each `ReactNodeView.mount()` instantiates a `ReactRenderer`, whose constructor calls `flushSync(() => this.render())` ([ReactRenderer.tsx#L188](https://github.com/ueberdosis/tiptap/blob/v3.23.1/packages/react/src/ReactRenderer.tsx#L188)) when `editor.isEditorContentInitialized` is true.
4. `render()` calls `editor.contentComponent.setRenderer(id, this)` ([ReactRenderer.tsx#L234](https://github.com/ueberdosis/tiptap/blob/v3.23.1/packages/react/src/ReactRenderer.tsx#L234) → [EditorContent.tsx#L65](https://github.com/ueberdosis/tiptap/blob/v3.23.1/packages/react/src/EditorContent.tsx#L65)), which synchronously notifies the `Portals` `useSyncExternalStore` subscriber.
5. Each NodeView's React component fires `setState` in a mount-time `useEffect`. That state update is "pending work" after the flushSync commit, so React increments `nestedUpdateCount`. After 50 of these in one update cycle, `getRootForUpdatedFiber` throws.

### Notable
Tiptap **already protects against this for initial mount** — [EditorContent.tsx#L123-L125](https://github.com/ueberdosis/tiptap/blob/v3.23.1/packages/react/src/EditorContent.tsx#L123-L125) calls `createNodeViews()` _before_ setting `isEditorContentInitialized = true`, so the initial population takes the `queueMicrotask` branch in `ReactRenderer` and never trips the limit even with hundreds of NodeViews. The same protection just isn't available for bulk dispatches that happen later (paste, `setContent`, programmatic insert).

### Stack trace from the CodeSandbox repro
```
at getRootForUpdatedFiber              (react-dom)
at enqueueConcurrentRenderForLane      (react-dom)
at forceStoreRerender                  (useSyncExternalStore subscriber)
at Set.forEach (subscribers)
at Object.setRenderer                  (EditorContent.tsx, line 65)
at ReactRenderer.render                (ReactRenderer.tsx, line 234)
at flushSync                           (ReactRenderer.tsx, line 188)
```

(Vite minifies/hoists the deeper frames above `flushSync`. In a Next.js + React 19 dev build the same throw shows the rest of the chain: `new ReactRenderer` → `ReactNodeView.mount` → `NodeViewDesc.create` → `iterDeco` / `updateChildren` → `EditorView.updateStateInner` → `Editor.dispatchTransaction`, with `doPaste` at the bottom when the trigger is a real paste event.)

### Workaround
Toggle `editor.isEditorContentInitialized = false` from `editorProps.handlePaste`, restore in a `queueMicrotask`. Forces `ReactRenderer` down its `else` branch (`queueMicrotask` instead of `flushSync`) so subscriber notifications batch into a single `Portals` re-render. The flag is internal — a public API for "render NodeViews asynchronously during this transaction" would be appreciated, since the alternative is reaching into private state.

```tsx
  const editor = useEditor({
    // ...
    editorProps: {
      handlePaste: (view) => {
        const editor: ReactRenderer["editor"] | undefined = (
          view.dom as TiptapEditorHTMLElement
        ).editor;
        if (!editor) return false;
        const wasInitialized = editor.isEditorContentInitialized;
        editor.isEditorContentInitialized = false;
        queueMicrotask(() => {
          editor.isEditorContentInitialized = wasInitialized;
        });
        return false;
      },
    },
  });
```

### Related (closed)
- #3764 — same flushSync mechanism
- #3976 — closed against v2 with "fixed in 2.6.0", but `flushSync` is still present in 3.22.x and 3.23.1
- #4695 — same as above
- #2985 — PR that introduced `flushSync`
- #3188 — PR that moved part of it to `queueMicrotask` (the branch the workaround exploits)

bdbch:
## Changes Overview

Batch React NodeView portal updates per ProseMirror view update, instead of flushing each NodeView separately. Also restore the early `contentDOM` attachment needed for selection-sensitive NodeView inserts.

## Implementation Approach

Added a protected `updateViewState` override point in core, then implemented the React batching behavior in `@tiptap/react`. React `EditorContent` now batches portal-store notifications, and React NodeViews keep their editable content container attached early so ProseMirror and React stay aligned.

## Testing Done

- `pnpm vitest run packages/react/src/Editor.spec.ts`
- `pnpm vitest run packages/react/src/EditorContent.spec.ts`
- `pnpm test:unit`

## Verification Steps

- Test the ReactComponentContent demo and press `Mod+Enter` inside the editable NodeView.
- Paste or `setContent` a large number of React NodeViews and confirm the maximum update depth error no longer occurs.

## Additional Notes

This adds a new core override seam for framework-specific view-update coordination, while keeping the batching logic in `@tiptap/react`.

## Checklist

## Related Issues

- Fixes #7811

### DIFF
--- a/.changeset/fair-forks-protect.md
+++ b/.changeset/fair-forks-protect.md
@@ -1,0 +1,7 @@
+---
+'@tiptap/core': minor
+---
+
+Add a protected `updateViewState` method to `Editor` as an override point for framework-specific editor subclasses that need to coordinate work around ProseMirror view updates.
+
+This allows integrations such as `@tiptap/react` to hook into editor view updates without adding framework-specific logic to the core package.
--- a/.changeset/quiet-mice-promise.md
+++ b/.changeset/quiet-mice-promise.md
@@ -1,5 +1,5 @@
 ---
-'@tiptap/react': patch
+'@tiptap/react': minor
 ---
 
 Batch React NodeView portal updates for each ProseMirror view update instead of triggering a separate `flushSync`-driven portal update for every NodeView.
--- a/.changeset/quiet-mice-promise.md
+++ b/.changeset/quiet-mice-promise.md
@@ -1,0 +1,7 @@
+---
+'@tiptap/react': patch
+---
+
+Batch React NodeView portal updates for each ProseMirror view update instead of triggering a separate `flushSync`-driven portal update for every NodeView.
+
+This reduces React update churn during bulk NodeView creation, such as paste, `setContent`, or other transactions that insert many React-backed NodeViews at once, while preserving selection-sensitive NodeView behavior.
--- a/packages/core/src/Editor.ts
+++ b/packages/core/src/Editor.ts
@@ -277,7 +277,7 @@ export class Editor extends EventEmitter<EditorEvents> {
       this.view.setProps(this.options.editorProps)
     }
 
-    this.view.updateState(this.state)
+    this.updateViewState(this.state)
   }
 
   /**
@@ -377,7 +377,7 @@ export class Editor extends EventEmitter<EditorEvents> {
 
     const state = this.state.reconfigure({ plugins })
 
-    this.view.updateState(state)
+    this.updateViewState(state)
 
     return state
   }
@@ -415,7 +415,7 @@ export class Editor extends EventEmitter<EditorEvents> {
       plugins,
     })
 
-    this.view.updateState(state)
+    this.updateViewState(state)
 
     return state
   }
@@ -556,7 +556,7 @@ export class Editor extends EventEmitter<EditorEvents> {
       plugins: this.extensionManager.plugins,
     })
 
-    this.view.updateState(newState)
+    this.updateViewState(newState)
 
     this.prependClass()
     this.injectCSS()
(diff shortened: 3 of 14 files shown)


bdbch:
I'll make this a WIP again as I'm not sure about the API changes and the implications for existing users.

PR #7869, opened 2026-05-25
Do not run selector when editor instance is `null` in current editor

### LINKED ISSUE #7849: TypeError: Cannot read properties of null (reading 'can')
### Affected Packages

@tiptap/extension-text-align

### Version(s)

3.23.1

### Bug Description

```
```

Place where it fails:
```ts
    editor,
    selector: (ctx) => {
      const editor = ctx.editor;

      return {
```

Current Workaround:
```ts
const editor = useEditor({
    immediatelyRender: false,
```

### Browser Used

Chrome

### Code Example URL

_No response_

### Expected Behavior

similiar issue

### Dependency Updates

arnaugomez:
## Changes Overview

Fix `useEditorState` so nullable editor snapshots return `null` instead of invoking selectors with `ctx.editor === null`.

## Implementation Approach

`useEditorState` now wraps the user selector and equality function to handle `null` editor snapshots. The state manager also notifies subscribers when the editor instance changes, including transitions to or from `null`.

## Testing Done

Added regression test in `packages/react/src/useEditorState.spec.ts`

## Verification Steps

Run the React package tests and verify a toolbar selector that calls `ctx.editor.can()` does not throw when `useEditorState` receives a `null` editor during development StrictMode lifecycle updates.

## Additional Notes

Fixes #7849.

## AI Usage

  - Used AI to inspect the issue, identify the `useEditorState` null-editor path, implement the fix, add regression coverage, and run verification commands.

## Checklist

## Related Issues

Fixes #7849

### DIFF
--- a/packages/react/src/useEditorState.ts
+++ b/packages/react/src/useEditorState.ts
@@ -80,11 +80,21 @@ class EditorStateManager<TEditor extends Editor | null = Editor | null> {
     }
   }
 
+  private notify() {
+    this.transactionNumber += 1
+    this.subscribers.forEach(callback => callback())
+  }
+
   /**
    * Watch the editor instance for changes.
    */
   watch(nextEditor: Editor | null): undefined | (() => void) {
+    const previousEditor = this.editor
+
     this.editor = nextEditor as TEditor
+    if (previousEditor !== nextEditor) {
+      this.notify()
+    }
 
     if (this.editor) {
       /**
@@ -93,8 +103,7 @@ class EditorStateManager<TEditor extends Editor | null = Editor | null> {
        * This could be more efficient, but it's a good trade-off for now.
        */
       const fn = () => {
-        this.transactionNumber += 1
-        this.subscribers.forEach(callback => callback())
+        this.notify()
       }
 
       const currentEditor = this.editor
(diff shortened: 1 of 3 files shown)


arnaugomez:
The user-provided `equalityFn` is typed for selected non-null values:

```ts
(a: TSelectorResult, b: TSelectorResult | null) => boolean
```

Is this correct? Or would it be more accurate to type it like this (breaking change):

```ts
(a: TSelectorResult | null, b: TSelectorResult | null) => boolean
```

packages/react/src/useEditorState.ts
```diff
@@ -153,14 +162,28 @@ export function useEditorState<TSelectorResult>(
...
+
+    return options.selector(snapshot as EditorStateSnapshot<Editor>)
+  }
+  const equalityFn = (a: TSelectorResult | null, b: TSelectorResult | null) => {
+    if (a === null || b === null) {
+      return a === b
+    }
+
+    return (options.equalityFn ?? deepEqual)(a, b)
+  }
```
arnaugomez, line 178: This change was necessary because the first argument of the user-provided equalityFn is typed as non-nullable:
```
(a: TSelectorResult, b: TSelectorResult | null) => boolean
```
Is this correct? Or would it be more accurate to type it like this (breaking change):
```
(a: TSelectorResult | null, b: TSelectorResult | null) => boolean
```

alexvcasillas (approved):

packages/react/src/useEditorState.ts
```diff
@@ -153,14 +162,28 @@ export function useEditorState<TSelectorResult>(
   options: UseEditorStateOptions<TSelectorResult, Editor> | UseEditorStateOptions<TSelectorResult, Editor | null>,
 ): TSelectorResult | null {
   const [editorStateManager] = useState(() => new EditorStateManager(options.editor))
+  const selector = (snapshot: EditorStateSnapshot<Editor | null>) => {
+    if (snapshot.editor === null) {
+      return null
```
bdbch, line 167: This is a runtime change right? Before - `useEditorState` would not return null as far as I remember which now is possible. This could break apps directly destructuring the useEditorState output in case the result of the selector is null.

```js
const { myValue } = useEditorState(...) // this causes a problem with a nullable return value
```

Before that, myValue would had a fallback value - where now the object couldn't even destruct.

packages/react/src/useEditorState.ts
```diff
@@ -80,11 +80,21 @@ class EditorStateManager<TEditor extends Editor | null = Editor | null> {
     }
   }
 
+  private notify() {
+    this.transactionNumber += 1
```
bdbch, line 84: I think the way you set this up now introduces a side-effect where the transactionNumber increases while the actual notify was caused because the editorInstance changed (as before, the notify would only be called when a transaction was handled)

bdbch: It's fine because in theory it's not changing a lot but maybe we rename this to something more descriptive like `updateCount` then?

bdbch (changes requested):
I'm not sure if we should merge this. The type change smells breaking to me because people who destructure and expect their editor objects to not be null would now suddenly get type errors or worse - runtime errors.

packages/react/src/useEditorState.ts
arnaugomez: I checked and the signature of useEditorState can return null.
```ts
export function useEditorState<TSelectorResult>(
  options: UseEditorStateOptions<TSelectorResult, Editor> | UseEditorStateOptions<TSelectorResult, Editor | null>,
): TSelectorResult | null
```
Originally, it would return null only before editor mount, when editor is null.
With this PR, it may return null anytime.
So this is, in my opinion, what might cause a breaking change.

bdbch: Yep I think so. I'm not really sure how we want to proceed. Is there another way to solve #7849 without including this right now?

arnaugomez: Yes, I think so. I'll create another PR tomorrow

PR #7870, opened 2026-05-25
fix(markdown): preserve unknown angle-bracket text as literal and add knownHtmlTags option

### LINKED ISSUE #7861: bug(markdown): Text inside angle brackets silently swallowed during markdown parsing
### Affected Packages

@tiptap/markdown

### Version(s)

3.23.6

### Bug Description

When markdown content contains text inside angle brackets (e.g. `<enter existing CID here if available>`), the `@tiptap/markdown` extension passes it through `marked` as raw HTML. The browser's DOM parser interprets it as an unknown HTML opening tag, causing:

1. **Text silently disappears** — the angle-bracket text is never rendered
2. With complex schemas (e.g. Mention, Variable, Table extensions), the resulting invalid DOM can crash ProseMirror with `Error: Called contentMatchAt on a node with invalid content`

### Steps to reproduce

```typescript
import { Editor } from '@tiptap/core';
import StarterKit from '@tiptap/starter-kit';
import { Markdown } from '@tiptap/markdown';

const editorDiv = document.createElement('div');

const editor = new Editor({
  element: editorDiv,
  extensions: [
    StarterKit,
    Markdown.configure({ markedOptions: { gfm: true, breaks: true } }),
  ],
  content: '<enter CID here if available>\n\nother text after',
  contentType: 'markdown',
  editorProps: {
    attributes: {
      class: 'Editor',
    },
  },
});

document.getElementById('app').appendChild(editorDiv);
```

**Observe:**
- The `<enter CID here if available>` text is **completely missing** from the rendered output
- With additional extensions (Mention, Variable, Table), clicking in the editor crashes with "Called contentMatchAt on a node with invalid content"

### Root cause

`marked` treats `<enter existing CID here if available>` as raw inline HTML and passes it through. The browser's DOM parser creates an unknown `<enter>` element (with no closing tag), which wraps all subsequent content. ProseMirror's DOMParser then produces a document with content that doesn't match the schema.

### Stacktrace

```
Error: Called contentMatchAt on a node with invalid content
  at contentMatchAt (prosemirror-model/dist/index.js:1455)
  at canReplace (prosemirror-model/dist/index.js:1466)
  at replaceRange (prosemirror-transform/dist/index.js:1964)
  at deleteRange (prosemirror-transform/dist/index.js:1979)
```

### Browser Used

Chrome

### Code Example URL

https://stackblitz.com/edit/vitejs-vite-ovefoo4u?file=src%2Fmain.js

### Expected Behavior

Since TipTap's ProseMirror schema has no node types for arbitrary HTML elements, unrecognized HTML-like text in markdown should be escaped and rendered as plain text (e.g. `&lt;enter existing CID here if available&gt;`).

_No response_

### Dependency Updates

bdbch:
## Changes Overview

Fix markdown parsing where HTML-like angle-bracket text (e.g. `<enter existing CID here>`) was silently swallowed. Uses `HTMLUnknownElement` (with schema parseDOM fallback) to detect unknown tags and preserves them as literal text.

Also adds `knownHtmlTags` option to Markdown extension for declaring custom tag names without a DOM API.

Closes #7861

## Checklist

### DIFF
--- a/.changeset/fix-markdown-unknown-html-tags.md
+++ b/.changeset/fix-markdown-unknown-html-tags.md
@@ -1,0 +1,5 @@
+---
+'@tiptap/markdown': patch
+---
+
+Fix text inside unknown angle-bracket tags being silently swallowed during markdown parsing. HTML-like content such as `<enter existing CID here if available>` that does not map to any known element is now preserved as literal text instead of disappearing, which also prevents downstream `contentMatchAt` errors when complex schemas (Mention, Variable, Table) are in use.
--- a/packages/markdown/__tests__/unknown-html-tags.spec.ts
+++ b/packages/markdown/__tests__/unknown-html-tags.spec.ts
@@ -2,6 +2,7 @@
  * @vitest-environment happy-dom
  */
 
+import { Node } from '@tiptap/core'
 import { Document } from '@tiptap/extension-document'
 import { Heading } from '@tiptap/extension-heading'
 import { Italic } from '@tiptap/extension-italic'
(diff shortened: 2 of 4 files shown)


bdbch:
@coderabbitai review full

PR #7871, opened 2026-05-25
fix(core): guard editor getters against null managers after destroy

### LINKED ISSUE #7849: TypeError: Cannot read properties of null (reading 'can')

jordanh:
## Description

The 3.23.0 memory-leak fix introduced this cleanup inside `Editor.destroy()`:

```ts
this.extensionManager.destroy()
this.extensionManager = null as any
this.schema = null as any
this.commandManager = null as any
```

The matching getters (`commands`, `chain`, `can`, `getHTML`, `getText`) were not updated to handle the null state. Any post-destroy access now throws `TypeError: Cannot read properties of null (reading 'commands')` instead of the silent no-op behavior of 3.18 and earlier.

This is a real-world hazard, not a theoretical one: in React, `@tiptap/react`'s `useEditor` schedules `destroy()` via `setTimeout(…, 1)` on unmount and also destroys + recreates when its dep array changes. Any `useEffect`, Relay/Apollo subscription callback, Yjs awareness update, or other async handler that holds a reference to the previous editor and reaches for `editor.commands.X(...)` between the schedule and the next mount will crash the entire React tree. We hit this in production after upgrading from 3.18.0 → 3.23.6 — multiple users could no longer load pages containing reflections rendered with TipTap.

## Fix

Restore the prior forgiveness by returning safe no-op values from the affected getters when the backing manager is null:

- `commands` / `can()` → a `Proxy` whose properties resolve to `() => false`
- `chain()` → a chainable `Proxy` whose `.run()` returns `false`
- `getHTML()` / `getText()` → empty string

No behavior change while the editor is alive — the guards only kick in when the manager has been nulled by `destroy()`.

## Test plan

Adds `packages/core/__tests__/destroyed.spec.ts` covering each getter after `destroy()`:

```
 ✓ does not throw when commands is accessed after destroy
 ✓ returns false when invoking a command after destroy
 ✓ does not throw when chain() is accessed after destroy
 ✓ returns false from chain().run() after destroy
 ✓ does not throw when can() is accessed after destroy
 ✓ returns "" from getHTML() after destroy
 ✓ returns "" from getText() after destroy
```

`pnpm run test:unit packages/core/__tests__/destroyed.spec.ts` → **7/7 passed**. Full suite remains green.

## Notes

- I deliberately did not change `destroy()` itself — keeping the memory-leak fix intact is important. The fix lives where the references are read.
- Happy to add a deprecation-style warning (`console.warn` once) if you'd prefer the getters to alert in dev when accessed post-destroy; let me know.

### DIFF
--- a/packages/core/src/Editor.ts
+++ b/packages/core/src/Editor.ts
@@ -247,6 +247,25 @@ export class Editor extends EventEmitter<EditorEvents> {
   }
 
   /**
+   * Create a no-op command chain used after `destroy()`, where every chainable method
+   * returns the chain itself and `.run()` returns `false`.
+   */
+  private createNoopChain(): ChainedCommands {
+    const noopChain: ChainedCommands = new Proxy(
+      { run: () => false },
+      {
+        get(target: { run: () => boolean }, prop: string | symbol) {
+          if (prop === 'run') {
+            return target.run
+          }
+          return () => noopChain
+        },
+      },
+    ) as unknown as ChainedCommands
+    return noopChain
+  }
+
+  /**
    * Create a command chain to call multiple commands at once.
    *
    * After `destroy()` returns a no-op chain where every chainable method returns the
@@ -254,18 +273,7 @@ export class Editor extends EventEmitter<EditorEvents> {
    */
   public chain(): ChainedCommands {
     if (!this.commandManager) {
-      const noopChain: ChainedCommands = new Proxy(
-        { run: () => false },
-        {
-          get(target: { run: () => boolean }, prop: string | symbol) {
-            if (prop === 'run') {
-              return target.run
-            }
-            return () => noopChain
-          },
-        },
-      ) as unknown as ChainedCommands
-      return noopChain
+      return this.createNoopChain()
     }
     return this.commandManager.chain()
   }
@@ -273,11 +281,24 @@ export class Editor extends EventEmitter<EditorEvents> {
   /**
    * Check if a command or a command chain can be executed. Without executing it.
    *
-   * After `destroy()` returns a no-op proxy whose checks always resolve to `false`.
+   * After `destroy()` returns a no-op proxy whose single-command checks resolve to
+   * `false` and whose `chain()` returns a no-op chain (so the
+   * `editor.can().chain().<cmd>().run()` pattern stays safe instead of throwing).
    */
   public can(): CanCommands {
     if (!this.commandManager) {
-      return new Proxy({}, { get: () => () => false }) as CanCommands
+      const createNoopChain = () => this.createNoopChain()
+      return new Proxy(
+        {},
+        {
+          get: (_target, prop) => {
+            if (prop === 'chain') {
+              return createNoopChain
+            }
+            return () => false
+          },
+        },
+      ) as CanCommands
     }
     return this.commandManager.can()
   }
--- a/packages/core/src/Editor.ts
+++ b/packages/core/src/Editor.ts
@@ -233,22 +233,52 @@ export class Editor extends EventEmitter<EditorEvents> {
 
   /**
    * An object of all registered commands.
+   *
+   * After `destroy()` the underlying `commandManager` is released, so this returns a no-op
+   * proxy whose properties resolve to functions returning `false`. This prevents
+   * `TypeError: Cannot read properties of null (reading 'commands')` when stale async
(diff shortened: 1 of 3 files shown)


bdbch (approved):

bdbch (changes requested):
Please resolve copilot review comments, thanks 🙏

jordanh:
@bdbch I have rebased this PR and addressed the automated review comments. Please let me know if there is anything else you'd like me to do before merging.

arnaugomez (dismissed):

jordanh:
@bdbch and @arnaugomez anything else blocking this merger? Happy to rebase again if necessary

bdbch:
@coderabbitai full review

bdbch:
@jordanh looks good from my end - I'll just do one last coderabbit review, then we're good I think!

jordanh:
Rebased onto current main (no conflicts) and addressed the CodeRabbit finding from the latest review round.

Fixed: no-op proxies were accidentally thenable. The post-destroy fallbacks for commands, chain(), and can() used catch-all get traps, so proxy.then resolved to a function — meaning await editor.commands (or any Promise.resolve() over these values) after `destroy()` would never settle. Each trap now returns undefined for then, so awaiting these values resolves immediately.

Added two regression tests: one verifying then is not exposed on any of the no-op proxies, and one verifying `Promise.resolve()` over them actually settles.

I also went ahead and cleaned up a few comments I had written to only include the comments I felt were necessary.

@bdbch and @arnaugomez please have another look. I think this is likely ready to be merged now.

jordanh:
Hi @bdbch and @arnaugomez, is this ready to merge from your perspective? Many thanks!

packages/core/src/Editor.ts
```diff
@@ -235,20 +235,74 @@ export class Editor extends EventEmitter<EditorEvents> {
...
+          }
+          if (prop === 'then') {
+            return undefined
+          }
+          return () => noopChain
+        },
+      },
+    ) as unknown as ChainedCommands
+    return noopChain
+  }
```
bdbch, line 269: maybe we can move this into the commandManager as a static function? That way we can move it out of the editor class that should not create command chains (even if they're noop)

packages/core/src/Editor.ts
```diff
@@ -235,20 +235,74 @@ export class Editor extends EventEmitter<EditorEvents> {
    * An object of all registered commands.
    */
   public get commands(): SingleCommands {
+    if (!this.commandManager) {
+      return new Proxy(
+        {},
+        {
+          get: (_target, prop) => (prop === 'then' ? undefined : () => false),
+        },
+      ) as SingleCommands
```
bdbch, line 244: Move this also into the commandManager as a static function so we can call it from the class.

packages/core/src/Editor.ts
```diff
@@ -235,20 +235,74 @@ export class Editor extends EventEmitter<EditorEvents> {
...
+            if (prop === 'chain') {
+              return createNoopChain
+            }
+            if (prop === 'then') {
+              return undefined
+            }
+            return () => false
+          },
+        },
+      ) as CanCommands
```
bdbch, line 304: Could probably also live on the commandManager as a createNoopCan() or something.

bdbch (changes requested):
Almost there, sorry for the long turnover. My notifications are overflowing and I'm trying to catch up right now. :)

bdbch:
Also I think this PR fixes #7849

bdbch:
Actually I've seen https://github.com/ueberdosis/tiptap/pull/8015 which solves the same problem but the approach is a bit more straight forward - where the commandManager is not nulled but kept alive but receives an isDestroyed flag.

I'd close this PR in favor of https://github.com/ueberdosis/tiptap/pull/8015. Sorry for the long wait @jordanh. I appreciate that you asked us again for reviews, kept the PR up to date, etc.

jordanh:
Not a worry @bdbch – glad to see it fixed!

PR #7873, opened 2026-05-25
Do not run useEditorState selector if editor is null (approach 2)

### LINKED ISSUE #7849: TypeError: Cannot read properties of null (reading 'can')

arnaugomez:
## Changes Overview

Fix `useEditorState` so selectors are not called when the editor snapshot briefly becomes `null` during teardown (for example in React StrictMode).

## Implementation Approach

Wrap the user selector and cache the last successful result in a discriminated `CachedSelectedState` ref (`stored: false` | `stored: true` with `value`). When the snapshot editor is `null` and nothing is cached yet, return `null` without calling the selector. When a cached value exists, return it via an `isCachedSelectedStateStored` type guard.

The selector `useCallback` depends on `options.selector`. A wrapped `equalityFn` treats `null` selections consistently during transient null snapshots.

This avoids the broader changes proposed in #7869 (no subscriber notifications on editor instance changes).

## Testing Done

Added `packages/react/src/useEditorState.spec.ts` covering:

- StrictMode remounts: selectors are never called with a `null` editor
- Permanently `null` editor: returns `null` without calling the selector

Ran `pnpm exec vitest run packages/react/src/useEditorState.spec.ts`.

## Verification Steps

Run the React package tests and verify a toolbar using `useEditorState` with `editor.can()` does not throw during StrictMode remounts.

## Additional Notes

Alternative to #7869 with a smaller surface area.

## AI Usage

  - Used AI to investigate #7849 and #7869, implement the fix, add regression tests, run verification commands, and refine the cache typing.

## Checklist

## Related Issues

Fixes #7849

### DIFF
--- a/packages/react/src/useEditorState.ts
+++ b/packages/react/src/useEditorState.ts
@@ -25,6 +25,9 @@ export type UseEditorStateOptions<TSelectorResult, TEditor extends Editor | null
   editor: TEditor
   /**
    * A selector function to determine the value to compare for re-rendering.
+   *
+   * When the editor snapshot is briefly `null`, this function is not called. If a
+   * previous selection was cached, the last successful result is returned instead.
    */
   selector: (context: EditorStateSnapshot<TEditor>) => TSelectorResult
   /**
@@ -149,6 +152,12 @@ export function useEditorState<TSelectorResult>(
 /**
  * This hook allows you to watch for changes on the editor instance.
  * It will allow you to select a part of the editor state and re-render the component when it changes.
+ *
+ * @remarks When the editor snapshot is briefly `null` (for example during React StrictMode
+ * remounts), the `selector` is not called. If a previous selection exists, the last successful
+ * result is returned instead. This can surface a briefly stale value until the editor is
+ * available again. If no selection was cached yet, `null` is returned.
+ *
  * @example
  * ```tsx
  * const editor = useEditor({...options})
--- a/packages/react/src/useEditorState.ts
+++ b/packages/react/src/useEditorState.ts
@@ -1,6 +1,6 @@
 import type { Editor } from '@tiptap/core'
 import { deepEqual } from 'fast-equals'
-import { useDebugValue, useEffect, useLayoutEffect, useState } from 'react'
+import { useCallback, useDebugValue, useEffect, useLayoutEffect, useRef, useState } from 'react'
 import { useSyncExternalStoreWithSelector } from 'use-sync-external-store/shim/with-selector.js'
 
 const useIsomorphicLayoutEffect = typeof window !== 'undefined' ? useLayoutEffect : useEffect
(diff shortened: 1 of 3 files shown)


arnaugomez:
This could be a potential fix. Another idea I have to fix this is make the signature of the selector accept a null value for the editor. Instead of only a non-null value.

This would be a breaking change. Perhaps part of Tiptap 4?

packages/react/src/useEditorState.ts
```diff
@@ -153,14 +161,43 @@ export function useEditorState<TSelectorResult>(
   options: UseEditorStateOptions<TSelectorResult, Editor> | UseEditorStateOptions<TSelectorResult, Editor | null>,
 ): TSelectorResult | null {
   const [editorStateManager] = useState(() => new EditorStateManager(options.editor))
+  const lastSelectedStateRef = useRef<CachedSelectedState<TSelectorResult>>({ stored: false })
+
+  const selector = useCallback(
+    (snapshot: EditorStateSnapshot<Editor | null>) => {
+      if (snapshot.editor === null) {
+        if (isCachedSelectedStateStored(lastSelectedStateRef.current)) {
+          return lastSelectedStateRef.current.value
```
bdbch, line 179: Should we add a notice around this behavior to the JSDoc? I think it's correct, but maybe it's worth to let users know that the selector can return the old, stale value in some special cases.

bdbch (dismissed):
LGTM

arnaugomez:
After having another look, I'm hesitant to merge this. I think this might lead to a breaking change.

Currently, the selector's `editor` parameter can already be null, when the `editor` option of the hook has type `Editor | null`

This situation where the `editor` parameter  is `null` is already handled by many developers in their apps, where they provide a fallback.

```ts
// editor can be null
const editor = useEditor()
// Default to fallback
const valueOrFallback = useEditorState({selector: (editor) => editor?.getJSON() ?? "fallback})
```

This PR removes that path and it causes the selector to return `null` when the `editor` is null. If we merged it, the fallback provided by developers would not work and this would be a breaking change.

Therefore, to me, the best solution is to always make the type of the `editor` parameter nullable. Which would be a breaking change (Tiptap 4), but easier to migrate

packages/react/src/useEditorState.ts
arnaugomez: Done

PR #7747, opened 2026-04-16
#feat: #7730 - First-class async fetching, positioning, and UX patterns for @tiptap/suggestion

gethari:
## Changes Overview

See #7730

## Implementation Approach

Already discussed/proposed in #7730 

## Testing Done

Verifed the positioning & updated the tiptap-docs https://github.com/ueberdosis/tiptap-docs/pull/732

## Checklist

- [ ] I have created a [changeset](https://github.com/changesets/changesets) for this PR if necessary.
- [ ] My changes do not break the library.
- [x] I have added tests where applicable.
- [X] I have followed the project guidelines.
- [x] I have fixed any lint issues.

PR #7874, opened 2026-05-25
feat(suggestion): async lifecycle and floating UI support

### LINKED ISSUE #7730: Feature Request: First-class async fetching, positioning, and UX patterns for `@tiptap/suggestion`
### Discussed in https://github.com/ueberdosis/tiptap/discussions/7646

<div type='discussions-op-text'>

<sup>Originally posted by **gethari** March 25, 2026</sup>
### Overview

After building a production suggestion/mention system on top of `@tiptap/suggestion`, we identified **nine gaps** between what the plugin provides and what a real-world async dropdown requires. Each gap forced us to implement workarounds inside `render()` — a lifecycle hook that was never designed to be a data-fetching or UI management layer.

This issue documents each gap as a distinct use case, the current workaround, and what a first-class API could look like. Addressing even a subset of these would significantly reduce the boilerplate every consumer has to write.

---

### Gap 1 — `items()` has no cancellation mechanism (stale results on fast typing)

**Problem**

`items()` is called on every keystroke. Tiptap awaits each call and fires `onUpdate` for every resolution — including stale ones from superseded queries. A user typing `@john` fires 5 concurrent `items()` calls; whichever resolves *last* wins, regardless of whether its query is still current.

**Workaround**

Bypass `items()` entirely, return `Promise.resolve([])`, and manage all fetching inside `render()` with a hand-rolled `AbortController` + `debounce`:

```ts
{
  // items() bypassed — Tiptap provides no cancellation
  items: () => Promise.resolve([]),
  render: () => {
    let abortController: AbortController | null = null;
    const debouncedFetch = debounce(async (query, cb) => {
      abortController?.abort();
      abortController = new AbortController();
      const results = await fetch(`/api/users?q=${query}`, {
        signal: abortController.signal,
      }).then(r => r.json());
      cb(results);
    }, 300);
    return {
      onStart(props) { debouncedFetch(props.query, update); },
      onUpdate(props) { debouncedFetch(props.query, update); },
      onExit() { debouncedFetch.cancel(); abortController?.abort(); },
    };
  },
}
```

**Proposed API**

```ts
// Option A — pass AbortSignal to items()
async items({ query, signal }) {
  return fetch(`/api/users?q=${query}`, { signal }).then(r => r.json());
}

// Option B — built-in debounce option
Suggestion({ debounce: 300, async items({ query }) { ... } })
```

---

### Gap 2 — No popup positioning

**Problem**

Tiptap provides `clientRect` (a getter for the decoration node's bounding rect) in `SuggestionProps`, but provides no mechanism to actually position a popup element relative to it. Every consumer must integrate their own positioning library.

**Workaround**

Integrate Floating UI manually inside `render()`, using `clientRect` as the virtual anchor:

```ts
onStart(props) {
  currentClientRect = props.clientRect;
  void computePosition(
    { getBoundingClientRect: () => currentClientRect?.() ?? new DOMRect() },
    popupElement,
    { strategy: 'fixed', placement: 'bottom-start', middleware: [offset(6), flip(), shift()] }
  ).then(({ x, y }) => {
    Object.assign(popupElement.style, { left: `${x}px`, top: `${y}px`, visibility: 'visible' });
  });
}
```

**Proposed API**

A `placement` option and a `onPosition` callback (or Floating UI middleware support) would let consumers skip the positioning boilerplate entirely:

```ts
Suggestion({
  placement: 'bottom-start',  // NEW — resolves position automatically from clientRect
  offset: 6,                  // NEW
})
```

---

### Gap 3 — Popup doesn't reposition on scroll or viewport resize

**Problem**

Even with custom positioning wired up, the popup becomes misaligned whenever the user scrolls the page or resizes the viewport while the suggestion is open. Tiptap does not emit any event for this.

**Workaround**

Manually attach `scroll` (capture) and `resize` listeners inside `onStart`, forward them to the positioning function, and remove them in `onExit`:

```ts
onStart() {
  window.addEventListener('scroll', reposition, { capture: true, passive: true });
  window.addEventListener('resize', reposition, { passive: true });
},
onExit() {
  window.removeEventListener('scroll', reposition, { capture: true });
  window.removeEventListener('resize', reposition);
}
```

**Proposed API**

If the plugin owns positioning (see Gap 2), repositioning on scroll/resize would be automatic. Alternatively, a `repositionOn?: string[]` option listing DOM events that should trigger a reposition.

---

### Gap 4 — Popup renders into `document.body`, breaking dialog/modal containment

**Problem**

Portaling the popup element into `document.body` places it outside any `<dialog>` or Radix/Headless UI modal that contains the editor. This causes z-index conflicts and (in browsers that implement the top-layer for `<dialog>`) the popup appears *behind* the dialog's backdrop.

**Workaround**

Walk up the DOM from `props.decorationNode` (or `props.editor.view.dom`) to find the nearest `[role="dialog"]` ancestor, and portal the popup inside that element instead:

```ts
onStart(props) {
  const dialogAncestor =
    props.decorationNode?.closest('[role="dialog"]') ??
    props.editor.view.dom.closest('[role="dialog"]');
  const container = dialogAncestor ?? document.body;
  container.appendChild(popupElement);
}
```

**Proposed API**

A `container` option accepting an `HTMLElement` or a resolver function:

```ts
Suggestion({
  container: (props) =>
    props.decorationNode?.closest('[role="dialog"]') ?? document.body,
})
```

---

### Gap 5 — Escape key is intercepted by dialogs before ProseMirror sees it

**Problem**

`handleKeyDown` in the ProseMirror plugin runs at the editor's DOM level. A `<dialog>` element (or any overlay with `keydown` capture) intercepts `Escape` *before* it reaches ProseMirror, so `onKeyDown` is never called and the suggestion popup is never dismissed. The dialog closes instead.

**Workaround**

Attach a `capture`-phase `keydown` listener on `document` that intercepts `Escape` before any overlay:

```ts
onStart() {
  document.addEventListener('keydown', handleEscapeKey, { capture: true });
},
function handleEscapeKey(event: KeyboardEvent) {
  if (event.key !== 'Escape' || !isOpen) return;
  event.preventDefault();
  event.stopPropagation();
  cleanup(); // close suggestion, NOT the dialog
}
```

**Proposed API**

The plugin should attach its own capture-phase Escape listener so suggestions inside dialogs work out of the box, without consumers needing to hook into `document` themselves.

---

### Gap 6 — No outside-click dismissal

**Problem**

Tiptap dismisses the suggestion when the editor loses focus (blur), but not when the user clicks on an unrelated part of the page *while the editor retains focus* (e.g. clicking a sidebar element scrolls the page but doesn't blur the editor). The popup stays open indefinitely.

**Workaround**

Attach a `capture`-phase `pointerdown` listener on `document` and check whether the click target is outside both the popup element and the editor:

```ts
document.addEventListener('pointerdown', (event) => {
  const isInsidePopup = popupElement.contains(event.target);
  const isInsideEditor = editor.view.dom.contains(event.target);
  if (!isInsidePopup && !isInsideEditor) {
    editor.commands.blur();
    cleanup();
  }
}, { capture: true });
```

**Proposed API**

A `dismissOnOutsideClick: boolean` option (defaulting to `true`) that handles this automatically.

---

### Gap 7 — No `charLimit` / minimum-query-length gating

**Problem**

For large user directories or expensive APIs, it is not practical to fetch suggestions on a single character (e.g. `@a` potentially matching thousands of users). There is no built-in way to suppress the fetch until the query reaches a minimum length, or to show the user a hint like *"Type at least 2 characters to search"*.

**Workaround**

Gate the `items()` bypass manually: check `query.length < charLimit` before calling the debounced fetch, and push a different UI state to the renderer component:

```ts
if (query.length < charLimit) {
  component.updateProps({ isCharLimitNotMet: true, users: [] });
  return;
}
// proceed with fetch
```

**Proposed API**

```ts
Suggestion({
  minQueryLength: 2,  // NEW — suppress items() until query reaches this length
})
```

When `minQueryLength` is set and not yet met, Tiptap should pass `items: []` to `onStart`/`onUpdate` along with a new `queryTooShort: true` flag in `SuggestionProps` so renderers can show a hint.

---

### Gap 8 — No pre-populated / instant-open items

**Problem**

A common UX pattern is to show a pre-populated list immediately when the trigger character is typed (e.g. `@` opens a dropdown with recently-mentioned users before any characters are typed). This requires items to be shown before any async fetch, instantly on trigger. There is no way to distinguish `onStart` with an empty query from `onUpdate` triggered by the user typing and deleting back to empty.

**Workaround**

Carry a separate `initialItems` config value, resolve it synchronously inside `onStart`, and inject it before the first fetch:

```ts
onStart(props) {
  const prePopulated = getInitialItems(); // synchronous getter
  if (props.query.length === 0 && prePopulated.length > 0) {
    component.updateProps({ users: prePopulated, isOpen: true });
    return; // skip fetch
  }
  // normal fetch path
}
```

**Proposed API**

```ts
Suggestion({
  // NEW — synchronous items shown immediately on trigger, before any query is typed
  initialItems: () => getRecentlyMentionedUsers(),
})
```

`SuggestionProps` would expose an `isInitial: boolean` flag so renderers can differentiate between "just opened" and "user has typed".

---

### Gap 9 — No loading state signal

**Problem**

Between the user typing a character and the `items()` promise resolving, there is nothing in `SuggestionProps` to indicate that a fetch is in progress. Renderers cannot show a spinner or skeleton without tracking this state themselves.

**Workaround**

Manually update the renderer component with a `isLoading: true` prop immediately before calling the debounced fetch, and `isLoading: false` in the callback:

```ts
component.updateProps({ isLoading: true }); // before fetch
debouncedFetch(query, (items) => {
  component.updateProps({ items, isLoading: false }); // after fetch
});
```

**Proposed API**

`SuggestionProps` should include a `loading: boolean` field that is `true` from the moment `items()` is called until it resolves:

```ts
interface SuggestionProps {
  // ... existing fields
  loading: boolean; // NEW — true while items() is in-flight
}
```

---

### Summary Table

| Gap | Workaround LoC | Proposed API |
|-----|:-:|---|
| 1. `items()` cancellation (stale results) | ~20 | `AbortSignal` in `items()` props, or `debounce` option |
| 2. Popup positioning | ~20 | `placement` + `offset` options |
| 3. Reposition on scroll/resize | ~8 | Automatic when plugin owns positioning |
| 4. Dialog/modal portal containment | ~6 | `container` resolver option |
| 5. Escape capture in dialogs | ~10 | Capture-phase listener in plugin |
| 6. Outside-click dismissal | ~10 | `dismissOnOutsideClick` option |
| 7. `charLimit` / minimum query length | ~15 | `minQueryLength` option + `queryTooShort` flag |
| 8. Pre-populated / instant-open items | ~15 | `initialItems` option + `isInitial` flag |
| 9. Loading state | ~6 | `loading: boolean` in `SuggestionProps` |
| **Total** | **~110 lines** | |

All nine of these are currently implemented by consumers in `render()` — a hook that was designed for UI lifecycle management, not data fetching or DOM event orchestration. First-class support for any subset would make `@tiptap/suggestion` significantly more self-contained for async use cases.

---

### Environment

- `@tiptap/suggestion`: `3.20.x`
- `@tiptap/core`: `3.x`
- Framework: React (`ReactRenderer`)

---

### Related

- [Discussion #1449 — suggestions when typing fast](https://github.com/ueberdosis/tiptap/issues/1449)
- [`suggestion.ts` — `items()` is awaited with no cancellation](https://github.com/ueberdosis/tiptap/blob/main/packages/suggestion/src/suggestion.ts)

### Use Case

All `@/#` or suggestion decorations rendered in the editor.

### Type

New feature</div>

bdbch:
Related - see #7747

## Changes Overview

Refactors `@tiptap/suggestion` into smaller internal modules, adds async request handling, and exposes resolved Floating UI config through `SuggestionProps`. Also updates the `SuggestionPositioning` and mention/emoji demos to use the shared positioning flow.

## Implementation Approach

Kept the public `Suggestion` API intact while moving lifecycle, async requests, props creation, and Floating UI config into dedicated helpers. The demo positioning logic now uses a shared helper so the examples follow the cursor consistently.

## Testing Done

- `pnpm exec vitest run packages/suggestion/src/__tests__/suggestion.test.ts`
- `pnpm exec playwright test demos/src/Examples/SuggestionPositioning/index.spec.ts --project=chromium`
- `pnpm --filter @tiptap/suggestion lint`

## Verification Steps

- Open the `SuggestionPositioning` demo and type `@` to confirm the popup appears in the expected position.
- Type after `@` and confirm the popup tracks the current cursor position instead of staying anchored to the trigger character.
- Click a suggestion item and confirm the mention is inserted.
- In the empty-query state, confirm the popup remains `top-start` and flips when there is not enough space above.

## Additional Notes

A changeset was added for `@tiptap/suggestion`.

## AI Usage

  - I used AI to help analyze the diff, refactor the suggestion internals, update the demos, and draft the PR description.

## Checklist

## Related Issues

- Fixes #7730

## Documentation

- https://github.com/ueberdosis/tiptap-docs/pull/806


alexvcasillas (dismissed):

bdbch:
Added documentation:
https://github.com/ueberdosis/tiptap-docs/pull/806

gethari:
Looks like a much better version of mine, Couple suggestions are missed ? Or is it intentional ?

#### Reposition on scroll/resize
- The plugin exposes positioning config, but doesn't automatically watch scroll/resize events. Consumers must still manually attach listeners.

#### Dialog/modal portal containment
- container option is exposed in SuggestionProps, but the plugin doesn't auto-portal. Consumers must handle element placement themselves.

#### Escape key in dialogs
- The plugin has escape handling, but doesn't use capture-phase listeners. In dialog contexts, parent dialogs may still intercept Escape first.

#### Outside-click dismissal
- No dismissOnOutsideClick option. Consumers must still wire up their own pointerdown listeners.

bdbch:
You're right @gethari however I think we should add those in different PRs to keep the PR easier to review & don't add too many things at once. We should try to keep this PR focused on the general refactor + async behavior and look into adding the other things via separate PRs.

packages/suggestion/src/helpers.ts
```diff
@@ -0,0 +1,129 @@
...
+ * Gets the DOM rectangle corresponding to the current editor cursor anchor position.
+ * Calculates screen coordinates based on Tiptap's cursor position and converts to a DOMRect object.
+ */
+export function getAnchorClientRect(editor: Editor): () => DOMRect | null {
+  return () => {
+    const pos = editor.state.selection.$anchor.pos
+    const coords = editor.view.coordsAtPos(pos)
+    const { top, right, bottom, left } = coords
+
+    try {
```
alexvcasillas, line 37: Question: why do we need a `try/catch` block here? `new DOMRect` does not throw because we're already using `new` and if any of the values cannot be parsed it will return `NaN` instead and not throw

alexvcasillas (dismissed):
Looks good to me 👍🏻 Just a small thing but not blocking at all :)

packages/suggestion/src/helpers.ts
bdbch: Hey, this is just a safety net for SSR environments. Some without dom polyfills dont know the reference DOMRect and will throw an undefined reference error.

bdbch:
I updated this PR to include:

- automatic mounting and positioning support. this should make wiring for mounting & auto positioning way easier
- An option to dismiss the option when clicked outside

For the last point that @gethari mentioned I'd rather do this on another PR.

alexvcasillas (approved):

PR #7875, opened 2026-05-26
fix(core): exclude unselected cells from plain-text copy of cell selections

### LINKED ISSUE #6867: [Bug]: Copying table cells includes text from other cells when pasting as plain text
### Affected Packages

extension-table, extension-table

### Version(s)

2.11.2

This issue is reproducing also in the Tiptap [documentation example](https://tiptap.dev/docs/editor/extensions/nodes/table-cell) with version 3.3.0.

### Bug Description

When I select two cells (one below the other) and paste the content as plain text (in Tiptap or another simple editor), the pasted text also includes the text from the other cells in between, based on the JSON/HTML ordering.

When pasting in Tiptap without plain-text mode, everything works correctly. 👌🏻

### Browser Used

Chrome

### Code Example URL

_No response_

### Expected Behavior

When pasting (both as plain text and not), I would expect to see only the content of the selected cells, without additional text from other cells.

### Dependency Updates

ozdemircibaris:
## Changes Overview

Fixes #6867: copying a table `CellSelection` and pasting as plain text included text from cells between the selected ones.

## Implementation Approach

`clipboardTextSerializer` no longer flattens `selection.ranges` to a single `min/max` window — it serializes each range independently and joins them with `blockSeparator`, sorted by document position so reverse drags also emit in doc order.

## Testing Done

Added 6 vitest specs (3 in core, 3 in extension-table) covering single `TextSelection` regression paths, the bug scenario (non-adjacent rows), reverse drag ordering, and adjacent cells; full suite passes 1064/1064.

## Verification Steps

Run `pnpm dev`, open the Table React demo, drag-select two cells in the same column and paste as plain text (Cmd+Shift+V) — only the selected cells should appear.

## Additional Notes

`@tiptap/core` patch; no public API change. `NodeRangeSelection` (the other multi-range `Selection` consumer) still produces identical output because its ranges are contiguous in document order.

## AI Usage

  - I used Claude to investigate the root cause in `clipboardTextSerializer.ts`, plan the per-range fix, write unit tests, and audit blast radius across `Selection` subclasses.

## Related Issues

Fixes #6867

### DIFF
--- a/packages/core/src/extensions/clipboardTextSerializer.ts
+++ b/packages/core/src/extensions/clipboardTextSerializer.ts
@@ -26,18 +26,25 @@ export const ClipboardTextSerializer = Extension.create<ClipboardTextSerializerO
             const { editor } = this
             const { state, schema } = editor
             const { doc, selection } = state
-            const { ranges } = selection
-            const from = Math.min(...ranges.map(range => range.$from.pos))
-            const to = Math.max(...ranges.map(range => range.$to.pos))
             const textSerializers = getTextSerializersFromSchema(schema)
-            const range = { from, to }
-
-            return getTextBetween(doc, range, {
-              ...(this.options.blockSeparator !== undefined
-                ? { blockSeparator: this.options.blockSeparator }
-                : {}),
+            const { blockSeparator } = this.options
+            const options = {
+              ...(blockSeparator !== undefined ? { blockSeparator } : {}),
               textSerializers,
-            })
+            }
+
+            // Serialize each selection range independently and join the results.
+            // CellSelection exposes one range per selected cell; flattening to
+            // min(from)/max(to) would pull in unselected cells between them.
+            // Sort by document position so reverse selections (e.g. dragging
+            // upward) still emit text in document order.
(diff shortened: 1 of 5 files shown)


packages/core/src/extensions/clipboardTextSerializer.ts
```diff
@@ -26,16 +26,23 @@ export const ClipboardTextSerializer = Extension.create<ClipboardTextSerializerO
...
+            const options = {
+              ...(blockSeparator !== undefined ? { blockSeparator } : {}),
               textSerializers,
-            })
+            }
+
+            // Serialize each selection range independently and join the results.
+            // CellSelection exposes one range per selected cell; flattening to
+            // min(from)/max(to) would pull in unselected cells between them
+            // (issue #6867). Sort by document position so reverse selections
```
bdbch: Remove explicit issue mention

bdbch (changes requested):

packages/core/src/extensions/clipboardTextSerializer.ts
ozdemircibaris: my bad sorry

bdbch (approved):
LGTM

PR #7876, opened 2026-05-26
fix(extension-collaboration,extension-collaboration-caret): prevent memory leaks on destroy

janthurau:
The extension subscribed to provider.awareness 'update' inside addProseMirrorPlugins but never unsubscribed. With multiple editors sharing a single provider, destroying an editor left the listener attached to the long-lived awareness emitter; the closure pinned the editor and prevented garbage collection.

Move the subscription into a dedicated PM plugin whose view().destroy() removes the listener, clears the local awareness user field, and resets storage.users. Adds a regression test.


demos/src/Experiments/CollaborationMemoryLeak/JS/index.html
```diff
@@ -0,0 +1,197 @@
+<div class="control-group">
```
bdbch: I was a bit confused about this demo, rarely see JS demos being used. Does having content in the index.html work - because I think it was stripped out last time I tried this. :D

bdbch (dismissed):

demos/src/Experiments/CollaborationMemoryLeak/JS/index.html
janthurau: yea I wasnt sure if I should keep this, but wanted to keep it for the review in case anyone wants to see it. I've removed it now.

arnaugomez (approved):
Added a small possible issue I found. I did not find any blockers.

packages/extension-collaboration-caret/src/collaboration-caret.ts
```diff
@@ -168,24 +169,39 @@ export const CollaborationCaret = Extension.create<
...
+
+        provider.awareness.setLocalStateField('user', user)
+        storage.users = awarenessStatesToArray(provider.awareness.states)
+        provider.awareness.on('update', onAwarenessUpdate)
+
+        return {
+          destroy: () => {
+            provider.awareness.off('update', onAwarenessUpdate)
+            storage.users = []
+          },
```
arnaugomez, line 194: The changeset says awareness state is cleared 

> The extension's awareness `update` listener is now removed on destroy, and the local awareness `user` state is cleared, so the editor can be garbage collected.

If the PR intends to actually clear the provider’s local awareness user, the destroy hook needs to clear awareness too:

```ts
destroy: () => {
  provider.awareness.off('update', onAwarenessUpdate)
  provider.awareness.setLocalStateField('user', null)
  storage.users = []
},
```

Otherwise, you can drop that claim in the changeset

arnaugomez (dismissed):

packages/extension-collaboration-caret/src/collaboration-caret.ts
janthurau: true! thx. I think we should not clear the awareness by default, because you may not want this if you connected multiple editors to the same awareness. I've removed the sentence from the changeset.

PR #7884, opened 2026-05-29
fix(markdown): preserve pipe characters inside inline code spans in table cells

### LINKED ISSUE #7858: special inline code in table parsing error
### Affected Packages

core, react, @tiptap/markown

### Version(s)

3.23.6

### Bug Description

```md
| Header |  Header | Header |
| ---- | ------ | --- |
| `||` | or |  `a || b`   |
| `&&` | and |  `a && b`   |
```
Can not  parsing `||` in markdown table

Image

### Browser Used

Edge

### Code Example URL

_No response_

### Expected Behavior

preserve `||` in markdown table cells as an inline code

_No response_

### Dependency Updates

sahiee-dev:
## Changes Overview

Fixes a parsing bug where pipe characters (`|`) inside backtick inline code spans were incorrectly treated as table column delimiters, causing table rows with cells like `` `||` `` or `` `a || b` `` to split into the wrong number of columns and lose their code formatting.

## Implementation Approach

`marked`'s internal `splitCells` function splits table rows by replacing every `|` with a split marker, only exempting backslash-escaped pipes (`\|`). It has no awareness of backtick code spans, so `` `||` `` in a table cell was split into multiple columns.

The fix adds a preprocessing step in `MarkdownManager.parse()`, before the markdown string is passed to marked's lexer, any `|` characters found inside backtick code spans on table-row lines (lines starting with `|`) are escaped to `\|`.

Since `splitCells` already converts `\|` back to `|` after splitting, the cell content is restored correctly with no second pass needed.

The preprocessing handles single, double, and triple backtick spans and is a no-op for lines that are not table rows or contain no backtick spans, so it does not affect any other markdown parsing.

## Testing Done

Added 5 tests to `packages/extension-table/__tests__/tableMarkdown.spec.ts`:

- Parses `` `||` `` in a table cell as a single cell with a code mark
- Parses a full 3-column row with `` `||` ``, `or`, `` `a || b` `` correctly
- Regression guard for `&&` (was also affected)
- Full parse → serialize → re-parse roundtrip preserving code marks
- Guard confirming tables without backtick spans are unaffected

All 6 tests in the file pass (1 pre-existing, 5 new).

## Verification Steps

1. Clone the repo and run:

```bash
vitest run packages/extension-table/tests/tableMarkdown.spec.ts
````

2. Start the demos dev server (`vite` from `demos/`) and open:

```text
/preview/Markdown/Full/React
```

3. Paste the following into the markdown input and click **Parse Markdown**:

```md
| Header | Header | Header |
| ------ | ------ | ------ |
| `||` | or | `a || b` |
| `&&` | and | `a && b` |
```

4. Confirm the table renders with 3 columns and inline code formatting on each code span.

5. Click **Extract Markdown** and confirm the output preserves the backtick wrappers.

## Additional Notes

The root cause is entirely inside `marked`'s `splitCells`, it is not something Tiptap can fix by overriding the table tokenizer without reimplementing the full table block rule.

The preprocessing approach is the minimal, safe intervention point within Tiptap's own code.

## Related Issues

Fixes #7858

### DIFF
--- a/demos/src/Markdown/Full/React/content.ts
+++ b/demos/src/Markdown/Full/React/content.ts
@@ -150,6 +150,15 @@ Nested content is also supported!
 
 :::
 
+### Inline Code in Tables (issue #7858)
+
+Pipe characters inside backtick code spans in tables should be preserved:
+
+| Expression | Meaning | Example |
+| ---------- | ------- | ------- |
+| \`||\` | or | \`a || b\` |
+| \`&&\` | and | \`a && b\` |
+
 ### Try editing the markdown on the left:
 
 1. Edit the markdown text
--- a/demos/src/Markdown/Full/React/content.ts
+++ b/demos/src/Markdown/Full/React/content.ts
@@ -150,7 +150,7 @@ Nested content is also supported!
 
 :::
 
-### Inline Code in Tables (issue #7858)
+### Inline Code in Tables
 
 Pipe characters inside backtick code spans in tables should be preserved:
 
--- a/packages/extension-table/__tests__/tableMarkdown.spec.ts
+++ b/packages/extension-table/__tests__/tableMarkdown.spec.ts
@@ -1,9 +1,79 @@
+import { Code } from '@tiptap/extension-code'
 import Document from '@tiptap/extension-document'
 import Paragraph from '@tiptap/extension-paragraph'
 import { TableKit } from '@tiptap/extension-table'
 import Text from '@tiptap/extension-text'
 import { MarkdownManager } from '@tiptap/markdown'
 import { describe, expect, it } from 'vitest'
+
+describe('table markdown — inline code with pipe characters (issue #7858)', () => {
+  const manager = new MarkdownManager({
+    extensions: [Document, Paragraph, Text, Code, TableKit],
+  })
+
+  it('should parse `||` inside a code span as a single cell', () => {
+    const markdown = '| Header |\n| ------ |\n| `||` |'
+    const parsed = manager.parse(markdown)
+    const bodyRow = parsed.content?.[0]?.content?.[1]?.content || []
+    expect(bodyRow).toHaveLength(1)
+    const textNode = bodyRow[0]?.content?.[0]?.content?.[0]
+    expect(textNode?.text).toBe('||')
+    expect(textNode?.marks?.[0]?.type).toBe('code')
+  })
+
+  it('should parse `a || b` inside a code span as a single cell', () => {
+    const markdown = '| H | H | H |\n| - | - | - |\n| `||` | or | `a || b` |'
+    const parsed = manager.parse(markdown)
+    const table = parsed.content?.[0]
+    expect(table?.type).toBe('table')
+    const bodyRow = table?.content?.[1]?.content || []
+    expect(bodyRow).toHaveLength(3)
+
+    const cell1 = bodyRow[0]?.content?.[0]?.content?.[0]
+    expect(cell1?.text).toBe('||')
+    expect(cell1?.marks?.[0]?.type).toBe('code')
+
+    const cell2 = bodyRow[1]?.content?.[0]?.content?.[0]
+    expect(cell2?.text).toBe('or')
+    expect(cell2?.marks).toBeUndefined()
+
+    const cell3 = bodyRow[2]?.content?.[0]?.content?.[0]
+    expect(cell3?.text).toBe('a || b')
+    expect(cell3?.marks?.[0]?.type).toBe('code')
+  })
+
+  it('should parse `&&` inside a code span correctly (regression guard)', () => {
+    const markdown = '| H | H | H |\n| - | - | - |\n| `&&` | and | `a && b` |'
+    const parsed = manager.parse(markdown)
+    const bodyRow = parsed.content?.[0]?.content?.[1]?.content || []
+    expect(bodyRow).toHaveLength(3)
+    const cell1 = bodyRow[0]?.content?.[0]?.content?.[0]
+    expect(cell1?.text).toBe('&&')
+    expect(cell1?.marks?.[0]?.type).toBe('code')
+    const cell3 = bodyRow[2]?.content?.[0]?.content?.[0]
+    expect(cell3?.text).toBe('a && b')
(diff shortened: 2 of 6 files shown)


demos/src/Markdown/Full/React/content.ts
```diff
@@ -150,6 +150,15 @@ Nested content is also supported!
 
 :::
 
+### Inline Code in Tables (issue #7858)
```
bdbch: I don't see why the issue should be mentioned in a demo.

packages/extension-table/__tests__/tableMarkdown.spec.ts
```diff
@@ -1,10 +1,80 @@
+import { Code } from '@tiptap/extension-code'
 import Document from '@tiptap/extension-document'
 import Paragraph from '@tiptap/extension-paragraph'
 import { TableKit } from '@tiptap/extension-table'
 import Text from '@tiptap/extension-text'
 import { MarkdownManager } from '@tiptap/markdown'
 import { describe, expect, it } from 'vitest'
 
+describe('table markdown — inline code with pipe characters (issue #7858)', () => {
```
bdbch: same here, our tests should be as generic as possible.

packages/markdown/src/MarkdownManager.ts
```diff
@@ -356,7 +356,7 @@ export class MarkdownManager {
     try {
       // Use a parse-scoped lexer so follow-up inline tokenization can reuse
       // the same configured lexer state without sharing it across parses.
-      const tokens = parseLexer.lex(markdown) as MarkdownToken[]
+      const tokens = parseLexer.lex(this.preprocessTablePipes(markdown)) as MarkdownToken[]
```
bdbch: Does it make sense to move this into the table extension's [parseMarkdown](https://github.com/ueberdosis/tiptap/blob/ae17f178e3e03b97b9292d1f8d82f73ae8623679/packages/extension-table/src/table/table.ts#L318) function instead? Throwing a table pipe fix into a generic parsing function feels very off...

packages/markdown/src/MarkdownManager.ts
```diff
@@ -371,6 +371,82 @@ export class MarkdownManager {
...
+      }
+      if (!found) {
+        // No matching closer on this line — emit the opening backtick run as-is
+        result += line.slice(i, i + runLen)
+        i += runLen
+      }
+    }
+    return result
+  }
+
```
bdbch: Same here, this should not live on the markdown manager at all.

bdbch (changes requested):

demos/src/Markdown/Full/React/content.ts
sahiee-dev: mistake on my side. Removed.

packages/extension-table/__tests__/tableMarkdown.spec.ts
sahiee-dev: removed.

packages/markdown/src/MarkdownManager.ts
sahiee-dev: Agreed. I looked at parseMarkdown, but it runs after the row has already been split into cells. I've moved the fix into the table extension's markdown tokenizer instead.

packages/markdown/src/MarkdownManager.ts
sahiee-dev: Agreed. it no longer does. All the logic is in `packages/extension-table/src/table/utilities/markdown.ts` wired through the Table extension's `markdownTokenizer`. The MarkdownManager diff is empty.

sahiee-dev:
Updated the PR based on your feedback.

The fix is now entirely inside `extension-table` via a block-level `markdownTokenizer`; `MarkdownManager` is untouched.

I also found and fixed a subtle bug in the first revision: `helper.blockTokens` was being called on the full remaining source, which caused `token.raw` to be longer than the original table content and could silently consume characters after the table.

The fix now extracts only the table lines before re-lexing, and I've added a regression test that would have caught this issue.

bdbch (changes requested):
Thanks for the PR.

I don’t think we should merge this yet.

The fix only handles tables where each row starts with a pipe.

For example:

    | A | B |
    | - | - |
    | a | b |

Markdown tables can also be written without the leading pipe:

    A | B
    - | -
    a | b

In that case the bug still exists, because the new tokenizer only checks lines that start with a pipe.

So the fix is incomplete.

Please update the logic to support tables without leading pipes too, and add a test for that case.

Also, please check the changeset. The code change is in `@tiptap/extension-table`, but the changeset only mentions `@tiptap/markdown`. The bot also reports a very large package release, which seems off for this PR.

Once these two things are fixed, we can review again.

sahiee-dev:
Both points addressed.

**Pipeless tables**: `start()` now does a two line lookahead, checks that the second line matches the GFM separator pattern (`/^[ \t|:]*-[ \t|:-]*$/`) and contains at least one `|` before activating. Covers both pipe prefixed and pipeless tables, and rejects setext headings (`---` without `|`) before `tokenize` is reached. Added two tests: `||` in a pipeless table cell, and a content-after-table regression guard.

**Changeset**: Already pointing at `@tiptap/extension-table` from a previous commit.

**Package count**: The 72 package figure is a pre existing artifact of this monorepo's changeset dependency graph, any `extension-table` patch produces the same number.

sahiee-dev:
Hi @bdbch, following up on this one.

I've addressed the requested changes:

Added support for both leading-pipe and pipeless GFM tables.
Corrected the changeset to target `@tiptap/extension-table`.
Tightened `start()` detection with a two-line lookahead and separator-row validation to reduce false positives and avoid activating on cases such as setext headings.
Added regression coverage for setext headings, content following tables, inline code containing pipes, and pipeless table cases.
Limited re-lexing to the candidate table block to avoid consuming subsequent content.

All tests are passing locally. I'd appreciate another review when you have time. Thanks!

sahiee-dev:
Hi @bdbch, just a friendly follow-up on this PR. Thanks!

bdbch (approved):
Thanks, looks good now. Sorry for the wait - I got blocked by other things.

Issue #6398, opened 2025-05-29
[PRO]: Comments on inline code and code blocks are deleted


tylercallahan:
### Affected Packages

extension-comments

### Version(s)

2.20.2

### Description of the Bug

When a user attempts to add a comment to a text selection within inline code or a code block, the comment is immediately deleted (and the `<span>` marking the comment is removed).

Similarly, when text in the editor has an associated comment, and that text becomes formatted as inline code or a code block, the associated comment is deleted.

This is reproducible in [Tiptap's commenting demo](https://next-comments-template.vercel.app/comments).

![Image](https://github.com/user-attachments/assets/698ebdb0-2527-4e78-a6d3-d2b070979bd9)

I'm wondering if this may also be related to issue #6348, if a `code` format disallows any other "marks".

### Browser Used

Chrome

### Code Example (Preferred)

https://next-comments-template.vercel.app/comments/

### Expected Behavior

* It should be possible to add a comment to text within inline code or a code block.
* When text with an associated comment is formatted as code, the comment should be preserved.

### Dependency Updates

- [x] Yes, I've updated all my dependencies.

bdbch:
Pretty sure this is a problem with the schema as code usually doesn't allow marks. Did you try allowing inline marks in comment blocks?

tylercallahan:
Thanks @bdbch - I investigated the extensions a bit more and think I determined how we can change the config so the schema allows comments on code.

**For the inline `Code` extension**, I see `excludes: '_'`, [meaning](https://tiptap.dev/docs/editor/core-concepts/schema#excludes) that all marks are excluded.

https://github.com/ueberdosis/tiptap/blob/8f296138216dc175c631a7ae1edf1076dd87298c/packages/extension-code/src/code.ts#L60

We can extend our use of this extension to remove the exclusion, e.g.

```js
extensions: [
  Code.extend({ excludes: '' }), // or excludes: everything except 'inlineThread'
  ...
]
```
... and then inline code will allow comments.

**For the `CodeBlock` extension**, similarly the setup includes `marks: ''`, [meaning](https://tiptap.dev/docs/editor/core-concepts/schema#marks) no marks are allowed.

https://github.com/ueberdosis/tiptap/blob/8f296138216dc175c631a7ae1edf1076dd87298c/packages/extension-code-block/src/code-block.ts#L82

We can similarly extend this extension to allow `inlineThread` (or all) marks, e.g.

```js
extensions: [
[...]

PR #7887, opened 2026-05-29
fix(core): add clearable mark option for unsetAllMarks

### LINKED ISSUE #6348: [PRO]: unsetAllMarks deletes comments
### Affected Packages

extension-comments

### Version(s)

2.17.5

### Description of the Bug

Our Tiptap editor has a "clear formatting" button that calls [`unsetAllMarks`](https://tiptap.dev/docs/editor/api/commands/nodes-and-marks/unset-all-marks) to remove formatting from the selection.

We noticed, however, that if the selection includes content associated with a comment, the comment is also deleted (presumably because the `<span>` element wrapping the content is removed).

Is this a bug? I can see how we might want `unsetAllMarks` to include comments, if we consider the comment tag a "mark," but at the same time I imagine users would generally not expect `unsetAllMarks` to delete comments.

For now, as a workaround, we are calling `unsetMark` separately for all non-comment marks, e.g.

```js
editor.chain().focus().unsetMark('link').unsetMark('bold').unsetMark('strike') ...
```

... but would be nice to be able to use `unsetAllMarks` again for formatting-specific marks. Thanks!

### Browser Used

Chrome

### Code Example (Preferred)

_No response_

### Expected Behavior

`unsetAllMarks` will remove formatting marks ([listed here](https://tiptap.dev/docs/editor/extensions/marks)) but not remove comment tags or comments.

_No response_

### Dependency Updates

bdbch:
## Changes Overview

Adds a `clearable` mark extension option (default `true`) and updates `unsetAllMarks` to only remove clearable marks. This fixes bulk "clear formatting" unintentionally removing semantic marks such as comments (see #6348).

Pro packages can opt out by setting `clearable: false` on comment/suggestion marks in a follow-up PR.

## Implementation Approach

- Added `clearable` to `MarkConfig`, mirroring the existing `keepOnSplit` / `splittableMarks` pattern.
- `ExtensionManager` collects mark names with `clearable: false` into `nonClearableMarks`.
- `unsetAllMarks` removes marks per type (skipping non-clearable) and returns `false` from `can()` when the selection has no clearable marks.
- `unsetMark` remains explicit and unchanged — targeted removal of a named mark type still works.

## Testing Done

- Added `packages/core/__tests__/unsetAllMarks.spec.ts` covering:
  - Clearing bold/italic via `unsetAllMarks`
  - Preserving marks with `clearable: false`
  - `can().unsetAllMarks()` behavior
  - Explicit `unsetMark` on non-clearable marks
- Ran `pnpm exec vitest run unsetAllMarks`
- Ran `pnpm -w -F @tiptap/core build`

## Verification Steps

1. Create an editor with a custom mark using `clearable: false` alongside `Bold`.
2. Select text with both marks and run `editor.commands.unsetAllMarks()`.
3. Confirm formatting is removed but the non-clearable mark remains.
4. Confirm `editor.can().unsetAllMarks()` is `false` when only non-clearable marks are selected.
5. Confirm `editor.commands.unsetMark('yourMark')` still removes the non-clearable mark when called explicitly.

## Additional Notes

- Changeset: `.changeset/2026-05-29-mark-clearable-unset-all-marks.md` (`@tiptap/core` patch)
- Docs update for `clearable` / `unsetAllMarks` in tiptap-docs can follow separately.

## AI Usage

  - Used AI to investigate the issue, design the `clearable` approach, implement core changes, write unit tests, and draft this PR description.

## Checklist

## Related Issues

Fixes #6348
Related #6398

### DIFF
--- a/packages/core/src/commands/unsetAllMarks.ts
+++ b/packages/core/src/commands/unsetAllMarks.ts
@@ -1,6 +1,3 @@
-import type { Node as ProseMirrorNode } from '@tiptap/pm/model'
-import type { SelectionRange } from '@tiptap/pm/state'
-
 import type { RawCommands } from '../types.js'
 
 declare module '@tiptap/core' {
@@ -16,27 +13,9 @@ declare module '@tiptap/core' {
   }
 }
 
-function selectionHasClearableMarks(
-  doc: ProseMirrorNode,
-  ranges: readonly SelectionRange[],
-  nonClearableMarks: string[],
-): boolean {
-  return ranges.some(({ $from, $to }) => {
-    let hasClearableMarks = false
-
-    doc.nodesBetween($from.pos, $to.pos, node => {
-      if (node.marks.some(mark => !nonClearableMarks.includes(mark.type.name))) {
-        hasClearableMarks = true
-      }
-    })
-
-    return hasClearableMarks
-  })
-}
-
 export const unsetAllMarks: RawCommands['unsetAllMarks'] =
   () =>
-  ({ tr, dispatch, editor, state }) => {
+  ({ tr, dispatch, editor }) => {
     const { selection } = tr
     const { empty, ranges } = selection
 
@@ -46,12 +25,8 @@ export const unsetAllMarks: RawCommands['unsetAllMarks'] =
 
     const { nonClearableMarks } = editor.extensionManager
 
-    if (!selectionHasClearableMarks(state.doc, ranges, nonClearableMarks)) {
-      return false
-    }
-
     if (dispatch) {
-      const clearableMarkTypes = Object.values(state.schema.marks).filter(
+      const clearableMarkTypes = Object.values(editor.schema.marks).filter(
         markType => !nonClearableMarks.includes(markType.name),
       )
 
--- a/packages/core/src/commands/unsetAllMarks.ts
+++ b/packages/core/src/commands/unsetAllMarks.ts
@@ -1,10 +1,14 @@
+import type { Node as ProseMirrorNode } from '@tiptap/pm/model'
+import type { SelectionRange } from '@tiptap/pm/state'
+
 import type { RawCommands } from '../types.js'
 
 declare module '@tiptap/core' {
   interface Commands<ReturnType> {
     unsetAllMarks: {
       /**
-       * Remove all marks in the current selection.
+       * Remove all clearable marks in the current selection.
+       * Marks with `clearable: false` are preserved
        * @example editor.commands.unsetAllMarks()
        */
       unsetAllMarks: () => ReturnType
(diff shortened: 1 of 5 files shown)


packages/core/src/commands/unsetAllMarks.ts
```diff
@@ -1,30 +1,64 @@
...
 declare module '@tiptap/core' {
   interface Commands<ReturnType> {
     unsetAllMarks: {
       /**
-       * Remove all marks in the current selection.
+       * Remove all clearable marks in the current selection.
+       * Marks with `clearable: false` are preserved
        * @example editor.commands.unsetAllMarks()
        */
       unsetAllMarks: () => ReturnType
```
arnaugomez: I wonder if there could be an option to unset unclearable marks too, that defaults to `false`

packages/core/src/commands/unsetAllMarks.ts
```diff
@@ -1,30 +1,64 @@
...
+  nonClearableMarks: string[],
+): boolean {
+  return ranges.some(({ $from, $to }) => {
+    let hasClearableMarks = false
+
+    doc.nodesBetween($from.pos, $to.pos, node => {
+      if (node.marks.some(mark => !nonClearableMarks.includes(mark.type.name))) {
+        hasClearableMarks = true
+      }
+    })
```
arnaugomez: consider returning early to break the iteration in `nodesBetween`

packages/core/src/commands/unsetAllMarks.ts
```diff
@@ -1,30 +1,64 @@
...
+      const clearableMarkTypes = Object.values(state.schema.marks).filter(
+        markType => !nonClearableMarks.includes(markType.name),
+      )
+
       ranges.forEach(range => {
-        tr.removeMark(range.$from.pos, range.$to.pos)
+        clearableMarkTypes.forEach(markType => {
+          tr.removeMark(range.$from.pos, range.$to.pos, markType)
+        })
       })
```
arnaugomez, line 40: Consider for..of loop which is faster than forEach method

arnaugomez (dismissed):
LGTM

Depending on how you look at it, this could be seen as a fix or as a regression/breaking change.

`unsetAllMarks()` now returns `false` whenever the current selection contains only `clearable: false` marks. Before this branch, the command returned `true` for any non-empty selection. That is a behavioral change, and it can break chained command flows or callers that treat `false` as a failed execution even though this is really just a no-op. 

But I think there is great value, and it is very useful for comments and track changes. So I think we should make this change, yes.

bdbch:
Thanks @arnaugomez - updated the PR.

bdbch:
Docs:
https://github.com/ueberdosis/tiptap-docs/pull/821

alexvcasillas (approved):

PR #7891, opened 2026-05-29
feat(extension-blockquote): lift and merge on backspace at child boundaries

iamgio:
## Changes Overview

Make Backspace at the start of a blockquote child interactively split or merge the blockquote, matching the affordance Enter on a blank child already provides.

Before:

- Inside `<blockquote><p>A</p><p>B</p></blockquote>` with the caret before `B`, Backspace did nothing useful — `joinBackward` stayed inert.
- With `B` standing as a top-level paragraph after a blockquote, Backspace silently wrapped `B` back inside the blockquote as another child paragraph, instead of merging into the existing one.

https://github.com/user-attachments/assets/1a2864b6-8b76-422f-83d3-8ccc72371ddd

After:

- Inside a blockquote, Backspace at the start of a non-first child lifts that child out, splitting the blockquote around it.
- At the start of a textblock whose previous sibling is a blockquote (with a textblock as its last child), Backspace merges the textblock's inline content into the blockquote's last textblock.

The first-child-of-a-blockquote case is intentionally left to the default `joinBackward` so the existing "pull the blockquote apart from the top" interaction is preserved.

https://github.com/user-attachments/assets/e7894778-76a9-4f30-b9da-bfa12d3e6644

## Implementation Approach

`packages/extension-blockquote/src/blockquote.tsx` wires a `Backspace` keymap on the Blockquote node. The handler:

1. Bails on non-empty selections or any cursor not at `parentOffset === 0`.
2. If the cursor's parent textblock is a non-first child of a blockquote → calls `editor.commands.lift('blockquote')`. ProseMirror's lift splits the blockquote naturally because the lifted block sits between siblings.
3. Otherwise, if the previous sibling is a blockquote whose last child is a textblock → runs a small custom transaction:
   - `tr.delete(blockStart, blockEnd)` removes the current paragraph.
   - `tr.insert(targetPos, $from.parent.content)` appends its inline content onto the end of the blockquote's last textblock (`targetPos = blockStart - 2`, stepping past the textblock & blockquote closing tags).
   - Selection lands at `targetPos`, at the seam between the original and merged content.

Deleting before inserting (since `blockStart > targetPos`) keeps `targetPos` valid without going through `tr.mapping.map`.

This sidesteps `joinBackward`'s `deleteBarrier` path, which would otherwise hit case 1 (wrap-and-move) and pull the paragraph back into the blockquote.

## Testing Done

Added four Playwright tests in `demos/src/Nodes/Blockquote/index.spec.ts`, exercised against both the React and Vue demos:

- Backspace at start of non-first child → `<blockquote><p>A</p></blockquote><p>B</p>`.
- Backspace at start of a middle child → `<blockquote><p>A</p></blockquote><p>B</p><blockquote><p>C</p></blockquote>` (split around).
- Backspace at start of a top-level paragraph following a blockquote → `<blockquote><p>AB</p></blockquote>` (merge).
- Two backspaces in a row (lift then merge) → `<blockquote><p>AB</p></blockquote>`.

`pnpm test:e2e --grep "Blockquote"` → `27 passed`.

## Verification Steps

1. `pnpm install`
2. `pnpm dev`
3. Open the Blockquote demo (React or Vue).
4. Type `> A`, press Enter, type `B` — you get a blockquote with two paragraphs.
5. Place the caret before `B` and press Backspace — `B` lifts out of the blockquote.
6. Press Backspace again — `B` merges into the blockquote's last paragraph.

## Additional Notes

None.

## Checklist


packages/extension-blockquote/src/blockquote.tsx
```diff
@@ -126,6 +127,45 @@ export const Blockquote = Node.create<BlockquoteOptions>({
   addKeyboardShortcuts() {
     return {
       'Mod-Shift-b': () => this.editor.commands.toggleBlockquote(),
+      Backspace: () => {
+        const { state, view } = this.editor
```
bdbch: I think extracting this into it's own function with a declarative name would be a good improvement.

bdbch (changes requested):
Looks good to me and a nice improvement - however I think moving the blockquote keyhandler function into it's own function would be a good improvement.

bdbch (approved):

PR #7893, opened 2026-05-30
feat(extension-list-keymap): sink paragraph into previous list item on tab

iamgio:
## Changes Overview

> **Note:** This PR is stacked on top of #7892. Once #7892 merges, the diff will narrow to just the actual Tab commit.

Add a `Tab` shortcut to `ListKeymap` that sinks a top-level textblock into the previous list's last item. Pressing Tab at the start of a paragraph immediately following a bullet/ordered/task list moves the paragraph (and its inline content) inside the last list item, where it becomes an additional block child. The handler stays out of the way when the cursor is already inside a list item (so `sinkListItem`'s nesting behavior is preserved), when there is no list before the paragraph, or when the caret is mid-textblock.

**Use case:** a user types a continuation under a list item that should belong inside it. Today the paragraph stays at the top level and `ListKeymap` makes it impossible to nest.

### Round-trip example

```
Before:                          After Tab:
<ul><li>A</li></ul>             <ul><li></ul>
<p>|</p>                          <p>A</p>
                                  <p>|</p>
                                </li></ul>
```

https://github.com/user-attachments/assets/e321b70e-a9e1-4a83-a0c9-abba53eb3611

## Implementation Approach

New helper `listHelpers/handleTab.ts`:

```ts
const handleTab = (editor, name, parentListTypes) => {
  // bail if selection isn't empty, cursor isn't at parentOffset 0,
  // cursor is already inside a list item, the previous sibling isn't a
  // list, or the list's last child isn't an item of the expected type
  ...
  const block = $from.parent
  const blockStart = $from.before()
  const blockEnd = $from.after()
  const insideLastItemEnd = blockStart - 2

  const tr = state.tr
  tr.delete(blockStart, blockEnd).insert(insideLastItemEnd, Fragment.from(block))
  tr.setSelection(TextSelection.create(tr.doc, insideLastItemEnd + 1))
  view.dispatch(tr.scrollIntoView())
  return true
}
```

`insideLastItemEnd = blockStart - 2` lands inside the previous list's last item at its end (one position back for the list's closing token, one more for the last item's closing token). Same PM-coordinate reasoning as the merge step in the blockquote backspace PR (#7891).

`handleBackspace` gains a guard that checks whether the cursor's textblock is the first child of its list item ancestor; if not, return `false` and let PM handle it.

The `ListKeymap` extension registers `Tab` alongside the existing Backspace / Delete handlers. `ListItem.Tab` (which calls `sinkListItem`) continues to win when the cursor is already inside a list item, because `handleTab` bails in that case.

## Testing Done

Added Playwright tests in `demos/src/Extensions/ListKeymap/index.spec.ts`:

- Tab in a paragraph after a list sinks it into the last item.
- Tab in an empty paragraph sinks the empty paragraph in.
- Tab targets the last item of a multi-item list.
- Tab in a paragraph not preceded by a list does nothing.
- Tab in the first paragraph of the document does nothing.
- Tab from mid-paragraph does nothing.
- Tab inside a non-first list item still nests via `sinkListItem` (regression check for `ListItem.Tab`).
- Backspace at the start of a sunken paragraph merges into the previous textblock.
- Tab then Backspace round-trips an empty paragraph back to the original state.

All existing list specs (BulletList, OrderedList, TaskList) continue to pass; 56 chromium tests pass overall.

`pnpm check` (lint + format) is clean.

## Verification Steps

1. `pnpm install`
2. `pnpm test:e2e --grep "ListKeymap|BulletList|OrderedList|TaskList"`
3. Manually in the ListKeymap demo at `/src/Extensions/ListKeymap/React/`: set content to `<ul><li>A</li></ul><p></p>`, focus the empty paragraph, press Tab. The paragraph sinks inside the last list item. Press Backspace. The state returns to `<ul><li>A</li></ul>`.

## Additional Notes

- The Backspace guard in `handleBackspace` is a real bug fix that would also matter on `main`. I bundled it here because the Tab feature is the most obvious way for users to encounter it.
- Out of scope here, but related: there are still some brittle position offsets in `hasListBefore` / `hasListItemBefore`. Happy to follow up in a separate PR.

## Checklist


.changeset/2026-05-29-list-keymap-backspace-lift-first.md
```diff
@@ -0,0 +1,5 @@
+---
+'@tiptap/extension-list': minor
+---
+
+`ListKeymap`'s Backspace handler now lifts the current list item before merging. At the start of a non-first list item, the item is lifted out of its wrapping list (splitting the list around it) instead of immediately joining its content into the previous item. A second Backspace then hits the existing "paragraph after a list" branch and merges the lifted textblock's content into the previous list's last item. Mirrors the two-step behavior introduced for blockquote in #7891.
```
bdbch: is this changeset correct in here since this is the PR revolving sinking paragraphs into list-items?

demos/src/Extensions/ListKeymap/index.spec.ts
```diff
@@ -0,0 +1,173 @@
...
+        page,
+      }) => {
+        const editor = await getEditor(page)
+        await placeCaretBefore(editor, '<ul><li><p>A</p></li><li><p>B</p></li></ul>', 'B')
+        await editor.press('Backspace')
+        const html = await editor.evaluate((el: any) => el.editor.getHTML())
+        expect(html).toBe('<ul><li><p>A</p></li></ul><p>B</p>')
+      })
+
+      test('backspace at the start of a middle item splits the list around it', async ({
```
bdbch, line 70: I think you mixed up your other PR into this one, could it be? I think this is also rather lift-behavior which doesn't fit your PR description

packages/extension-list/src/keymap/listHelpers/handleTab.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  const { $from } = selection
+  if ($from.parentOffset !== 0) return false
+
+  // Bail when the cursor is already inside a list item. ListItem and TaskItem
+  // own Tab themselves (sinkListItem) and we should not double-handle it.
+  if (isNodeActive(state, name)) return false
+
+  const parentDepth = $from.depth - 1
+  const parent = $from.node(parentDepth)
```
bdbch: I've seen this pattern in your PRs a few times - is it worth moving this into a reusable utility?

bdbch (changes requested):
LGTM - however there are some changes in here that I think were meant to be in another PR?

iamgio:
@bdbch yes, it's written in the blockquote at the beginning. These changes were dependent on those of my previous PR. It'll look correct once the other PR is merged

bdbch:
Alright - lets get your blockquote PR merged first then.

iamgio:
@bdbch diff is clean and added a reusable `getPreviousBlockSibling()` function. I'll let #7891 adopt it once merged

bdbch (changes requested):
Sorry for coming back again - some small nitpicks:

1. Could you implement a way to check if the sank content actually fits into the schema? otherwise this could throw an error when the content won't fit the list schema
2. Can you add a test for the new getPreviousBlockSibling util?

Otherwise this looks fine.

iamgio:
Hi @bdbch, sorry for pushing but could I get another pass?

iamgio:
pinging; I applied all the changes as asked, and this is nice-to-have for #7892 -- @bdbch @arnaugomez

packages/extension-list/src/keymap/listHelpers/handleTab.ts
```diff
@@ -0,0 +1,47 @@
...
+import { Fragment } from '@tiptap/pm/model'
+import { TextSelection } from '@tiptap/pm/state'
+
+export const handleTab = (editor: Editor, name: string, parentListTypes: string[]) => {
+  const { state, view } = editor
+  const { selection } = state
+  if (!selection.empty) return false
+
+  const { $from } = selection
+  if ($from.parentOffset !== 0) return false
```
arnaugomez, line 12: Blocking: `handleTab` has no textblock guard, so a GapCursor can trigger it with a non-textblock parent.

`selection.empty` and `parentOffset === 0` are both satisfiable by a GapCursor, e.g. `<ul><li><p>A</p></li></ul><blockquote><hr></blockquote>` with the gap cursor before the `<hr>`: `$from.parent` is the blockquote, `parentOffset` is 0, the previous sibling is the list, and `listItem` accepts a blockquote as a non-first child, so `canReplace` passes. 

Tab would then sink the entire blockquote and `TextSelection.create(tr.doc, insideLastItemEnd + 1)` points at a non-inline position (before the `<hr>`), which throws a RangeError at runtime. 

One-line fix: bail unless `selection instanceof TextSelection` (or `$from.parent.isTextblock`).

packages/extension-list/src/keymap/listHelpers/handleTab.ts
```diff
@@ -0,0 +1,47 @@
...
+  // its end (one for the closing token of the list, one for the closing
+  // token of the last item).
+  const insideLastItemEnd = blockStart - 2
+
+  const tr = state.tr
+  tr.delete(blockStart, blockEnd).insert(insideLastItemEnd, Fragment.from(block))
+  // Cursor lands right inside the inserted block at its start: one position
+  // past the insertion point steps over the block's opening token.
+  tr.setSelection(TextSelection.create(tr.doc, insideLastItemEnd + 1))
+  view.dispatch(tr.scrollIntoView())
```
arnaugomez: Prefer using an API like editor.commands.command() to run commands instead of view.dispatch, which is a lower-level ProseMirror API.

packages/core/src/helpers/getPreviousBlockSibling.ts
```diff
@@ -0,0 +1,34 @@
+import type { Node, ResolvedPos } from '@tiptap/pm/model'
+
+/**
+ * Returns the block-level sibling immediately before the cursor's textblock
+ * (or null when the cursor is at the first child of its block parent).
```
arnaugomez, line 5: getPreviousBlockSibling's docstring says "cursor's textblock" though the function is defined for any resolved position — its semantics at non-textblock positions are subtle enough to merit a sentence in the JSDoc, since it's now public core API.

packages/extension-list/src/keymap/list-keymap.ts
```diff
@@ -99,6 +99,21 @@ export const ListKeymap = Extension.create<ListKeymapOptions>({
...
+        this.options.listTypes.forEach(({ itemName, wrapperNames }) => {
+          if (editor.state.schema.nodes[itemName] === undefined) {
+            return
+          }
+
+          if (handleTab(editor, itemName, wrapperNames)) {
+            handled = true
+          }
+        })
+
```
arnaugomez, line 115: Blocking: The `Tab` keymap loop doesn't short-circuit, so one keypress can sink twice. The `forEach` over `listTypes` continues after `handled = true`, and `handleTab` re-reads `editor.state` fresh each call. If a bullet list's last item ends with a taskList — `<ul><li><p>A</p><taskList>…</taskList></li></ul><p>|B</p>` — the `listItem` pass sinks B into the `<li>`, after which the cursor sits at the start of B with a `taskList` as its previous sibling, so the `taskItem` pass matches and sinks B again into the taskItem. Contrived content, but valid and reachable. The existing Backspace/Delete handlers share the pattern, but Tab is the one where the handler's own output re-satisfies the next handler's preconditions. Fix by returning early once handled (`.some()` or a `handled` check at the top of the loop body).

arnaugomez: Prefer for..of loop rather than forEach method.

arnaugomez (dismissed):
Found 2 blocking bugs. 

Thanks for your patience 🙏

bdbch:
@coderabbitai review

packages/extension-list/src/keymap/listHelpers/handleBackspace.ts
```diff
@@ -59,6 +59,16 @@ export const handleBackspace = (editor: Editor, name: string, parentListTypes: s
     return false
   }
 
+  // Only intercept at the very start of the list item (its first child).
+  // For a cursor at the start of a later child (e.g. a second paragraph
+  // inside the same item), let joinBackward merge upward within the item
+  // instead of lifting the whole item out.
```
bdbch: I think we could minimize those comments - 4 lines for saying that a backspace is only handled at the beginning of a list item feels pretty verbose

bdbch (dismissed):
Looks good to me - one minor comment but not a blocker

iamgio:
Sorry, agents do get needlessly verbose indeed 🤦‍♂️ 

I've compacted comments and docs -- @arnaugomez I'll need your review in order to merge, or maybe @bdbch you can merge for me if you get a chance. Thanks!

iamgio:
fyi: seems like `[chromium] › demos/src/Examples/Default/index.spec.ts:42:7 › /src/Examples/Default › Svelte › should apply the paragraph style when the keyboard shortcut is pressed` is a flaky test that's been failing often

iamgio:
@arnaugomez 🙏

bdbch (approved):

PR #7898, opened 2026-05-30
fix(markdown): add backslash-escape handling in parser and serializer

### LINKED ISSUE #7258: Markdown extension misinterprets some escaped characters as formatting
### Affected Packages

@tiptap/markdown

### Version(s)

3.10.7

## Issue

When using a Markdown editor, if a user inputs two escape characters, the editor appears to interpret this as a request to format any text in between those two escape characters. It is fairly easy to replicate this issue using the [live demo](https://tiptap.dev/docs/editor/markdown) provided on Tiptap Docs.

As an aside, it looks like the Markdown extension only interprets double-backlashes as escapes. Markdown characters with single backslashes, such as `\*`, do not appear at all when parsed. This may be intentional.

## Example

**Input:** `\\*This sentence is surrounded with escaped asterisks.\\*`

**Expected Output:** \*This sentence is surrounded with escaped asterisks.\*

**Actual Output:** _This sentence is surrounded with escaped asterisks._

I do not encounter this issue if the escaped character cannot be "paired" with another escape character. For example, `\\*text` will render as: *text.  

### Browser Used

Chrome

### Code Example URL

_No response_

### Expected Behavior

All escaped characters are correctly interpreted and displayed by the Markdown editor.

_No response_

### Dependency Updates

bdbch:
## Changes Overview

Fix backslash-escape handling in `@tiptap/markdown`. Backslash-escaped markdown characters (e.g. `\*`, `\_`, `\\`) were silently dropped during parsing, and markdown-significant characters in text nodes were not escaped during serialization.

## Implementation Approach

Two-pronged fix in `MarkdownManager.ts`:

**Parsing (Markdown → ProseMirror):** Added handling for marked.js `escape` tokens in both `parseInlineTokens` and `parseFallbackToken`. The `escape` token's `text` property (the escaped character) is used directly to create a text node. Adjacent text nodes are merged to match ProseMirror's expectations.

**Serialization (ProseMirror → Markdown):** Added `escapeMarkdownSyntax()` method that backslash-escapes markdown-significant characters (`\`, `` ` ``, `*`, `_`, `[`, `]`, `~`) in text nodes during serialization. This runs after HTML entity encoding and respects the existing `isInsideCode` guard so code blocks/marks remain untouched.

The solution is generic — it works with all extensions since escapes are handled at the manager level, not per-extension.

## Testing Done

- Added 18 new test cases covering: parsing various escaped chars, mixed escaped + real formatting, serialization escaping for all 7 chars, round-trip stability, and code block/mark exemption
- All 1090 tests pass (119 test files)
- Lint and build pass cleanly

## Verification Steps

1. Enable the `@tiptap/markdown` extension in the demo
2. Type `\*text\*` → should display literal `*text*`, not italic
3. Type `\\` → should display a single backslash
4. Type `\_text\_` → should display literal `_text_`, not italic
5. Verify round-trip: the markdown output re-parses correctly

## Additional Notes

N/A

## AI Usage

  - AI assisted with code implementation and test scaffolding
  - AI generated this PR description

## Checklist

## Related Issues

Closes #7258


packages/markdown/src/MarkdownManager.ts
```diff
@@ -814,6 +820,26 @@ export class MarkdownManager {
...
+    // lives in a single text node.
+    for (let i = result.length - 1; i > 0; i -= 1) {
+      const current = result[i]
+      const previous = result[i - 1]
+
+      if (current.type === 'text' && previous.type === 'text') {
+        const currentMarks = current.marks || []
+        const previousMarks = previous.marks || []
+
+        if (JSON.stringify(currentMarks) === JSON.stringify(previousMarks)) {
```
alexvcasillas: Isn't better to do object equality check instead of string checks? This can fail for logically equal objects with different key order 🤔  also it allocates much more memory due to string serialization. Is there a reason we can't go with something like `lodash/isEqual`, or `fast-deep-equal`, or similar instead?

alexvcasillas (dismissed):
Looks good 👍🏻 Just dropped a small question about the `JSON.stringify` equal comparison

packages/markdown/src/MarkdownManager.ts
bdbch: I think performance wise this should be fine because this is not run very fast / often - but I agree we should either sort the keys before. Pulling in a dependency for a deep equal feels a bit overkill for a simple mark comparision here

bdbch: I added a new `marksEqual` utility for cases like that.

StevenVanDijk:
Thank you for this fix! Native `escape` token handling closes the round-trip gap for the 7 characters covered by `escapeMarkdownSyntax` (`\`, `` ` ``, `*`, `_`, `[`, `]`, `~`).

However, the [CommonMark spec](https://spec.commonmark.org/0.31.2/#backslash-escapes) lists 18 escapable characters (`/([\\` `` ` ``*_{}\[\]()<>#\+\-\.!\|])/g`) in total. The current regex `/([\\` `` ` ``*_[\]~])/g` leaves 11 of them unhandled.

I have a use case where my document contained `\<class\>`, marked parses both `\<` and `\>` as `escape` tokens, so `token.text` is `<` and `>` respectively. The serializer then outputs `<class>` — which marked re-parses as an HTML tag.

Was there a specific reason to limit the regex to those 7 characters? Extending it to the full CommonMark set would close all 18 cases.

bdbch:
Hey @StevenVanDijk I tried to extend the list of escapable characters but that went to broad and cause a lot of regressions across a lot of tests. I think we need to implement context-aware escape character handling to not regress half the codebase haha.

PR #7899, opened 2026-05-30
fix: prevent cursor jump during IME composition in colored text

semimikoh:
## Problem

When typing with an IME (Korean, Japanese, Chinese) inside text that has a color mark applied, the composition is interrupted and the cursor jumps. This happens because color attribute values are re-serialized inconsistently (e.g. `#FF0000` vs `rgb(255, 0, 0)` vs `#ff0000`), causing ProseMirror to see a mark change mid-composition and reset the selection.

## Solution

Normalize color attribute values at the extension boundary so the same color always serializes to the same canonical string. This prevents spurious mark diffs during IME composition.

The normalization runs in two places via a ProseMirror plugin:
- `view()` — normalizes initial JSON/HTML content on editor mount
- `appendTransaction` — uses `findDiffStart`/`findDiffEnd` to compute the changed range and only walks those nodes (avoids full-document scans)

## Testing

- Added regression tests for IME composition with colored text
- Added a test case for programmatic/collaborative (Yjs) updates being normalized via `appendTransaction`

## Related

Re-submission of #7626, which was auto-closed when I accidentally deleted my fork during cleanup. The commits are identical (the `appendTransaction` refactor suggested by @bdbch is included).


bdbch (dismissed):
Looks good to me! Two things:

**1. Rebase**

I think this needs a rebase to main - there are some merge conflicts

**2. Make this opt-in and customizable**

I'm concerned about this technically being a breaking change for a small subset of users because it will change how color values will be retrieved and stored from the editor. For some users, the exact color property may be important.

I think it makes sense to make the normalization an opt-in option on the extension and also defining the normalization format so people can pick what colors should be normalized to. What do you think?

---

Also if we add new options we will need new docs over on [ueberdosis/tiptap-docs](https://github.com/ueberdosis/tiptap-docs)

semimikoh:
Thanks for the feedback! I’ve addressed the requested changes in `a054fecfa` (`feat(extension-text-style): make color normalization opt-in`).

What changed:
- Added an opt-in `colorNormalizer` callback to both `Color` and `BackgroundColor`; it defaults to `null`, so existing color values remain unchanged unless explicitly enabled.
- Exported the built-in `normalizeColor` helper for browser-canonical normalization, while allowing custom callbacks for formats such as hex.
- Applied the configured normalizer consistently across HTML parsing/rendering, commands, initial JSON content, and programmatic/collaborative updates.
- Only registers the ProseMirror normalization plugin when normalization is enabled.
- Rebased onto the latest `main` and resolved the text-style conflicts while preserving the newer `getStyleProperty` behavior.
- Fixed the async initial normalization to build transactions from the latest editor state, added timer cleanup on plugin destruction, and added a regression test covering simultaneous color/background-color normalization.
- Updated the changeset for the new opt-in API.

Validation:
- lint and formatting pass
- `@tiptap/extension-text-style` build passes
- extension-text-style unit tests pass (54 tests across 14 files)

Documentation PR: https://github.com/ueberdosis/tiptap-docs/pull/985

semimikoh:
@bdbch The requested opt-in and customizable color normalization changes are now implemented, and the related documentation PR is open at https://github.com/ueberdosis/tiptap-docs/pull/985. When you have a chance, could you please take another look? Thanks!

PR #7903, opened 2026-06-01
fix(static-renderer): type JSON renderers so JSONContent works without casts

### LINKED ISSUE #7175: renderJSONContentToReactElement Type mismatch
### Affected Packages

@tiptap/static-renderer

### Version(s)

3.10.2

### Bug Description

In node mapping, there is no `node.text` inside the text node. Also, every prop of the node, e.g., type, is `any`.
`Property 'text' does not exist on type 'NodeType'.ts(2339)`

There is a type error if you render the element and pass `content` as props to it.
`Type 'JSONContent' is not assignable to type 'NodeType'.
  Property 'type' is optional in type 'JSONContent' but required in type 'NodeType'.ts(2322)
index.d.ts(186, 5): The expected type comes from property 'content' which is declared here on type '{ content: NodeType; parent?: NodeType | undefined; }'`

I assume the renderJSONContentToReactElement input is also typed as `JSONContent`.

### Browser Used

Chrome

### Code Example URL

https://codesandbox.io/p/devbox/jdz2vg

### Expected Behavior

Look at the type highlighting in the codesandbox at `/src/index.ts`.

There isn't much documentation on that useful function. I don't want to import the PM static renderers to save on bundle size, so I hope to find help here. https://tiptap.dev/docs/editor/api/utilities/static-renderer#technical-details

_No response_

### Dependency Updates

ozdemircibaris:
## Changes Overview

`renderJSONContentToReactElement` and `renderJSONContentToString` defaulted their node/mark generics to the ProseMirror-oriented `NodeType`/`MarkType`. As a result, in JSON node mappings `node.text` errored, every node prop was typed `any`, and passing a value typed as `JSONContent` failed (`type` is optional in `JSONContent` but was required). They now default to the newly exported `JSONNodeType`/`JSONMarkType`, so `node.text` is typed `string | undefined` and a `JSONContent` document is accepted directly.

## Implementation Approach

Added `JSONNodeType<TMark>` and `JSONMarkType` to `json/renderer.ts` and used them as the generic defaults in both JSON wrappers. `JSONNodeType` keeps `marks` tied to the mark type parameter so the node↔mark relationship stays sound — a plain `JSONContent` default hits a `readonly TMarkType[]` bivariance error. The node `type` constraint was relaxed to optional, and node-type resolution now tolerates a missing `type` (falling back to the normal "missing handler" error instead of a raw TypeError). The ProseMirror `pm/*` renderers are unaffected since they parameterize explicitly with `Mark`/`Node`.

## Testing Done

`pnpm -F @tiptap/static-renderer build` (incl. DTS), `pnpm test:unit` (34/34 pass), `pnpm lint`, and `pnpm format` all pass. The three existing tests that cast via `as unknown as TextType` were simplified to read `node.text` directly (a regression signal), and two tests were added covering the missing-`type` behavior.

## Verification Steps

In a consumer `.ts` file: `const render = renderJSONContentToString({ nodeMapping: { text: ({ node }) => node.text ?? '' }, markMapping: {} })` then `render({ content })` with `content: JSONContent`. This type-checks with the fix and errors without it (TS2339 on `node.text`, TS2322 on `content`).

## Additional Notes

New exported types `JSONNodeType` and `JSONMarkType` are additive to the package's public API.

## AI Usage

  - I used Claude to investigate and reproduce the issue, design and adversarially review the type fix, implement it, and run the validation (build/tests/lint/format).

## Checklist

## Related Issues

Closes #7175

### DIFF
--- a/.changeset/chatty-rules-occur.md
+++ b/.changeset/chatty-rules-occur.md
@@ -1,0 +1,5 @@
+---
+'@tiptap/static-renderer': patch
+---
+
+Fix the types of `renderJSONContentToReactElement` and `renderJSONContentToString`. They defaulted their node and mark generics to the ProseMirror-oriented `NodeType`/`MarkType`, so `node.text` did not exist in a node mapping, every node prop was typed `any`, and passing a value typed as `JSONContent` failed (`type` is optional in `JSONContent` but was required). They now default to the newly exported `JSONNodeType`/`JSONMarkType`, so `node.text` is typed `string | undefined` and a `JSONContent` document is accepted directly. The ProseMirror `pm/*` renderers are unaffected, since they parameterize explicitly with `Mark`/`Node`.
--- a/.changeset/chatty-rules-occur.md
+++ b/.changeset/chatty-rules-occur.md
@@ -2,4 +2,4 @@
 '@tiptap/static-renderer': patch
 ---
 
-Fix the types of `renderJSONContentToReactElement` and `renderJSONContentToString`. They defaulted their node and mark generics to the ProseMirror-oriented `NodeType`/`MarkType`, so `node.text` did not exist in a node mapping, every node prop was typed `any`, and passing a value typed as `JSONContent` failed (`type` is optional in `JSONContent` but was required). They now default to the newly exported `JSONNodeType`/`JSONMarkType`, so `node.text` is typed `string | undefined` and a `JSONContent` document is accepted directly. The ProseMirror `pm/*` renderers are unaffected, since they parameterize explicitly with `Mark`/`Node`.
+Fix the types of the JSON static renderers (`renderJSONContentToReactElement` and `renderJSONContentToString`): you can now pass `JSONContent` directly and read node fields like `node.text` in your mappings without type errors or casts.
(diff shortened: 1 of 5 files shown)


.changeset/chatty-rules-occur.md
```diff
@@ -0,0 +1,5 @@
+---
+'@tiptap/static-renderer': patch
+---
+
+Fix the types of `renderJSONContentToReactElement` and `renderJSONContentToString`. They defaulted their node and mark generics to the ProseMirror-oriented `NodeType`/`MarkType`, so `node.text` did not exist in a node mapping, every node prop was typed `any`, and passing a value typed as `JSONContent` failed (`type` is optional in `JSONContent` but was required). They now default to the newly exported `JSONNodeType`/`JSONMarkType`, so `node.text` is typed `string | undefined` and a `JSONContent` document is accepted directly. The ProseMirror `pm/*` renderers are unaffected, since they parameterize explicitly with `Mark`/`Node`.
```
bdbch: It's a quite long changeset - I would reduce it to something that describes quickly to the user what changed for them. I don't think they care about what exact types were changed around but rather what improved/changed for them directly.

bdbch (changes requested):

bdbch (changes requested):
invalid

bdbch (approved):
lgtm, thanks!

PR #7906, opened 2026-06-01
Bump prosemirror-tables so that findTable is always available on @tiptap/pm

arnaugomez:
## Changes Overview

Bump `prosemirror-tables` in `@tiptap/pm` and the demos workspace so `findTable` is available, and align the related ProseMirror dependencies required by the newer tables package.

## Implementation Approach

Updated the dependency pins owned by `@tiptap/pm`, mirrored them in the demos workspace, refreshed the lockfile, and added a patch changeset for `@tiptap/pm`.

## Testing Done

I also had to update a few test expectations because the newer dependency versions serialize HTML a little differently. The editor still behaves the same, but the exact strings in the tests changed.

For example, `text-align: left` is now serialized as `text-align: left;`, so the TextAlign demo tests now expect `<p style="text-align: left;">...` instead of `<p style="text-align: left">...`.

The same thing happened for table markup and text-style commands: styles such as `min-width`, `color`, `background-color`, `font-family`, `font-size`, and `line-height` now include the trailing semicolon in the generated HTML, so those assertions were updated to match the new output.

One whitespace case also changed: when preserving whitespace during `insertContent`, the content is now rendered with `<br>` instead of a doubled space in the HTML output, so that expectation was updated too.

The CustomDocument demo test also needed a small interaction fix. The test was clicking the editor wrapper before pressing `End` and `Enter`; under parallel Playwright runs that did not always put the caret inside the heading, so the generated paragraph existed but typing did not land in it. The test now clicks the heading itself, then presses `End` and `Enter`, and types through the current keyboard focus. That keeps the test aligned with the user interaction it is checking: pressing Enter after the heading should move into the paragraph.

## Verification Steps

Build or import `@tiptap/pm` and confirm `findTable` resolves from `prosemirror-tables`.

## Additional Notes

The lockfile update is included in the branch as requested.

## AI Usage

  - I used AI to draft the PR body and summarize the dependency bump.
  - I used AI to investigate and fix the build failure caused by mismatched ProseMirror dependency versions.

## Checklist

## Related Issues

Closes TT-497


arnaugomez:
Unfortunately the version bump breaks the tests because of how HTML is generated in the new versions of ProseMirror:
```
    Expected: "<p style=\"text-align: right\">Example Text</p>"
    Received: "<p style=\"text-align: right;\">Example Text</p>"
```

bdbch:
> For example, text-align: left is now serialized as text-align: left;, so the TextAlign demo tests now expect <p style="text-align: left;">... instead of <p style="text-align: left">....

Is there a way to inspect HTML or the computed style object via playwright? Maybe that's safer than using the actual style tag to check for it? That way we don't need to worry about it breaking in the future again. :)

bdbch (dismissed):

arnaugomez:
> For example, `text-align: left` is now serialized as `text-align: left;`, so the TextAlign demo tests now expect `<p style="text-align: left;">...` instead of `<p style="text-align: left">...`.

Yes. you can assert computed CSS:
```ts
const paragraph = page.locator('p').filter({ hasText: '...' })

await expect(paragraph).toHaveCSS('text-align', 'left')

```

bdbch:
Nice, thanks!

PR #7915, opened 2026-06-03
refactor(extensions): refactor the clear selection logic on blur

bdbch:
## Changes Overview

Fixes a bug where the native browser selection could remain visible (and sometimes extend beyond the actual range) after blur when using the `Selection` extension alongside global `::selection` styles. The extension now clears the native selection on blur and restores it on focus, so only the `.selection` decoration is shown while the editor is unfocused.

Replaces the previous approach of injecting CSS to hide `::selection` while blurred.

## Implementation Approach

- On **blur**, when the editor has a non-empty text selection and is editable, call `window.getSelection()?.removeAllRanges()` so global `::selection` rules no longer paint a ghost highlight on top of the decoration spans.
- On **focus**, restore the DOM selection from ProseMirror state via `view.focus()` (same mechanism as `EditorView#focus` / `selectionToDOM`), deferred with `requestAnimationFrame`.
- Introduced `shouldSyncDomSelection()` shared by blur/focus handlers (non-empty, not a node selection, editor is editable). `shouldPreserveSelection()` builds on that for decoration rendering (also requires blurred and not dragging).
- Removed the injected stylesheet (`.ProseMirror:not(.ProseMirror-focused) *::selection { … }`) since clearing/restoring the native selection addresses the root cause more reliably when apps use strong `::selection` styles.

## Testing Done

- Added `packages/extensions/__tests__/selection.spec.ts`:
  - Clears native selection on blur (editable editor)
  - Renders `.selection` decorations when blurred with a non-empty selection
  - Skips clear/restore when `editable: false`
- Ran `pnpm test:unit -- packages/extensions/__tests__/selection.spec.ts` (all tests passing)

## Verification Steps

1. Open the Selection demo (`Extensions/Selection/React` or `Vue`).
2. Style native selection globally, e.g. `*::selection { background: purple; }`.
3. Select text across multiple lines or up to a link, then blur the editor (click outside).
4. Confirm only `.selection` decoration highlights appear — no extra native highlight past the intended range.
5. Focus the editor again and confirm the native text selection/caret is restored.
6. Optional: set `editable: false` and repeat — native selection should not be cleared or forcibly restored by the extension.

## Additional Notes

- `removeAllRanges()` is global (same as `editor.commands.blur()`). We only invoke it when the editor has a non-empty ProseMirror text selection and is editable, which matches when decorations are shown.
- ProseMirror does not expand `state.selection` on blur; the ghost was leftover native DOM selection, not incorrect document positions.

## AI Usage

  - Investigated ProseMirror blur/selection sync behavior and root cause of the ghost selection
  - Wrote unit tests and this PR description

## Checklist


packages/extensions/src/selection/selection.ts
```diff
@@ -20,7 +13,44 @@ export type SelectionOptions = {
...
+function shouldPreserveSelection(
+  state: EditorState,
+  editor: {
+    isEditable: boolean
+    isFocused: boolean
+    view: EditorView
+  },
+): boolean {
+  return shouldSyncDomSelection(state, editor) && !editor.isFocused && !editor.view.dragging
+}
```
arnaugomez: Perhaps the parameters can be simplified by removing state (can be obtained with `editor.state`) and using the `Editor` type for the type of the editor.

Or if the EditorState can't be obtained with `editor.state`, explain why.

packages/extensions/src/selection/selection.ts
```diff
@@ -20,7 +13,44 @@ export type SelectionOptions = {
 }
 
 /**
- * This extension allows you to add a class to the selected text.
+ * Checks if the selection should be preserved when the editor is blurred.
+ * @param state The editor state.
+ * @param editor The editor instance.
+ * @returns `true` if the selection should be preserved when the editor is blurred.
+ */
+function shouldSyncDomSelection(
```
arnaugomez: The comment says "Checks if the selection should be preserved" but the function is called `shouldSyncDomSelection`. Meanwhile, there is a `shouldPreserveSelection` function with no comment

arnaugomez (dismissed):

packages/extensions/src/selection/selection.ts
bdbch: We could take the state from editor - however I believe it's clearer to pass an explicit state through here to make sure we actually pass the correct state. It allows for more control & is more clear.

packages/extensions/src/selection/selection.ts
```diff
@@ -20,7 +13,34 @@ export type SelectionOptions = {
...
+}
+
+function clearDomSelection() {
+  window.getSelection()?.removeAllRanges()
+}
+
+function restoreDomSelection(view: EditorView) {
+  // Sync the native selection from the editor state (see prosemirror-view `EditorView#focus`).
+  view.focus()
+}
```
arnaugomez, line 41: This comment could be a documentation comment, and link to https://prosemirror.net/docs/ref/#view.EditorView.focus

.gitignore
```diff
@@ -45,6 +45,9 @@ packages/**/*.tgz
 packages-deprecated/**/*.tgz
 demos/*.tgz
 
+# pnpm
+.pnpm-store
```
arnaugomez, line 50: Is this change related to this PR?

arnaugomez (dismissed):

.gitignore
bdbch: No - the pnpm-store however sometimes is created when running the pnpm CLI. We could also move this out into a separate PR - but I think it's fine just merging it with this.

PR #7917, opened 2026-06-03
fix(core): allow ids and classes starting with a number in parseAttributes

### LINKED ISSUE #7345: parseAttributes strips ids starting with number
### Affected Packages

core, markdown

### Version(s)

3.13

### Bug Description

```js
console.log(parseAttributes('#2123 .foo'))
console.log(parseAttributes('#hello .foo'))
console.log(parseAttributes('#5hello .foo'))
```

produces

```
{ class: 'foo' }
{ class: 'foo', id: 'hello' }
{ class: 'foo' }
```

As `parseAttributes` is used by default this affects all markdown parsing more or less, at least nodes created by the official helper utils.

### Browser Used

Firefox

### Code Example URL

_No response_

### Expected Behavior

I.e. any id with a leading number gets stripped. Reading https://github.com/jgm/pandoc/issues/7920 i'm unsure what to conclude. 

1. If pandoc now allows ids to start with number, then it should be the same here
2. If not, then a warning should be added [here](https://tiptap.dev/docs/editor/markdown/api/utilities#parseattributes) and elsewhere, as this behaviour is not necessariliy known to people. 

_No response_

### Dependency Updates

tugkanpilka:
## Changes Overview

`parseAttributes` silently stripped ids and classes whose first character is a digit (e.g. `#2123` or `.2xl`):

```js
parseAttributes('#2123 .foo')  // before: { class: 'foo' } — id silently dropped
parseAttributes('#2123 .foo')  // after:  { class: 'foo', id: '2123' }
```

This PR loosens the class/id parsing regexes from `[a-zA-Z][\w-]*` (a leftover from the HTML4 id rule) to `[\w-]+`, since HTML5 allows ids to start with any non-whitespace character.

## Implementation Approach

- Updated the class and id parsing regexes in `packages/core/src/utilities/markdown/attributeUtils.ts`, including their copies in the boolean-attribute cleanup step (otherwise `#2123` would leak into the boolean-attribute stage).
- Attribute-name regexes (key-value pairs and boolean attributes) are intentionally left untouched — HTML attribute names must still start with a letter.

## Testing Done

- Added `packages/core/__tests__/parseAttributes.spec.ts`: first direct unit coverage of `parseAttributes`, written test-first (the numeric id/class cases failed before the fix, all 10 pass after). Includes characterization tests for existing behavior (classes, ids, key-value pairs, boolean attributes, quoted strings) to guard against regressions.
- Full monorepo unit suite passes: 129 files, 1173 tests.
- `pnpm lint` and `pnpm format` are clean.

## Verification Steps

1. Run `pnpm vitest run packages/core/__tests__/parseAttributes.spec.ts`
2. Or revert `attributeUtils.ts` and observe the three numeric-prefix tests fail, matching the issue's repro.

## Additional Notes

The same first-character restriction also affected classes (`.2xl` was silently dropped), so this PR fixes both, while the issue only mentions ids.

## AI Usage

  - I used Claude Code to analyze the issue, write the tests (test-first), apply the regex fix, and run the validation checklist, reviewing each step manually.

## Checklist

## Related Issues

Fixes #7345

### DIFF
--- a/.changeset/2026-06-03-parse-attributes-numeric-ids.md
+++ b/.changeset/2026-06-03-parse-attributes-numeric-ids.md
@@ -2,4 +2,4 @@
 '@tiptap/core': patch
 ---
 
-`parseAttributes` no longer silently strips ids and classes that start with a number (e.g. `#2123` or `.2xl`). HTML5 allows ids to start with any non-whitespace character, so Pandoc-style attribute strings like `{#2123 .foo}` now parse to `{ id: '2123', class: 'foo' }` instead of dropping the id.
+`parseAttributes` now supports any non-white character at the start of classes or id attributes.
--- a/.changeset/2026-06-03-parse-attributes-numeric-ids.md
+++ b/.changeset/2026-06-03-parse-attributes-numeric-ids.md
@@ -1,0 +1,5 @@
+---
+'@tiptap/core': patch
+---
+
+`parseAttributes` no longer silently strips ids and classes that start with a number (e.g. `#2123` or `.2xl`). HTML5 allows ids to start with any non-whitespace character, so Pandoc-style attribute strings like `{#2123 .foo}` now parse to `{ id: '2123', class: 'foo' }` instead of dropping the id.
--- a/.changeset/2026-06-03-parse-attributes-numeric-ids.md
+++ b/.changeset/2026-06-03-parse-attributes-numeric-ids.md
@@ -2,4 +2,4 @@
 '@tiptap/core': patch
 ---
 
-`parseAttributes` now supports any non-white character at the start of classes or id attributes.
+`parseAttributes` now supports any word characters at the start of classes or id attributes.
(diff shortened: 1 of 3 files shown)


arnaugomez (dismissed):

.changeset/2026-06-03-parse-attributes-numeric-ids.md
```diff
@@ -0,0 +1,5 @@
+---
+'@tiptap/core': patch
+---
+
+`parseAttributes` no longer silently strips ids and classes that start with a number (e.g. `#2123` or `.2xl`). HTML5 allows ids to start with any non-whitespace character, so Pandoc-style attribute strings like `{#2123 .foo}` now parse to `{ id: '2123', class: 'foo' }` instead of dropping the id.
```
bdbch:
```suggestion
`parseAttributes` now supports any non-white character at the start of classes or id attributes.
```

bdbch (dismissed):

PR #7923, opened 2026-06-07
docs: clarify common React and text-style setup

ketpatil77:
## Summary
- document the null-editor guard pattern when immediatelyRender: false is used in React
- add a short companion-extension note for richer TextStyle content

## Testing
- not run (README-only change)


arnaugomez (changes requested):
As far as I know, all the extensions' README files have the same content. Should we update it? I believe the documentation you added is already mentioned in the docs website (tiptap-docs repository), so this change is not necessary.

ketpatil77:
Thanks for the review. Since this guidance already belongs on the documentation site and package READMEs are kept consistent, I’m closing this PR rather than duplicating the content here.

PR #7925, opened 2026-06-08
fix(extension-table): sync node attribute changes to TableView DOM

matthiask:
## Changes Overview

`TableView.update()` re-rendered column widths but never re-applied `HTMLAttributes` when node attributes changed after the initial render. Attributes added via `addGlobalAttributes` and updated with `tr.setNodeAttribute()` were correctly reflected in the ProseMirror state and serialized HTML but the live `<table>` DOM element was not updated.

This is a follow-up to #7896, which fixed the constructor path (initial render) but left the update path unhandled.

## Implementation Approach

1. **`TableView.ts`** — Added an optional `getHTMLAttributes` callback as a 5th constructor parameter. When provided, `update()` calls it to obtain fresh merged `HTMLAttributes` for the updated node and syncs all non-`style` attributes to `this.table`. The `style` attribute is left to `updateColumns()` which already manages `width`/`minWidth`.

2. **`table.ts` (`addNodeView`)** — Passes a callback that recomputes the merged `HTMLAttributes` via `getRenderedAttributes(updatedNode, extensionAttributes)` merged with `this.options.HTMLAttributes`, mirroring exactly what the constructor receives at creation time.

## Testing Done

- Extended `tableHTMLAttributes.spec.ts` with 3 new tests under `Table node attribute updates (addGlobalAttributes)`:
  - reflects a class set via `setNodeAttribute` on the `<table>` DOM element
  - removes the class from the `<table>` DOM element when the attribute is cleared
  - merges a dynamic class with the static `HTMLAttributes` class
- All 1166 existing tests continue to pass.

## Verification Steps

1. `pnpm test:unit` — all tests pass
2. Create an editor with an extension that adds a `class` attribute to the `table` type via `addGlobalAttributes`, insert a table, then call `tr.setNodeAttribute(pos, 'class', 'my-class')` — the `<table>` element should immediately reflect the class in the live DOM.

## Additional Notes

The fix is backward-compatible: `getHTMLAttributes` is optional, so existing `TableView` subclasses are unaffected. The `style` attribute is deliberately excluded from syncing since `updateColumns()` manages `width`/`minWidth` on the element's style.

## AI Usage

  - Used Claude to help identify the root cause, write the fix, and write the tests.

## Checklist

## Related Issues

Follow-up to #7896.

### DIFF
--- a/.changeset/tableview-sync-node-attrs.md
+++ b/.changeset/tableview-sync-node-attrs.md
@@ -1,0 +1,5 @@
+---
+"@tiptap/extension-table": patch
+---
+
+Fix `TableView` not syncing node attribute changes (e.g. `class` set via `addGlobalAttributes`) to the `<table>` DOM element after the initial render. Previously `TableView.update()` only re-rendered column widths, so attribute changes made via `tr.setNodeAttribute()` were reflected in the ProseMirror state and serialized HTML but invisible in the live editor DOM.
--- a/packages/extension-table/__tests__/tableHTMLAttributes.spec.ts
+++ b/packages/extension-table/__tests__/tableHTMLAttributes.spec.ts
@@ -1,6 +1,6 @@
 // @vitest-environment happy-dom
 
-import { Editor } from '@tiptap/core'
+import { Editor, Extension } from '@tiptap/core'
 import Document from '@tiptap/extension-document'
 import Paragraph from '@tiptap/extension-paragraph'
 import { Table, TableCell, TableHeader, TableRow } from '@tiptap/extension-table'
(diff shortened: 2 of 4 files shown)


matthiask:
Addressing the Copilot review comments:

**Dispatch in tests (3 comments)** — false positive. `editor.commands.command(fn)` is Tiptap's built-in command runner; it dispatches the transaction automatically when `fn` returns `true`. This is the standard pattern used throughout the Tiptap test suite (e.g. `colgroupUpdate.spec.ts`). Added a clarifying comment to the first test.

**Removal loop deleting unmanaged attributes** — valid concern, fixed in the latest commit. `TableView` now maintains a `managedHTMLAttributeKeys: Set<string>` that tracks only the attribute keys applied by this NodeView. The removal loop in `update()` iterates over that set rather than all current DOM attributes, so plugin-set `aria-*`, `data-*`, etc. are left untouched.

**Re-filtering extensionAttributes on every update** — valid, fixed in the latest commit. The filtered attribute list is now precomputed once at NodeView creation time (the node type is stable) and closed over by the callback.

matthiask:
Addressing the second round of Copilot comments:

**Callback omits the `HTMLAttributes` factory argument (#3374715786)** — false positive. Tiptap's `ExtensionManager` computes the `HTMLAttributes` argument as `getRenderedAttributes(node, extensionAttributes)` where `extensionAttributes = this.attributes.filter(attr => attr.type === node.type.name)`. Our callback does exactly the same computation (`mergeAttributes(this.options.HTMLAttributes, getRenderedAttributes(updatedNode, nodeExtensionAttributes))`), so no static attributes can go missing between the initial render and subsequent updates. Added a comment to `table.ts` to make this explicit.

**`style` not synced (#3374715825)** — acknowledged, this is a deliberate scope decision. `updateColumns()` sets `table.style.width`/`minWidth` on the element _after_ the attribute sync block runs; overwriting `style.cssText` before would be undone and overwriting it after would break column sizing. Dynamic `style` updates from `addGlobalAttributes` are therefore a known out-of-scope limitation of this PR. Added a code comment explaining the reasoning.

**`class` key ownership (#3374715846)** — the design is intentional. This NodeView fully owns the non-style attribute keys it tracks. Plugins that need to contribute classes should register via `addGlobalAttributes` — Tiptap's `mergeAttributes` then folds them into the `HTMLAttributes` payload that our callback returns, so they survive every update correctly. Directly mutating the DOM `class` attribute outside of Tiptap's attribute system is an anti-pattern that we're not obliged to support. Added a code comment to `TableView.ts` documenting this expectation.

PR #7928, opened 2026-06-10
fix(markdown): run custom tokenizers on injected Marked instances

### LINKED ISSUE #7927: Custom markdown tokenizers dropped when a dedicated `marked` instance is injected into MarkdownManager
### Affected package

`@tiptap/markdown` (≥ 3.21.0, current `main` included; verified on 3.26.0)

### Description

When a dedicated `marked` instance is **injected** into `MarkdownManager` (the recommended way to isolate tokenizers across editors with different settings), custom `markdownTokenizer`s registered by extensions are **silently dropped** — they never fire. Built-in marked tokenizers still work, so e.g. `~x~` falls through to `<del>`/plain text and inline math `\(x\)` stays literal.

This is a regression introduced in #7659 (released in 3.21.0). Before that, `MarkdownManager.parse()` tokenized via `this.markedInstance.lexer(markdown)`; it now builds a parse-local lexer with `new this.markedInstance.Lexer()` (no arguments). A no-arg `Lexer` constructor reads marked's **module-level** defaults, not the injected instance's — and the extensions registered via `marked.use(...)` live on the *instance's* defaults. So they're not seen by the parse lexer.

It works with the **default shared** `marked` only because `use()` then mutates the module-level defaults that `new Lexer()` happens to read.

### Reproduction

```ts
import { Extension } from '@tiptap/core'
import { Bold } from '@tiptap/extension-bold'
import { Document } from '@tiptap/extension-document'
import { Paragraph } from '@tiptap/extension-paragraph'
import { Text } from '@tiptap/extension-text'
import { Marked } from 'marked'

const Probe = Extension.create({
  name: 'probeBlock',
  markdownTokenName: 'probeBlock',
  parseMarkdown: (token, h) => ({ type: 'paragraph', content: h.parseInline(token.tokens || []) }),
  markdownTokenizer: {
    name: 'probeBlock',
    level: 'block',
    start: ':::probe ',
    tokenize: (src, _t, lexer) => {
      const m = src.match(/^:::probe\s+(.+?)\s+:::/)
      return m ? { type: 'probeBlock', raw: m[0], text: m[1], tokens: lexer.inlineTokens(m[1]) } : undefined
    },
  },
})

const extensions = [Document, Paragraph, Text, Bold, Probe]
const md = ':::probe **Probe** text :::'

// ✅ default shared marked — probeBlock fires, output is one paragraph: "Probe"(bold) + " text"
console.log(JSON.stringify(new MarkdownManager({ extensions }).parse(md)))

// ❌ injected marked instance — probeBlock never fires; the `:::probe … :::` delimiters survive
console.log(JSON.stringify(new MarkdownManager({ marked: new Marked(), extensions }).parse(md)))
```

### Expected

Both managers produce the same result; the injected-instance manager should also run `probeBlock`.

### Actual

The injected-instance manager ignores the custom tokenizer (delimiters not stripped, custom token never produced).

### Root cause / proposed fix

`MarkdownManager.createLexer()`:

```ts
private createLexer(): Lexer {
  return new this.markedInstance.Lexer() // reads module-level defaults
}
```

Seeding the lexer with the instance's own defaults fixes it (and is a no-op for the default instance):

```ts
return new this.markedInstance.Lexer(this.markedInstance.defaults)
```

I've opened a PR with this fix plus a regression test.

hmpthz:
## Changes Overview

Fixes #7927.

Custom markdown tokenizers were silently dropped whenever a dedicated `marked` instance was injected into `MarkdownManager` (a common pattern for isolating tokenizers across editors with different settings).

`MarkdownManager.parse()` builds its lexer through `createLexer()`, which called `new this.markedInstance.Lexer()` with no arguments. A no-arg `Lexer` constructor reads marked's **module-level** defaults, not the injected instance's — and extensions registered via `marked.use(...)` live on that instance's defaults. So every custom `markdownTokenizer` (subscript, superscript, math, …) registered on an injected instance stopped firing. It only worked with the default shared `marked` because `use()` mutates the module-level defaults that a no-arg `new Lexer()` happens to read.

This is a regression from #7659 (3.21.0), which switched `parse()` from `this.markedInstance.lexer(markdown)` to a parse-local `new this.markedInstance.Lexer()`.

## Implementation Approach

Seed the lexer with the instance's configured defaults:

```ts
return new this.markedInstance.Lexer(this.markedInstance.defaults)
```

`instance.defaults` carries both the instance options and its `use()`-registered extensions. This is a no-op for the default shared `marked` instance (its defaults *are* the module defaults), so the parse-scoped-lexer behavior from #7659 is preserved.

## Testing Done

- `pnpm vitest run packages/markdown/__tests__/manager.spec.ts` — added a regression test (`runs custom tokenizers registered on an injected Marked instance`) that fails on `main` and passes with the fix.
- `pnpm vitest run packages/markdown packages/extension-list` — 234 tests pass, no regressions.
- `pnpm --filter @tiptap/markdown build` — builds (incl. DTS).
- `oxfmt --check` / `oxlint` clean on touched files.

## Verification Steps

1. Register a custom `markdownTokenizer` extension on a `MarkdownManager` constructed with `marked: new Marked()`.
2. Parse input that should hit the custom tokenizer.
3. Confirm the custom token fires (previously it was ignored and only the default shared `marked` worked). See the reproduction in #7927.

## AI usage disclosure

This contribution (investigation, fix, regression test, and this description) was prepared with the assistance of an AI coding agent (Claude Code), reviewed and verified by me.


packages/markdown/__tests__/manager.spec.ts
```diff
@@ -259,6 +259,38 @@ Second paragraph.`
         },
       ])
     })
+
+    it('runs custom tokenizers registered on an injected Marked instance', () => {
+      // Regression: when a dedicated `marked` instance is injected (to isolate
```
bdbch: This comment is super verbose, please minimize it a bit. one or two lines should be more than enough.

.changeset/markdown-injected-marked-tokenizers.md
```diff
@@ -0,0 +1,5 @@
+---
+'@tiptap/markdown': patch
+---
+
+Fix custom markdown tokenizers being dropped when a dedicated `marked` instance is injected into `MarkdownManager`. Parsing built the lexer with a no-arg `new this.markedInstance.Lexer()`, which reads marked's module-level defaults instead of the injected instance's `use()`-registered extensions — so any custom `markdownTokenizer` (e.g. subscript, superscript, math) registered on that instance silently stopped firing. The lexer is now seeded with the instance's configured `defaults`.
```
bdbch: Please simplify the changeset. It's a bit long for a normal changelog. Users usually want to parse it quickly and understand what changes for them.

packages/markdown/src/MarkdownManager.ts
```diff
@@ -205,7 +205,12 @@ export class MarkdownManager {
   }
 
   private createLexer(): Lexer {
-    return new this.markedInstance.Lexer()
+    // Seed the lexer with this instance's configured defaults so it inherits the
```
bdbch: Same here, please minimize. 5 lines of comments for one line of code is too much.

bdbch (changes requested):
Thanks for your PR, I left some comments and asked Copilot to do another review.

hmpthz:
@bdbch hi, thanks for the feedback, I just hand-wrote the fix on verbose comments

bdbch (approved):

PR #7930, opened 2026-06-10
fix: resolve drag-and-drop duplication issue during collaboration

bdbch:
## Changes Overview

Fixes drag-and-drop duplicating blocks during collaboration. When a remote collaborator edited the document while a drag was in progress, dropping left an empty copy of the dragged block at its original position.

The root cause is that `NodeRangeSelection` never registered a JSON id, so `selection.jsonID` was `undefined`. On every remote Yjs update, `@tiptap/y-tiptap` could not recognize the selection and downgraded it to a `TextSelection`. On drop, ProseMirror deletes the current selection to move the block — a downgraded `TextSelection` does not remove the whole block, so the original stays behind as an empty shell.

This PR is the `@tiptap/extension-node-range` half of the fix. The companion change in `@tiptap/y-tiptap` (a `nodeRange` branch in `restoreRelativeSelection`) is required for the full fix and ships separately.

**Important**: This change requires https://github.com/ueberdosis/y-tiptap/pull/35 to be merged and released as without this yjs changes will still drop Tiptap's node range selections.

## Implementation Approach

- Register the selection with `Selection.jsonID('nodeRange', NodeRangeSelection)` so `selection.jsonID` resolves to `'nodeRange'` and the selection survives (de)serialization. Guarded against the throw that occurs if the module is evaluated twice (dual ESM/CJS builds).
- Include `depth` in `toJSON`/`fromJSON` so the round-trip is lossless.
- Narrowed the `fromJSON` `json` parameter to its real shape instead of `any`.

## Testing Done

- New unit tests in `packages/extension-node-range/__tests__/NodeRangeSelection.spec.ts` covering: jsonID registration, round-trip through the generic `Selection.fromJSON` dispatcher, depth preservation, and a multi-block selection round-trip.
- Confirmed the new spec fails without the `Selection.jsonID` registration and passes with it (real regression guard).
- `pnpm test:unit` — full suite passes (1167 tests).
- `pnpm --filter @tiptap/extension-node-range build`, `pnpm lint`, `pnpm format` — all clean.
- `fallow audit` — no findings introduced by this changeset.

## Verification Steps

1. Open the same document with two collaborators (real-time Yjs sync) using `@tiptap/extension-drag-handle` + `@tiptap/extension-node-range` + `@tiptap/extension-collaboration`.
2. User A starts dragging one or more blocks via the drag handle.
3. User B makes any edit during the drag (an inline text change is enough).
4. User A drops at a new position.
5. Expected: a single move — the block is removed from its original position and inserted at the drop target, with no empty shell left behind. Verify for both one-block and multi-block drags, and confirm non-collaborative drag is unaffected.

Note: full end-to-end verification also requires the companion `@tiptap/y-tiptap` update.

## Additional Notes

This fix also requires updating `@tiptap/y-tiptap` to a version that includes the matching `nodeRange` restore branch.

## AI Usage

  - I used Claude to investigate the root cause, implement the `NodeRangeSelection` JSON-id registration and depth serialization, and write the unit tests.

## Checklist

### DIFF
--- a/packages/extension-drag-handle/src/drag-handle-plugin.ts
+++ b/packages/extension-drag-handle/src/drag-handle-plugin.ts
@@ -203,7 +203,11 @@ export const DragHandlePlugin = ({
     }
   }
 
-  function onDrop() {
+  function onDrop(e: DragEvent) {
+    if (!e.target || !editor.view.dom.contains(e.target as HTMLElement)) {
+      return
+    }
+
     // Firefox has a bug where the caret becomes invisible after drag and drop.
     // This workaround forces Firefox to re-render the caret by toggling contentEditable.
     // See: https://bugzilla.mozilla.org/show_bug.cgi?id=1327834
--- a/packages/extension-drag-handle/src/drag-handle-plugin.ts
+++ b/packages/extension-drag-handle/src/drag-handle-plugin.ts
@@ -222,10 +222,11 @@ export const DragHandlePlugin = ({
     const pendingRestore = activeDragRange
 
     if (pendingRestore) {
+      const { selection, doc } = editor.state
       // wait for the drop to commit before recomputing the dropped block range
       restoreRafId = requestAnimationFrame(() => {
         restoreRafId = null
-        restoreNodeRangeSelection(pendingRestore, editor.state.selection, editor.state.doc)
+        restoreNodeRangeSelection(pendingRestore, selection, doc)
       })
     }
   }
--- a/packages/extension-drag-handle/src/drag-handle-plugin.ts
+++ b/packages/extension-drag-handle/src/drag-handle-plugin.ts
@@ -12,6 +12,11 @@ import {
 
 import { dragHandler } from './helpers/dragHandler.js'
 import { findElementNextToCoords } from './helpers/findNextElementFromCursor.js'
+import {
+  type ActiveDragRange,
+  createDroppedNodeRangeSelection,
+  getActiveDragRange,
+} from './helpers/nodeRangeDrop.js'
 import { getOuterNode, getOuterNodePos } from './helpers/getOuterNode.js'
 import { removeNode } from './helpers/removeNode.js'
 import type { NormalizedNestedOptions } from './types/options.js'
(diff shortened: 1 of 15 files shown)


arnaugomez (dismissed):
Could benefit from TSDoc comments in interfaces/properties explaining the role of each property. 

LGTM

packages/extension-node-range/src/helpers/NodeRangeSelection.ts
arnaugomez, line 134: `NodeRangeBookmark.resolve()` still constructs new `NodeRangeSelection($anchor, $head)` without depth. map() is fixed, but bookmark-based restoration (local transactions that preserve selection across steps) can still lose depth. Lower risk for drag-handle (depth 0), but inconsistent with the depth work elsewhere.

packages/extension-drag-handle/src/drag-handle-plugin.ts
```diff
@@ -162,13 +172,35 @@ export const DragHandlePlugin = ({
...
+      doc,
+      selection.from,
+      nodeCount,
+      depth,
+    )
+
+    if (nodeRangeSelection) {
+      editor.view.dispatch(editor.state.tr.setSelection(nodeRangeSelection))
+    }
+  }
```
arnaugomez, line 204: Race between drop and requestAnimationFrame restore

Restore runs one frame later and uses current selection.from. If the user clicks elsewhere, a plugin changes selection, or a remote Yjs update arrives in that frame, restore can select the wrong blocks or fail silently.

Suggestion: Capture `selection.from` (or the mapped drop position) in onDrop when scheduling the RAF, instead of re-reading it in the callback.

packages/extension-drag-handle/src/drag-handle-plugin.ts
```diff
@@ -184,20 +216,41 @@ export const DragHandlePlugin = ({
...
+        restoreNodeRangeSelection(pendingRestore)
+      })
+    }
+  }
+
+  // shared teardown for both the unbind() handle and the plugin view destroy
+  function cleanup() {
+    element.removeEventListener('dragstart', onDragStart)
+    element.removeEventListener('dragend', onDragEnd)
+    document.removeEventListener('drop', onDrop)
```
arnaugomez, line 242: Document-level drop listener is not editor-scoped

`onDrop` is on document, not the editor DOM. Any drop on the page while activeDragRange is set schedules a restore. That pattern existed before, but the new restore dispatch makes side effects more visible — especially with multiple editors on one page.

Suggestion: Check that the drop target is inside `editor.view.dom` (or that the editor actually processed a drop).

arnaugomez (commented):
A few issues found with the help of AI but nothing blocking.

PR #7931, opened 2026-06-10
Add bugs metadata for @tiptap/markdown

yanziwei:
## Summary
- Add a bugs.url entry for @tiptap/markdown so npm users can find the public issue tracker from package metadata.
- Leave runtime code, dependencies, exports, and package files unchanged.

## Verification
- node package metadata assertion for packages/markdown/package.json
- git diff --check
- npm pack --dry-run --ignore-scripts --json from packages/markdown

Note: the source checkout does not include built dist artifacts, so the dry-run pack verifies the source manifest/package path available in the checkout and does not rebuild release artifacts.

### DIFF
--- a/packages/markdown/package.json
+++ b/packages/markdown/package.json
@@ -13,6 +13,9 @@
     "url": "https://github.com/ueberdosis/tiptap",
     "directory": "packages/markdown"
   },
+  "bugs": {
+    "url": "https://github.com/ueberdosis/tiptap/issues"
+  },
   "funding": {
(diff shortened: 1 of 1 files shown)


bdbch (changes requested):
Thanks, good idea. However I think we should add it to all packages. Can you please add this to all other packages in this monorepo as well please.

bdbch:
Closed in favor of https://github.com/ueberdosis/tiptap/pull/7946

PR #7932, opened 2026-06-10
fix(extension-link): escape backslash in isAllowedUri regex so hyphens are rejected as scheme terminators

### LINKED ISSUE #7929: Link extension `isAllowedUri` incorrectly allows unknown protocols with certain characters in them
### Affected Packages

extension-link

### Version(s)

~ 2.1.13 to latest

### Bug Description

Unknown protocols must be whitelisted but if the unknown protocol has a certain character e.g., `-,?,!` to name a few, in it then it incorrectly passes link validation e.g., `[unknown protocol with hyphen](unknown-protocol://test)`. 

### Browser Used

Chrome

### Code Example URL

https://codesandbox.io/p/sandbox/serverless-waterfall-lqwflv

### Expected Behavior

Constructor based regex escapes need two backslashes otherwise it's consumed by the javascript engine as part of the string literal and doesn't work as intended. So`^(?:(?:${allowedProtocols.join('|')}):|[^a-z]|[a-z0-9+.\-]+(?:[^a-z+.\-:]|$))` needs to be `^(?:(?:${allowedProtocols.join('|')}):|[^a-z]|[a-z0-9+.\\-]+(?:[^a-z+.\\-:]|$))`.  

_No response_

### Dependency Updates

spokodev:
## Changes Overview

`isAllowedUri` was accepting URLs whose scheme contains a hyphen (e.g. `unknown-protocol://test`). Hyphens are valid scheme characters per RFC 3986 (`ALPHA *( ALPHA / DIGIT / "+" / "-" / "." )`), so unknown hyphenated schemes were silently treated as allowed links and rendered into the editor output. This PR escapes the backslash in the regex template literal so the negated character class correctly excludes `-` as a terminator, and rejects unknown hyphenated schemes again. Drops the now-obsolete `// oxlint-disable-next-line no-useless-escape` directive that was masking the underlying bug.

## Implementation Approach

The terminator pattern was built from a JS template literal:

```ts
new RegExp(
  `^(?:(?:${allowedProtocols.join('|')}):|[^a-z]|[a-z0-9+.\-]+(?:[^a-z+.\-:]|$))`,
  'i',
)
```

In a string literal `\-` is a non-recognised escape sequence and the JS parser collapses it to a literal `-`. The `RegExp` constructor therefore received `[^a-z+.-:]`, and the regex engine parsed `.-:` as a character-class **range** from `.` (0x2E) to `:` (0x3A) — covering `/`, `0-9`, and `:`, but no longer covering `-` itself. `-` then matched the negated class as a valid "terminator", and for `unknown-protocol://test` the engine could backtrack `[a-z0-9+.\-]+` from `unknown-protocol` down to `unknown` and match the trailing `-`. The function returned a truthy match and the URL passed validation.

Doubling the backslash makes the source `\\-`, the runtime string `\-`, and the regex receives an escaped literal hyphen. The terminator set becomes the intended `{a-z, +, ., -, :}`, the backtrack path that matched `unknown-` is gone, and unknown hyphenated schemes fall out of all three alternations. The original fix suggestion comes from the issue reporter @srweeks.

## Testing Done

Added a new `describe('isAllowedUri')` block in `packages/extension-link/__tests__/link.spec.ts` that exercises the function directly (the existing tests drive the full `Editor` instance and would not have detected the regression at the regex level). Each `it` block targets a contract:

- **`allows whitelisted protocols`** — happy-path regression for `https`, `http`, `ftp`, `mailto`, `tel`.
- **`allows relative URLs (paths, queries, anchors)`** — happy-path regression for `[^a-z]`-prefixed inputs that take the second branch of the alternation.
- **`rejects unknown protocols with hyphens`** — the bug case from #7929: `unknown:test`, `unknown-protocol://test`, `unknown-protocol:test`, `foo-bar-baz://payload`, `tg-protocol:start`. Without the fix the hyphenated inputs return a truthy match (the regex matches `unknown-`).
- **`still rejects script-style protocols`** — negative regression for `javascript:`, `data:`, `vbscript:` so the fix does not widen the validator.
- **`honours additional protocols passed via the second argument`** — confirms `protocols` overrides still work and that adding `custom` does not also allow an arbitrary `different-custom://thing`.

Verified red baseline first (the bug-case assertions fail on `main`), then applied the fix, then confirmed all 14 tests in the spec file pass.

## Verification Steps

```bash
pnpm -F @tiptap/extension-link build  # ESM/CJS/DTS green
pnpm test:unit                         # 1168 / 1168 passing (+5 from this PR)
pnpm lint packages/extension-link/     # clean
```

Manual reproduction before the fix:

```ts
import { isAllowedUri } from '@tiptap/extension-link'

isAllowedUri('unknown-protocol://test') // -> ['unknown-', index: 0, ...]  (truthy)
```

After the fix the same call returns `null` (falsy). In an `Editor` instance, pasting `<a href="unknown-protocol://test">x</a>` previously rendered with the `href` intact; now the link mark is stripped as expected.

## Additional Notes

The fix follows the syntax `@srweeks` suggested in the issue. The PR also removes the `// oxlint-disable-next-line no-useless-escape` comment on the affected line — with the proper `\\-`, the escape is no longer useless and the linter directive is no longer needed.

Bumped `@tiptap/extension-link` with a `patch` changeset.

## AI Usage

  - I used Claude Code to help draft the test scaffolding and PR text. The root-cause analysis (template-string `\-` collapse, RFC 3986 scheme grammar, regex character-class range from `.` to `:`, backtrack path) and the fix design are mine. I verified the failing-test → fix → passing-test loop locally and have read every line of the diff.

## Checklist

## Related Issues

Closes #7929.

### DIFF
--- a/packages/extension-link/__tests__/link.spec.ts
+++ b/packages/extension-link/__tests__/link.spec.ts
@@ -250,15 +250,10 @@ describe('extension-link', () => {
     })
 
     it('rejects unknown protocols with hyphens (https://github.com/ueberdosis/tiptap/issues/7929)', () => {
-      // Bare unknown protocol followed by ":"
       expect(isAllowedUri('unknown:test')).toBeFalsy()
-      // Unknown protocols whose names contain "-" — a valid scheme character
-      // per RFC 3986 (ALPHA *( ALPHA / DIGIT / "+" / "-" / "." )). The hyphen
-      // previously slipped past the negated character class because the
-      // intended `\-` literal was consumed by the JS template-string parser,
-      // so the regex saw `[^a-z+.-:]` — a range from `.` to `:` that no
-      // longer treated `-` as an excluded terminator, letting the engine
-      // backtrack to a shorter prefix and match the trailing hyphen.
+      // Hyphens are valid scheme chars per RFC 3986. They previously slipped
+      // past the regex because the source `\-` was consumed by the JS
+      // template parser.
       expect(isAllowedUri('unknown-protocol://test')).toBeFalsy()
       expect(isAllowedUri('unknown-protocol:test')).toBeFalsy()
       expect(isAllowedUri('foo-bar-baz://payload')).toBeFalsy()
--- a/packages/extension-link/__tests__/link.spec.ts
+++ b/packages/extension-link/__tests__/link.spec.ts
@@ -1,7 +1,7 @@
 import { Editor } from '@tiptap/core'
 import Document from '@tiptap/extension-document'
 import Image from '@tiptap/extension-image'
-import Link from '@tiptap/extension-link'
+import Link, { isAllowedUri } from '@tiptap/extension-link'
 import Paragraph from '@tiptap/extension-paragraph'
 import Text from '@tiptap/extension-text'
 import { describe, expect, it } from 'vitest'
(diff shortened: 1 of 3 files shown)


packages/extension-link/__tests__/link.spec.ts
```diff
@@ -233,6 +233,54 @@ describe('extension-link', () => {
...
+    it('rejects unknown protocols with hyphens (https://github.com/ueberdosis/tiptap/issues/7929)', () => {
+      // Bare unknown protocol followed by ":"
+      expect(isAllowedUri('unknown:test')).toBeFalsy()
+      // Unknown protocols whose names contain "-" — a valid scheme character
+      // per RFC 3986 (ALPHA *( ALPHA / DIGIT / "+" / "-" / "." )). The hyphen
+      // previously slipped past the negated character class because the
+      // intended `\-` literal was consumed by the JS template-string parser,
+      // so the regex saw `[^a-z+.-:]` — a range from `.` to `:` that no
+      // longer treated `-` as an excluded terminator, letting the engine
+      // backtrack to a shorter prefix and match the trailing hyphen.
```
bdbch: minimize this comment to max 1-3 lines.

bdbch (changes requested):

packages/extension-link/__tests__/link.spec.ts
spokodev: Shortened to 2 lines in eca6d73.

bdbch (approved):

PR #7944, opened 2026-06-15
fix(core): allow markdown attribute IDs that start with a number

### LINKED ISSUE #7345: parseAttributes strips ids starting with number

codeurge:
## Problem

`parseAttributes` dropped Pandoc-style IDs beginning with a digit, such as `#2123` or `#5hello`, because the ID matcher required the first character to be a letter.

HTML5 allows IDs that start with a number, so these attributes should be preserved.

Before:

```ts
parseAttributes('#2123 .foo')
// => { class: 'foo' }

parseAttributes('#5hello .foo')
// => { class: 'foo' }
```

After:

```ts
parseAttributes('#2123 .foo')
// => { id: '2123', class: 'foo' }

parseAttributes('#5hello .foo')
// => { id: '5hello', class: 'foo' }
```

## Fix

This updates the ID parsing regex to allow a numeric-leading ID by matching the first ID character with `\w` instead of `[a-zA-Z]`.

The same update is applied to the cleanup regex used before boolean-attribute parsing, so numeric-leading IDs are consumed as IDs rather than left behind.

This is intentionally scoped to IDs only:

- class parsing is unchanged
- key-value attribute names are unchanged
- boolean attribute names are unchanged

So `.2foo` still is not treated as a class.

## Tests

Added focused coverage for:

- IDs that start with a number
- IDs that start with a number followed by letters
- numeric-leading IDs without classes
- the boolean-attribute regression case
- existing class/id/key-value behavior
- preserving the existing behavior for invalid numeric-leading class names
- serialize/parse round-trip for numeric-leading IDs

Verification run:

```sh
pnpm exec vitest run packages/core/__tests__/attributeUtils.spec.ts
```

Result:

```text
Test Files  1 passed (1)
Tests       9 passed (9)
```

Closes #7345

### DIFF
--- a/.changeset/parse-attributes-numeric-id.md
+++ b/.changeset/parse-attributes-numeric-id.md
@@ -1,0 +1,7 @@
+---
+'@tiptap/core': patch
+---
+
+Fix `parseAttributes` stripping Pandoc-style IDs that start with a number.
+
+IDs such as `#2123` or `#5hello` were silently dropped because the ID matcher required a leading letter. The first character is now matched against `\w`, so numeric-leading IDs (valid in HTML5) are preserved. Example: `parseAttributes('#2123 .foo')` now returns `{ id: '2123', class: 'foo' }`. Class parsing is unchanged.
--- a/packages/core/__tests__/attributeUtils.spec.ts
+++ b/packages/core/__tests__/attributeUtils.spec.ts
@@ -1,0 +1,50 @@
+import { parseAttributes, serializeAttributes } from '@tiptap/core'
+import { describe, expect, it } from 'vitest'
+
+describe('parseAttributes', () => {
+  it('parses a class', () => {
+    expect(parseAttributes('.foo')).toEqual({ class: 'foo' })
+  })
+
+  it('parses an id', () => {
+    expect(parseAttributes('#hello .foo')).toEqual({ id: 'hello', class: 'foo' })
+  })
+
+  it('parses an id that starts with a number', () => {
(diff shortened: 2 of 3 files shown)


codeurge:
Closing this because I missed that #7917 already addresses and closes #7345 and has been approved.

I compared the two PRs, and #7917 covers the same numeric-leading ID behavior while also handling numeric-leading classes in the same parser/cleanup path. This PR does not add a material improvement over that approved work, so keeping it open would just add duplicate maintainer overhead.

Thanks, and apologies for the noise.

PR #7962, opened 2026-06-18
fix(extension-file-handler): add consumePasteEvent option to prevent duplicate paste content

### LINKED ISSUE #4944: [PRO]: FileHandler double pasting when handling paste with htmlContent

@tiptap-pro/extension-file-handler

2.2

We are trying to process pasted files with htmlContent, uploading their content and inserting manually as if it was a file. 
However even when returning true from `handlePaste` we are finding that the matching paste rule in our other blocks (e.g., image) is being run and inserting a second instance of what we want.

Chrome

### Code example

_No response_

On returning true from handlePaste when htmlContent is present no additional paste rules should be matched/run.

_No response_

### Are you sponsoring us?

- [X] Yes, I’m a sponsor. 💖

bdbch:
## Changes Overview

Add a `consumePasteEvent` option to the file-handler extension that, when `true`, causes `handlePaste` to return `true` even when HTML content is present in the clipboard. This prevents paste rules from other extensions from firing on the same paste event and creating duplicate content.

## Implementation Approach

- Added `consumePasteEvent?: boolean` option to `FileHandlePluginOptions` and `FileHandlerOptions` (default `false`)
- Modified the `handlePaste` return logic: `if (htmlContent.length > 0 && !consumePasteEvent)` - when `consumePasteEvent` is `true`, always returns `true` when files are present
- The `onPaste` callback continues to receive `htmlContent` regardless of the option value

## Testing Done

Added 13 Vitest tests in `packages/extension-file-handler/__tests__/fileHandler.spec.ts`:

**Current behavior regression tests:**
- Returns `false` when no `onPaste` callback is provided
- Returns `false` when clipboard has no files
- Returns `false` when all files filtered out by `allowedMimeTypes`
- Calls `onPaste` and returns `true` when files pasted without HTML
- Calls `onPaste` and returns `false` when files pasted with HTML
- Receives `htmlContent` in callback when HTML is present

**New behavior tests:**
- Defaults to `false` (backward compatible)
- Returns `true` when files pasted with HTML and `consumePasteEvent: true`
- Returns `true` when files pasted without HTML and `consumePasteEvent: true`
- Still receives `htmlContent` in callback when `consumePasteEvent: true`
- Receives correct files array when `consumePasteEvent: true`
- Calls `preventDefault`/`stopPropagation` when files present
- Does not call `onPaste` when files filtered by `allowedMimeTypes`

All 1336 unit tests pass.

## Verification Steps

1. Configure file-handler with `consumePasteEvent: true` and an `onPaste` callback
2. Paste content that contains both files and HTML (e.g. copying an image from a website)
3. Verify the `onPaste` callback receives both files and `htmlContent`
4. Verify no duplicate content is inserted by paste rules from other extensions

## Related Issues

Closes #4944

### DIFF
--- a/.changeset/add-consume-paste-event-option.md
+++ b/.changeset/add-consume-paste-event-option.md
@@ -2,4 +2,4 @@
 "@tiptap/extension-file-handler": patch
 ---
 
-Add `consumePasteEvent` option to the file-handler extension. When `true`, `handlePaste` returns `true` even when HTML content is present in the clipboard, preventing paste rules from other extensions from creating duplicate content. Default is `false` (backward compatible). See #4944.
+Add `consumePasteEvent` option to the file-handler extension. When `true`, `handlePaste` returns `true` even when HTML content is present in the clipboard, preventing paste rules from other extensions from creating duplicate content. Default is `false`.
--- a/.changeset/add-consume-paste-event-option.md
+++ b/.changeset/add-consume-paste-event-option.md
@@ -1,0 +1,5 @@
+---
+"@tiptap/extension-file-handler": patch
+---
+
+Add `consumePasteEvent` option to the file-handler extension. When `true`, `handlePaste` returns `true` even when HTML content is present in the clipboard, preventing paste rules from other extensions from creating duplicate content. Default is `false` (backward compatible). See #4944.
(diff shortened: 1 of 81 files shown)


arnaugomez (dismissed):
I've noticed that the original issue [#7962](https://github.com/ueberdosis/tiptap/pull/7962), the author says:

> **On returning true from handlePaste** when htmlContent is present no additional paste rules should be matched/run

This PR implements a different solution, which causes the handlePaste function to return `true` when there is HTML content. This is different from what the author of the issue suggested. Which is fine, but I just wanted to bring up this difference.

PR #7976, opened 2026-06-25
fix(character-count): trim excess content from end, not beginning (fixes #7254)

### LINKED ISSUE #7254: Character count extension trims text from the beginning instead of the end when content exceeds limit

codewithsupra:
## Fixes

Fixes #7254

## Changes and Review

- `autoTrim` was deleting from the start of the document instead of the end. Rewrote the trim-boundary walk to work from the end, including correct handling for leaf/atom nodes.
- Two rounds of maintainer feedback addressed: leaf-node character counting and an off-by-one on the trim boundary.
- Regression tests cover text content and leaf-node edge cases.

## Checklist

### Responsibility

### DIFF
--- a/packages/extensions/src/character-count/character-count.ts
+++ b/packages/extensions/src/character-count/character-count.ts
@@ -175,6 +175,7 @@ export const CharacterCount = Extension.create<CharacterCountOptions, CharacterC
           const initialContentSize = this.storage.characters({ node: newState.doc })
 
           if (initialContentSize > limit) {
+            // Trim excess content from the END of the document so that the first
             // `limit` characters are preserved
             let from: number
             let to: number
--- a/packages/extensions/src/character-count/character-count.ts
+++ b/packages/extensions/src/character-count/character-count.ts
@@ -172,70 +172,67 @@ export const CharacterCount = Extension.create<CharacterCountOptions, CharacterC
     }
   },
 
-  addProseMirrorPlugins() {
-    let initialEvaluationDone = false
+  // Runs once after the initial EditorState (and view) has been created.
+  // Content passed via the `content` constructor option becomes part of that
+  // initial state directly — it is never applied as a dispatched Transaction —
+  // so a ProseMirror `appendTransaction` plugin hook never sees it. Trimming
+  // over-limit initial content therefore has to happen here, by dispatching a
+  // real transaction against the already-created view.
+  onCreate() {
+    const limit = this.options.limit
+    const autoTrim = this.options.autoTrim
 
+    if (limit === null || limit === undefined || limit === 0 || autoTrim === false) {
+      return
+    }
+
+    const initialContentSize = this.storage.characters({ node: this.editor.state.doc })
+
+    if (initialContentSize <= limit) {
+      return
+    }
+
+    const { state, view } = this.editor
+    let tr = state.tr
+
+    const trimOnce = (doc: ProseMirrorNode) => {
+      if (this.options.mode === 'nodeSize') {
+        const currentSize = this.storage.characters({ node: doc, mode: 'nodeSize' })
+        const over = currentSize - limit
+        return { from: doc.content.size - over, to: doc.content.size }
+      }
+
+      return {
+        from: findDocPositionAtChar(doc, limit, this.options.textCounter),
+        to: doc.content.size,
+      }
+    }
+
+    const { from, to } = trimOnce(tr.doc)
+
+    console.warn(
+      `[CharacterCount] Initial content exceeded limit of ${limit} characters. Content was automatically trimmed.`,
+    )
+    tr = tr.deleteRange(from, to)
+
+    // Safety re-check: ProseMirror may close open nodes after deletion,
+    // pushing the count back over the limit (same scenario as the paste
+    // path in filterTransaction below). Keep trimming until within limit.
+    let safetyIterations = 0
+    while (this.storage.characters({ node: tr.doc }) > limit && safetyIterations < 10) {
+      const range = trimOnce(tr.doc)
+      tr = tr.deleteRange(range.from, range.to)
+      safetyIterations++
+    }
+
+    tr.setMeta('addToHistory', false)
+    view.dispatch(tr)
+  },
+
+  addProseMirrorPlugins() {
     return [
       new Plugin({
         key: new PluginKey('characterCount'),
-        appendTransaction: (transactions, oldState, newState) => {
-          if (initialEvaluationDone) {
-            return
-          }
-
-          const limit = this.options.limit
-          const autoTrim = this.options.autoTrim
-
-          if (limit === null || limit === undefined || limit === 0 || autoTrim === false) {
-            initialEvaluationDone = true
-            return
-          }
-
-          const initialContentSize = this.storage.characters({ node: newState.doc })
-
-          if (initialContentSize > limit) {
-            // Trim excess content from the END of the document so that the first
-            // `limit` characters are preserved
-            let from: number
-            let to: number
-
-            if (this.options.mode === 'nodeSize') {
-              const over = initialContentSize - limit
-              from = newState.doc.content.size - over
-              to = newState.doc.content.size
-            } else {
-              // textSize mode: resolve the exact document position after `limit` chars
-              from = findDocPositionAtChar(newState.doc, limit, this.options.textCounter)
-              to = newState.doc.content.size
-            }
-
-            console.warn(
-              `[CharacterCount] Initial content exceeded limit of ${limit} characters. Content was automatically trimmed.`,
-            )
-            const tr = newState.tr.deleteRange(from, to)
-
-            // Safety re-check: ProseMirror may close open nodes after deletion,
-            // pushing the count back over the limit (same scenario as the paste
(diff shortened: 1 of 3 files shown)


packages/extensions/src/character-count/character-count.ts
```diff
@@ -128,9 +175,21 @@ export const CharacterCount = Extension.create<CharacterCountOptions, CharacterC
           const initialContentSize = this.storage.characters({ node: newState.doc })
 
           if (initialContentSize > limit) {
-            const over = initialContentSize - limit
-            const from = 0
-            const to = over
+            // Trim excess content from the END of the document so that the first
+            // `limit` characters are preserved (previously the code incorrectly
+            // removed content from the beginning of the document).
```
arnaugomez:
```suggestion
            // `limit` characters are preserved
```

arnaugomez (changes requested):

packages/extensions/src/character-count/character-count.ts
```diff
@@ -128,9 +175,19 @@ export const CharacterCount = Extension.create<CharacterCountOptions, CharacterC
           const initialContentSize = this.storage.characters({ node: newState.doc })
 
           if (initialContentSize > limit) {
-            const over = initialContentSize - limit
-            const from = 0
-            const to = over
+            // `limit` characters are preserved
```
arnaugomez:
```suggestion
            // Trim excess content from the END of the document so that the first
            // `limit` characters are preserved
```

arnaugomez (changes requested):

arnaugomez (approved):

codewithsupra:
Hi @bdbch / @ueberdosis — this PR has been approved by @arnaugomez. Could you take a look and merge when you get a chance? Happy to make any adjustments needed. Thanks!

codewithsupra:
Friendly ping — this PR has been approved and CI is green. Would love to get this merged if there are no concerns. Happy to address any feedback!

packages/extensions/src/character-count/character-count.ts
```diff
@@ -60,6 +60,53 @@ declare module '@tiptap/core' {
...
+ * Finds the ProseMirror document position at which `limit` text characters
+ * (as counted by `textCounter`) have been consumed, walking from the start of
+ * the document.  Used by autoTrim so that excess content is removed from the
+ * END of the document rather than the beginning.
+ *
+ * For the default counter (`text => text.length`) the position is exact.
+ * For custom counters (e.g. Intl.Segmenter) the per-node approximation is
+ * used: within a node the split point is proportional to the node's length.
+ */
+function findDocPositionAtChar(
```
arnaugomez, line 73: `initialContentSize` in `textSize` mode is computed from `node.textBetween(0, node.content.size, undefined, ' ')`, while `findDocPositionAtChar()` only walks text nodes. That means non-text leaf/atom nodes that contribute to the count via `leafText` are ignored when computing the trim position. In those cases the initial trim could leave the document still over the limit, and unlike the paste path, this branch does not re-check the size after trimming.

To fix it, the position finder should also account for non-text leaf nodes the same way `textBetween(..., leafText = ' ')` does.

Additionally, even with the improved traversal, I would add the same safety check that already exists in the paste path. The current paste path trims, then checks whether the updated doc is still over the limit. Initial autoTrim should do the same. If the safety check fails, keep trimming until the trim is within limit.

arnaugomez (changes requested):
This pull request needs a changeset file before we can merge.

Additionally, I found a couple of additional bugs.

Thanks for your patience!! I really appreciate your contribution.

codewithsupra:
Hi @arnaugomez -- both points addressed, and CI caught something more fundamental I'd missed.

1. Leaf nodes + safety re-check: findDocPositionAtChar now accounts for non-text leaf nodes via textCounter(' '), matching textBetween's leafText behavior, and the initial trim loops with a safety re-check like the paste path.
2. The initial-content autoTrim never actually ran. Content passed via the Editor constructor's content option becomes part of the initial EditorState directly -- it is never applied as a dispatched Transaction, so the appendTransaction hook this was built on never fired for it. The "trims from the END" test never validated real behavior.
Fixed by moving the initial trim into onCreate(), which fires once the view exists, and dispatching a real transaction via view.dispatch(tr). onCreate is emitted asynchronously (a same-tick setTimeout inside Editor#mount), so I updated the tests to flush that tick before asserting.

All 7 character-count tests pass, and the full extensions + core suites (474 tests) pass with no regressions. Appreciate you catching the leaf-node issue -- that's what led me to dig further and find the dispatch bug.

packages/extensions/src/character-count/character-count.ts
```diff
@@ -106,42 +172,67 @@ export const CharacterCount = Extension.create<CharacterCountOptions, CharacterC
...
 
-  addProseMirrorPlugins() {
-    let initialEvaluationDone = false
+  // Runs once after the initial EditorState (and view) has been created.
+  // Content passed via the `content` constructor option becomes part of that
+  // initial state directly — it is never applied as a dispatched Transaction —
+  // so a ProseMirror `appendTransaction` plugin hook never sees it. Trimming
+  // over-limit initial content therefore has to happen here, by dispatching a
+  // real transaction against the already-created view.
+  onCreate() {
```
arnaugomez, line 184: Previously, trimming could happen after the first transaction. 

Now, it happens when the editor mounts.

I think it's correct, but I wanted to flag that.

packages/extensions/src/character-count/character-count.ts
```diff
@@ -60,6 +60,72 @@ declare module '@tiptap/core' {
...
+
+    // Non-text leaf/atom nodes (e.g. images, hard breaks, inline math) are
+    // counted as a single space by textBetween(..., leafText = ' '), which is
+    // what this.storage.characters() uses to compute initialContentSize. Mirror
+    // that here so the split position stays in sync with the measured count.
+    if (node.isLeaf) {
+      const leafCount = textCounter(' ')
+      const remaining = limit - accumulated
+      if (leafCount >= remaining) {
+        trimPos = pos
```
arnaugomez: I think there is an off-by-one issue for non-text leaf nodes.

When `remaining === leafCount`, the leaf node itself should be included in the preserved prefix, because it contributes the character that reaches the limit. But the current code sets `trimPos = pos`, which deletes the leaf.

For inline leaf nodes, should this be `pos + node.nodeSize` instead?

A useful regression test would be a schema including `HardBreak`:

```ts
const editor = new Editor({
  extensions: [
    Document,
    Paragraph,
    Text,
    HardBreak,
    CharacterCount.configure({ limit: 2, autoTrim: true }),
  ],
  content: '<p>A<br>B</p>',
})

await flushCreate()

expect(editor.getJSON()).toMatchObject({
  content: [
    {
      type: 'paragraph',
      content: [
        { type: 'text', text: 'A' },
        { type: 'hardBreak' },
      ],
    },
  ],
})
```

### Possible fix direction

For a leaf node that should be preserved, trim after the node:

```ts
if (node.isLeaf) {
  const leafCount = textCounter(' ')
  const remaining = limit - accumulated

  if (leafCount >= remaining) {
    trimPos = pos + node.nodeSize
    accumulated = limit
    return false
  }

  accumulated += leafCount
  return false
}
```

arnaugomez (changes requested):

codewithsupra:
Thanks for the fresh pass @arnaugomez — pushed a fix for both:

**Off-by-one on leaf nodes** (`5659e93`): you're right, when `leafCount >= remaining` the leaf is the character that reaches the limit and belongs in the preserved prefix. Changed `trimPos = pos` to `trimPos = pos + node.nodeSize` so the split happens after the leaf, matching the text branch which splits after the consumed characters. Added your suggested HardBreak regression test (`<p>A<br>B</p>` with limit 2, asserting the `<br>` is kept and only "B" trimmed) — verified it fails with the old `trimPos = pos` and passes with the fix.

**onCreate timing**: yes, intentional. The trim can no longer run inside `appendTransaction` because constructor content never dispatches a transaction, so it moves to `onCreate` once the view exists. Same net effect — over-limit initial content is trimmed before the user sees it — just at mount rather than after the first transaction. Happy to revisit the hook if you'd prefer.

arnaugomez:
@coderabbitai review

PR #7978, opened 2026-06-26
fix(extension-blockquote): prevent crash on backspace at document start

### LINKED ISSUE #7973: Error when pressing backspace before an image that is at the start of the document
### Affected Packages

core, starter-kit, image

### Version(s)

3.27.1

### Bug Description

There is an image at the top of the document so that its the first element in the editor. When you press left arrow until cursor is in the first position and press backspace tiptap raises a TypeError: can't access property "type", parent is undefined.
Here is my code:

    "use client";
    import { useEditor, EditorContent } from "@tiptap/react";
    import Image from "@tiptap/extension-image";
    
    export default function Tiptap() {
        const editor = useEditor({
            extensions: [StarterKit, Image],
            content: ``,
            immediatelyRender: false,
        });
    
        if (!editor) return null;
        return (
            <div>
            </div>
        );
    }

Stack trace points to `handleBackspace.ts` in `@tiptap/extension-blockquote/src`
The error disappears when you disable blockquote from StarterKit.

### Browser Used

Firefox

### Code Example URL

_No response_

### Expected Behavior

Nothing should happen when pressing backspace at the very start of the document

_No response_

### Dependency Updates

Jian-Zhang08:
## Changes Overview

Pressing <kbd>Backspace</kbd> at the very start of the document with a leading block atom (e.g. an image) crashed the editor with `TypeError: Cannot read properties of undefined (reading 'type')`. This adds a guard so backspace is a no-op at the document start instead of crashing.

## Implementation Approach

The blockquote `handleBackspace` helper assumed the caret sits inside a textblock and read `$from.node($from.depth - 1)`. At the very start of the document the caret resolves to the top (doc) level, where `$from.depth` is `0`, so `parentDepth` is `-1` and `$from.node(-1)` is `undefined`. Added `if (parentDepth < 0) return false`.

## Testing Done

Regression test added; fails on main with the exact TypeError, passes with fix. Build (73 pkgs), blockquote tests, lint, format, fallow audit all green.

## AI Usage

- [x] I have used AI tools in creating this PR.
  - Used Claude to locate the root cause, write the regression test, and implement the guard; verified locally.

## Checklist

- [x] Changeset created.
- [x] Changes do not break the library.
- [x] Tests added.
- [x] Followed project guidelines.
- [x] Lint issues fixed.

## Related Issues

Fixes #7973

### DIFF
--- a/.changeset/fix-blockquote-backspace-doc-start.md
+++ b/.changeset/fix-blockquote-backspace-doc-start.md
@@ -2,4 +2,4 @@
 "@tiptap/extension-blockquote": patch
 ---
 
-Fix a crash when pressing backspace at the very start of the document with a leading image. The blockquote backspace handler dereferenced an undefined parent at the top (doc) level, throwing `TypeError: Cannot read properties of undefined (reading 'type')`. It now bails out so backspace is a no-op at the document start. (#7973)
+Fix a crash when pressing backspace at the very start of the document with a leading image. The blockquote backspace handler dereferenced an undefined parent at the top (doc) level, throwing `TypeError: Cannot read properties of undefined (reading 'type')`. It now bails out so backspace is a no-op at the document start.
--- a/.changeset/fix-blockquote-backspace-doc-start.md
+++ b/.changeset/fix-blockquote-backspace-doc-start.md
@@ -1,0 +1,5 @@
+---
+"@tiptap/extension-blockquote": patch
+---
+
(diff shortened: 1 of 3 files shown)


arnaugomez (dismissed):

arnaugomez:
Good catch, thank you very much for your contribution ❤️

bdbch (approved):

bdbch:
Thanks for your contribution!

PR #7979, opened 2026-06-26
fix(react): do not throw when the Tiptap provider editor is null

### LINKED ISSUE #7529: `Tiptap` component throws when `editor` is null despite optional type definition — undocumented breaking change from v3.18
### Affected Packages

react

### Version(s)

3.19.0, 3.20.0

### Bug Description

The `Tiptap` component throws an error when `editor` is `null` (the initial return value of `useEditor()`), even though the TypeScript type defines it as optional (`editor?: Editor`).

This is also an **undocumented breaking change** from v3.18, where `TiptapWrapper` handled `null` gracefully internally.

**Runtime error:**
```
Error: Tiptap: An editor instance is required. Pass a non-null `editor` prop.
    at TiptapWrapper
```

**Type definition allows `undefined`:**
```ts
type TiptapWrapperProps = {
  editor?: Editor; // optional — implies undefined is acceptable
}
```

**But the runtime throws immediately:**
```js
const resolvedEditor = editor != null ? editor : instance;
if (!resolvedEditor) {
  throw new Error("Tiptap: An editor instance is required. Pass a non-null `editor` prop.");
}
```

`useEditor()` returns `null` on the first render before the editor is initialized. This makes the following pattern throw on every mount:

```tsx
const editor = useEditor({ ... }); // returns null on first render

return (
  <Tiptap editor={editor ?? undefined}> {/* throws */}
    <Tiptap.Content />
  </Tiptap>
);
```

**Why this is a breaking change from v3.18:**

In v3.18, `TiptapWrapper` handled `null` gracefully with internal `isReady` state:

```js
// v3.18 — null was handled internally, no throw
function TiptapWrapper({ instance, children }) {
  const [isReady, setIsReady] = useState(instance?.isInitialized ?? false);
  useEffect(() => {
    if (!instance) {
      setIsReady(false);
      return;
    }
    instance.on('create', () => setIsReady(true));
  });
}
```

In v3.19 this was replaced with a hard throw. The changelog mentions *"guaranteed non-null editor instance from useTiptap hook"* but does not document that callers must now guard against `null` before rendering `<Tiptap>`. The official React installation docs for SPAs also make no mention of this requirement.

### Browser Used

Chrome

### Code Example URL

_No response_

### Expected Behavior

Either:

1. **Fix the type** to reflect the actual requirement: `editor: Editor` (required, not optional)
2. **Or restore graceful null handling** — render nothing until the editor is ready, as v3.18 did

And add a **migration note** for v3.18 → v3.19 that explicitly states the null guard is now the caller's responsibility:

```tsx
// Required guard in v3.19+ — not mentioned in docs
const editor = useEditor({ ... });

if (!editor) return null;

return (
  <Tiptap editor={editor}>
    <Tiptap.Content />
  </Tiptap>
);
```

_No response_

### Dependency Updates

Jian-Zhang08:
## Changes Overview

`<Tiptap editor={editor}>` (the `@tiptap/react` provider) threw `Tiptap: An editor instance is required` whenever `editor` was `null`. Because `useEditor()` returns `null` on the first render (and with `immediatelyRender: false`), the canonical `<Tiptap editor={useEditor(...)}>` usage crashed on mount. Regression from v3.18.

This renders nothing until the editor is ready instead of throwing — matching the optional `editor?` type, `EditorProvider`, and `EditorContent`. Fixes #7529.

## Implementation Approach

`TiptapWrapper` threw before its `useMemo` hooks. (1) Move the guard after the hooks and `return null` (Rules of Hooks), mirroring `EditorProvider`. (2) Widen `editor`/`instance` prop types to `Editor | null` to match `useEditor()`.

## Testing Done

- Regression unit test `packages/react/src/Tiptap.spec.ts`: fails on main with the exact error, passes with the fix.
- `pnpm vitest run packages/react` -> 5 passing (no regression)
- `pnpm build` (73 packages) -> success
- lint / format / `fallow:audit` -> clean

## AI Usage

- [x] I have used AI tools in creating this PR.
  - Used Claude to locate the throw, confirm the pattern against `EditorProvider`, write the test, and implement the fix; verified locally.

## Checklist

- [x] Changeset created.
- [x] Changes do not break the library.
- [x] Tests added.
- [x] Followed project guidelines.
- [x] Lint fixed.

## Related Issues

Fixes #7529

### DIFF
--- a/.changeset/fix-react-tiptap-null-editor.md
+++ b/.changeset/fix-react-tiptap-null-editor.md
@@ -1,0 +1,5 @@
+---
+"@tiptap/react": patch
+---
+
+Fix `<Tiptap>` throwing `Tiptap: An editor instance is required` when `editor` is `null`. `useEditor()` returns `null` on the first render (and when `immediatelyRender: false` is used), so the canonical `<Tiptap editor={useEditor(...)}>` usage threw. The provider now renders nothing until the editor is ready, matching `EditorProvider` and the optional `editor` prop type. The `editor` and `instance` props are also typed as `Editor | null` to match what `useEditor()` returns. (#7529)
--- a/packages/react/src/Tiptap.spec.ts
+++ b/packages/react/src/Tiptap.spec.ts
@@ -1,0 +1,23 @@
+import { render } from '@testing-library/react'
+import React from 'react'
+import { describe, expect, it } from 'vitest'
+
+import { Tiptap } from './Tiptap.js'
+
+describe('Tiptap provider', () => {
+  // Regression test for https://github.com/ueberdosis/tiptap/issues/7529
+  // `useEditor()` returns `null` on the first render (and with
+  // `immediatelyRender: false`), so the canonical `<Tiptap editor={editor}>`
+  // usage must not throw. It should render nothing until the editor is ready.
+  it('renders nothing instead of throwing when the editor is null', () => {
+    let container: HTMLElement | undefined
+
+    expect(() => {
+      container = render(
+        React.createElement(Tiptap, { editor: null }, React.createElement(Tiptap.Content)),
+      ).container
+    }).not.toThrow()
+
+    expect(container?.firstChild).toBeNull()
+  })
+})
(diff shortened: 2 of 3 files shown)


arnaugomez:
We want the `editor` props to be required, it should not be optional. See this comment: https://github.com/ueberdosis/tiptap/issues/7529#issuecomment-4808026716

I've created a new pull request now that throws a type error when the editor field is not provided:

https://github.com/ueberdosis/tiptap/pull/7983

So I close this PR. Thanks anyway for your contribution, it helped me understand the issue.

PR #7983, opened 2026-06-26
Make either editor or instance field required

### LINKED ISSUE #7529: `Tiptap` component throws when `editor` is null despite optional type definition — undocumented breaking change from v3.18

arnaugomez:
## Changes Overview

Fixes #7529

Although the `Tiptap` component types accept `undefined` as a possible value of the `editor` props, it throws an error when the `editor` props is `undefined`.

Currently, this is like this because there are two fields that allow you to pass the editor: the `editor` field an the deprecated `instance` field. Developers are expected to pass the editor instance through either one field or the other. 

This PR changes the type definition so that either the `editor` field or the deprecated `instance` field are required.

## Implementation Approach

Uses a TypeScript type union to make the types stricter.

The error is no longer thrown at runtime, because it is caught ahead of time by the TypeScript compiler.

## Testing Done

Tests still pass. There is no change in functionality, only the types change.

## AI Usage

No AI assistance used

## Checklist

## Related Issues

#7529

### DIFF
--- a/packages/react/src/Tiptap.tsx
+++ b/packages/react/src/Tiptap.tsx
@@ -93,21 +93,25 @@ export function useTiptapState<TSelectorResult>(
   })
 }
 
+export type TiptapWrapperEditorInstanceProps =
+  | {
+      /**
+       * The editor instance to provide to child components.
+       * Use `useEditor()` to create this instance.
+       */
+      editor: Editor
+    }
+  | {
+      /**
+       * @deprecated Use `editor` instead. Will be removed in the next major version.
+       */
+      instance: Editor
+    }
+
 /**
  * Props for the `Tiptap` root/provider component.
  */
-export type TiptapWrapperProps = {
-  /**
-   * The editor instance to provide to child components.
-   * Use `useEditor()` to create this instance.
-   */
-  editor?: Editor
-
-  /**
-   * @deprecated Use `editor` instead. Will be removed in the next major version.
-   */
-  instance?: Editor
-
+export type TiptapWrapperProps = TiptapWrapperEditorInstanceProps & {
   children: ReactNode
 }
 
@@ -138,12 +142,8 @@ export type TiptapWrapperProps = {
  * }
  * ```
  */
-export function TiptapWrapper({ editor, instance, children }: TiptapWrapperProps) {
-  const resolvedEditor = editor ?? instance
-
-  if (!resolvedEditor) {
-    throw new Error('Tiptap: An editor instance is required. Pass a non-null `editor` prop.')
-  }
+export function TiptapWrapper({ children, ...props }: TiptapWrapperProps) {
+  const resolvedEditor = 'editor' in props ? props.editor : props.instance
 
   const tiptapContextValue = useMemo<TiptapContextType>(
     () => ({ editor: resolvedEditor }),
--- a/packages/react/src/Tiptap.tsx
+++ b/packages/react/src/Tiptap.tsx
@@ -144,6 +144,10 @@ export type TiptapWrapperProps = TiptapWrapperEditorInstanceProps & {
  */
 export function TiptapWrapper({ children, ...props }: TiptapWrapperProps) {
   const resolvedEditor = 'editor' in props ? props.editor : props.instance
+  
+  if (!resolvedEditor) {
+    throw new Error('Tiptap: An editor instance is required. Pass a non-null `editor` prop.')
+  }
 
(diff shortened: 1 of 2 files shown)


packages/react/src/Tiptap.tsx
```diff
@@ -138,12 +142,8 @@ export type TiptapWrapperProps = {
  * }
  * ```
  */
-export function TiptapWrapper({ editor, instance, children }: TiptapWrapperProps) {
-  const resolvedEditor = editor ?? instance
-
-  if (!resolvedEditor) {
-    throw new Error('Tiptap: An editor instance is required. Pass a non-null `editor` prop.')
```
ozdemircibaris, line 145: AI found something -> Nice use of the union to force editor-or-instance at compile time. One gap worth a look: `useEditor({ extensions })` (no `immediatelyRender`) is typed `Editor` but returns `null` on the first render under Next.js auto-detection, so `<Tiptap editor={editor}>` compiles yet passes `null` here. With the throw removed, that null reaches `TiptapContext` (typed non-null) and crashes consumers like `useTiptap().editor.chain()` instead of the old clear error. Could we keep a small runtime guard (or render nothing while null) for that case?

arnaugomez: Ok I'll bring back the clause that throws the error

packages/react/src/Tiptap.tsx
```diff
@@ -138,12 +142,8 @@ export type TiptapWrapperProps = {
...
-  }
+export function TiptapWrapper({ children, ...props }: TiptapWrapperProps) {
+  const resolvedEditor = 'editor' in props ? props.editor : props.instance
```
arnaugomez, line 146:
```suggestion
  const resolvedEditor = 'editor' in props ? props.editor : props.instance
  
  if (!resolvedEditor) {
    throw new Error('Tiptap: An editor instance is required. Pass a non-null `editor` prop.')
  }
```

ozdemircibaris (approved):
LGTM

PR #7985, opened 2026-06-27
fix(extension-emoji): fix arrow key navigation past emoji nodes in Firefox

### LINKED ISSUE #7980: Emoji at the start of a paragraph block the cursor
### Affected Packages

core, extension-emoji

### Version(s)

3.27.1

### Bug Description

Really a problem with any `inline: true` & `selectable: false` node, but the emoji extension displays this most conveniently.

When such a node appears at the beginning of a paragraph, pressing the left arrow key with the cursor at the right of the node does not allow the cursor to go to the start of the line:

Image

### Browser Used

Firefox

### Code Example URL

https://tiptap.dev/docs/editor/extensions/nodes/emoji

### Expected Behavior

Pressing the left arrow should skip over the node and place the cursor at the beginning of the paragraph. Pressing the Home key does give the expected behavior:

Image

Maybe related: you also can't use arrow keys to navigate between emoji if they are right next to each other (no space in between); pressing an arrow key just skips over the whole group.

Mouse navigation works as expected in all cases, but this odd keyboard navigation behavior is tripping my users up 😢

### Dependency Updates

huzaifahGoher:
## Changes Overview

Fix arrow key navigation past inline non-selectable emoji nodes in Firefox. Added a `handleKeyDown` plugin prop that intercepts ArrowLeft/ArrowRight and manually resolves the correct cursor position when adjacent to an emoji node.

## Implementation Approach

Added a Firefox-only `handleKeyDown` handler to the existing emoji ProseMirror plugin. When the cursor is adjacent to a non-selectable inline node and the user presses an arrow key, the handler computes the correct `TextSelection` and dispatches it manually — bypassing Firefox's broken native cursor movement. Uses the existing `isFirefox()` utility from `@tiptap/core`. Bounds-checked with `$pos.start()` and `$pos.end()` to avoid jumping outside the parent paragraph.

## Testing Done

- Unit tests pass (`pnpm vitest run packages/extension-emoji/__tests__` — 6/6 passing)
- Added 2 e2e tests in `demos/src/Nodes/Emoji/index.spec.ts` for arrow left/right navigation
- Manually verified in Firefox on the emoji demo at `localhost:3000/preview/Nodes/Emoji`
- Chrome behavior unchanged (handler early-returns via `isFirefox()`)

## Verification Steps

1. Run `pnpm dev` and open http://localhost:3000/preview/Nodes/Emoji/React/ in Firefox
2. Insert an emoji at the start of a paragraph (type `:smile:`)
3. Place cursor right after the emoji
4. Press ArrowLeft — cursor should now move to position 0 (before the emoji)
5. Also test with adjacent emojis — arrow keys should navigate between them one at a time

## Additional Notes

This is a known browser-level issue with Firefox's handling of `contenteditable` and inline non-editable elements. ProseMirror acknowledged it in [issue #1220](https://github.com/ProseMirror/prosemirror/issues/1220) as a browser bug they cannot fix at their level. This fix handles it at the Tiptap extension level.

## AI Usage

  - I used AI to research the root cause across ProseMirror issues and forums
  - I used AI to assist with implementation of the handleKeyDown fix
  - I used AI to draft the e2e test cases

## Checklist

## Related Issues

Closes #7980

### DIFF
--- a/demos/src/Nodes/Emoji/index.spec.ts
+++ b/demos/src/Nodes/Emoji/index.spec.ts
@@ -27,6 +27,54 @@ test.describe(`${demoPath}/${demoName}`, () => {
         await page.getByRole('button', { name: 'Insert ⚡' }).click()
         await expect(page.locator('.tiptap [data-type="emoji"][data-name="zap"]')).toHaveCount(1)
       })
+
+      test('arrow left navigates past emoji at start of paragraph', async ({ page }) => {
+        // Set content with an emoji at the start followed by text
+        await setEditorContent(
+          page,
+          '<p><span data-type="emoji" data-name="smile">😄</span>hello</p>',
+        )
+        const editor = await getEditor(page)
+
+        // Place cursor right after the emoji (before "hello")
+        await editor.click()
+        await page.keyboard.press('Home')
+        await page.keyboard.press('ArrowRight')
+
+        // Now press ArrowLeft — cursor should move before the emoji (position 0)
+        await page.keyboard.press('ArrowLeft')
+
+        // Type a character to verify cursor is at position 0 (before emoji)
+        await page.keyboard.type('X')
+
+        // The "X" should appear before the emoji
+        const html = await editor.evaluate((el: any) => el.editor.getHTML())
+        expect(html).toMatch(/X.*data-type="emoji"/)
+      })
+
+      test('arrow right navigates past emoji node', async ({ page }) => {
+        // Set content with text followed by an emoji
+        await setEditorContent(
+          page,
+          '<p>hi<span data-type="emoji" data-name="zap">⚡</span>there</p>',
+        )
+        const editor = await getEditor(page)
+
+        // Place cursor at start and move right past "hi" to be just before the emoji
+        await editor.click()
+        await page.keyboard.press('Home')
+        await page.keyboard.press('ArrowRight')
+        await page.keyboard.press('ArrowRight')
+
+        // Now press ArrowRight — cursor should skip over the emoji
+        await page.keyboard.press('ArrowRight')
+
+        // Type a character to verify cursor is after the emoji (before "there")
+        await page.keyboard.type('Y')
+
+        const html = await editor.evaluate((el: any) => el.editor.getHTML())
+        expect(html).toMatch(/data-type="emoji".*Y.*there/)
+      })
     })
   })
 })
--- a/packages/extension-emoji/src/emoji.ts
+++ b/packages/extension-emoji/src/emoji.ts
@@ -4,13 +4,14 @@ import {
   findChildrenInRange,
   getChangedRanges,
   InputRule,
+  isFirefox,
   mergeAttributes,
   Node,
   nodeInputRule,
   PasteRule,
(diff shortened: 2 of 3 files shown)


demos/src/Nodes/Emoji/index.spec.ts
arnaugomez, line 1: E2E tests only run in Chromium in CI. Did you run this e2e test locally with Firefox as the browser?

packages/extension-emoji/src/emoji.ts
```diff
@@ -385,8 +386,72 @@ export const Emoji = Node.create<EmojiOptions, EmojiStorage>({
...
+            if (isLeft) {
+              const nodeBefore = $pos.nodeBefore
+
+              if (
+                !nodeBefore ||
+                !nodeBefore.isInline ||
+                nodeBefore.type.spec.selectable !== false
+              ) {
+                return false
+              }
```
arnaugomez: The code handles any inline node with selectable: false, not just emoji. That means editors with Emoji installed may get changed Firefox arrow behavior for mentions/hard breaks/custom inline atoms, while editors without Emoji still have the same bug for those nodes. Since the linked issue says this affects any inline and unselectable node, this probably belongs in a core/keymap-level workaround, or the emoji extension should restrict the check to `node.type === this.type`.

packages/extension-emoji/src/emoji.ts
```diff
@@ -385,8 +386,72 @@ export const Emoji = Node.create<EmojiOptions, EmojiStorage>({
...
+          // at paragraph boundaries using arrow keys. We intercept and
+          // manually resolve the correct position.
+          // See: https://github.com/ProseMirror/prosemirror/issues/1220
+          handleKeyDown(view, event) {
+            if (!isFirefox()) {
+              return false
+            }
+
+            const isLeft = event.key === 'ArrowLeft'
+            const isRight = event.key === 'ArrowRight'
```
arnaugomez, line 401: Minor issue:

The handler only checks `event.key === 'ArrowLeft'/'ArrowRight'`, so Shift+Arrow, Alt+Arrow, Ctrl+Arrow, and Meta+Arrow are treated like plain arrows when adjacent to an emoji/non-selectable inline node. That can break expected selection extension and word/line navigation by dispatching a collapsed TextSelection and returning true.

Suggestion: skip this behavior if a modifier key is used. Document it in a comment.

arnaugomez (changes requested):

demos/src/Nodes/Emoji/index.spec.ts
huzaifahGoher: Yes, e2e testing was done for both, chromium and firefox.

packages/extension-emoji/src/emoji.ts
huzaifahGoher: Noted, this check has been added now.

packages/extension-emoji/src/emoji.ts
huzaifahGoher: Noted, In case of modifier keys, this will be skipped.

arnaugomez (approved):

PR #7989, opened 2026-06-29
fix(bubble-menu,floating-menu): apply middlewareData.arrow coordinates to position arrow element

### LINKED ISSUE #7964: BubbleMenu: arrow option does not apply middlewareData.arrow coordinates to position the arrow element
### Affected Packages

extension-bubble-menu

### Version(s)

3.27.1

### Bug Description

The BubbleMenu (and FloatingMenu) plugin accepts an arrow option that feeds the @floating-ui/dom arrow middleware. However, after computePosition resolves, the plugin never reads middlewareData.arrow to position the arrow element (see https://floating-ui.com/docs/arrow#usage).

This means passing `arrow: { element: myArrowEl }` only affects placement calculations (preventing the floating element from overlapping the arrow), but the arrow element itself is never repositioned and stays at its initial CSS position.

**Suggested fix**

After applying x/y to the floating element, apply arrow coordinates:

```ts
if (middlewareData.arrow && this.floatingUIOptions.arrow) {
  const { x: arrowX, y: arrowY } = middlewareData.arrow
  const arrowEl = this.floatingUIOptions.arrow?.element

  if (arrowEl instanceof HTMLElement) {
    arrowEl.style.left = arrowX != null ? `${arrowX}px` : ''
    arrowEl.style.top = arrowY != null ? `${arrowY}px` : ''
  }
}
```

Additionally, the @tiptap/react package could export a convenience component that wraps FloatingArrow from @floating-ui/react, but that would require exposing the Floating UI context.

### Browser Used

Other

### Code Example URL

_No response_

### Expected Behavior

The arrow element should be positioned using middlewareData.arrow.x / middlewareData.arrow.y after each computePosition call, similar to what FloatingArrow from @floating-ui/react does internally.

_No response_

### Dependency Updates

codewithsupra:
## Fixes

Fixes #7964

## Changes and Review

- Bubble/floating menu arrows now read `middlewareData.arrow` from Floating UI instead of ignoring it, and use `offsetWidth`/`offsetHeight` correctly per placement axis.
- Regression test asserts correct arrow offset for each placement.

## Checklist

### Responsibility

### DIFF
--- a/packages/extension-bubble-menu/src/bubble-menu-plugin.ts
+++ b/packages/extension-bubble-menu/src/bubble-menu-plugin.ts
@@ -548,8 +548,14 @@ export class BubbleMenuView implements PluginView {
 
           if (!staticSide) return
 
+          // Offset the arrow by half its size along the static side's axis.
+          // For top/bottom the static side is vertical, so use the height;
+          // for left/right it is horizontal, so use the width.
+          const arrowSize =
+            side === 'left' || side === 'right' ? arrowEl.offsetWidth : arrowEl.offsetHeight
+
           const styles: Partial<CSSStyleDeclaration> = {
-            [staticSide]: `${-(arrowEl.offsetHeight / 2)}px`,
+            [staticSide]: `${-(arrowSize / 2)}px`,
           }
 
           if (arrowX != null) {
--- a/packages/extension-bubble-menu/src/bubble-menu-plugin.ts
+++ b/packages/extension-bubble-menu/src/bubble-menu-plugin.ts
@@ -483,7 +483,7 @@ export class BubbleMenuView implements PluginView {
       placement: this.floatingUIOptions.placement,
       strategy: this.floatingUIOptions.strategy,
       middleware: this.middlewares,
-    }).then(({ x, y, strategy, middlewareData }) => {
+    }).then(({ x, y, strategy, placement, middlewareData }) => {
       if (!this.isVisible || this.editor.isDestroyed || !this.element.isConnected) {
         return
       }
@@ -499,6 +499,25 @@ export class BubbleMenuView implements PluginView {
       this.element.style.position = strategy
       this.element.style.left = `${x}px`
       this.element.style.top = `${y}px`
+
+      // Apply arrow position if the arrow middleware is configured
+      if (this.floatingUIOptions.arrow && middlewareData.arrow) {
+        const arrowEl = (this.floatingUIOptions.arrow as Parameters<typeof arrow>[0]).element as HTMLElement
+
+        if (arrowEl) {
+          const { x: arrowX, y: arrowY } = middlewareData.arrow
+          const side = placement.split('-')[0] as 'top' | 'right' | 'bottom' | 'left'
+          const staticSide = { top: 'bottom', right: 'left', bottom: 'top', left: 'right' }[side]
+
+          Object.assign(arrowEl.style, {
+            left: arrowX != null ? `${arrowX}px` : '',
+            top: arrowY != null ? `${arrowY}px` : '',
+            right: '',
+            bottom: '',
+            [staticSide]: `${-(arrowEl.offsetHeight / 2)}px`,
+          })
+        }
+      }
 
       if (this.isVisible && this.floatingUIOptions.onUpdate) {
         this.floatingUIOptions.onUpdate()
--- a/packages/extension-bubble-menu/src/bubble-menu-plugin.ts
+++ b/packages/extension-bubble-menu/src/bubble-menu-plugin.ts
@@ -500,9 +500,11 @@ export class BubbleMenuView implements PluginView {
       this.element.style.left = `${x}px`
       this.element.style.top = `${y}px`
 
-      // Apply arrow position if the arrow middleware is configured
-      if (this.floatingUIOptions.arrow && middlewareData.arrow) {
-        const arrowEl = (this.floatingUIOptions.arrow as Parameters<typeof arrow>[0]).element as HTMLElement
+      // Apply arrow position if the arrow middleware is configured with a static element.
+      // Skip Derivable (function) forms — the caller manages positioning in that case.
+      const arrowOption = this.floatingUIOptions.arrow
+      if (arrowOption && typeof arrowOption !== 'function' && middlewareData.arrow) {
+        const arrowEl = arrowOption.element as HTMLElement | null
 
         if (arrowEl) {
           const { x: arrowX, y: arrowY } = middlewareData.arrow
--- a/packages/extension-bubble-menu/src/bubble-menu-plugin.ts
+++ b/packages/extension-bubble-menu/src/bubble-menu-plugin.ts
@@ -175,6 +175,13 @@ export class BubbleMenuView implements PluginView {
 
   private scrollTarget: HTMLElement | Window = window
 
+  // Captures the resolved `.element` when `options.arrow` is a Derivable
+  // (a function of MiddlewareState). We can't re-invoke that function ourselves
+  // outside the middleware pipeline — it needs `rects`/`elements`/`platform`,
+  // which aren't available after `computePosition()` resolves — so instead we
+  // intercept it at the point floating-ui itself calls it.
+  private resolvedArrowElement: HTMLElement | null = null
+
   private floatingUIOptions: NonNullable<BubbleMenuPluginProps['options']> = {
     strategy: 'absolute',
     placement: 'top',
@@ -254,7 +261,25 @@ export class BubbleMenuView implements PluginView {
     }
 
     if (this.floatingUIOptions.arrow) {
-      middlewares.push(arrow(this.floatingUIOptions.arrow))
+      const arrowOption = this.floatingUIOptions.arrow
+
+      if (typeof arrowOption === 'function') {
+        // Wrap the Derivable so we can capture the ArrowOptions it resolves to
+        // (specifically `.element`) at the moment floating-ui invokes it with
+        // the real MiddlewareState — we have no other way to get that element.
+        middlewares.push(
+          arrow(state => {
+            const resolved = arrowOption(state)
+
+            this.resolvedArrowElement = resolved.element as HTMLElement | null
+
+            return resolved
+          }),
+        )
+      } else {
+        this.resolvedArrowElement = arrowOption.element as HTMLElement | null
+        middlewares.push(arrow(arrowOption))
+      }
     }
 
     if (this.floatingUIOptions.size) {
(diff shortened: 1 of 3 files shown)


packages/extension-bubble-menu/src/bubble-menu-plugin.ts
```diff
@@ -500,6 +500,27 @@ export class BubbleMenuView implements PluginView {
...
+      // Apply arrow position if the arrow middleware is configured with a static element.
+      // Skip Derivable (function) forms — the caller manages positioning in that case.
+      const arrowOption = this.floatingUIOptions.arrow
+      if (arrowOption && typeof arrowOption !== 'function' && middlewareData.arrow) {
+        const arrowEl = arrowOption.element as HTMLElement | null
+
+        if (arrowEl) {
+          const { x: arrowX, y: arrowY } = middlewareData.arrow
+          const side = placement.split('-')[0] as 'top' | 'right' | 'bottom' | 'left'
+          const staticSide = { top: 'bottom', right: 'left', bottom: 'top', left: 'right' }[side]
```
arnaugomez: Would it be possible to handle the case where side is not one of `'top' | 'right' | 'bottom' | 'left'` ?

packages/extension-floating-menu/src/floating-menu-plugin.ts
```diff
@@ -507,6 +507,27 @@ export class FloatingMenuView {
...
+      // Apply arrow position if the arrow middleware is configured with a static element.
+      // Skip Derivable (function) forms — the caller manages positioning in that case.
+      const arrowOption = this.floatingUIOptions.arrow
+      if (arrowOption && typeof arrowOption !== 'function' && middlewareData.arrow) {
+        const arrowEl = arrowOption.element as HTMLElement | null
+
+        if (arrowEl) {
+          const { x: arrowX, y: arrowY } = middlewareData.arrow
+          const side = placement.split('-')[0] as 'top' | 'right' | 'bottom' | 'left'
+          const staticSide = { top: 'bottom', right: 'left', bottom: 'top', left: 'right' }[side]
```
arnaugomez: Would it be possible to handle the case where side is not one of 'top' | 'right' | 'bottom' | 'left' ?

packages/extension-floating-menu/src/floating-menu-plugin.ts
```diff
@@ -507,6 +507,27 @@ export class FloatingMenuView {
...
+
+          Object.assign(arrowEl.style, {
+            left: arrowX != null ? `${arrowX}px` : '',
+            top: arrowY != null ? `${arrowY}px` : '',
+            right: '',
+            bottom: '',
```
arnaugomez: Why are these styles applied?

packages/extension-floating-menu/src/floating-menu-plugin.ts
```diff
@@ -507,6 +507,27 @@ export class FloatingMenuView {
...
+        const arrowEl = arrowOption.element as HTMLElement | null
+
+        if (arrowEl) {
+          const { x: arrowX, y: arrowY } = middlewareData.arrow
+          const side = placement.split('-')[0] as 'top' | 'right' | 'bottom' | 'left'
+          const staticSide = { top: 'bottom', right: 'left', bottom: 'top', left: 'right' }[side]
+
+          Object.assign(arrowEl.style, {
+            left: arrowX != null ? `${arrowX}px` : '',
+            top: arrowY != null ? `${arrowY}px` : '',
```
arnaugomez: Wouldn't it be better to not assign `left` and `right` altogether, instead of assigning them the `""` value?

packages/extension-bubble-menu/src/bubble-menu-plugin.ts
```diff
@@ -500,6 +500,27 @@ export class BubbleMenuView implements PluginView {
       this.element.style.left = `${x}px`
       this.element.style.top = `${y}px`
 
+      // Apply arrow position if the arrow middleware is configured with a static element.
+      // Skip Derivable (function) forms — the caller manages positioning in that case.
+      const arrowOption = this.floatingUIOptions.arrow
+      if (arrowOption && typeof arrowOption !== 'function' && middlewareData.arrow) {
```
arnaugomez: I see that the case where arrowOption is a function is not handled. Why is that?

arnaugomez (changes requested):

packages/extension-bubble-menu/src/bubble-menu-plugin.ts
```diff
@@ -500,6 +500,37 @@ export class BubbleMenuView implements PluginView {
...
+            [staticSide]: `${-(arrowEl.offsetHeight / 2)}px`,
+          }
+
+          // Reset the opposite axes so stale values don't linger
+          const opposites: Record<string, string> = { top: 'bottom', bottom: 'top', left: 'right', right: 'left' }
+          ;(['top', 'bottom', 'left', 'right'] as const).forEach(prop => {
+            if (prop !== staticSide && prop !== opposites[staticSide]) {
+              arrowEl.style[prop] = ''
+            }
+          })
```
arnaugomez:
```suggestion
          // Reset the opposite axes so stale values don't linger
          const sidePairs = [
            ['top', 'bottom'],
            ['bottom', 'top'],
            ['left', 'right'],
            ['right', 'left'],
          ] as const

          for (const [prop, opposite] of sidePairs) {
            if (prop !== staticSide && opposite !== staticSide) {
              arrowEl.style.removeProperty(prop)
            }
          }
```

arnaugomez (changes requested):

packages/extension-bubble-menu/src/bubble-menu-plugin.ts
arnaugomez: Refactored the code:
- Made opposites more explicit by encoding them as tuples instead of as an array and a map.
- Used for..of loop instead of forEach method, which is faster and contains less boilerplate.
- Used removeProperty instead of assigning property to "". This makes the intent more explicit.

packages/extension-bubble-menu/src/bubble-menu-plugin.ts
```diff
@@ -500,6 +525,49 @@ export class BubbleMenuView implements PluginView {
...
+
+          for (const [prop, opposite] of sidePairs) {
+            if (prop !== staticSide && opposite !== staticSide) {
+              arrowEl.style.removeProperty(prop)
+            }
+          }
+
+          Object.assign(arrowEl.style, styles)
+        }
+      }
```
arnaugomez, line 570: Arrow styles can remain stale when placement flips axes.

The cleanup keeps the opposite of staticSide, but that opposite is exactly the stale dynamic coordinate from the previous axis. Example: if placement changes from left to top, the old top value can remain while the code adds bottom, because assigning `undefined` via `Object.assign(arrowEl.style, { top: undefined })` does not reliably clear the style. This can leave both top/bottom or left/right set and misposition/stretch the arrow. The implementation should explicitly clear the inactive coordinate with `''`/`removeProperty`.

Suggested change:

```ts
// Apply arrow position if the arrow middleware is configured. For both the
// static-options and Derivable (function) forms, `resolvedArrowElement` is
// set in the `middlewares` getter — for Derivable it's captured from the
// function's return value at the point floating-ui itself calls it.
if (this.floatingUIOptions.arrow && middlewareData.arrow) {
  const arrowEl = this.resolvedArrowElement
  if (arrowEl) {
    const { x: arrowX, y: arrowY } = middlewareData.arrow
    const side = placement.split('-')[0]
    const staticSide: string | undefined = {
      top: 'bottom',
      right: 'left',
      bottom: 'top',
      left: 'right',
    }[side]

    for (const prop of ['top', 'right', 'bottom', 'left'] as const) {
      arrowEl.style.removeProperty(prop)
    }

    if (!staticSide) return
    const styles: Partial<CSSStyleDeclaration> = {
      [staticSide]: `${-(arrowEl.offsetHeight / 2)}px`,
    }
    if (arrowX != null) {
      styles.left = `${arrowX}px`
    }
    if (arrowY != null) {
      styles.top = `${arrowY}px`
    }
    Object.assign(arrowEl.style, styles)
  }
}
```

packages/extension-floating-menu/src/floating-menu-plugin.ts
```diff
@@ -507,6 +532,49 @@ export class FloatingMenuView {
...
+
+          for (const [prop, opposite] of sidePairs) {
+            if (prop !== staticSide && opposite !== staticSide) {
+              arrowEl.style.removeProperty(prop)
+            }
+          }
+
+          Object.assign(arrowEl.style, styles)
+        }
+      }
```
arnaugomez, line 577: Arrow styles can remain stale when placement flips axes.

The cleanup keeps the opposite of staticSide, but that opposite is exactly the stale dynamic coordinate from the previous axis. Example: if placement changes from left to top, the old top value can remain while the code adds bottom, because assigning `undefined` via `Object.assign(arrowEl.style, { top: undefined })` does not reliably clear the style. This can leave both top/bottom or left/right set and misposition/stretch the arrow. The implementation should explicitly clear the inactive coordinate with `''`/`removeProperty`.

Suggested change:

```ts
// Apply arrow position if the arrow middleware is configured. For both the
// static-options and Derivable (function) forms, `resolvedArrowElement` is
// set in the `middlewares` getter — for Derivable it's captured from the
// function's return value at the point floating-ui itself calls it.
if (this.floatingUIOptions.arrow && middlewareData.arrow) {
  const arrowEl = this.resolvedArrowElement
  if (arrowEl) {
    const { x: arrowX, y: arrowY } = middlewareData.arrow
    const side = placement.split('-')[0]
    const staticSide: string | undefined = {
      top: 'bottom',
      right: 'left',
      bottom: 'top',
      left: 'right',
    }[side]

    for (const prop of ['top', 'right', 'bottom', 'left'] as const) {
      arrowEl.style.removeProperty(prop)
    }

    if (!staticSide) return
    const styles: Partial<CSSStyleDeclaration> = {
      [staticSide]: `${-(arrowEl.offsetHeight / 2)}px`,
    }
    if (arrowX != null) {
      styles.left = `${arrowX}px`
    }
    if (arrowY != null) {
      styles.top = `${arrowY}px`
    }
    Object.assign(arrowEl.style, styles)
  }
}
```

arnaugomez (changes requested):
Add a changeset file

codewithsupra:
Hi @arnaugomez -- addressed in the latest commit:

1. Stale arrow styles: replaced the sidePairs cleanup with an upfront removeProperty loop over all four sides (top, right, bottom, left) before computing the new styles, so no stale coordinate from a previous axis can linger regardless of which axis the placement flips to. Applied identically in both bubble-menu-plugin.ts and floating-menu-plugin.ts.
2. Changeset: added .changeset/fix-bubble-floating-menu-arrow-positioning.md with a patch bump for both @tiptap/extension-bubble-menu and @tiptap/extension-floating-menu.
Thanks again for the thorough review!

packages/extension-bubble-menu/src/bubble-menu-plugin.ts
```diff
@@ -500,6 +525,44 @@ export class BubbleMenuView implements PluginView {
...
+          }[side]
+
+          for (const prop of ['top', 'right', 'bottom', 'left'] as const) {
+            arrowEl.style.removeProperty(prop)
+          }
+
+          if (!staticSide) return
+
+          const styles: Partial<CSSStyleDeclaration> = {
+            [staticSide]: `${-(arrowEl.offsetHeight / 2)}px`,
```
arnaugomez: The PR sets the static side with:

```
[staticSide]: `${-(arrowEl.offsetHeight / 2)}px`,
```

That is dimensionally correct for top / bottom, where the arrow is offset vertically. For left / right, the static side is horizontal (right / left), so offsetWidth / 2 is the appropriate dimension.

arnaugomez: Suggested fix:

```ts
const arrowSize =
  side === 'left' || side === 'right'
    ? arrowEl.offsetWidth
    : arrowEl.offsetHeight

const styles: Partial<CSSStyleDeclaration> = {
  [staticSide]: `${-(arrowSize / 2)}px`,
}
```

Same issue applies to the floating menu plugin

arnaugomez (changes requested):

codewithsupra:
Good catch @arnaugomez — fixed in `abe1547`. You're right that `offsetHeight / 2` is only dimensionally correct for top/bottom placements; for left/right the static side is horizontal, so it should use the width. Applied your suggested fix (pick `offsetWidth` for left/right, `offsetHeight` for top/bottom) identically in both the bubble-menu and floating-menu plugins.

arnaugomez:
@coderabbitai review

packages/extension-bubble-menu/src/bubble-menu-plugin.ts
```diff
@@ -500,6 +525,50 @@ export class BubbleMenuView implements PluginView {
...
+            arrowEl.style.removeProperty(prop)
+          }
+
+          if (!staticSide) return
+
+          // Offset the arrow by half its size along the static side's axis.
+          // For top/bottom the static side is vertical, so use the height;
+          // for left/right it is horizontal, so use the width.
+          const arrowSize =
+            side === 'left' || side === 'right' ? arrowEl.offsetWidth : arrowEl.offsetHeight
```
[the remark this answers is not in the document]
arnaugomez: Very good feedback. @codewithsupra you can fix this and I'll approve.

PR #7990, opened 2026-06-29
fix(core): re-check canSplit after deleteSelection to prevent cross-boundary crash

### LINKED ISSUE #7734: splitBlock crashes with "Inserted content deeper than insertion position" on cross-boundary selections
### Affected Packages

`core`

### Version(s)

`2.6.0` through `2.27.2`, likely also `3.22.3`

### Browser Used

All

### Bug Description

The `splitBlock` command throws `TransformError: Inserted content deeper than insertion position` when pressing Enter with a text selection that spans nested block boundaries (e.g. across list items).

The issue was introduced in v2.6.0 when `canSplit` was moved outside the `if (dispatch)` block to support the `can()` API. Before that change, both `deleteSelection` and `canSplit` ran sequentially inside `dispatch`, operating on the same document state. After the refactor, `canSplit` validates the pre-deletion document, but `split` runs on the post-deletion document. When the deletion collapses nested nodes, the pre-checked position becomes invalid.

**v2.5.0 (correct):**
```js
if (dispatch) {
    tr.deleteSelection();          // 1. delete first
    can = canSplit(tr.doc, ...);   // 2. check on resulting doc
    if (can) {
        tr.split(...);             // 3. split on same doc ✓
    }
}
```

**v2.6.0+ (broken):**
```js
can = canSplit(tr.doc, ...);       // 1. check on original doc
if (dispatch) {
    if (can) {
        tr.deleteSelection();      // 2. delete changes doc structure
        tr.split(...);             // 3. split on different doc ✗
    }
}
```

### Steps to reproduce

1. Create an editor with list support
2. Add a bullet list with multiple items
3. Click and drag to select across two or more list items
4. Press Enter
5. Console throws `TransformError: Inserted content deeper than insertion position`

### Expected Behavior

Pressing Enter with a cross-boundary selection should delete the selection and split the block (or gracefully decline to split), without throwing.

### Additional Context

A possible fix would be to move `deleteSelection` before the `canSplit` re-check inside the `dispatch` block, though this would need thorough testing:

```js
if (dispatch) {
    if (selection instanceof TextSelection) {
        tr.deleteSelection();
    }
    // Re-check after deletion since document structure may have changed
    can = canSplit(tr.doc, tr.mapping.map($from.pos), 1, types);
    if (can) {
        tr.split(tr.mapping.map($from.pos), 1, types);
        // ...
    }
}
```

This should restore the v2.5.0 ordering while preserving the `can()` pre-check behavior (the initial `canSplit` before `dispatch` still runs for that path).

It looks like the bug may also exist on v3 (`3.22.3`) because the `splitBlock` code is identical. It would be great if both branches could be patched — we're currently on v2 and the upgrade to v3 may take a while.

### Dependency Updates

- [x] I have updated all `@tiptap/*` packages and their dependencies (within v2 branch).

codewithsupra:
## Fixes

Fixes #7734

## Changes and Review

- `splitBlock` re-checks `canSplit` after `deleteSelection` mutates the document — the original check ran against now-stale positions, letting an invalid split through on cross-boundary selections.
- Added a regression test reproducing the crash; it fails with the fix reverted.

## Checklist

### Responsibility

### DIFF
--- a/packages/core/src/commands/splitBlock.ts
+++ b/packages/core/src/commands/splitBlock.ts
@@ -99,12 +99,27 @@ export const splitBlock: RawCommands['splitBlock'] =
     }
 
     if (dispatch) {
-      if (can) {
-        if (selection instanceof TextSelection) {
-          tr.deleteSelection()
-        }
+      if (selection instanceof TextSelection) {
+        tr.deleteSelection()
+      }
 
-        tr.split(tr.mapping.map($from.pos), 1, types)
+      // Re-check canSplit after deletion: deleteSelection may merge or remove nodes,
+      // making the pre-deletion position invalid for split. This restores the v2.5.0
+      // ordering where both checks operated on the same document state.
+      const mappedPos = tr.mapping.map($from.pos)
+      let canAfterDelete = canSplit(tr.doc, mappedPos, 1, types)
+
+      if (
+        !types &&
+        !canAfterDelete &&
+        canSplit(tr.doc, mappedPos, 1, deflt ? [{ type: deflt }] : undefined)
+      ) {
+        canAfterDelete = true
+        types = deflt ? [{ type: deflt, attrs: newAttributes }] : undefined
+      }
+
+      if (canAfterDelete) {
+        tr.split(mappedPos, 1, types)
 
         if (deflt && !atEnd && !$from.parentOffset && $from.parent.type !== deflt) {
           const first = tr.mapping.map($from.before())
--- a/packages/core/__tests__/splitBlock.spec.ts
+++ b/packages/core/__tests__/splitBlock.spec.ts
@@ -1,0 +1,61 @@
+import { Editor } from '@tiptap/core'
+import BulletList from '@tiptap/extension-bullet-list'
+import Document from '@tiptap/extension-document'
+import ListItem from '@tiptap/extension-list-item'
+import Paragraph from '@tiptap/extension-paragraph'
+import Text from '@tiptap/extension-text'
+import { afterEach, describe, expect, it } from 'vitest'
+
+describe('splitBlock', () => {
+  let editor: Editor
+
+  afterEach(() => {
+    editor?.destroy()
+  })
+
+  it('does not throw when splitting across list item boundaries (cross-boundary selection)', () => {
+    editor = new Editor({
+      extensions: [Document, Paragraph, Text, BulletList, ListItem],
+      content: {
+        type: 'doc',
+        content: [
+          {
+            type: 'bulletList',
+            content: [
+              {
+                type: 'listItem',
+                content: [{ type: 'paragraph', content: [{ type: 'text', text: 'first item' }] }],
+              },
+              {
+                type: 'listItem',
+                content: [{ type: 'paragraph', content: [{ type: 'text', text: 'second item' }] }],
+              },
+            ],
+          },
+        ],
+      },
+    })
+
+    // Select from inside first list item to inside second list item (cross-boundary)
+    const { doc } = editor.state
+    const firstTextPos = 3 // inside "first item"
+    const secondTextPos = doc.content.size - 4 // inside "second item"
+
+    editor.commands.setTextSelection({ from: firstTextPos, to: secondTextPos })
+
+    // This should not throw "TransformError: Inserted content deeper than insertion position"
+    expect(() => editor.commands.splitBlock()).not.toThrow()
+  })
+
+  it('splits a normal paragraph without error', () => {
+    editor = new Editor({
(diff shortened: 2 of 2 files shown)


arnaugomez:
@coderabbitai review

packages/core/src/commands/splitBlock.ts
```diff
@@ -99,12 +99,27 @@ export const splitBlock: RawCommands['splitBlock'] =
     }
 
     if (dispatch) {
-      if (can) {
-        if (selection instanceof TextSelection) {
-          tr.deleteSelection()
-        }
+      if (selection instanceof TextSelection) {
+        tr.deleteSelection()
+      }
```
arnaugomez, line 104: Why have you removed `if(can)`? Now it does not check if it can delete the selection, before it does so.

aribn:
We hit this independently and filed #8156 before finding #7734 and this PR — apologies for the duplicate. Closing ours in favor of #7734. Adding what we have here in case it's useful for review, since this has been open a while.

**It reaches production, not just lists.** We saw it from a real user in a shipped app. Our repro needs no list at all — `<p>one</p><p>two</p>` with a selection from 1 to 6 (start of line 1 to start of line 2) is enough:

```js
editor.commands.setTextSelection({ from: 1, to: 6 })
editor.commands.splitBlock() // TransformError
```

The user gesture is: click at the start of a line, extend the selection into a later line, press Enter. `deleteSelection` removes the block containing `$from`, so the position maps to the deletion's start, which resolves at depth 0 — and `split` builds a `Slice` with `openStart: 1`.

**Upgrading is not a workaround.** `splitBlock` in `3.29.2` (current `latest`) is byte-identical to `3.29.0` — we diffed the published tarballs. Anyone hitting this on 3.x has no released version to move to.

**The failure is silent.** The throw escapes before the transaction is dispatched, so there is no delete and no split — the keypress is simply lost, with no user-visible error. That is likely why it is underreported; we only found it because an uncaught-exception reporter caught the throw.

**A second message from the same line:** `TransformError: Invalid content for node orderedList`, when the selection crosses a list.

**Measurements.** We swept every valid text selection whose endpoints lie in inline content (the only kind a browser produces) across paragraph / heading / list / blockquote documents — 737 selections. 27 of them throw on current `main`.

One caveat on how that applies here: we measured a **more conservative variant** than this PR — we kept `deleteSelection` inside the `if (can)` gate and only guarded the `split` call. On that variant the sweep showed 27 crashes fixed, 0 remaining, and no behavioral change on the 710 selections that never crashed. This PR additionally moves `deleteSelection` out of the `if (can)` gate, so when the pre-delete `can` is `false` the selection now gets deleted where previously nothing happened. That is closer to `prosemirror-commands` and may well be the behavior you want — but it is a change our numbers do not cover, so please do not read them as validating this patch exactly.

One question for whoever reviews: `return can` still returns the pre-delete value, so `splitBlock` can now delete the selection and return `true` without having split. Given #5363 / #5371 went the other way on that, is that worth handling here or separately?

Happy to help however is useful — more test cases, or re-running the sweep against whichever variant you prefer.

_(Written with AI assistance.)_

packages/core/src/commands/splitBlock.ts
```diff
@@ -99,12 +99,27 @@ export const splitBlock: RawCommands['splitBlock'] =
     }
 
     if (dispatch) {
-      if (can) {
-        if (selection instanceof TextSelection) {
-          tr.deleteSelection()
-        }
+      if (selection instanceof TextSelection) {
```
bdbch, line 102: `instanceof` can cause a lot of issues. In cases where a consumer - for whatever reasons - has two different versions of prosemirror packages, this can fail and cause invalid behavior.

We should update [`isTextSelection`](https://github.com/ueberdosis/tiptap/blob/168338ba8cea30febe72a54a5190d845398b8b6e/packages/core/src/helpers/isTextSelection.ts#L3) to not use instanceof & then use it here.

packages/core/src/commands/splitBlock.ts
bdbch: I think the problem is that `can` needs to run after the deletion is applied on the transaction, so `tr.doc` includes the state after the deletion. However I believe this is also wrong to keep like that.

We can revert the step by storing the deletions replaceStep and inverse it if `canAfterDelete` is false. It's just important that the replace step is removed from the transaction and this function exits early after doing so.

bdbch (changes requested):

bdbch:
@codewithsupra I reviewed your code, the tr.delete problem is critical and we should make sure that the deletion is not run when the `canAfterDelete` is not true. Make sure to store the ReplaceStep and inverse it in the case where we exit early.

PR #7993, opened 2026-06-30
fix(static-renderer): respect unhandledNode/unhandledMark for unknown types

### LINKED ISSUE #6866: [Static Renderer]: `unhandledNode` and `unhandledMark` are ignored in `renderToReactElement`
### Affected Packages

static-renderer

### Version(s)

3.2.0

### Bug Description

The static renderer provides the options `unhandledNode` and `unhandledMark` to define a fallback when encountering an unknown node or mark. However, in the `renderToElement` function, the following code creates a Node from the content:

```js
  if (!(content instanceof Node)) {
    content = Node.fromJSON(getSchemaByResolvedExtensions(extensions), content);
  }
```

`Node.fromJSON` throws a runtime exception when encountering unknown nodes or marks. It does not use the provided `unhandledNode` and `unhandledMark`.

### Browser Used

Chrome

### Code Example URL

_No response_

### Expected Behavior

`renderToReactElement` should respect `unhandledNode` and `unhandledMark`. There is currently no point in providing these fallbacks, because `Node.fromJSON` crashes before the content is being rendered.

_No response_

### Dependency Updates

ozdemircibaris:
## Changes Overview

Fix `renderToReactElement`, `renderToHTMLString`, and `renderToMarkdown` ignoring the `unhandledNode` / `unhandledMark` options when the content contains node or mark types missing from the schema.

## Implementation Approach

`renderToElement` now attempts `Node.fromJSON` and, only when it fails because of unknown types that have matching fallbacks, renders the original JSON so the renderer routes those types to the fallbacks; structural/schema errors and unknown types without a matching fallback are re-thrown unchanged.

## Testing Done

Added `packages/static-renderer/__tests__/pm-unhandled-types.spec.ts` covering the fallback path, the preserved throw behaviour (no fallback, and a node-only unknown with only `unhandledMark`), and that structural errors are not masked; `test:unit`, `lint`, `build`, and `fallow:audit` all pass.

## Verification Steps

Call `renderToReactElement` (or `renderToHTMLString`) with JSON containing a node/mark type absent from the provided extensions plus `unhandledNode`/`unhandledMark` options — it now renders the fallbacks instead of throwing a `RangeError`.

## Additional Notes

In the fallback path the document is rendered from raw JSON, so schema attribute defaults are not materialised; this only affects documents that previously threw outright.

## AI Usage

  - I used Claude to reproduce the issue, design the fix approach, implement it, write the regression tests, and draft this description.

## Checklist

## Related Issues

Closes #6866

### DIFF
--- a/packages/static-renderer/__tests__/pm-unhandled-types.spec.ts
+++ b/packages/static-renderer/__tests__/pm-unhandled-types.spec.ts
@@ -4,7 +4,9 @@ import Document from '@tiptap/extension-document'
 import Heading from '@tiptap/extension-heading'
 import Paragraph from '@tiptap/extension-paragraph'
 import Text from '@tiptap/extension-text'
+import StarterKit from '@tiptap/starter-kit'
 import { renderToHTMLString } from '@tiptap/static-renderer/pm/html-string'
+import { renderToMarkdown } from '@tiptap/static-renderer/pm/markdown'
 import { renderToReactElement } from '@tiptap/static-renderer/pm/react'
 import React from 'react'
 import { describe, expect, it } from 'vitest'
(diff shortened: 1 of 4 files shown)


packages/static-renderer/__tests__/pm-unhandled-types.spec.ts
arnaugomez, line 1: Nitpick: Could we add a regression test for `renderToMarkdown` + unknown mark + `unhandledMark`? The PR title and description mention Markdown and mark fallbacks, but the Markdown coverage currently only exercises an unknown node.

packages/static-renderer/src/pm/extensionRenderer.ts
```diff
@@ -182,6 +182,198 @@ export function mapMarkExtensionToReactNode<T>(
...
+/**
+ * Presents a placeholder node/mark to a fallback renderer under its original type
+ * name and attributes — including a `toJSON()` that returns the restored original
+ * JSON — while keeping the live ProseMirror children/methods intact.
+ */
+function withOriginalIdentity<T extends { attrs: Record<string, any>; toJSON: () => JSONContent }>(
+  target: T,
+): T {
+  const overrides: Record<string | symbol, unknown> = Object.assign(Object.create(null), {
+    type: { name: target.attrs[ORIGINAL_TYPE_ATTR] },
```
arnaugomez, line 349: This proxy restores `type.name`, but `node.type` / `mark.type` is no longer a real ProseMirror `NodeType` / `MarkType`. That may be surprising for fallback renderers that inspect `type.spec`, `type.schema`, `isInline`, etc. Can we either document that limitation or make the fallback type shape explicit?

packages/static-renderer/src/pm/extensionRenderer.ts
```diff
@@ -182,6 +182,198 @@ export function mapMarkExtensionToReactNode<T>(
...
+// restored when it reaches `unhandledNode`/`unhandledMark`.
+const UNHANDLED_NODE_TYPE = '__tiptapUnhandledNode__'
+const UNHANDLED_MARK_TYPE = '__tiptapUnhandledMark__'
+const ORIGINAL_TYPE_ATTR = '__originalType'
+const ORIGINAL_ATTRS_ATTR = '__originalAttrs'
+
+const placeholderAttrs = {
+  [ORIGINAL_TYPE_ATTR]: { default: '' },
+  [ORIGINAL_ATTRS_ATTR]: { default: {} },
+}
```
arnaugomez, line 197: Nitpick: Potential edge case: the internal placeholder names could collide with user-defined node/mark names. But this is, to my mind, a minor issue.

packages/static-renderer/src/pm/extensionRenderer.ts
```diff
@@ -182,6 +182,198 @@ export function mapMarkExtensionToReactNode<T>(
...
+
+/**
+ * Returns true if `content` references any node or mark type absent from `schema`.
+ * Pure; does no I/O.
```
arnaugomez: Can we remove the`Pure` and `Pure; does no I/O.` comments? It seems to me that they do not add any useful information.

arnaugomez (changes requested):

arnaugomez:
My only blocking change request is removing "Pure, does not IO" comments, the rest of my suggestions are optional.

arnaugomez (approved):

PR #7995, opened 2026-07-02
fix(core): recursively convert nested lists when toggling list type

### LINKED ISSUE #7970: Toggling between list types breaks nested list structure
### Affected Packages

@tiptap/extension-list

### Version(s)

3.27.1

### Bug Description

When converting between list types using commads `toggleBulletList`, `toggleOrderedList`, `toggleTaskList`, nested list indentation/structure is not preserved.

https://github.com/user-attachments/assets/d4492491-31b2-4465-8400-6ffbc4aa6788

Example list:
```md
• List item 1
• List item 2
  • List item 2-1
  • List item 2-2
• List item 3
```

Actual result after converting to ordered list:
```md
1. List item 1
  • List item 2
2. List item 2-1
3. List item 2-2
4. List item 3
```

Expected result:
```md
1. List item 1
2. List item 2
  a. List item 2-1
  b. List item 2-2
3. List item 3
```

### Browser Used

Chrome

### Code Example URL

_No response_

### Expected Behavior

Changing list type should preserve the original nesting hierarchy and indentation.

_No response_

### Dependency Updates

codewithsupra:
## Problem

Fixes #7970.

When converting between list types on a selection that spans a list containing nested sublists (e.g. toggleBulletList to toggleOrderedList), only the top-level list node gets its type changed. Nested sublists keep their original type, breaking the visual hierarchy — matches the repro in the issue exactly.

## Root cause

In toggleList's "change list type" branch (packages/core/src/commands/toggleList.ts), tr.setNodeMarkup(currentList.pos, listType) is only called on the single node returned by findParentNode — the nearest matching ancestor list. It never descends into nested sublists inside the list items, so they keep their old node type.

## Fix

Added convertNestedLists(), which walks the list subtree via node.descendants() after the top-level conversion and converts any nested list node still on the original type. Since setNodeMarkup only changes a node's type/attrs (not its size), positions captured before the walk stay valid throughout — no remapping needed between conversions.

## Tests

Added packages/extension-list/__tests__/toggleListNestedStructure.spec.ts:
- 2-level nested list conversion (matches the issue's repro)
- - 3-level deeply nested list conversion
- - Regression check that single-level (non-nested) conversion still works as before
All three fail on main (reproducing the reported bug) and pass with this fix. Full extension-list + core suites (522 tests) pass with no regressions.

## Changeset

Added a patch-level changeset for @tiptap/core.


arnaugomez (dismissed):
LGTM but I see that the fix only applies to bulletList and orderedList. Could a fix be included for taskList, too? The issue https://github.com/ueberdosis/tiptap/issues/7970 also mentions taskList.

codewithsupra:
Good catch, thanks -- you're right, I only tested bulletList/orderedList and missed that the issue names toggleTaskList too. Pushed a second commit that covers it.

taskList has a different content model (taskItem+) than bulletList/orderedList (listItem+), so taskList <-> bulletList/orderedList conversions go through toggleList's wrapInList + clearNodes fallback branch, not the setNodeMarkup branch the first commit fixed. Digging into that branch turned up something more serious than a wrong list type: clearNodes() was lifting every non-text node in the selection (including listItem/taskItem wrapper nodes, not just textblocks) in document order, using positions mapped against a doc that earlier iterations in the same pass had already mutated. For a taskList nested inside a taskItem inside a taskList, this actually ejected the nested item's paragraph out of the list entirely -- real content loss, not just a formatting mismatch.

Fixed clearNodes() to only lift textblocks, processed in reverse document order so a deeper node's liftTarget is computed against a still-intact ancestor chain. This fixes the content-loss for all three commands named in #7970, and top-level type conversion now works correctly for taskList cross-family toggles.

One gap I want to flag rather than paper over: when a cross-family conversion has a nested sublist of the other family (e.g. a taskList nested inside what's becoming a bulletList), the nested sublist itself keeps its original type -- only the outer list converts. taskItem and listItem have incompatible content models, so fixing that requires reconstructing the nested list (new list + item nodes), not just setNodeMarkup, and I didn't want to ship an under-tested version of that today. Happy to take that on as a follow-up if you'd like it in this PR rather than a separate one.

Added packages/core/__tests__/clearNodes.spec.ts for the fixed command directly, and packages/extension-list/__tests__/toggleListTaskListCrossFamily.spec.ts for the taskList scenarios from #7970. Full monorepo suite (1368 tests) passes.

codewithsupra:
@arnaugomez the gap I flagged in my last comment is now closed — nested cross-family sublists convert fully. Latest commit adds rebuildListNode(): since taskItem and listItem are incompatible node types, setNodeMarkup can't convert across families, so the subtree is recreated node by node (items recreated as the target item type, deeper nested lists converted recursively) and swapped in with a single size-preserving replaceWith. The rebuild validates against the target content models and leaves the original untouched rather than ever producing an invalid document.

convertNestedLists now classifies each nested sublist: same family (bulletList <-> orderedList) keeps the cheap setNodeMarkup path; cross family takes the rebuild path. The wrapInList fallback branches (used when the top-level toggle itself crosses families) run the same converter over the selection afterward, so sublists inside the newly wrapped items convert too.

One deliberate behavior choice to flag for review: taskItem's checked state is dropped when converting away from the task family (a bullet list has no checked state to carry it), and converting back yields unchecked defaults. Covered by a test so the behavior is pinned either way — happy to change it if you'd prefer something else.

Tests upgraded from content-preservation to full-conversion assertions, plus new cases: cross-family sublist nested inside a same-family toggle, checked-state round-trip, and 3-level deep nesting. All 6 fail with only the test change applied (verified by reverting just toggleList.ts) and pass with the fix. Full monorepo suite: 1371 tests green. This should now cover all three commands named in #7970 completely.

bdbch:
Could you please not send in those huge LLM generated chunks of text? It's really hard to read and follow. There's no need to describe all tests by name or include all filenames in your updates. Otherwise I could just prompt my Claude to check this PR for updates.

codewithsupra:
Fair feedback — I'll keep updates shorter and more to the point going forward. Thanks for flagging it.

Issue #7958, opened 2026-06-17
@tiptap/markdown: heading inline tokens are missing after ordered list


maxpaleo:
### Affected Packages

@tiptap/markdown, @tiptap/starter-kit, @tiptap/core

### Version(s)

3.20.4

### Bug Description

When parsing Markdown with `@tiptap/markdown`, a heading that comes directly after an ordered list is parsed with its inline Markdown syntax as literal text.

For example, this Markdown:

````md
1. Plain text search
2. Exact phrase search
3. Structured filters

### **Plain text search**

```txt
conference chair
```
````

is parsed so that the heading text becomes the literal string `**Plain text search**`.

That means the editor displays:

```txt
**Plain text search**
```

inside the heading instead of displaying a bold heading:

```txt
Plain text search
```

The `###` heading marker is handled correctly, but the inline bold syntax inside the heading is not parsed.

### Browser Used

Chrome

## Expected Behavior

The heading should parse as a level 3 heading containing bold text.

Expected ProseMirror/Tiptap JSON shape:

```json
{
  "type": "heading",
  "attrs": {
    "level": 3
  },
  "content": [
    {
      "type": "text",
      "text": "Plain text search",
      "marks": [
        {
          "type": "bold"
        }
      ]
    }
  ]
}
```

## Actual Behavior

The heading is parsed with the Markdown syntax as plain text.

Actual shape:

```json
{
  "type": "heading",
  "attrs": {
    "level": 3
  },
  "content": [
    {
      "type": "text",
      "text": "**Plain text search**"
    }
  ]
}
```

### Additional Context (Optional)

This was found while building a Markdown-first editor using the native `@tiptap/markdown` package.

The issue is visible in the editor because the heading renders with literal `**` characters. If the user edits the heading manually, the live editor behavior can recover and render bold correctly, so the problem appears to be in the initial Markdown parse/import path.

### Dependency Updates

- [x] Yes, I've updated all my dependencies.

maxpaleo:
<img width="745" height="464" alt="Image" src="https://github.com/user-attachments/assets/e19691a8-0f3a-47f6-828a-54663e146b76" />

devvaannsh:
This is fixed via #7997

PR #7997, opened 2026-07-04
Heading after list item now terminates the list and gets parsed as headings

devvaannsh:
BUG: If a heading is placed on the line directly after an ordered list item (with no blank line in between), the heading gets pulled into the list item as literal text (e.g. a visible ###) instead of being parsed as a heading. 
Reference PR: https://github.com/ueberdosis/tiptap/issues/7958

FIX: A heading line now terminates lazy continuation, so the list ends and the heading is parsed properly. This matches how other markdown parsers handle it. 
Also, an indented heading inside a list item is now parsed as a real heading within that item instead of literal text. Lines that aren't valid headings for ex: (#hashtag, 7+ #s) are unaffected and remain continuation text.

Visual reference:
**Before**
Screenshot 2026-07-04 at 9 46 25 PM

**After**
Screenshot 2026-07-04 at 7 05 59 PM

## AI Usage

## Checklist

### DIFF
--- a/.changeset/fix-ordered-list-heading-lazy-continuation.md
+++ b/.changeset/fix-ordered-list-heading-lazy-continuation.md
@@ -1,0 +1,5 @@
+---
+'@tiptap/extension-list': patch
+---
+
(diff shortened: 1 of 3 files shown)


arnaugomez (approved):
Thanks for flagging and fixing this issue.

Could this also happen with other types of Markdown elements?

arnaugomez:
Could a similar bug be happening in https://tiptap.dev/docs/editor/extensions/nodes/task-list ?

devvaannsh:
Thanks for approving and merging the PR, @arnaugomez.

I don't think this bug affects other Markdown elements, as I was unable to reproduce it with any of them. It also seems to work fine with task lists.

PR #7998, opened 2026-07-05
fix(extension-list): use item's own indent as base for ordered-list markdown nesting

### LINKED ISSUE #7966: Indented ordered list markdown does not parse inline formatting correctly
### Affected Packages

ordered-list

### Version(s)

3.18, main

### Bug Description

Contents of indented ordered list items, e.g. inside an unordered list, don't parse correctly.

A minimal example is `" 1. **bold** item"` (note the leading space).

```ts
import {Node} from '@tiptap/core';
import Paragraph from '@tiptap/extension-paragraph';
import Text from '@tiptap/extension-text';
import Bold from '@tiptap/extension-bold';
import BulletList from '@tiptap/extension-bullet-list';
import OrderedList from '@tiptap/extension-ordered-list';
import ListItem from '@tiptap/extension-list-item';
import {Markdown, MarkdownManager} from '@tiptap/markdown';

const Document = Node.create({name: 'doc', topNode: true, content: 'block+'});
const manager = new MarkdownManager({
  extensions: [Document, Paragraph, Text, Bold, BulletList, OrderedList, ListItem, Markdown],
});

// no space
console.log(JSON.stringify(manager.parse('1. **bold** item').content));
// with space
console.log(JSON.stringify(manager.parse(' 1. **bold** item').content));

// output:
// [{"type":"orderedList","content":[{"type":"listItem","content":[{"type":"paragraph","content":[{"type":"text","text":"bold","marks":[{"type":"bold"}]},{"type":"text","text":" item"}]}]}]}]
// [{"type":"orderedList","content":[{"type":"listItem","content":[{"type":"paragraph","content":[{"type":"text","text":"**bold** item"}]}]}]}]
```

### Browser Used

Chrome

### Code Example URL

_No response_

### Expected Behavior

Contents of indented ordered list items should parse correctly: both examples should produce the same output.

_No response_

### Dependency Updates

codewithsupra:
Fixes #7966.

Indented ordered list items lose inline formatting when parsed as markdown. Minimal repro from the issue:

manager.parse('1. **bold** item')   -- correct: "bold" gets a bold mark
manager.parse(' 1. **bold** item')  -- broken: literal text "**bold** item", no mark

The only difference is a single leading space before the marker (the case that occurs when a top-level ordered list is itself nested inside another list).

Root cause: OrderedList's custom markdown tokenizer (packages/extension-list/src/ordered-list/ordered-list.ts) calls buildNestedStructure(listItems, 0, lexer) with a hardcoded baseIndent of 0. collectOrderedListItems() correctly records each item's actual indentation (indent: 1 for a line with one leading space), but buildNestedStructure() only includes an item if item.indent === baseIndent. With the hardcoded 0, an indented top-level item never matches its own list's base indent, so buildNestedStructure() silently returns zero items and the tokenizer bails out. That fallback path skips this extension's inline tokenizer, so bold/italic markup ends up as literal text.

I confirmed this by logging the intermediate arrays: for ' 1. **bold** item', listItems[0].indent is 1, buildNestedStructure(listItems, 0, lexer) returns [], and the tokenizer falls back -- reproducing the exact broken output from the issue.

Fix: derive baseIndent from the first collected item's own indent instead of hardcoding it to 0.

Tests: added a test to packages/extension-list/__tests__/orderedListType.spec.ts asserting that the indented and non-indented inputs parse to identical JSON, both with the bold mark applied. Verified the test actually catches the bug by reverting only the one-line fix (keeping the test) and confirming it fails with the exact broken output described in the issue, then restoring the fix and confirming it passes. Full monorepo suite (1365 tests) passes with no regressions.

Added a patch-level changeset for @tiptap/extension-list.

### DIFF
--- a/packages/extension-list/src/ordered-list/ordered-list.ts
+++ b/packages/extension-list/src/ordered-list/ordered-list.ts
@@ -243,7 +243,14 @@ export const OrderedList = Node.create<OrderedListOptions>({
         return undefined
       }
 
-      const items = buildNestedStructure(listItems, 0, lexer)
+      // The first collected item defines the indentation level of this list.
+      // Using a hardcoded 0 here meant an indented top-level list (e.g. one
+      // space before the marker, as can happen when nested inside another
+      // list) never matched its own indent in buildNestedStructure, silently
+      // produced zero items, and made this tokenizer bail out — falling back
+      // to marked's default list handling, which does not run this list's
+      // items back through this extension's inline tokenizer.
+      const items = buildNestedStructure(listItems, listItems[0].indent, lexer)
 
       if (items.length === 0) {
         return undefined
--- a/packages/extension-list/src/ordered-list/ordered-list.ts
+++ b/packages/extension-list/src/ordered-list/ordered-list.ts
@@ -243,13 +243,11 @@ export const OrderedList = Node.create<OrderedListOptions>({
         return undefined
       }
 
-      // The first collected item defines the indentation level of this list.
-      // Using a hardcoded 0 here meant an indented top-level list (e.g. one
-      // space before the marker, as can happen when nested inside another
-      // list) never matched its own indent in buildNestedStructure, silently
-      // produced zero items, and made this tokenizer bail out — falling back
-      // to marked's default list handling, which does not run this list's
-      // items back through this extension's inline tokenizer.
+      // buildNestedStructure() only includes an item when item.indent matches
+      // the base indent it's given, so the base must be this list's own
+      // indentation level (that of its first collected item, e.g. 1 for a
+      // line with a single leading space) rather than always 0 — a nested
+      // list is not necessarily flush with the start of the line.
       const items = buildNestedStructure(listItems, listItems[0].indent, lexer)
 
       if (items.length === 0) {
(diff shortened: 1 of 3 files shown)


arnaugomez (approved):

packages/extension-list/src/ordered-list/ordered-list.ts
```diff
@@ -243,7 +243,14 @@ export const OrderedList = Node.create<OrderedListOptions>({
...
 
-      const items = buildNestedStructure(listItems, 0, lexer)
+      // The first collected item defines the indentation level of this list.
+      // Using a hardcoded 0 here meant an indented top-level list (e.g. one
+      // space before the marker, as can happen when nested inside another
+      // list) never matched its own indent in buildNestedStructure, silently
+      // produced zero items, and made this tokenizer bail out — falling back
+      // to marked's default list handling, which does not run this list's
+      // items back through this extension's inline tokenizer.
+      const items = buildNestedStructure(listItems, listItems[0].indent, lexer)
```
arnaugomez: In the comment, do not mention how it was configured in the past

> Using a hardcoded 0 here meant...

Instead, explain how it is configured now and why. For example, you can explain why it passes the indentation level of the list and not a hardcoded 0, and what would happen if we passed a hardcoded 0.

arnaugomez (changes requested):

arnaugomez (approved):

PR #7999, opened 2026-07-05
Fix (extension-table): when parsing HTML respect col width from colgroup

### LINKED ISSUE #7800: the RichTextEditor content prop fails to respect the width attribute of <col> tags within the provided HTML snippet
### Affected Packages

extension-table

### Version(s)

3.22.4

### Bug Description

html snippet: 
```html
<table>
  <colgroup>
    <col width="64"  />
    <col />
  </colgroup>
  <tr><td>hello </td></tr>
</table>
```
 richTextEditor props content set html let I given, width attribute is ignored.

### Browser Used

Chrome

### Code Example URL

_No response_

### Expected Behavior

The editor should map the width attribute to the rendered table column's style.

_No response_

### Dependency Updates

devvaannsh:
Bug: The first column's width was always dropped because of an incorrect truthiness check for index 0. 
Screenshot 2026-07-06 at 1 36 19 AM
Here, in the line ```if (cellIndex && ...)``` , cellIndex for index 0 will always return false.

Secondly, header cells (`<th>`) never read the colgroup at all, so tables that start with a header row lost every column width.

Reference issue: Fixes https://github.com/ueberdosis/tiptap/issues/7800

fix: Added a new ```parseColWidth``` utility that is used by both ```TableCell``` and ```TableHeader```. It prefers the cell's own ```colWidth``` attribute and only when unavailable falls back to the ```width``` of the matching ```col``` in the ```colgroup```.

---

Visual Reference: Focus on the ```parsed val``` section
**Before**
Screenshot 2026-07-06 at 1 30 51 AM
Screenshot 2026-07-06 at 1 29 10 AM

---

**After**
Screenshot 2026-07-06 at 1 25 01 AM
Screenshot 2026-07-06 at 1 26 52 AM

## Checklist


packages/extension-table/src/utils/parseColwidth.ts
arnaugomez, line 1: Rename utilities folder to utils (just to follow repository conventions)

arnaugomez (changes requested):

packages/extension-table/src/utils/parseColwidth.ts
devvaannsh: Done!

arnaugomez (approved):

PR #8001, opened 2026-07-06
Fix (extension-list): add paragraph interrupters so that block elements after list doesn't come up as literal text

devvaannsh:
BUG: If a thematic break (`---`, `***`, `___`), a fenced code block (```` ``` ````/`~~~`), or an unindented bullet (`- item`) is placed on the line directly after an ordered list item (with no blank line in between), it gets pulled into the list item instead of ending the list. 
Thematic breaks are the worst case: they end up as literal text inside the item, and every line after them gets swallowed too. Code fences and bullets get parsed, but nested inside the list item instead of after the list.

This is the same lazy continuation issue as #7997 , which fixed it for headings only.

FIX: Any line that starts a block that can interrupt a paragraph now rightly terminates the lazy continuation. This change is regression free.

---
Visual Reference
**Before**
lazy-continuation-before

---

**After**
lazy-continuation-after

## AI Usage

## Checklist

### DIFF
--- a/.changeset/fix-ordered-list-block-start-lazy-continuation.md
+++ b/.changeset/fix-ordered-list-block-start-lazy-continuation.md
@@ -1,0 +1,11 @@
+---
+'@tiptap/extension-list': patch
+---
+
+Fix markdown parsing bugs where block elements right after an ordered list item (with no blank line in between) were wrongly treated as lazy continuation of the list item, instead of terminating the list the way other markdown parsers do:
+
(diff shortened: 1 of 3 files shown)


arnaugomez (approved):

arnaugomez (approved):
Awesome. Could this bug also be affecting other kinds of lists, like unordered lists or checkboxes?

devvaannsh:
No, this bug affected only ordered lists!

PR #8008, opened 2026-07-07
fix(markdown): recover blank lines absorbed into block token raw

### LINKED ISSUE #8007: Markdown parse drops blank lines after headings, tables and HTML blocks
### Affected Packages

@tiptap/markdown

### Version(s)

3.27.3

### Bug Description

`@tiptap/markdown` drops a blank line after certain block tokens on every parse. When a heading (also a table or an HTML block) is followed by one or more blank lines and then text, `MarkdownManager.parse()` fails to reconstruct the empty paragraph(s) that represent those blank lines, because marked's tokenizer for those blocks matches a trailing `\n+` and absorbs the blank lines into the block token's `raw`, emitting no `space` token. Empty paragraphs are only reconstructed from `space` tokens, so the blank line is lost.

Because editors that persist content as markdown re-parse on every load/save, this erodes one blank line per round-trip until all blank lines between the block and the following text are gone. Blocks whose tokenizer emits a standalone `space` token (paragraph, list, code block, blockquote, thematic break) round-trip correctly.

Minimal reproduction: https://codesandbox.io/p/sandbox/elegant-jackson-z5wn9l

### Browser Used

Chrome

### Code Example URL

https://codesandbox.io/p/devbox/eager-goodall-l7ny89?workspaceId=ws_HqmVPpujLHyRQfvNS7iocR

### Expected Behavior

Blank lines between a block (heading/table/HTML block) and following content survive a parse/serialize round-trip, the same way they do after a paragraph.

Surfaced when stored as markdown and re-loaded into a Tiptap editor — silently lost their spacing over successive edits.

### Dependency Updates

sravanth-space:
marked's heading, table and html-block tokenizers match a trailing `\n+`, so the blank lines that follow such a block are absorbed into the token's own `raw` and no standalone `space` token is emitted. Empty paragraphs were only reconstructed from `space` tokens, so those blank lines were dropped on parse — eroding one blank line per parse/serialize round-trip in editors that persist content as markdown.

Normalize the marked token stream before parsing: `extractAbsorbedBlankLines` splits an absorbed trailing blank-line run back into an explicit `space` token, so the existing single reconstruction path handles every block uniformly. Adds a regression spec and a changeset.

## Changes Overview

Fix `@tiptap/markdown` dropping blank lines after headings, tables and HTML blocks on parse. Empty paragraphs are now recovered from blank lines that a block tokenizer absorbed into its own `raw` (marked's heading/table/html rules do this), not only from standalone `space` tokens.

## Implementation Approach

The root inconsistency is that marked represents blank lines between blocks as a standalone `space` token for most blocks, but folds them into the block token's own `raw` for a few tokenizers whose rules match a trailing `\n+` (heading, table, html blocks). Rather than add a second recovery path, I normalize marked's output first: a small pre-pass (`extractAbsorbedBlankLines` in `utils.ts`) splits an absorbed trailing blank-line run back into an explicit `space` token, and `parseTokens` then runs its existing single reconstruction path over the normalized stream. So there's exactly one place that turns blank lines into empty paragraphs. No behaviour change for tokens that already emit a `space` token; nothing fires for a single trailing newline (the regex requires ≥2); and trimming `raw` is safe because those handlers parse from `token.tokens`/`text`/`depth`.

## Testing Done

 Added `packages/markdown/__tests__/block-token-blank-line.spec.ts` (4 tests): parse preserves the blank line after a heading and after a table; no spurious paragraph when a heading is directly followed by text; parse/serialize round-trip is stable (no erosion).
- Full markdown suite passes: 249 tests (245 existing + 4 new).
- `oxlint` clean.

## Verification Steps

```js
// with Document, Paragraph, Text, Heading extensions:
new MarkdownManager({ extensions }).parse('# Title\n\n\n\nBody').content
// → [heading, paragraph([]), paragraph('Body')]  (empty paragraph preserved)
```
Or run `pnpm vitest run packages/markdown`.

## Additional Notes

Affected block tokenizers are heading, setext heading, table and HTML block (all match a trailing `\n+`); paragraph/list/code/blockquote/thematic-break already emit a `space` token and are unaffected. There is a separate, pre-existing lossiness for **odd-length** newline runs between plain paragraphs (`countParagraphSeparators` counts non-overlapping `/\n\n/g`); that is intentionally left out of scope for this PR.

## AI Usage

## Checklist

## Related Issues

Closes #8007


packages/markdown/src/utils.ts
```diff
@@ -4,6 +4,43 @@ import type { Fragment, Node } from '@tiptap/pm/model'
 
 import type { ContentType } from './types.js'
 
+/**
+ * Matches a run of two or more consecutive line breaks (a blank line) at the
+ * end of a string, allowing horizontal whitespace between the breaks.
+ */
+const TRAILING_BLANK_LINES = /\n[^\S\n]*(?:\n[^\S\n]*)+$/
```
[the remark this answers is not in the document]
sravanth-space: Thanks — I checked this and it's not reachable in practice. Parsing goes through parseLexer.lex(markdown), and marked's Lexer.lex() normalizes line endings as its very first step (src = src.replace(/\r\n|\r/g, '\n'), confirmed in marked@17.0.1). So every token.raw is already \n-only by the time it reaches extractAbsorbedBlankLines — CRLF can't survive to this regex, even on Windows input.

The \r\n → \n replace in countParagraphSeparators is redundant defensive code for the same reason; it isn't evidence that tokens carry CRLF. Adding CRLF branches here would just be dead code guarding an unreachable input, so I'm leaving the regex as-is.

sravanth-space:
Thanks so much for jumping in and helping polish this, @bdbch! Really appreciate the updates to the changeset, the spec refinements, and the comment cleanup in utils.ts.

Just wanted to check in on the status: are there any further changes you'd like to see, or does this look good to go for a review/merge?

bdbch:
Hey @sravanth-space - thanks for contributing. I still need to do a final review but iirc it looked very promising already. I'll come back later to you when I did the final review.

arnaugomez:
@coderabbitai review

packages/markdown/src/utils.ts
```diff
@@ -74,11 +103,8 @@ export function findMarksToOpen(
 }
 
 /**
- * Determines which marks need to be closed at the end of the current text node.
- * This handles cases where marks end at node boundaries or when transitioning
- * to nodes with different mark sets.
- * Compares both mark type and attributes — two marks of the same type with
- * different attributes are treated as distinct and trigger a close/reopen.
+ * Determines which marks to close at the node end, treating same-type marks
+ * with different attributes as distinct (close + reopen).
```
arnaugomez, line 111: Why did you update the comments here? Seems unrelated to this PR

packages/markdown/src/MarkdownManager.ts
```diff
@@ -365,7 +366,19 @@ export class MarkdownManager {
     tokens: MarkdownToken[],
     parseImplicitEmptyParagraphs = false,
   ): JSONContent[] {
-    const nonSpaceTokenIndexes = tokens.reduce<number[]>((indexes, token, index) => {
+    // Blank lines between blocks are reconstructed below from `space` tokens.
+    // marked emits them consistently for most blocks, but some tokenizers
```
arnaugomez:
```suggestion
    // Marked emits them consistently for most blocks, but some tokenizers
```

arnaugomez (changes requested):
Looks promising. Thanks for your contrib :)

packages/markdown/src/utils.ts
sravanth-space: This is from @bdbch in this commit  374078d60c1d41e192d4ab70430e4a5076e15c8f

arnaugomez (dismissed):

sravanth-space:
Hey @bdbch, just checking in on this one — no rush at all, just wanted to see if you've had a chance to get to the final review yet, or if there's anything else you'd like me to address in the meantime. Thanks again for taking the time to look at it!

packages/markdown/src/MarkdownManager.ts
```diff
@@ -365,7 +366,19 @@ export class MarkdownManager {
...
   ): JSONContent[] {
-    const nonSpaceTokenIndexes = tokens.reduce<number[]>((indexes, token, index) => {
+    // Blank lines between blocks are reconstructed below from `space` tokens.
+    // Marked emits them consistently for most blocks, but some tokenizers
+    // (notably heading, table and html blocks, whose rules match a trailing
+    // `\n+`) absorb the following blank lines into the block token's own `raw`
+    // and emit no `space` token. Normalize those into explicit `space` tokens
+    // first so a single reconstruction path handles every block uniformly —
+    // otherwise the blank line is dropped, eroding one empty paragraph per
+    // parse/serialize round-trip.
```
bdbch:
```suggestion
    // Normalize absorbed blank lines into `space` tokens so they survive round-trips.
```

packages/markdown/src/utils.ts
```diff
@@ -4,6 +4,38 @@ import type { Fragment, Node } from '@tiptap/pm/model'
 
 import type { ContentType } from './types.js'
 
+/**
+ * Matches a run of two or more consecutive line breaks (a blank line) at the
+ * end of a string, allowing horizontal whitespace between the breaks.
+ */
```
bdbch, line 14:
```suggestion
/**
 * Matches a run of two or more consecutive line breaks (a blank line) at the
 * end of a string, allowing horizontal whitespace between the breaks.
 *
 * @example
 * TRAILING_BLANK_LINES.test('paragraph\n  \n')
 * // => true
 */
```

bdbch (dismissed):
Looks good - added a few comment refactors that ill apply.

bdbch (approved):

bdbch (changes requested):
`Blank lines after a block token > round-trips explicit &nbsp; paragraphs` seems to fail

sravanth-space:
> `Blank lines after a block token > round-trips explicit &nbsp; paragraphs` seems to fail

Fixed — that test wrongly expected &nbsp; to survive verbatim, but a lone `&nbsp; paragraph` normalizes to blank-line spacing (as paragraph.spec.ts already asserts), so I updated it to check no blank-line erosion across round-trips instead.

bdbch (approved):

PR #8016, opened 2026-07-10
fix(core): merge task-item wrapped continuation lines instead of dropping them

### LINKED ISSUE #7909: Markdown: task-item wrapped/indented continuation lines render as code blocks
### Affected Packages

@tiptap/core, @tiptap/markdown, @tiptap/extension-list

### Version(s)

3.23.2, 3.24.0

### Bug Description

We have a local patch for this issue, which I ran into when opening a markdown note containing an LLM-generated task list, and which **Claude Code** then diagnosed, so please take this bug report with a grain of salt:

When a task list item has a **wrapped (soft-continued) or nested line indented more than `baseIndentSize` (2) spaces past the marker**, that content renders as an **indented code block** instead of joining the item's paragraph.

This is a very common authoring style — aligning the continuation under the checkbox **text column** (`- [ ] ` is 6 columns wide), which is also what many LLMs emit. Plain bullet (`-`) lists are unaffected (they go through `marked`'s native list tokenizer); only task items hit `parseIndentedBlocks`. CommonMark renderers (remark/react-markdown) handle the same input correctly, because an indented code block cannot interrupt a paragraph.

**Input markdown:**

```markdown
- [ ] Agent tab renders the composer in all states (no session
      empty transcript + composer).
```

**Result:** the second line (`empty transcript + composer).`) becomes a code block under the task item instead of being merged into the item's text.

**Root cause** — in `@tiptap/core`'s `parseIndentedBlocks` (`packages/core/src/utilities/markdown/parseIndentedBlocks.ts`), nested content is dedented by a *fixed* amount:

```js
const dedentedNested = nestedContent
    .map(nestedLine => nestedLine.slice(indentLevel + baseIndentSize)) // Remove base indent + 2 spaces
    .join("\n");
```

When the continuation is indented more than `baseIndentSize` past the marker, residual leading whitespace (≥4 spaces) survives the dedent, and the downstream `marked.blockTokens(...)` (or the recursive task parser's fallback) parses it as an indented code block.

**Minimal, DOM-free repro of the root cause:**

```js
import { marked } from "marked";

// Continuation line collected by parseIndentedBlocks for the item above
// (indented to the checkbox text column = 6 spaces).
const nested = ["      empty transcript + composer)."];
const indentLevel = 0;     // from the item pattern match
const baseIndentSize = 2;  // default

// CURRENT behavior: dedent by a FIXED indentLevel + baseIndentSize (=2) -> 4 spaces survive
const dedented = nested.map((l) => l.slice(indentLevel + baseIndentSize)).join("\n");
console.log(JSON.stringify(dedented)); // "    empty transcript + composer)."
console.log(new marked.Lexer().blockTokens(dedented).map((t) => t.type));
// -> [ 'code' ]   ❌  expected a paragraph
```

**Suggested fix** — dedent by the *actual minimum indentation* of the non-empty nested lines (preserving relative nesting), instead of a fixed `indentLevel + baseIndentSize`. The mixed bullet/task path in `@tiptap/markdown` already does exactly this (`Math.min(...indents)`), so this also removes an internal inconsistency:

```js
const nestedIndents = nestedContent
    .filter(nestedLine => nestedLine.trim())
    .map(nestedLine => nestedLine.length - nestedLine.trimStart().length);
const nestedDedent = nestedIndents.length > 0
    ? Math.min(...nestedIndents)
    : indentLevel + baseIndentSize;
const dedentedNested = nestedContent
    .map(nestedLine => (nestedLine.trim() ? nestedLine.slice(nestedDedent) : ""))
    .join("\n");
```

With this change the repro yields a `paragraph`, and genuinely-nested sub-lists/sub-tasks still dedent correctly (relative indentation preserved); fenced code blocks (```` ``` ````) are unaffected.

_Related: #7220 / PR #7225 — same function (`parseIndentedBlocks`), different defect._

### Browser Used

Chrome

### Code Example URL

https://codesandbox.io/p/devbox/adoring-wright-kmqhlt

### Expected Behavior

The wrapped/over-indented task-item continuation joins the item's paragraph (matching CommonMark and marked's native list handling), rather than rendering as an indented code block.

_No response_

### Dependency Updates

codewithsupra:
## Fixes

Fixes #7909

## Changes and Review

- Task-item continuation lines indented under `- [ ] ` were dedented incorrectly, leaving 4+ leading spaces — which CommonMark parses as an indented code block and silently drops.
- Continuation lines now merge into the item's paragraph instead of being treated as nested content.
- Regression tests added; they fail with the fix reverted.

## Checklist

### Responsibility

### DIFF
--- a/packages/core/src/utilities/markdown/parseIndentedBlocks.ts
+++ b/packages/core/src/utilities/markdown/parseIndentedBlocks.ts
@@ -6,6 +6,28 @@
  * for lists, task lists, and other indented block types.
  */
 
+/**
+ * Whether a (leading-whitespace-trimmed) line begins a new markdown block —
+ * a list item (bullet, ordered, or task), blockquote, fenced code block, ATX
+ * heading, or thematic break. Such a line is not a lazy paragraph continuation
+ * and must not be merged into a preceding paragraph.
+ */
+function isBlockStart(trimmedLine: string): boolean {
+  return (
+    // oxlint-disable-next-line prefer-string-starts-ends-with
+    /^[-+*](\s|$)/.test(trimmedLine) || // bullet / task-item marker
+    // oxlint-disable-next-line prefer-string-starts-ends-with
+    /^\d+[.)](\s|$)/.test(trimmedLine) || // ordered list marker
+    // oxlint-disable-next-line prefer-string-starts-ends-with
+    /^>/.test(trimmedLine) || // blockquote
+    // oxlint-disable-next-line prefer-string-starts-ends-with
+    /^(```|~~~)/.test(trimmedLine) || // fenced code block
+    // oxlint-disable-next-line prefer-string-starts-ends-with
+    /^#{1,6}(\s|$)/.test(trimmedLine) || // ATX heading
+    /^(-{3,}|\*{3,}|_{3,})$/.test(trimmedLine) // thematic break
+  )
+}
+
 export interface ParsedBlock {
   type: string
   raw: string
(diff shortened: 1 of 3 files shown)


arnaugomez:
@coderabbitai review

packages/core/src/utilities/markdown/parseIndentedBlocks.ts
```diff
@@ -160,12 +182,51 @@ export function parseIndentedBlocks(
...
+    // content. A soft-wrapped line that directly follows the item's text — no
+    // blank line separating it, and not itself the start of a new block — is a
+    // continuation of the item's paragraph in CommonMark, not new nested
+    // content. Authors (and LLMs) commonly align such wrapped lines under the
+    // marker's *text* column (e.g. six columns for `- [ ] `). Previously these
+    // were treated as nested content, dedented by a fixed amount that left >= 4
+    // residual spaces, and parsed by `marked` as an indented code block — which
+    // cannot interrupt a paragraph, so the line was silently dropped. Merging
+    // them into `mainContent` (which is tokenized inline by the caller) yields a
+    // single paragraph, matching CommonMark renderers.
```
arnaugomez, line 197: Make the explanatory comments more concise. Explanatory comments should help understand the code. If they are too long, they have the opposite effect of decreasing legibility. 

Do not mention the previous behavior: "Previously these were treated as nested content..."

Do not mention irrelevant info: "Authors (and LLMs)..." is irrelevant, it doesn't matter who wrote the markdown content.

Make the language simple and clear to understand. A beginner developer would not know what "Peel off" means.

packages/core/src/utilities/markdown/parseIndentedBlocks.ts
```diff
@@ -160,12 +182,51 @@ export function parseIndentedBlocks(
...
+    }
 
     if (nestedContent.length > 0) {
-      // Remove the base indentation from nested content
+      // Dedent by the smallest leading indentation shared by the non-blank
+      // nested lines rather than a fixed `indentLevel + baseIndentSize`, so a
+      // genuinely-nested block indented past `indentLevel + baseIndentSize` is
+      // not left with residual leading whitespace (which `marked` would parse
+      // as an indented code block). Deeper content keeps its relative
+      // indentation.
```
arnaugomez, line 222: Same here. This comment is very hard to understand. Make it clearer, make the sentences shorter. TWrite the explanatory comment yourself, do not have an LLM write it for you.

arnaugomez (changes requested):

PR #8017, opened 2026-07-10
fix(markdown): never parse to an empty document (fixes setContent crash)

### LINKED ISSUE #7914: Markdown: `setContent` issues when content only contains whitespace and slashes
### Affected Packages

markdown

### Version(s)

3.25.0

### Bug Description

If you try to parse `           /               ` with `contentType: markdown` it throws following error:
```
RangeError: Invalid content for node doc: <>
    at _NodeType.checkContent (chunk-X36TMPZB.js?v=9ef6b9c8:2197:13)
    at _Node.check (chunk-X36TMPZB.js?v=9ef6b9c8:1504:15)
    at chunk-ZQYF7ISI.js?v=9ef6b9c8:1425:12
    at Array.forEach (<anonymous>)
    at chunk-ZQYF7ISI.js?v=9ef6b9c8:1424:11
    at Object.insertContentAt (chunk-ZQYF7ISI.js?v=9ef6b9c8:820:57)
    at chunk-ZQYF7ISI.js?v=9ef6b9c8:1754:19
    at Object.chainedCommand [as setContent] (chunk-ZQYF7ISI.js?v=9ef6b9c8:777:47)
    at input.tsx:123:12
```

Also if you try to parse `               \` it just returns empty string for some reason.

### Browser Used

Chrome

### Code Example URL

_No response_

### Expected Behavior

It should parse it as text node

_No response_

### Dependency Updates

codewithsupra:
## Fixes

Fixes #7914

## Changes and Review

- `parse()` could return a `doc` with no valid block content for certain inputs, which crashes `setContent`.
- Fallback now checks for actual block content rather than a non-empty array, so a custom extension returning a bare inline node can't slip through.
- Addressed two rounds of maintainer review feedback; regression tests cover both the original bug and the review findings.

## Checklist

### Responsibility


packages/markdown/src/MarkdownManager.ts
```diff
@@ -348,10 +348,17 @@ export class MarkdownManager {
...
+      // throw `RangeError: Invalid content for node doc: <>`. This happens for
+      // input that yields no renderable blocks — whitespace-only markdown, or
+      // markdown whose only token has no registered handler (e.g. an indented
+      // code block when no code-block extension is present, as with a line of
+      // leading whitespace followed by text). Fall back to a single empty
+      // paragraph, matching how an empty markdown string is represented.
       return {
         type: 'doc',
-        content,
+        content: content.length > 0 ? content : [{ type: 'paragraph' }],
```
arnaugomez: This would only work on editors that support the paragraph node. However, we assume that anybody using Markdown conversion uses an editor that supports paragraph nodes, because paragraphs are an integral part of Markdown. So LGTM.

arnaugomez (approved):

packages/markdown/src/MarkdownManager.ts
```diff
@@ -348,10 +348,17 @@ export class MarkdownManager {
...
 
-      // Return a document node containing the parsed content
+      // A `doc` node requires at least one block child (`block+`), so an empty
+      // `content` array produces an invalid document that makes `setContent`
+      // throw `RangeError: Invalid content for node doc: <>`. This happens for
+      // input that yields no renderable blocks — whitespace-only markdown, or
+      // markdown whose only token has no registered handler (e.g. an indented
+      // code block when no code-block extension is present, as with a line of
+      // leading whitespace followed by text). Fall back to a single empty
+      // paragraph, matching how an empty markdown string is represented.
```
arnaugomez: Move this comment to just before "content", and re-write it to make it shorter.

packages/markdown/src/MarkdownManager.ts
```diff
@@ -348,10 +348,17 @@ export class MarkdownManager {
       // Convert tokens to Tiptap JSON
       const content = this.parseTokens(tokens, true)
 
-      // Return a document node containing the parsed content
```
arnaugomez, line 352: Do not delete this comment

arnaugomez (changes requested):

arnaugomez (dismissed):

packages/markdown/__tests__/empty-doc-fallback.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+  it('setContent with whitespace+slash markdown does not throw', () => {
+    expect(() => {
+      editor = new Editor({
+        extensions: [Document, Paragraph, Text, Markdown],
+        content: '           /               ',
+        contentType: 'markdown',
+      })
+    }).not.toThrow()
+    expect(editor!.getJSON().content![0].type).toBe('paragraph')
+  })
```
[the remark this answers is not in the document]
codewithsupra: Fixed — added a separate test that constructs an empty `Editor` and calls `editor.commands.setContent(markdown, { contentType: 'markdown' })` directly, alongside the existing constructor-init test. See `packages/markdown/__tests__/empty-doc-fallback.spec.ts`.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

packages/markdown/src/MarkdownManager.ts
```diff
@@ -351,7 +351,8 @@ export class MarkdownManager {
       // Return a document node containing the parsed content
       return {
         type: 'doc',
-        content,
+        // A document requires at least one block child.
+        content: content.length > 0 ? content : [{ type: 'paragraph' }],
```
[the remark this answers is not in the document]
codewithsupra: Fixed — the fallback now checks that at least one top-level node is actually a registered block node type, not just that `content.length > 0`. Added a regression test where a custom extension's `parseMarkdown` handler returns a bare `{ type: 'text' }` node at the top level; it now correctly falls back to a paragraph instead of producing an invalid doc.

I deliberately avoided `getSchema()` for the block-type check: it requires every registered extension to combine into one globally valid ProseMirror schema, which is a stricter bar than this needs — schema construction can throw for reasons unrelated to a given parse (e.g. duplicate/incompatible extension names across an app's full extension set), and I confirmed via the existing `conversion.spec.ts` extension set that this happens in practice. Swallowing that error would have silently discarded all valid parsed content. Instead I derive block-ness per-extension (mirroring ProseMirror's own `inline`/`topNode`/`text` checks), which doesn't require the full set to be schema-valid.

See `MarkdownManager.ts` (`getSchemaBlockNodeNames`) and `empty-doc-fallback.spec.ts`.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

packages/markdown/src/MarkdownManager.ts
```diff
@@ -1051,6 +1059,54 @@ export class MarkdownManager {
...
+
+      const isInline = !!callOrReturn(getExtensionField(extension, 'inline', context))
+      const isTopNode = !!callOrReturn(getExtensionField(extension, 'topNode', context))
+
+      if (!isInline && !isTopNode && extension.name !== 'text') {
+        names.add(extension.name)
+      }
+    })
+
+    this.schemaBlockNodeNamesCache = names
```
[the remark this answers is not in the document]
codewithsupra: Fixed — `getSchemaBlockNodeNames()` now reads `this.extensions` (every extension ever passed to `registerExtension`) instead of `this.baseExtensions`, and `registerExtension()` invalidates the cache on each call. Added a regression test that registers a block node directly via `registerExtension()` after construction and confirms its content survives instead of being replaced by the fallback paragraph.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

packages/markdown/src/MarkdownManager.ts
```diff
@@ -1052,6 +1059,54 @@ export class MarkdownManager {
...
+   * Derived per-extension (mirroring ProseMirror's own `NodeType.isBlock`:
+   * not inline, not the top node, not `text`) rather than via `getSchema`,
+   * which requires every registered extension to combine into one globally
+   * valid ProseMirror schema. That's a stricter bar than markdown parsing
+   * needs — schema construction can fail for reasons unrelated to a given
+   * parse (duplicate/incompatible node names across an app's full extension
+   * set), and treating that failure as "no block nodes exist" would discard
+   * otherwise-correct parsed content.
+   */
+  private getSchemaBlockNodeNames(): Set<string> {
```
arnaugomez: Shouldn't this function check for the `group: "block"` property of the Node extension?

arnaugomez (dismissed):

arnaugomez:
I added these fixes:
- Use the "group" property to check if a node is in the "block" group. This is the way to know if a node is a valid children of the `doc` node.
- Fix failing test, because the "Heading" extension was not registered.

arnaugomez (approved):

PR #8025, opened 2026-07-11
fix(markdown): extend CommonMark escape handling

bdbch:
## Fixes

Addresses the follow-up review comment on #7898 about incomplete Markdown escape coverage.

## Changes and Review

- Expanded literal text serialization to escape the remaining punctuation supported by Marked’s CommonMark-compatible tokenizer.
- Preserved HTML entity encoding for angle brackets so escaped text such as `\\<class\\>` cannot be re-parsed as HTML.
- Added serializer and round-trip regression tests, plus updated exact-output fixtures.

## Checklist

### Responsibility

- [x] I have reviewed and understand these changes, and I take responsibility for this PR, even if an AI agent created them.


bdbch:
Closing this PR for now. The current approach broadens escaping too much and causes unrelated unit-test regressions; we’ll revisit with a context-aware implementation later.

Issue #7833, opened 2026-05-15
[Bug]: InputRules calculate incorrect ranges when inline atom nodes such as Mention are present


jjongsu:
### Affected Packages

core, extension-mention

### Version(s)

3.23.4(latest)

### Bug Description

First of all, thank you for maintaining Tiptap and for all the work that goes into this project.

I am not completely sure whether this is the intended behavior or an actual bug, but I encountered a few cases where inputRules seem to behave unexpectedly when used together with the Mention extension.

I tried to investigate the issue and found that it may be related to how inputRules calculate ranges when inline atom nodes are involved. I would appreciate it if you could confirm whether this is a bug, an expected limitation, or if there is a recommended way to handle this case.

---

InputRules appear to calculate ranges from rendered text content instead of safely mapping back to actual ProseMirror document positions. This causes incorrect replacements when inline atom nodes such as Mention are included in or near the matched text.

Since Mention's nodeSize differs from the length of its renderText output, the matched text range and the actual document range can diverge. As a result, inputRules may apply transformations to the wrong range or unexpectedly remove mention content.

Examples:

1. '\` [@developer]\`'
   When an inputRule converts this into inline code, the wrong range is converted.

2. '[@developer<]-'
   The `<-` inputRule can be triggered around the mention boundary, and the mention text is unexpectedly removed.

3. '[@\d \`eveloper] text \`'
   Similar to the first case, the calculated range is incorrect when a mention and inline code-like text are involved.

---

I suspect the issue may be related to a few areas in the codebase:

1. `core/src/InputRule.ts`

When `InputRule` calculates the range, inline atom/block-like nodes may need to be handled differently from plain text nodes.

In particular, I suspect that the logic that sets the `const range` may be incorrect in this case. It seems to calculate the range based on the length of the matched/rendered text, but inline atom nodes such as Mention can have a `nodeSize` that differs from the length of their rendered text representation.

Because of this mismatch, the calculated `range.from` and `range.to` may not correspond to the actual ProseMirror document positions. As a result, when the inputRule applies a transaction, it can transform or replace the wrong part of the document.

Currently, the matched text range and the actual ProseMirror document range can diverge when inline nodes such as Mention are included in the match.

2. `extension-mention`

One possible workaround or improvement might be to safely mark the boundaries of the mention node by inserting zero-width spaces on both sides in the text representation.

   For example:
```ts
[...]

PR #8026, opened 2026-07-11
fix(core): input rules no longer crash when the match spans an inline atom node

### LINKED ISSUE #7933: Input rules throw "RangeError: Position -N out of range" when the textblock contains inline atom nodes (#5298 still reproduces in 3.26.0)
### Affected Packages

core

### Version(s)

3.26.0 (latest, verified), also reproduced on 3.20.5

### Bug Description

When a textblock contains inline **atom** nodes (mentions, custom link chips, …), input rules can compute a **negative document position** and crash with `RangeError: Position -N out of range`.

**Root cause — two different coordinate systems are mixed:**

`getTextContentFromNodes` builds `textBefore` by expanding each atom node to `node.type.spec.toText(...) || node.textContent || "%leaf%"` — i.e. an atom contributes **6 characters** (`%leaf%`) to the string while it only occupies **1 position** in the document.

`run()` (inputRule handler) then subtracts a **string length** from a **document position**:

```js
// packages/core/src/inputRules/run (dist line ~3286 in 3.26.0)
const range = {
  from: from - (match[0].length - text.length),
  to,
}
```

If the regex match spans the atom expansions, `match[0].length` (string chars) is much larger than the actual positional distance, so `range.from` becomes negative. `markInputRule` then calls `tr.delete(range.from + startSpaces, textStart)` → `doc.resolve(-N)` → `RangeError`.

With N atoms between the opening and closing backtick, `range.from = 1 - 5N`, so even **one** atom crashes.

**Minimal reproduction** (jest + jsdom, only `@tiptap/core` + default `@tiptap/extension-code`, no custom input rule needed):

```ts
import { Editor, Node } from "@tiptap/core";
import { Document } from "@tiptap/extension-document";
import { Paragraph } from "@tiptap/extension-paragraph";
import { Text } from "@tiptap/extension-text";
import { Code } from "@tiptap/extension-code";

// minimal stand-in for a mention-like node
const InlineAtom = Node.create({
  name: "inlineAtom",
  group: "inline",
  inline: true,
  atom: true,
  parseHTML: () => [{ tag: "span[data-atom]" }],
  renderHTML: () => ["span", { "data-atom": "" }],
});

const editor = new Editor({
  extensions: [Document, Paragraph, Text, InlineAtom, Code],
  content: "<p>`a<span data-atom></span></p>",
});

// simulate typing the closing backtick at the end of the paragraph
const pos = editor.state.doc.content.size - 1;
editor.commands.setTextSelection(pos);
editor.view.someProp("handleTextInput", (f) =>
  f(editor.view, pos, pos, "`", () => editor.state.tr.insertText("`", pos, pos))
);
// => RangeError: Position -4 out of range
```

In a real browser this is simply: have a paragraph containing a stray backtick followed by a mention, then type a closing backtick.

**Real-world impact:** we see this in production (Sentry) as `RangeError: Position -101 out of range` in an editor that uses mention/link atom nodes heavily — a paragraph with ~20 atoms between backticks matches the observed magnitude (`1 - 5×21 = -104`).

### Browser Used

Chrome (reproduces in any environment; also in jsdom)

### Code Example URL

_No response_ (self-contained reproduction above)

### Expected Behavior

Typing should never crash the editor. Either the rule should not match across atom boundaries, or `range` should be computed in position space so the rule applies correctly.

**Suggested fix:** make string offsets equal position offsets by having `getTextContentFromNodes` represent every atom/leaf as **exactly `node.nodeSize` characters** (e.g. a single object replacement character `￼`, like ProseMirror does elsewhere), instead of `%leaf%`/`toText()` of arbitrary length. Then `from - (match[0].length - text.length)` is always a valid position. (Matching against `￼` would also naturally stop `[^`]+`-style patterns from producing semantically wrong matches across atoms — though that part is debatable; the crash itself is not.)

- #5298 described exactly this (`getTextContentFromNodes not handling atoms correctly`) and was closed as resolved in July 2024, but the reproduction above still crashes on 3.26.0 — both the `%leaf%` expansion and the string-length subtraction are still present in `dist/index.js`.
- #2946 (open) is the `renderText`-flavored symptom of the same underlying coordinate mismatch.
- Workaround we ship today: wrapping input rules to bail out when `range.from < 0` and swallowing `RangeError` from `handleTextInput`.

### Dependency Updates

devvaannsh:
Fixes #7933 and #7833

## Changes and Review

Input rules compute their range from matched text where atoms expand to `%leaf%` (6 chars for 1 doc position), so a match spanning an inline atom produced an invalid range and crashed.
For a detailed description: please refer to the issues linked above. 

Now, the rule checks the matched text against the actual doc content and skips itself when they don't line up, so the typed character just inserts normally. 

To verify: put a backtick + text + mention in a paragraph and type a closing backtick.

## Checklist

### Responsibility

### DIFF
--- a/.changeset/fix-input-rule-inline-atom-range.md
+++ b/.changeset/fix-input-rule-inline-atom-range.md
@@ -1,0 +1,5 @@
+---
+'@tiptap/core': patch
+---
+
+Input rules no longer crash or replace the wrong content when the matched text spans an inline atom node (like a mention). Input rules match against a string where leaf nodes get expanded to placeholders like `%leaf%`, so a match crossing an atom ended up with a range that didn't map back to real document positions, throwing `RangeError: Position -N out of range` or corrupting nearby content. Such matches are now skipped. Rules on plain text, including text after an atom, behave the same as before.
--- a/.changeset/fix-input-rule-inline-atom-range.md
+++ b/.changeset/fix-input-rule-inline-atom-range.md
@@ -2,4 +2,4 @@
 '@tiptap/core': patch
 ---
 
-Input rules no longer crash or replace the wrong content when the matched text spans an inline atom node (like a mention). Input rules match against a string where leaf nodes get expanded to placeholders like `%leaf%`, so a match crossing an atom ended up with a range that didn't map back to real document positions, throwing `RangeError: Position -N out of range` or corrupting nearby content. Such matches are now skipped. Rules on plain text, including text after an atom, behave the same as before.
+Fix input rules crashing when the matched text spans an inline atom node like a mention.
--- a/packages/core/src/InputRule.ts
+++ b/packages/core/src/InputRule.ts
@@ -133,6 +133,25 @@ function run(config: {
       return
     }
 
+    // textBefore expands leaf nodes into placeholders like %leaf% (check getTextContentFromNodes),
+    // so an atom takes up 6 chars in the string but just 1 position in the doc.
+    // if the match crosses a leaf node, the range computed below would land on the wrong
+    // doc positions (can even go negative), so in that case we skip the rule
+    // see: https://github.com/ueberdosis/tiptap/issues/7933
+    const matchedDocLength = match[0].length - text.length
+
+    if (matchedDocLength > 0) {
+      const matchStartOffset = $from.parentOffset - matchedDocLength
+
+      if (
+        matchStartOffset < 0 ||
+        $from.parent.textBetween(matchStartOffset, $from.parentOffset) !==
+          match[0].slice(0, matchedDocLength)
+      ) {
+        return
+      }
+    }
+
     const tr = view.state.tr
     const state = createChainableState({
       state: view.state,
--- a/packages/core/src/InputRule.ts
+++ b/packages/core/src/InputRule.ts
@@ -133,11 +133,8 @@ function run(config: {
       return
     }
 
-    // textBefore expands leaf nodes into placeholders like %leaf% (check getTextContentFromNodes),
-    // so an atom takes up 6 chars in the string but just 1 position in the doc.
-    // if the match crosses a leaf node, the range computed below would land on the wrong
-    // doc positions (can even go negative), so in that case we skip the rule
-    // see: https://github.com/ueberdosis/tiptap/issues/7933
+    // Leaf nodes have a different length in text and the document.
+    // Skip matches across them to avoid wrong document positions.
     const matchedDocLength = match[0].length - text.length
 
     if (matchedDocLength > 0) {
--- a/packages/core/src/__tests__/inputRuleInlineAtoms.test.ts
+++ b/packages/core/src/__tests__/inputRuleInlineAtoms.test.ts
@@ -8,7 +8,6 @@ import { afterEach, describe, expect, it } from 'vitest'
 // minimal stand-in for a mention-like node. inline atoms expand to %leaf% (6 chars)
 // in getTextContentFromNodes but only take up 1 position in the doc,
 // which used to break the input rule range math
-// https://github.com/ueberdosis/tiptap/issues/7933
 const InlineAtom = Node.create({
   name: 'inlineAtom',
   group: 'inline',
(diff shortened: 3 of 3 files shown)


packages/core/src/__tests__/inputRuleInlineAtoms.test.ts
```diff
@@ -0,0 +1,93 @@
...
+import Code from '@tiptap/extension-code'
+import Document from '@tiptap/extension-document'
+import Paragraph from '@tiptap/extension-paragraph'
+import Text from '@tiptap/extension-text'
+import { afterEach, describe, expect, it } from 'vitest'
+
+// minimal stand-in for a mention-like node. inline atoms expand to %leaf% (6 chars)
+// in getTextContentFromNodes but only take up 1 position in the doc,
+// which used to break the input rule range math
+// https://github.com/ueberdosis/tiptap/issues/7933
```
bdbch: Remove the issue mention here
```suggestion
```

packages/core/src/InputRule.ts
```diff
@@ -133,6 +133,25 @@ function run(config: {
       return
     }
 
+    // textBefore expands leaf nodes into placeholders like %leaf% (check getTextContentFromNodes),
+    // so an atom takes up 6 chars in the string but just 1 position in the doc.
+    // if the match crosses a leaf node, the range computed below would land on the wrong
+    // doc positions (can even go negative), so in that case we skip the rule
+    // see: https://github.com/ueberdosis/tiptap/issues/7933
```
bdbch: same here

```suggestion
```

packages/core/src/InputRule.ts
```diff
@@ -133,6 +133,25 @@ function run(config: {
       return
     }
 
+    // textBefore expands leaf nodes into placeholders like %leaf% (check getTextContentFromNodes),
+    // so an atom takes up 6 chars in the string but just 1 position in the doc.
+    // if the match crosses a leaf node, the range computed below would land on the wrong
+    // doc positions (can even go negative), so in that case we skip the rule
```
bdbch: simplify this comment please

```suggestion
    // Leaf nodes have a different length in text and the document.
    // Skip matches across them to avoid wrong document positions.
```

packages/core/src/__tests__/inputRuleInlineAtoms.test.ts
```diff
@@ -0,0 +1,93 @@
...
+  })
+
+  it('does not corrupt preceding text when the match spans an atom (#7833)', () => {
+    // with enough text before the backtick the miscalculated range stays positive,
+    // so instead of crashing it used to silently mark/replace the wrong content
+    editor = createEditor('<p>hello world `a<span data-atom></span></p>')
+
+    expect(() => typeAtEndOfParagraph(editor, '`')).not.toThrow()
+    expect(editor.getHTML()).not.toContain('<code>')
+    expect(editor.getText()).toContain('hello world')
```
bdbch: Could we assert the complete document content here? `toContain()` could still pass if some surrounding content was accidentally changed.

packages/core/src/__tests__/inputRuleInlineAtoms.test.ts
```diff
@@ -0,0 +1,93 @@
...
+import Document from '@tiptap/extension-document'
+import Paragraph from '@tiptap/extension-paragraph'
+import Text from '@tiptap/extension-text'
+import { afterEach, describe, expect, it } from 'vitest'
+
+// minimal stand-in for a mention-like node. inline atoms expand to %leaf% (6 chars)
+// in getTextContentFromNodes but only take up 1 position in the doc,
+// which used to break the input rule range math
+// https://github.com/ueberdosis/tiptap/issues/7933
+const InlineAtom = Node.create({
```
bdbch, line 11: Could we add one test with a custom `toText()` value? `getTextContentFromNodes()` uses that before falling back to `%leaf%`, so it would be good to cover that path too.

bdbch (changes requested):

packages/core/src/__tests__/inputRuleInlineAtoms.test.ts
devvaannsh: Done!

packages/core/src/__tests__/inputRuleInlineAtoms.test.ts
devvaannsh: Good one! Updated.

packages/core/src/__tests__/inputRuleInlineAtoms.test.ts
devvaannsh: Added!

bdbch:
@coderabbitai review before i take a look

.changeset/fix-input-rule-inline-atom-range.md
```diff
@@ -0,0 +1,5 @@
+---
+'@tiptap/core': patch
+---
+
+Input rules no longer crash or replace the wrong content when the matched text spans an inline atom node (like a mention). Input rules match against a string where leaf nodes get expanded to placeholders like `%leaf%`, so a match crossing an atom ended up with a range that didn't map back to real document positions, throwing `RangeError: Position -N out of range` or corrupting nearby content. Such matches are now skipped. Rules on plain text, including text after an atom, behave the same as before.
```
bdbch: Please reduce the changelog message to what is important to our end users

bdbch (changes requested):

.changeset/fix-input-rule-inline-atom-range.md
devvaannsh: Apologies, my bad! Updated.

bdbch (approved):

PR #8027, opened 2026-07-12
chore(ci): use explicit release config instead of automatic release type detection

bdbch:
## Fixes

N/A - infrastructure improvement, not a bug fix.

## Changes and Review

Replaced the branch-name-based magic in the publish workflow with an explicit `.github/publish-config.json`. A branch must be listed in both the workflow trigger and the config to publish. If no config entry exists, the workflow exits cleanly.

Some minor formatting changes sneaked themselves into the PR as well but they don't change any logic.

## Checklist

### Responsibility

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,6 +27,7 @@
     "test:e2e:open:firefox": "playwright test --ui --project=firefox",
     "test:e2e:open:all": "playwright test --ui",
     "test:e2e:report": "playwright show-report",
+    "test:publish-config": "node --test .github/scripts/__tests__/*.test.mjs",
     "test": "turbo run build && pnpm run test:unit",
     "build": "turbo build",
     "build:demos": "pnpm --prefix demos run build:demos",
--- a/.changeset/fix-blockquote-backspace-doc-start.md
+++ b/.changeset/fix-blockquote-backspace-doc-start.md
@@ -1,0 +1,5 @@
+---
+'@tiptap/extension-blockquote': patch
+---
+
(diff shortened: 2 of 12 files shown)


arnaugomez (changes requested):
It looks like this pull request also has formatting changes. It would be better to include the formatting changes in a separate pull request. 

There are also merge conflicts. 

Finally, check if the "magic" branch-based release rules were documented anywhere and, if so, remove it.

bdbch:
@arnaugomez resolved

package.json
```diff
@@ -27,6 +27,7 @@
     "test:e2e:open:firefox": "playwright test --ui --project=firefox",
     "test:e2e:open:all": "playwright test --ui",
     "test:e2e:report": "playwright show-report",
+    "test:publish-config": "node --test .github/scripts/__tests__/*.test.mjs",
```
arnaugomez, line 30: Why aren't these tests part of pnpm test:unit?

arnaugomez (dismissed):

package.json
bdbch: I don't really count CI tests as part of the unit test suite for the actual library code. Not sure if we should mix those up. I'd rather have them separated - we could run it before running the actual publish script if we want. I just ran it manually for now.

bdbch: Added an extra validation step that runs `node --test .github/scripts/__tests__/*.test.mjs` directly. The NPM script is just there for manual testing it now.

PR #8029, opened 2026-07-12
fix(core): deduplicate named extensions to prevent duplicate warnings

### LINKED ISSUE #2890: Warning when using extension-dropcursor with multiple editors

I am using multiple editors on the same page which use both image and dropcursor. When they are loaded, I get the following warning in the console:

```
[tiptap warn]: Duplicate extension names found: ['dropCursor']. This can lead to issues. 
```

I'm wondering if there is a way to configure this correctly to avoid the warning?

[Version 1.39.122 Chromium: 102.0.5005.115 (Official Build) (x86_64)](https://brave.com/latest/)

It's re-producable via the sandbox, or:
 1. Create a component called `Editor` which loads the tiptap editor
 2. Enable `extension-dropcursor` in the `Editor` component
 3. In `App.js`, load 2 or more `Editor` components
 4. Look at the console logs

https://codesandbox.io/s/dropcursor-multiple-editors-m0iti5

I expected to be able to find some configuration value which allows me to use `extension-dropcursor` with multiple editors without a warning (for example, a unique Id given to each editor instance which allows the `extension-dropcursor` to know which editor is being dropped on.

I appreciate this may not be a bug but rather a limitation due to the library being in beta, but any suggestions would be highly recommended.

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

bdbch:
## Fixes

Fixes #2890

## Changes and Review

- Deduplicate named extensions in `resolveExtensions` using last-defined precedence.
- This removes the `dropCursor` duplicate warning when `StarterKit` and `Dropcursor` are used together.
- It also stops duplicate ProseMirror plugins from being registered in the same editor.
- Added unit tests in `packages/core/__tests__/resolveExtensions.spec.ts`.

To verify: create an editor with `[StarterKit, Dropcursor]` and confirm there is no duplicate extension warning.

## Checklist

### Responsibility

### DIFF
--- a/packages/core/src/helpers/resolveExtensions.ts
+++ b/packages/core/src/helpers/resolveExtensions.ts
@@ -22,8 +22,10 @@ function deduplicateExtensions(extensions: Extensions): Extensions {
       seen.add(name)
     }
 
-    deduplicated.unshift(extension)
+    deduplicated.push(extension)
   }
+
+  deduplicated.reverse()
 
   return deduplicated
 }
@@ -37,5 +39,5 @@ function deduplicateExtensions(extensions: Extensions): Extensions {
 export function resolveExtensions(extensions: Extensions): Extensions {
   const flattenedExtensions = flattenExtensions(extensions)
 
-  return deduplicateExtensions(sortExtensions(flattenedExtensions))
+  return sortExtensions(deduplicateExtensions(flattenedExtensions))
 }
--- a/packages/core/src/helpers/resolveExtensions.ts
+++ b/packages/core/src/helpers/resolveExtensions.ts
@@ -1,25 +1,41 @@
 import type { Extensions } from '../types.js'
-import { findDuplicates } from '../utilities/findDuplicates.js'
 import { flattenExtensions } from './flattenExtensions.js'
 import { sortExtensions } from './sortExtensions.js'
 
 /**
- * Returns a flattened and sorted extension list while
- * also checking for duplicated extensions and warns the user.
- * @param extensions An array of Tiptap extensions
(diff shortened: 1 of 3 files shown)


packages/core/src/helpers/resolveExtensions.ts
```diff
@@ -1,25 +1,43 @@
...
+    }
+
+    if (name) {
+      seen.add(name)
+    }
+
+    deduplicated.push(extension)
+  }
+
+  deduplicated.reverse()
```
arnaugomez, line 28: I wonder if the order of the extensions in the list matters here, so that we have to reverse.

arnaugomez (approved):
I find this great, although I wonder if it is a breaking change, because now resolveExtensions returns a different value than before.

arnaugomez (commented):
I have another concern here. Following your example, if I declare DropCursor before StarterKit, then the DropCursor extension from StarterKit prevails?

And what if I initialize the same extension twice, but with different configurations?

packages/core/src/helpers/resolveExtensions.ts
bdbch: I'm doing the reverse here because pushing the extension instead of unshift is faster and doesn't increase loop complexity and because we can reverse the array it will do the same thing in the end. :)

But yes the order matters - things will be registered in registration order.

bdbch:
> I have another concern here. Following your example, if I declare DropCursor before StarterKit, then the DropCursor extension from StarterKit prevails?
> 
> And what if I initialize the same extension twice, but with different configurations?

I'm not sure about the fix myself. It depends on how strict we want to be here. Having multiple extensions with

PR #8037, opened 2026-07-13
Feat(extension-youtube): allow string width/height for videos

devvaannsh:
## Fixes
Implements #7372 

## Changes and Review
Before: width and height for the Youtube extension were typed as number only, so passing something like width: "100%" needed a type cast. On top of that, the HTML parser internally auto converted width="100%" into 100 when content was loaded or pasted. This behaviour feels broken.

Now, plain numeric values are still parsed as numbers, and any non-numeric string like "100%", "100px" or "20rem" is preserved as is.

## Checklist

### Responsibility

### DIFF
--- a/packages/extension-youtube/__tests__/youtube.spec.ts
+++ b/packages/extension-youtube/__tests__/youtube.spec.ts
@@ -213,6 +213,121 @@ describe('extension-youtube', () => {
     },
   )
 
+  describe('iframe dimension handling', () => {
+    it('preserves percentage width and height when parsing HTML', () => {
+      editor = new Editor({
+        element: createEditorEl(),
+        extensions: [Document, Text, Paragraph, Youtube],
+        content:
+          '<div data-youtube-video><iframe src="https://www.youtube.com/embed/dQw4w9WgXcQ" width="100%" height="56.25%"></iframe></div>',
+      })
+
+      const attrs = editor.getJSON().content?.[0]?.attrs ?? {}
+
+      expect(attrs.width).toBe('100%')
+      expect(attrs.height).toBe('56.25%')
+
+      editor?.destroy()
+      getEditorEl()?.remove()
+    })
+
+    it('keeps percentage dimensions when content is loaded back from rendered HTML', () => {
+      editor = new Editor({
+        element: createEditorEl(),
+        extensions: [Document, Text, Paragraph, Youtube],
+        content: {
+          type: 'doc',
+          content: [
+            {
+              type: 'youtube',
+              attrs: {
+                src: 'https://www.youtube.com/watch?v=dQw4w9WgXcQ',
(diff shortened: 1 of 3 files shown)


arnaugomez (dismissed):

packages/extension-youtube/src/youtube.ts
```diff
@@ -12,7 +12,14 @@ const getParsedDimension = (value: string | null) => {
...
 
-  const parsedValue = Number.parseInt(value, 10)
+  const trimmedValue = value.trim()
+
+  // keep percentage values like "100%" as strings
+  if (/^\d+(\.\d+)?%$/.test(trimmedValue)) {
+    return trimmedValue
+  }
+
+  const parsedValue = Number.parseInt(trimmedValue, 10)
```
bdbch: I think instead of only accepting percentage values via a regex, we can preserve any non-numeric string.

This would support values like `100%`, `100px`, `20rem`, or `calc(...)`, while still parsing plain numeric values as numbers.

```suggestion
  if (trimmedValue === '') {
    return null
  }

  const parsedValue = Number(trimmedValue)

  return Number.isNaN(parsedValue) ? trimmedValue : parsedValue
```

There's just one drawback that invalid values also are persisted but if strings should be allowed in general I think that's fine. (You could also validate if it's a valid css dimension value here).

packages/extension-youtube/__tests__/youtube.spec.ts
```diff
@@ -213,6 +213,121 @@ describe('extension-youtube', () => {
...
+      const attrs = editor.getJSON().content?.[0]?.attrs ?? {}
+
+      expect(attrs.width).toBe('100%')
+      expect(attrs.height).toBe('56.25%')
+
+      editor?.destroy()
+      getEditorEl()?.remove()
+    })
+
+    it('keeps percentage dimensions when content is loaded back from rendered HTML', () => {
```
bdbch, line 234: Would be nice to have a test with

1. a different string-based size like px/rem/em/vw
2. a invalid string like "not-a-size"

bdbch (changes requested):

packages/extension-youtube/__tests__/youtube.spec.ts
devvaannsh: Done!

packages/extension-youtube/src/youtube.ts
devvaannsh: Done! Funny enough, while testing I realized parseInt was silently turning 20rem into 20 and I'd assumed unit values would fail parsing and fall back to the default, but they were getting truncated into the document instead. 
Your suggestion was right on point! Thanks a lot. Ready for another look!

bdbch (approved):

PR #8084, opened 2026-07-16
fix(core): remove unreachable navigator.platform check in isAndroid (fixes #6560)

AndrianBalanescu:
## Problem

`navigator.platform === "Android"` in `isAndroid.ts` causes a TypeScript build error (TS2367):

```
Type error: This comparison appears to be unintentional because the types
MacIntel

PR #8088, opened 2026-07-16
fix(core): avoid TS literal-overlap error in isAndroid()

### LINKED ISSUE #6560: Type error in isAndroid.ts: navigator.platform comparison causes build failure in TypeScript
### Affected Packages

core

### Version(s)

^2.12.0

## 📝 Issue Description

Hi team,

I’m encountering a TypeScript build error in the file:

./node_modules/@tiptap/core/src/utilities/isAndroid.ts:2:10

The error message is:

Type error: This comparison appears to be unintentional because the types '"MacIntel" | "Win32" | "Linux x86_64"' and '"Android"' have no overlap.

1 | export function isAndroid(): boolean {

2 | return navigator.platform === 'Android' || /android/i.test(navigator.userAgent)
| ^
3 | }

lua
복사
편집

This causes the build to fail completely.

---

### 📦 My `package.json` relevant dependencies:

```json
{
  "@tiptap/extension-code-block-lowlight": "^2.12.0",
  "@tiptap/extension-color": "^2.12.0",
  "@tiptap/extension-heading": "^2.12.0",
  "@tiptap/extension-horizontal-rule": "^2.12.0",
  "@tiptap/extension-image": "^2.12.0",
  "@tiptap/extension-link": "^2.12.0",
  "@tiptap/extension-placeholder": "^2.12.0",
  "@tiptap/extension-text-style": "^2.12.0",
  "@tiptap/extension-typography": "^2.12.0",
  "@tiptap/extension-underline": "^2.12.0",
  "@tiptap/pm": "^2.12.0",
  "@tiptap/react": "^2.12.0",
  "@tiptap/starter-kit": "^2.12.0"
}

- navigator.platform is a string union that does not include "Android" as a possible value, which causes the type error during compilation.

### Browser Used

Chrome

### Code Example URL

_No response_

### Expected Behavior

Typescript build success

<img width="1157" height="105" alt="Image" src="https://github.com/user-attachments/assets/8cd18c4b-6dab-4c32-99a7-bd032bac935e" />

### Dependency Updates

### PR DESCRIPTION
**Pull request by Mohak-Agrawal**
## What was the bug

`isAndroid()` fails to compile under some `lib.dom.d.ts` typings:

```
Type error: This comparison appears to be unintentional because the types
'"MacIntel" | "Win32" | "Linux x86_64"' and '"Android"' have no overlap.
```

This breaks the build entirely for consumers who hit it.

## Root cause

```ts
export function isAndroid(): boolean {
  return navigator.platform === 'Android' || /android/i.test(navigator.userAgent)
}
```

`navigator.platform === 'Android'` is a direct `===` comparison between two string literal types. When TypeScript's `lib.dom.d.ts` (or a consumer's own narrower typing) narrows `Navigator.platform` to a union that doesn't include `'Android'`, this specific comparison shape is flagged as a literal type with no overlap.

The sibling utility `isiOS()` avoids this entirely by comparing against a plain array with `.includes()` instead of `===`, since `Array<string>.includes()` takes a widened `string` parameter rather than triggering literal-overlap narrowing:

```ts
export function isiOS(): boolean {
  return (
    ['iPad Simulator', 'iPhone Simulator', ...].includes(navigator.platform) || ...
  )
}
```

## What the fix does

Applies the same `.includes()` pattern to `isAndroid()`:

```ts
export function isAndroid(): boolean {
  return ['Android'].includes(navigator.platform) || /android/i.test(navigator.userAgent)
}
```

No runtime behavior change — purely a type-level fix, consistent with the existing convention in this file's sibling utility.

## Testing

Added `packages/core/src/__tests__/isAndroid.test.ts` covering both branches (userAgent-based detection, and a non-Android platform). Ran the full validation checklist:

- `pnpm vitest run packages/core/src/__tests__/isAndroid.test.ts` — pass
- `pnpm -w -F @tiptap/core build` — 74/74 tasks succeed
- `pnpm lint` — clean
- `pnpm fallow:audit` — no issues in changed files

Changeset included.

Fixes #6560

### DIFF
--- a/.changeset/2026-07-16-fix-isandroid-type-error.md
+++ b/.changeset/2026-07-16-fix-isandroid-type-error.md
@@ -1,0 +1,5 @@
+---
+'@tiptap/core': patch
+---
+
+Fix a TypeScript build error in `isAndroid()` where comparing `navigator.platform` against the literal `'Android'` with `===` could fail to compile under some `lib.dom.d.ts` typings ("types have no overlap"). Switched to the same `.includes()` pattern already used by `isiOS()`, which is not affected by this TypeScript narrowing issue. No runtime behavior change.
--- a/packages/core/src/__tests__/isAndroid.test.ts
+++ b/packages/core/src/__tests__/isAndroid.test.ts
@@ -16,6 +16,11 @@ describe('isAndroid', () => {
     expect(isAndroid()).toBe(true)
   })
 
+  it('detects Android via platform', () => {
+    mockNavigator('Android', 'Mozilla/5.0 (X11; Linux x86_64)')
+    expect(isAndroid()).toBe(true)
+  })
+
   it('returns false for non-Android platforms', () => {
     mockNavigator('MacIntel', 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7)')
     expect(isAndroid()).toBe(false)
--- a/packages/core/src/__tests__/isAndroid.test.ts
+++ b/packages/core/src/__tests__/isAndroid.test.ts
@@ -1,0 +1,23 @@
+import { afterEach, describe, expect, it, vi } from 'vitest'
+
+import { isAndroid } from '../utilities/isAndroid.js'
+
+function mockNavigator(platform: string, userAgent: string) {
+  vi.stubGlobal('navigator', { platform, userAgent })
+}
+
+describe('isAndroid', () => {
+  afterEach(() => {
+    vi.unstubAllGlobals()
+  })
+
+  it('detects Android via userAgent', () => {
+    mockNavigator('Linux armv8l', 'Mozilla/5.0 (Linux; Android 13; Pixel 7)')
+    expect(isAndroid()).toBe(true)
+  })
+
+  it('returns false for non-Android platforms', () => {
+    mockNavigator('MacIntel', 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7)')
+    expect(isAndroid()).toBe(false)
+  })
+})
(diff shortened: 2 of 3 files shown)


Mohak-Agrawal:
Addressed the CodeRabbit nitpick — added a test case that exercises the `navigator.platform === 'Android'` branch directly (previously only covered via the userAgent regex path).

Also want to flag two things I found after opening this:

1. **Overlap with #8084** — that PR targets the same issue and file. Worth noting the approaches differ: #8084 drops the `navigator.platform` check entirely (a behavior change, flagged by @bdbch in that thread), while this PR keeps both original detection paths intact and only changes `===` to `.includes()` — mirroring the existing pattern in the sibling `isiOS()` utility, so it's a pure type-level fix with no behavior change.

2. **I couldn't reproduce the underlying TS2367 error in this repo's own build.** I temporarily reverted to the original `isAndroid.ts` and ran `pnpm -F @tiptap/core build` — it succeeds cleanly (74/74 tasks, DTS included) using this repo's own `tsconfig.json` (`lib: [esnext, dom, dom.iterable, scripthost]`). That lib doesn't narrow `Navigator.platform` to a literal union anywhere, so the comparison type-checks fine here. This lines up with @arnaugomez's comment on #8084 — consumers normally only see the published `.d.ts`, not this source file, so whatever's happening in the original reporter's project is likely coming from a non-standard `Navigator` type augmentation in their own setup, not this repo's build or published types.

Given that, this PR is safe to merge if you want a defensive, zero-behavior-change fix for whatever environment the reporter hit — but it's also fair to close both this and #8084 and ask for a minimal repro first, per @arnaugomez's suggestion on the other thread. Deferring to your call either way.

bdbch (approved):

PR #8092, opened 2026-07-16
test(e2e): deflake keyboard-shortcut test in Default example

### LINKED ISSUE #8080: Flaky tests in CI verify test stability across the suite
**Affected Packages**
tests (e2e / playwright)

**Tiptap Version**
latest (main)

**Browser Used**
Not browser-specific

**What happened?**
Several tests in the Playwright e2e suite are flaky and fail intermittently in CI without code changes. A user reported the following test is failing often:

> `[chromium] › demos/src/Examples/Default/index.spec.ts:42:7 › /src/Examples/Default › Svelte › should apply the paragraph style when the keyboard shortcut is pressed`

Ref: https://github.com/ueberdosis/tiptap/pull/7893#issuecomment-4983453516

Flaky tests erode confidence in CI — failures get ignored or retried, which masks real regressions. The test suite should be deterministic.

**Expected Behavior**
Tests pass or fail consistently — no intermittent failures due to timing, race conditions, or missing wait conditions.

**Additional Context (Optional)**
This seems like a broader issue beyond one test. We should audit the Playwright test suite for common flakiness patterns:
- Missing or insufficient `waitFor` / assertion retries
- Race conditions with editor initialization or async state
- Hard-coded timeouts instead of proper wait conditions
- Tests that depend on exact DOM timing rather than editor state

codewithsupra:
Fixes #8080

Reproduced the flake locally: 6/30 failures running the Svelte test with 4 parallel workers. Instrumenting each step showed the cause — the test clicks the editor, then programmatically selects all, then presses the shortcut. The mouse click fires an async `selectionchange` that under load lands *after* the programmatic selectAll (sometimes after the keydown), collapsing the selection into the trailing empty paragraph. The shortcut then no-ops on an already-empty paragraph, the h1 survives, and the assertion finds the empty trailing `<p>` — exactly the `""` failure CI reports.

Fix: drop the click and use the same `focus()` + `selectAll()` evaluate that every other shortcut spec already uses. After the change: 150/150 repeats pass under the same parallel load, and the full Default spec passes on chromium + firefox.

I also audited the rest of the suite for the patterns listed in the issue: no `waitForTimeout` sleeps anywhere, and all other keyboard-shortcut specs already use the programmatic focus/select idiom — this test was the one straggler still clicking first.

### DIFF
--- a/demos/src/Examples/Default/index.spec.ts
+++ b/demos/src/Examples/Default/index.spec.ts
@@ -45,8 +45,14 @@ test.describe(`${demoPath}/${demoName}`, () => {
         await expect(page.locator('.tiptap h1')).toBeVisible()
 
         const editor = await getEditor(page)
-        await editor.click()
-        await editor.evaluate((el: any) => el.editor.commands.selectAll())
+        // Focus and select programmatically instead of clicking first: a real
+        // mouse click emits an async `selectionchange` that can land after the
+        // programmatic selectAll (or even after the keydown) under CI load,
+        // collapsing the selection before the shortcut applies.
+        await editor.evaluate((el: any) => {
+          el.editor.commands.focus()
+          el.editor.commands.selectAll()
+        })
         await editor.press(process.platform === 'darwin' ? 'Meta+Alt+0' : 'Control+Alt+0')
 
         await expect(page.locator('.tiptap p').first()).toContainText('Example Text')


bdbch:
> **Link your pull request to an issue** — Pull requests must be linked to an existing issue that has been assigned to you. Before opening a PR, ensure there is an issue describing the bug or feature you're addressing. Trivial fixes (e.g. typos or broken links) are exempt.

**Thanks for the contribution, but I’m closing this PR because it does not follow our contribution guidelines.**

Pull requests must be linked to an existing issue that has been assigned to you before work begins. This helps us avoid duplicate work, align on the approach, and keep contributions manageable.

Please open an issue first and wait until it has been approved and assigned to you before submitting a new PR linked to that issue.

**Note:** Please use the official pull request template and keep the description concise and specific to the change. Large AI-generated descriptions do not make a PR easier to review and should be avoided.

PR #8095, opened 2026-07-17
fix(react): onCreate fired twice under React.StrictMode in useEditor

### LINKED ISSUE #7091: UseEditor onCreate event is called twice with two different editor instances in React.StrictMode

a-y-ibrahim:
## Fixes

Fixes #7091

## Changes and Review

`useEditor()`'s `EditorInstanceManager` is created via a `useState` lazy initializer. React 18/19 StrictMode invokes that initializer twice for a single mount, and both calls construct (and auto-mount, since `Editor` defaults to a detached div) a real `Editor`, so both used to schedule their own `create` event.

`onCreate` is now buffered until the owning `EditorInstanceManager` is confirmed mounted (i.e. its render effect actually runs), and only forwarded to the user's callback then. An instance whose effect never runs - the StrictMode duplicate - simply never forwards its buffered event. This doesn't depend on any cross-invocation persistence (e.g. a ref surviving StrictMode's double-invoke), which turned out not to be reliable across React versions: verified against real React 17.0.2, 18.2.0, 18.3.1, and 19.1.0 builds using the actual built `@tiptap/react` dist, not just this repo's own React 19 devDependency.

Added `useEditor.spec.ts` (`@testing-library/react`, matching `BubbleMenu.spec.ts`'s existing pattern) covering: single `onCreate` under StrictMode, independent sibling editors, `immediatelyRender: false`, a full unmount/remount cycle, and mounting without StrictMode.

## Checklist

### Responsibility

### DIFF
--- a/.changeset/fix-react-strictmode-double-oncreate.md
+++ b/.changeset/fix-react-strictmode-double-oncreate.md
@@ -1,0 +1,5 @@
+---
+'@tiptap/react': patch
+---
+
+Fix `useEditor` firing `onCreate` twice for two different `Editor` instances under React StrictMode. StrictMode invokes the hook's internal lazy initializer twice, and both calls used to construct and auto-mount a real `Editor`; the discarded instance is now destroyed synchronously, before its `create` event has a chance to fire.
--- a/.changeset/fix-react-strictmode-double-oncreate.md
+++ b/.changeset/fix-react-strictmode-double-oncreate.md
@@ -2,4 +2,4 @@
 '@tiptap/react': patch
 ---
 
-Fix `useEditor` firing `onBeforeCreate`/`onCreate` twice for two different `Editor` instances under React StrictMode. `useEditor` used a `useState` lazy initializer, which StrictMode invokes twice for a single mount, and both calls constructed (and auto-mounted) a real `Editor`. Replaced it with a guarded `useRef`, whose value isn't reset between StrictMode's double invocation, so only one `Editor` is ever constructed per mount.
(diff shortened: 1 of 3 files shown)


arnaugomez:
I'm of the opinion that onCreate running twice is not a bug - it should be the default behavior, because it helps enforce React best practices and it aligns with its philosophy of keeping initializers free from side-effects.

See my comment here: https://github.com/ueberdosis/tiptap/issues/7091#issuecomment-5003491946

@bdbch let me know if you agree on that. Perhaps I'm wrong.

PR #8096, opened 2026-07-19
fix(core): collapse selection after deleting an AllSelection

### LINKED ISSUE #7961: Phantom selection remains after Ctrl+A → Backspace clears editor content
### Affected Packages

core

### Version(s)

3.26.1

## Describe the bug

After typing content into a `useEditor` instance, selecting all with Ctrl+A (Cmd+A on Mac), and pressing Backspace to delete, a small phantom selection visual remains visible in the top-left corner of the editor. The editor content is correctly cleared, but a residual selection highlight/indicator persists until the user clicks or interacts with the editor again.

This is also reproducible on the TipTap website's own editor demos.

## To reproduce

1. Mount a basic TipTap editor with StarterKit
2. Type any text (e.g. "hello world")
3. Press Ctrl+A (Cmd+A) to select all
4. Press Backspace to delete the selection
5. Observe the small phantom selection indicator at the top-left of the editor

### Browser Used

Chrome

### Code Example URL

_No response_

### Expected Behavior

After the content is cleared, no selection visual should remain. The cursor should appear as a collapsed caret with no highlighted region.

## Root cause analysis

Ctrl+A triggers ProseMirror's `selectAll` command, which creates an `AllSelection` spanning the entire document. When Backspace processes the `AllSelection` and clears the content, ProseMirror's internal state is correct (empty document, cursor at position 1), but the browser's native selection (`window.getSelection()`) is not properly collapsed after the deletion. This leaves a ghost selection artifact at the original anchor point of the `AllSelection`.

The bug is specific to `AllSelection` — it does not occur when the same content is selected via click-and-drag or Shift+arrow keys (which create a `TextSelection`).

## Workaround
Override `Mod-a` with a custom extension to use `TextSelection` instead of `AllSelection`. A TextSelection collapses cleanly after deletion:

```js
import { Extension } from "@tiptap/core";

const SelectAllFix = Extension.create({
  name: "selectAllFix",
  priority: 1000,
  addKeyboardShortcuts() {
    return {
      "Mod-a": ({ editor }) => {
        const { doc } = editor.state;
        editor.commands.setTextSelection({ from: 1, to: doc.content.size - 1 });
        return true;
      },
    };
  },
});

// Add to useEditor extensions: [..., SelectAllFix]
```

### Dependency Updates

RIDAM7:
Fixes #7961

## Problem
After Ctrl/Cmd+A (an `AllSelection`) then Backspace, the document empties but a
selection highlight lingers in the top-left. `AllSelection.map()` returns another
`AllSelection`, so the custom `deleteSelection` command emptied the doc while the
selection stayed a whole-document range instead of collapsing to a cursor.

## Fix
`deleteSelection` now collapses any surviving non-empty selection to a text cursor
after the deletion, so an emptied editor shows a caret rather than a phantom
highlight.

## Testing
- Added a regression test in `packages/core/__tests__/delete.spec.ts` asserting the
  selection is empty and no longer an `AllSelection` after `selectAll().deleteSelection()`.
- All existing core / table / list / selection delete tests pass; `oxlint` clean.
- Included a changeset.

## AI disclosure
Claude (Anthropic) assisted in diagnosing the root cause and drafting this fix.

### DIFF
--- a/.changeset/fix-allselection-delete-phantom-selection.md
+++ b/.changeset/fix-allselection-delete-phantom-selection.md
@@ -1,0 +1,5 @@
+---
+'@tiptap/core': patch
+---
+
+Fixed a bug where deleting an `AllSelection` (for example right after Ctrl/Cmd+A) left a lingering "phantom" selection highlight over the emptied document instead of a text cursor. `deleteSelection` now collapses the selection to a cursor.
--- a/packages/core/__tests__/delete.spec.ts
+++ b/packages/core/__tests__/delete.spec.ts
@@ -3,6 +3,7 @@ import Bold from '@tiptap/extension-bold'
 import Document from '@tiptap/extension-document'
 import Paragraph from '@tiptap/extension-paragraph'
 import Text from '@tiptap/extension-text'
+import { AllSelection } from '@tiptap/pm/state'
 import { describe, expect, it, vi } from 'vitest'
 
 const InlineDocument = Document.extend({
@@ -140,6 +141,21 @@ describe('delete extension', () => {
 
     // The node should remain with remaining text
     expect(editor.getHTML()).toBe('<p>before <span data-test-node="">st</span> after</p>')
+    editor.destroy()
+  })
+
+  it('collapses the selection to a cursor after deleting an AllSelection', () => {
+    const editor = new Editor({
+      extensions: [Document, Paragraph, Text],
+      content: '<p>hello world</p>',
+    })
+
+    editor.chain().selectAll().deleteSelection().run()
+
+    expect(editor.getHTML()).toBe('<p></p>')
+    expect(editor.state.selection.empty).toBe(true)
+    expect(editor.state.selection instanceof AllSelection).toBe(false)
+
     editor.destroy()
   })
 
--- a/packages/core/src/commands/deleteSelection.ts
+++ b/packages/core/src/commands/deleteSelection.ts
@@ -1,4 +1,5 @@
 import type { ResolvedPos, Schema } from '@tiptap/pm/model'
+import { TextSelection } from '@tiptap/pm/state'
 
 import type { RawCommands } from '../types.js'
 
@@ -92,6 +93,13 @@ export const deleteSelection: RawCommands['deleteSelection'] =
         const { from, to } = expandSelectionForInlineText($from, $to, state.schema)
         tr.deleteRange(from, to)
       })
+
+      // after deleting the selection we restore a text selection
+      // near the from position to avoid having confusing selections
+      // after the deletion (for example after deleting via Ctrl/Cmd+A)
+      if (!tr.selection.empty) {
+        tr.setSelection(TextSelection.near(tr.doc.resolve(tr.selection.from)))
+      }
 
       tr.scrollIntoView()
       dispatch(tr)


.changeset/fix-allselection-delete-phantom-selection.md
```diff
@@ -0,0 +1,5 @@
+---
+'@tiptap/core': patch
+---
+
+Fixed a bug where deleting an `AllSelection` (for example right after Ctrl/Cmd+A) left the selection as a whole-document range over the now-empty document instead of collapsing to a text cursor. Because `AllSelection.map()` returns another `AllSelection`, the emptied editor kept a lingering "phantom" selection highlight. `deleteSelection` now collapses any surviving non-empty selection to a cursor.
```
bdbch: Can we please make this PR shorter? I think we don't need the "Because" - just what was fixed and what changes for the user.

packages/core/__tests__/delete.spec.ts
```diff
@@ -143,6 +144,24 @@ describe('delete extension', () => {
...
 
+  it('collapses the selection to a cursor after deleting an AllSelection (#7961)', () => {
+    const editor = new Editor({
+      extensions: [Document, Paragraph, Text],
+      content: '<p>hello world</p>',
+    })
+
+    // Ctrl/Cmd+A produces an AllSelection; deleting it used to leave the
+    // selection as a whole-document AllSelection over the now-empty doc, which
+    // renders as a lingering "phantom" highlight instead of a text cursor.
```
bdbch: I don't think we need this explanation here.


```suggestion
```

packages/core/__tests__/delete.spec.ts
```diff
@@ -143,6 +144,24 @@ describe('delete extension', () => {
     editor.destroy()
   })
 
+  it('collapses the selection to a cursor after deleting an AllSelection (#7961)', () => {
```
bdbch:
```suggestion
  it('collapses the selection to a cursor after deleting an AllSelection', () => {
```

packages/core/src/commands/deleteSelection.ts
```diff
@@ -93,6 +94,15 @@ export const deleteSelection: RawCommands['deleteSelection'] =
         tr.deleteRange(from, to)
       })
 
+      // A non-collapsing selection (e.g. the `AllSelection` from Ctrl/Cmd+A)
+      // maps to itself across the deletion: `AllSelection.map()` returns another
+      // `AllSelection`, so once the document is emptied the selection is still a
+      // whole-document range instead of a cursor, leaving a "phantom" highlight
+      // in the view (#7961). Collapse whatever range survives to a text cursor.
```
bdbch:
```suggestion
      // after deleting the selection we restore a text selection 
      // near the from position to avoid having confusing selections
      // after the deletion (for example after deleting via Ctrl/Cmd+A)
```

bdbch (changes requested):
Thanks for your contribution! I added a few nitpicks.

.changeset/fix-allselection-delete-phantom-selection.md
RIDAM7: Done — removed the "Because…" explanation. The changeset now just states what was fixed and the user-facing effect.

packages/core/__tests__/delete.spec.ts
RIDAM7: Removed, thanks.

packages/core/__tests__/delete.spec.ts
RIDAM7: Done — dropped the issue number from the test title.

packages/core/src/commands/deleteSelection.ts
RIDAM7: Applied your suggested wording, thanks.

RIDAM7:
Thanks for the quick review @bdbch! Addressed all of them. Trimmed the changeset (dropped the "Because" bit), removed the extra comments in the test and in deleteSelection, and took the issue number out of the test title. Let me know if there's anything else!

bdbch (approved):
Thanks ❤️

PR #8097, opened 2026-07-19
fix: stop vitest's global esbuild jsx default from overriding React JSX

a-y-ibrahim:
## Fixes

Not tied to a filed issue - found this while investigating a different issue and figured it was worth fixing directly since I already had a verified fix.

## Changes and Review

`vitest.config.ts` sets `esbuild.jsxImportSource` to `@tiptap/core` globally, but that's only meant for the few files (`bold.tsx`, `blockquote.tsx`) that already opt into it themselves via a local `@jsxImportSource` pragma. The global default silently overrides plain React JSX for every other `.tsx` file under vitest too, including `packages/react`'s own components - rendering `<EditorContent>` in a vitest test throws `Objects are not valid as a React child` immediately, since the compiled output is Tiptap's own PM-array format instead of real React elements.

Removing the global default fixes that. It doesn't touch real builds (tsup reads `tsconfig.build.json` directly), and the two files that need the custom JSX keep working via their own pragma. Added a small regression test for `<EditorContent>` and confirmed the full suite still passes (150 files, 1517 tests).

## Checklist

- [x] I have added a [changeset](https://github.com/changesets/changesets) if necessary. (Not needed - no published package is affected, this only changes test config.)

### Responsibility

### DIFF
--- a/packages/react/src/EditorContent.spec.ts
+++ b/packages/react/src/EditorContent.spec.ts
@@ -1,8 +1,13 @@
+import { Document } from '@tiptap/extension-document'
+import { Paragraph } from '@tiptap/extension-paragraph'
+import { Text } from '@tiptap/extension-text'
+import { render } from '@testing-library/react'
 import React from 'react'
-import { describe, expect, it, vi } from 'vitest'
+import { afterEach, describe, expect, it, vi } from 'vitest'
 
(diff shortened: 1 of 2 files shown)


bdbch:
Thanks for your PR - already fixing this in another PR right now together with react related tests.

PR #8099, opened 2026-07-20
Feat(extension-invisible-characters): show tabs and non-breaking spaces

devvaannsh:
## Changes and Review
Currently, the invisible characters extension can show spaces, paragraphs and line breaks, but tabs and non-breaking spaces stay invisible. This PR adds two new builders, `TabCharacter` and `NonBreakingSpaceCharacter`, so those show up too (`→` for tabs and `°` for non-breaking spaces).

They are opt-in, so nothing changes for existing setups.

---
**Before**
Screenshot 2026-07-20 at 6 25 31 PM

**After**
Screenshot 2026-07-20 at 6 25 48 PM

---

## Checklist

### Responsibility


alexvcasillas (approved):
Overall: Approve. Clean, consistent with existing patterns, well-tested.
The dedup change to InvisibleCharacter.createDecoration is a nice catch —
it also fixes a pre-existing duplicate-widget bug for the space character
on multi-step transactions, not just the new builders.

Nits (none blocking):

1. Missing JSDoc on the two new classes (TabCharacter,
   NonBreakingSpaceCharacter). The repo's CLAUDE.md asks for JSDoc on every
   class/type. The existing sibling builders (SpaceCharacter, HardBreakNode)
   also lack them, so it's consistent with current code — but a one-line
   JSDoc on each new public export would satisfy the standard.

2. Docs not updated. Out of scope for this OSS PR (docs live in the separate
   docs repo), but the new builders should be documented on the
   invisible-characters page as a follow-up. Worth noting in the PR description.

3. The `°` (degree sign) glyph for non-breaking space is unconventional —
   most editors use something like `⍽`, `⎵`, or a shaded box; `°` reads as
   "degree" to many users. Purely cosmetic and CSS-overridable, but worth
   confirming it's the intended symbol.

Non-issue, just flagging:
- The new find/remove/add runs for every matching character on every update,
  even single-step cases where nothing needs removing. It's bounded by the
  updated range (not the whole doc) and matches the existing InvisibleNode
  approach, so the cost is fine — just slightly more allocation per char.

PR #8101, opened 2026-07-21
feat(extension-find-and-replace): add find and replace extension

bdbch:
## Changes and Review

- Adds `@tiptap/extension-find-and-replace`, a new extension that searches the document, highlights matches with decorations, and supports replace current, replace all, and navigation between results.
- Supports case-sensitive and regex search modes via commands and storage.
- Includes React and Vue demos, unit tests, and a changeset.

Reviewers can run `pnpm test:unit -- packages/extension-find-and-replace` and try the FindAndReplace demos with `pnpm dev`.

## Checklist

### Responsibility


Aslam97:
## Here are all findings that I already verifed @bdbch 

### 1. Chained commands overwrite earlier settings

```ts
function setPluginMeta(meta: FindAndReplaceMeta): Command {
  return ({ tr, dispatch }) => {
    if (dispatch) {
      tr.setMeta(FindAndReplacePluginKey, meta)
    }

    return true
  }
}
```

Reproduction:

```ts
editor
  .chain()
  .setSearchTerm('hello')
  .setReplaceTerm('world')
  .setCaseSensitive(true)
  .run()
```

Actual state:

```ts
{
  searchTerm: '',
  replaceTerm: '',
  caseSensitive: true,
}
```

### 2. Initial configured results are missing from extension storage

Current storage initialization:

```ts
addStorage() {
  return {
    searchTerm: this.options.searchTerm,
    replaceTerm: this.options.replaceTerm,
    caseSensitive: this.options.caseSensitive,
    useRegex: this.options.useRegex,
    wholeWord: this.options.wholeWord,
    results: [],
    currentIndex: null,
  }
}
```

The plugin separately calculates the initial results:

```ts
function createState(doc: Node, options: FindAndReplaceOptions) {
  const results = searchDocument(doc, options.searchTerm, options)
  const currentIndex = results.length > 0 ? 0 : null

  return {
    // ...
    results,
    currentIndex,
  }
}
```

Storage is synchronized only inside `onTransaction()`, so it does not receive the plugin’s initial state immediately.

Reproduction:

```ts
const editor = new Editor({
  extensions: [
    FindAndReplace.configure({
      searchTerm: 'hello',
    }),
  ],
  content: '<p>Hello hello HELLO</p>',
})

editor.storage.findAndReplace.results
```

Expected:

```ts
[
  { from: 1, to: 6 },
  { from: 7, to: 12 },
  { from: 13, to: 18 },
]
```

Actual:

```ts
[]
```

`currentIndex` is also `null` instead of `0`.

### 3. An edit can create results without selecting a current result

```ts
function clampIndex(
  currentIndex: number | null,
  results: SearchResult[],
  meta: FindAndReplaceMeta | undefined,
): number | null {
  if (results.length === 0) {
    return null
  }

  if (isNewSearch(meta)) {
    return 0
  }

  if (currentIndex === null) {
    return null
  }

  return Math.min(currentIndex, results.length - 1)
}
```

Reproduction:

```ts
editor.commands.setSearchTerm('world')
// No results, currentIndex is null.

editor.commands.setContent('<p>world</p>')
```

Actual:

```ts
results: [{ from: 1, to: 6 }]
currentIndex: null
currentHighlightCount: 0
```

The result list is non-empty, but there is no current result.

Suggested fix: whenever results are non-empty, maintain a valid current index. If there was no previous result and no explicit index was requested, select index `0`.

### 4. The current occurrence changes when a new match is inserted before it

The current implementation keeps only a numeric array index:

```ts
return Math.min(currentIndex, results.length - 1)
```

Reproduction:

```ts
// Document: "foo bar foo"
editor.commands.setSearchTerm('foo')
editor.commands.goToNextResult() // Select the second "foo".

editor.commands.insertContentAt(1, 'foo ')
```

The new results are:

```ts
[
  { from: 1, to: 4 },
  { from: 5, to: 8 },
  { from: 13, to: 16 },
]
```

Expected current index:

```ts
2 // The original second occurrence.
```

Actual current index:

```ts
1 // Now points to a different occurrence.
```

Suggested fix: map the previous current result position through the transaction before choosing the new index.

For example:

```ts
const previousResult = previousState.results[previousState.currentIndex]
const mappedFrom = tr.mapping.map(previousResult.from, 1)
const nextIndex = findNextIndex(results, mappedFrom)
```

### 5. Regex can split one emoji into two partial matches

```ts
return new RegExp(source, caseSensitive ? 'g' : 'gi')
```

Reproduction:

```ts
// Document contains one emoji: "😀"
editor.commands.setUseRegex(true)
editor.commands.setSearchTerm('.')
```

Expected:

```ts
[{ from: 1, to: 3 }]
```

Actual:

```ts
[
  { from: 1, to: 2 },
  { from: 2, to: 3 },
]
```

This allows replacement to target individual UTF-16 surrogate halves.

Suggested fix:

```ts
return new RegExp(source, caseSensitive ? 'gu' : 'giu')
```

### 6. Whole-word matching is incorrect for Unicode words

Current implementation:

```ts
const escaped = escapeRegExp(term)
source = options.wholeWord ? `\\b${escaped}\\b` : escaped
```

JavaScript `\b` uses ASCII-style word boundaries and does not treat letters such as `é` as normal word characters.

Reproduction:

```ts
// Document: "café caféine café"
editor.commands.setSearchTerm('café')
editor.commands.setWholeWord(true)
```

Expected: two standalone `café` results.

Actual:

```ts
[{ from: 6, to: 10 }]
```

It misses both standalone words and incorrectly matches the `café` prefix inside `caféine`.

Suggested fix: use Unicode-aware boundaries, for example Unicode property escapes with negative lookaround, or `Intl.Segmenter` where appropriate.

A possible direction:

```ts
const wordCharacter = String.raw`[\p{L}\p{N}\p{M}_]`

source =
  `(?<!${wordCharacter})` +
  escaped +
  `(?!${wordCharacter})`
```

This requires Unicode regex mode.

### 7. The regex safety heuristic still allows catastrophic backtracking

Current check:

```ts
// Matches identical alternatives such as (a|a)+.
if (/\(([^|]+)\|(\1)\)[+*]/.test(pattern)) {
  return true
}
```

It does not detect overlapping alternatives such as:

```regex
^(a|aa)+$
```

Verification:

```ts
createSearchRegex('^(a|aa)+$', {
  caseSensitive: true,
  useRegex: true,
  wholeWord: false,
})
```

Expected: rejected or safely isolated.

Actual:

```ts
/^(a|aa)+$/g
```

A 40-character non-matching input took approximately 1.1 seconds in the benchmark, and runtime grows exponentially.

The existing heuristic is a useful guardrail, but it should not be treated as a complete ReDoS prevention mechanism.

Suggested options:

- Use a non-backtracking engine such as RE2.
- Run regex matching in a cancellable worker with a time budget.
- Add strict pattern/document-size limits.
- At minimum, document that the current check is heuristic and add overlapping-alternation tests.

### 8. Position lookup is quadratic in mark-heavy paragraphs

Current `offsetToPos` starts from the first text segment for every result:

```ts
function offsetToPos(segments: TextSegment[], offset: number): number {
  let remaining = offset

  for (const segment of segments) {
    if (remaining < segment.text.length) {
      return segment.pos + Math.min(remaining, segment.length)
    }

    remaining -= segment.text.length
  }

  // ...
}
```

Each result calls this twice:

```ts
results.push({
  from: offsetToPos(segments, match.index),
  to: offsetToPos(segments, match.index + match[0].length),
})
```

With 10,000 alternating marked segments and 10,000 matches, the search took approximately 69 ms in the isolated PR-head benchmark.

Suggested fix: store a cumulative text offset for every segment and use binary search when converting a match offset to a document position.

The same benchmark takes approximately 3 ms with that change.

### 9. Every document edit performs a complete search and decoration rebuild

Current plugin update:

```ts
if (docChanged || touchesSearch(meta)) {
  return refreshState(state, doc, meta)
}
```

`refreshState` then does:

```ts
const results = searchDocument(doc, state.searchTerm, state)
const currentIndex = clampIndex(state.currentIndex, results, meta)

return {
  ...state,
  results,
  currentIndex,
  decorations: createDecorations(doc, results, currentIndex),
}
```

Therefore, one character edit searches the complete document and recreates every decoration.

Verified benchmark:

```text
1 edit with 8,000 active results: ~29 ms
```

In the browser test with approximately 20,000 matches, one edit took around 98 ms.

Suggested fix: map existing decorations through the transaction and rescan only the textblocks affected by changed ranges.

### 10. Replace-all creates one transaction step per result

Current implementation:

```ts
const results = [...pluginState.results].reverse()

results.forEach(result => {
  tr.insertText(pluginState.replaceTerm, result.from, result.to)
})
```

Verified isolated benchmark with 20,000 matches:

```text
replaceAll command: ~1.274 seconds
```

After the command, the test worker exceeded roughly 3.7 GB of heap and terminated with an out-of-memory error. In the real browser test, the UI was blocked for several seconds before recovering.

Suggested fix: reduce the number of transaction steps, group compatible replacements, or introduce a documented result limit. Any batching approach should preserve undo and collaboration behavior.

alexvcasillas:
Really nice work on this, the overall design is clean and reads well.

A couple of things I ran into that might be worth a look before release:

### 1. The regex safety check (`search.ts`, `isUnsafeRegex`)

Since regex mode runs the pattern over the whole document on each keystroke, I get why there's a guard here, that's a sensible thing to protect against. A couple of edge cases I noticed while poking at it:

- Some catastrophic patterns still slip past it, for example `(a|aa)+`, `(.*a){20}`, or `a*a*a*b`, so it may not fully cover the freeze case it's aiming at.
- On the other side, a few safe patterns get caught by the nested-quantifier and `.`-in-group rules and return no results, for example `(ab+)+` or `(a.b)+`. Since there's no feedback in that case, it might be a little confusing for someone using a valid regex.

No strong opinion on the direction, but a couple of options could be: leaning into "best-effort" and documenting that regex mode runs user-supplied patterns, or handling it more robustly (time-bounding the search or moving it off the main thread). Even just relaxing the rules that reject valid patterns would probably help the common case.

### 2. Result count on large documents

Right now every match is collected and decorated, and the search re-runs on each doc change and keystroke. On a large document with a common term that could add up to a lot of decorations. Might be worth a `maxResults` cap, or a note suggesting callers debounce `setSearchTerm`, whatever feels right.

### Minor things (all non-blocking)

- Whole-word uses `\b`, which is ASCII-oriented, so results may be surprising for accented/CJK text. Could be worth a quick doc note.
- `replace` ends up searching the document twice per call (once in `replaceResult`, then again in the plugin's `apply` on `docChanged`). Works fine, just a small redundancy if it's easy to avoid.
- The injected `<style>` isn't removed on destroy. It's only injected once so it's harmless, and it matches what `extension-invisible-characters` does, just noting it.
- Small thing: the PR description mentions "case-sensitive and regex" but the feature also has whole-word.

Everything else I traced through looked correct: offset-to-position mapping, navigation wrap-around, `replace` skipping a still-matching replacement, `replaceAll` inserting in reverse, and index clamping. Great addition overall.

alexvcasillas (dismissed):
Looks good to me! Added a couple comments we might want to look into, nothing blocking though :)

bdbch:
@coderabbitai full review

alexvcasillas (dismissed):

alexvcasillas (approved):

PR #8108, opened 2026-07-22
fix(core): make editor usable in onContentError for invalid initial content (#7493)

### LINKED ISSUE #7493: onContentError behavior passes broken editor on first load
### Affected Packages

react

### Version(s)

3.18.0

### Bug Description

If the initial content is invalid, the `editor` that is passed to `onContentError()` in this case is in an invalid state and the content can't be updated.

### Browser Used

Chrome

### Code Example URL

https://codesandbox.io/p/sandbox/summer-river-mkj4zl

### Expected Behavior

`onContentError` should be invoked with an editor that can have content updated / fixed, even if it is the initial content that is invalid.

I'm trying to figure out how to handle invalid content. Based on #3437, it looks like the recommended way to do this is `onContentError()`.

If an editor has loaded content successfully, and an operation produces invalid content, and `emitContentError: true`, then `onContentError()` gets invoked as I would expect. This lets me then do something like this:

```
onContenterror: ({ editor }) => {
  const content = editor.getJSON();
  const cleanContent = sanitizeContent(content);
  editor.commands.setContent(cleanContent);
},
```

This doesn't work if the editor's initial load has invalid content, however. In this case, I need to explicitly set BOTH `emitContentError: true` and `enableContentCheck: true`. `onContentError()` then gets invoked, but the editor is in an invalid state so I can't sanitize and fix the content.

I have a repro case here:

https://codesandbox.io/p/sandbox/summer-river-mkj4zl

You can change this node type to trigger bad initial content:

Image

This then triggers an error about in invalid transaction, which means that this error is unrecoverable:

```
Cannot destructure property 'tr' of 'state' as it is undefined.
```

Image

Is there another way to handle this that I'm missing?

Even if there needed to be an extra `initialContentInvalid: boolean` in the `onContentError` params, I think that would be an improvement. As it is, I'm left with an invalid editor that I can't reset even to the empty string.

### Dependency Updates

alexvcasillas:
## Fixes

Fixes #7493

## Changes and Review

When an editor was created with invalid initial content plus `enableContentCheck` + `emitContentError`, `onContentError({ editor })` fired before the editor had a state, so calling `editor.commands.setContent(…)` (the documented recovery path) inside the handler threw `Cannot destructure property 'tr' of 'state'`.

- `createDoc()` now parses the stripped fallback document and seeds `editor.state` from it **before** emitting `contentError`, so the handler receives a usable editor.
- The constructor only builds the initial state itself when the error path didn't already seed one, so a handler's `setContent` (or any other command) is preserved and the fallback content isn't discarded.
- Scope is `@tiptap/core` only (patch); `disableCollaboration()` stays synchronous and unchanged.
- To verify: create an editor with invalid content + `enableContentCheck: true` + `emitContentError: true` and call `editor.commands.setContent(valid)` in `onContentError`, and it no longer throws and the content is fixed.

## Checklist

### Responsibility

### DIFF
--- a/.changeset/fix-oncontenterror-editor-usable-quiet-lake-drift.md
+++ b/.changeset/fix-oncontenterror-editor-usable-quiet-lake-drift.md
@@ -1,0 +1,5 @@
+---
+'@tiptap/core': patch
+---
+
+Fixed `onContentError` throwing when calling `editor.commands` from inside the handler on initial load with invalid content. The editor now has a usable state (seeded from the stripped fallback document) before `onContentError` fires.
--- a/packages/core/__tests__/onContentError.spec.ts
+++ b/packages/core/__tests__/onContentError.spec.ts
@@ -208,4 +208,79 @@ describe('onContentError', () => {
     expect(editor.getText()).toBe('Example Text')
     expect(editor.storage.collaboration.isDisabled).toBe(false)
   })
+
+  it('does not throw when calling setContent(valid) from the onContentError handler', () => {
+    const json = {
+      invalid: 'doc',
+      content: [
+        {
+          type: 'paragraph',
+          content: [
+            {
+              type: 'text',
+              text: 'Example Text',
+            },
+          ],
+        },
+      ],
+    }
+
+    const validJson = {
+      type: 'doc',
+      content: [
+        {
+          type: 'paragraph',
+          content: [
+            {
+              type: 'text',
+              text: 'Recovered',
+            },
+          ],
+        },
+      ],
+    }
+
+    let editor: Editor | undefined
+
+    expect(() => {
+      editor = new Editor({
+        content: json,
+        extensions: [Document, Paragraph, Text],
+        enableContentCheck: true,
+        onContentError: ({ editor: contentErrorEditor }) => {
+          contentErrorEditor.commands.setContent(validJson)
+        },
+      })
+    }).not.toThrow()
+
+    expect(editor?.getText()).toBe('Recovered')
+  })
+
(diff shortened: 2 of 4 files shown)


bdbch (dismissed):
Looks good to me, I'd just double check that collaboration doesn't clash somehow. But I think it shouldn't.

alexvcasillas:
Good call @bdbch. Double-checked with the real Collaboration extension: created an editor with invalid initial content + `enableContentCheck`, and had `onContentError` call `disableCollaboration()` then `setContent()`. No clash. The fallback state is seeded from the original (collab-inclusive) schema, and since Collaboration contributes nothing to the schema and `disableCollaboration()` only rebuilds the extension manager (not the schema), the seeded state stays valid through mount. Added a regression test in `filterInvalidContent.spec.ts` covering exactly this: handler disables collaboration and recovers content, collab storage is removed, content is applied, no throw.

bdbch (approved):

alexvcasillas:
@coderabbitai `emitContentError: true` isn't needed here. That option only gates the `insertContentAt` runtime path (`packages/core/src/commands/insertContentAt.ts`); it is not read by the `Editor` constructor's `createDoc()`, which emits `contentError` for invalid initial content based on `enableContentCheck` alone. This test exercises the initial-content path and already confirms the handler runs (it disables collaboration and recovers content). The existing `onContentError.spec.ts` follows the same convention (uses `enableContentCheck` only, no `emitContentError`). Leaving the fixture as-is.

bdbch:
@coderabbitai full review

consider @alexvcasillas's comments

PR #8109, opened 2026-07-22
feat: add the streaming-reveal extension to @tiptap/ai-toolkit

ozdemircibaris:
## Fixes

N/A (new feature)

## Changes and Review

- Adds the `AiInsertReveal` extension (`@tiptap/ai-toolkit/streaming-reveal`): fades in text as the AI streams it into a collaborative document.
- View-only decorations from Yjs insert deltas, so it adds no marks and does not mutate the document. Requires the Collaboration extension.
- Verify with the streaming demo (`revealAiText`) or `streaming-reveal.spec.ts`.

## Checklist

- [ ] I have made sure to test my changes myself.

### Responsibility


packages/ai-toolkit/src/streaming-reveal.ts
```diff
@@ -0,0 +1,259 @@
...
+   * How long (ms) each run keeps its reveal decoration before it is dropped.
+   * Keep it at or above your CSS animation duration so the fade can finish.
+   */
+  durationMs: number
+}
+
+/**
+ * Upper bound (chars) on a single revealed run. A normal streamed token is a
+ * few chars; this only guards against a mis-resolved relative position yielding
+ * an absurd range (e.g. spanning the whole document).
```
bdbch: what exactly is a revealed run? Sometimes I think these comments should describe things more general so someone who doesn't already have a large grasp of the context of the feature can understand the code.

packages/ai-toolkit/src/streaming-reveal.ts
```diff
@@ -0,0 +1,259 @@
...
+            const ystate = ySyncPluginKey.getState(state) as YSyncState | undefined
+            if (entries.length === 0 || !ystate?.binding) return null
+
+            const now = Date.now()
+            const decorations = entries
+              .map(entry => resolveRevealRange(ystate, entry, now, durationMs))
+              .filter((range): range is NonNullable<typeof range> => range !== null)
+              // Seed the CSS animation from the run's real age so a re-render
+              // (y-tiptap rebuilds the doc on every token) resumes the fade at the
+              // correct point instead of restarting it.
```
bdbch:
```suggestion
              // Offset the animation by the run's age so token updates resume the fade instead of restarting it.
```

bdbch (dismissed):
LGTM just a few nitpicks around comment lengths

packages/ai-toolkit/src/streaming-reveal.spec.ts
```diff
@@ -0,0 +1,187 @@
...
+      },
+    })
+  })
+}
+
+/**
+ * Applies a remote (non-local) insert into the first paragraph's text, mimicking
+ * an AI streaming into the shared document from another peer. Uses a second Y.Doc
+ * synced from `ydoc` so the resulting transaction has `local === false`.
+ */
```
bdbch:
```suggestion
/** Applies a remote insert via a synced second Y.Doc. */
```

packages/ai-toolkit/src/streaming-reveal.ts
```diff
@@ -0,0 +1,250 @@
...
+   */
+  durationMs: number
+}
+
+/**
+ * Upper bound (chars) on one revealed run: the text from a single streamed
+ * insert, usually a token of a few chars. This only guards against a
+ * mis-resolved relative position yielding an absurd range (e.g. spanning the
+ * whole document).
+ */
```
bdbch:
```suggestion
/** Limits each revealed streamed insert to guard against mis-resolved positions spanning the document. */
```

packages/ai-toolkit/src/streaming-reveal.ts
```diff
@@ -0,0 +1,250 @@
...
+const MAX_REVEAL_RANGE = 400
+
+/**
+ * One freshly-inserted text run, anchored by Yjs relative positions so it stays
+ * valid across the whole-document rebuild that y-tiptap applies on every remote
+ * sync (a plain ProseMirror position would be collapsed by that rebuild).
+ */
```
bdbch:
```suggestion
/** Anchors each inserted run with Yjs relative positions so it survives y-tiptap document rebuilds. */
```

packages/ai-toolkit/src/streaming-reveal.ts
```diff
@@ -0,0 +1,250 @@
...
+  }
+}
+
+/**
+ * Fades in text inserted by remote Yjs transactions using view-only inline
+ * decorations anchored by relative positions. It never mutates the document,
+ * so it stays inert to accept/reject and to persistence, and local edits are
+ * ignored so a user's own typing does not fade. Requires the Collaboration
+ * extension and CSS on the configured `className` (default `ai-insert-reveal`).
+ */
```
bdbch, line 133:
```suggestion
/**
 * Fades in remote Yjs inserts using view-only decorations, ignoring local edits and document persistence.
 * Requires Collaboration and CSS for `className` (default: `ai-insert-reveal`).
 */
```

bdbch (changes requested):
I'd shorten the comments.

packages/ai-toolkit/src/streaming-reveal.ts
ozdemircibaris: This part is outdated, but interestingly it doesn’t show as outdated

packages/ai-toolkit/package.json
```diff
@@ -39,6 +39,14 @@
...
       "require": "./dist/index.cjs"
+    },
+    "./streaming-reveal": {
+      "types": {
+        "import": "./dist/streaming-reveal.d.ts",
+        "require": "./dist/streaming-reveal.d.cts"
+      },
+      "import": "./dist/streaming-reveal.js",
+      "require": "./dist/streaming-reveal.cjs"
     }
```
bdbch, line 50: Just to be sure, this is meant to be an extra export yeah?

bdbch (approved):
LGTM - just a verification question

packages/ai-toolkit/src/streaming-reveal.ts
```diff
@@ -0,0 +1,233 @@
...
+            }, durationMs + 80)
+          }
+
+          const onChange = (
+            events: Array<Y.YEvent<Y.AbstractType<unknown>>>,
+            transaction: Y.Transaction,
+          ) => {
+            // Only remote edits, i.e. the AI. The local user's own typing is a
+            // local transaction and must not fade.
+            if (transaction.local) return
```
bdbch, line 225: Is this really only AI? What about collaborative changes that are not local transactions? I think this needs an `transaction.origin` filter as well to make sure that collab changes don't go through this reveal feature.

packages/ai-toolkit/src/streaming-reveal.ts
```diff
@@ -0,0 +1,233 @@
...
+    ystate.binding.mapping,
+  )
+  return from === null || to === null ? null : orderedSpan(from, to)
+}
+
+function orderedSpan(from: number, to: number): { from: number; to: number } | null {
+  const a = Math.min(from, to)
+  const b = Math.max(from, to)
+  return a >= b || b - a > MAX_REVEAL_RANGE ? null : { from: a, to: b }
+}
```
bdbch: I think we should clamp here to make sure those ranges never go outside the document ranges.

packages/ai-toolkit/src/streaming-reveal.ts
```diff
@@ -0,0 +1,233 @@
...
+  className: string
+  /**
+   * How long (ms) each run keeps its reveal decoration before it is dropped.
+   * Keep it at or above your CSS animation duration so the fade can finish.
+   */
+  durationMs: number
+}
+
+/** Limits each revealed streamed insert to guard against mis-resolved positions spanning the document. */
+const MAX_REVEAL_RANGE = 400
```
bdbch, line 22: As far as I can see content that exceeds the range just is silently dropping the run (correct me if I'm wrong).

should we just clamp the reveal to those characters and make sure the whole run still completes afterwards?

bdbch (dismissed):
some more comments

packages/ai-toolkit/package.json
ozdemircibaris: Yes, on purpose. It imports `@tiptap/y-tiptap` and yjs (optional peers), so it’s not in the root index. That way people who don’t use Collaboration don’t pull them in

packages/ai-toolkit/src/streaming-reveal.ts
ozdemircibaris: aah good point, it’s not only the `AI. local === false` just means the update came from `applyUpdate`, so other collaborators fade too. An origin filter won’t help though: Hocuspocus applies every remote update with the provider as origin, and origin isn’t sent over the wire. Only the clientID on the inserted items can tell AI and human edits apart. I’ll fix the comment for now. We can add clientID filtering if we want it strictly AI only

bdbch: Imho we would need to make sure that this is AI only - otherwise collaborative changes from other users would be run through the same reveal logic, right?

bdbch:
@ozdemircibaris I'd like to clarify the open questions before going forward.

packages/ai-toolkit/src/streaming-reveal.ts
```diff
@@ -16,18 +16,31 @@ export type AiInsertRevealOptions = {
    * Keep it at or above your CSS animation duration so the fade can finish.
    */
   durationMs: number
+  /**
+   * The collaboration provider, e.g. `HocuspocusProvider` or `TiptapCloudProvider`.
+   * Its awareness is what identifies the AI, so nothing is revealed without it.
+   */
+  provider: any
```
bdbch: Shouldn't we be able to look into the editor options / settings to find out if collaboration is enabled and find the provider from there? That way the user wouldn't need to pass it.

ozdemircibaris: actually collaboration has a provider option but the extension never reads it and nothing in the repo passes it. So in practice we can only find the provider when CollaborationCaret is used. Not everyone uses the caret, so we would still need the option. I would keep it explicit for now

packages/ai-toolkit/src/streaming-reveal.ts
ozdemircibaris: it's AI only right now <3

packages/ai-toolkit/src/streaming-reveal.ts
ozdemircibaris: You were right, the old code dropped the whole run. fixed thanks!

.changeset/new-pots-admire.md
```diff
@@ -0,0 +1,5 @@
+---
+"@tiptap/ai-toolkit": minor
+---
+
+Add the `AiInsertReveal` extension, exported from `@tiptap/ai-toolkit/streaming-reveal`, to fade in text as the AI streams it into a collaborative document. Pass your collaboration provider to it so only the AI's inserts are revealed, never another collaborator's.
```
bdbch: Keep this short. It's a user-facing changelog message, not an internal documentation.

packages/ai-toolkit/__tests__/streaming-reveal.spec.ts
```diff
@@ -0,0 +1,509 @@
...
+  const xmlText = paragraph.get(0) as Y.XmlText
+  xmlText.insert(index, text)
+  Y.applyUpdate(ydoc, Y.encodeStateAsUpdate(peer, Y.encodeStateVector(ydoc)))
+}
+
+/** Applies a whole new block from `peer`, the way the AI writes a heading. */
+function remoteInsertBlock(peer: Y.Doc, ydoc: Y.Doc, text: string): void {
+  Y.applyUpdate(peer, Y.encodeStateAsUpdate(ydoc, Y.encodeStateVector(peer)))
+  const fragment = peer.getXmlFragment('default')
+  const block = new Y.XmlElement('paragraph')
```
bdbch, line 108: The comment talks about a heading but this inserts a paragraph. Is this correct?

bdbch (dismissed):
LGTM.

I didn't verify all logic because honestly it's a lot, so I assume you tested, made sure that the tests are correct, etc.

I just have one notice about the changeset & a spec comment

packages/ai-toolkit/__tests__/streaming-reveal.spec.ts
ozdemircibaris: updated the comment, thanks

PR #8110, opened 2026-07-22
test(extension-link): guard non-link click selection with enableClickSelection (#7347)

alexvcasillas:
## Fixes

Refs #7347

## Changes and Review

#7347 (images hard to select when Link's `enableClickSelection` is on) was a regression introduced in v3.13.0 and already fixed in **v3.16.0** (#7359), which restored an unconditional `if (!link) return false` guard in the link click handler. This PR adds a regression test so that exact path can't silently break again.

- The existing "non-link elements" test only covered `openOnClick: false`, not `enableClickSelection: true`, which is the option named in #7347.
- The new test asserts the link plugin's `handleClick` reports **unhandled** (falsy) for a click on a non-link image, so the editor's own `NodeSelection` isn't blocked. Verified it fails if the guard is removed.
- Test-only change, so no changeset (no user-facing/behavior change).

AI usage disclosure: prepared with AI assistance (Claude Code); the author has reviewed and verified it.

## Checklist

- [ ] I have added a [changeset](https://github.com/changesets/changesets) if necessary. (n/a: test-only)

### Responsibility

### DIFF
--- a/packages/extension-link/__tests__/link.spec.ts
+++ b/packages/extension-link/__tests__/link.spec.ts
@@ -351,6 +351,51 @@ describe('extension-link', () => {
     getEditorEl()?.remove()
   })
 
+  // Regression guard for #7347: with enableClickSelection, clicking a non-link
+  // element (e.g. an image) must not be handled by the link plugin, otherwise
+  // ProseMirror's own NodeSelection is blocked.
+  it('should return false when clicking non-link elements with enableClickSelection', () => {
+    editor = new Editor({
+      element: createEditorEl(),
+      extensions: [
+        Document,
+        Text,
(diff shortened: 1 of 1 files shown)


alexvcasillas:
Thanks @coderabbitai, both addressed:
1. `Image` is now configured with `inline: true` so the image is valid content inside the paragraph (and genuinely at pos 1).
2. The click event's `target` is now set to the actual `img` element, so the test exercises a real non-link image click rather than a null target.

I also confirmed it's a true guard: it fails (`expected true to be falsy`) if the click handler's non-link early-return is removed, and passes on current code.

bdbch (approved):

Issue #7495, opened 2026-02-03
getMarkdown returns `&nbsp;` when editor is empty


Szorcu:
### Affected Packages

@tiptap/markdown

### Version(s)

3.18.0

### Bug Description

When the editor is visually empty, `editor.getMarkdown()` returns `&nbsp;` (or an equivalent non-breaking space), rather than an empty string. In contrast, `editor.getText()` returns an empty string for the same state.

On the other hand, `editor.getHtml()` returns `<p></p>`, which may indicate that it is an expected behaviour. 

### Browser Used

Firefox

### Expected Behavior

If the editor contains no user-entered content (i.e. `editor.isEmpty === true`), `editor.getMarkdown()` should return an empty string (`""`) rather than a placeholder entity like `&nbsp;`

### Additional Context (Optional)

You can observe this behaviour in Tiptap's Markdown docs: https://tiptap.dev/docs/editor/markdown.

Just delete all content from the Tiptap Editor, click `Extract Markdown` button and notice Markdown Input content.

### Dependency Updates

- [x] Yes, I've updated all my dependencies.

alexvcasillas:
Quick update: the case you reported (empty editor returning `&nbsp;`) is already fixed in #7497 (v3.16.0+), so on the latest version `editor.getMarkdown()` returns `""` for an empty editor.

While verifying, we found that the #7497 fix could go too far: it stripped non-breaking spaces from the output, so a document whose only content is a real non-breaking space was silently serialized to `""` (data loss). #8111 fixes that by deciding emptiness structurally (matching `editor.isEmpty`) instead of string-stripping, so real content is preserved while a truly empty editor still returns `""`.

Leaving this open until #8111 lands. Thanks for the report.

PR #8111, opened 2026-07-22
fix(markdown): keep non-breaking-space-only content instead of serializing to empty (#7495)

alexvcasillas:
## Fixes

Refs #7495

## Changes and Review

The literal report in #7495 (empty editor returning `&nbsp;`) was already fixed in #7497, but that fix used a string heuristic (`isEmptyOutput`) that stripped all `&nbsp;`/non-breaking-space characters from the rendered output. That caused silent data loss: a document whose only content is a real non-breaking space (`editor.isEmpty === false`) serialized to `""`.

- `MarkdownManager.serialize()` now decides emptiness structurally, before rendering, using `isNodeEmpty` (the same check behind `editor.isEmpty`), and returns the rendered output for anything non-empty. The string heuristic is removed.
- The manager is given the editor's actual `schema`, so the emptiness check matches `editor.isEmpty` exactly (no separately-rebuilt schema), and the shared cache also removes a per-call schema rebuild.
- Contract is now: `getMarkdown()` returns `""` if and only if `editor.isEmpty` is true.
- A bare content array passed to `serialize()` still renders normally (structural check falls through on non-doc input).

Behavior change worth a look: a document whose only content is an empty list item or empty blockquote now serializes to `""`, because `editor.isEmpty` is already `true` for those. Populated lists/blockquotes are unchanged. Three `paragraph.spec.ts` assertions were updated to reflect this.

## Checklist

### Responsibility

### DIFF
--- a/packages/markdown/__tests__/empty-content.spec.ts
+++ b/packages/markdown/__tests__/empty-content.spec.ts
@@ -85,4 +85,16 @@ describe('empty markdown content', () => {
     expect(textNode.text).toBe('bold')
     expect(textNode.marks?.[0]?.type).toBe('bold')
   })
+
+  it('should keep content when the document holds only a real non-breaking space (#7495)', () => {
+    // A lone non-breaking space is real content (editor.isEmpty is false), so
+    // getMarkdown() must not silently discard it by returning "".
+    editor = new Editor({
+      extensions: [Document, Paragraph, Text, Markdown],
+      content: '<p>&nbsp;</p>',
+    })
+
+    expect(editor.isEmpty).toBe(false)
+    expect(editor.getMarkdown()).toBe(' ')
+  })
 })
--- a/packages/markdown/__tests__/empty-content.spec.ts
+++ b/packages/markdown/__tests__/empty-content.spec.ts
@@ -86,7 +86,7 @@ describe('empty markdown content', () => {
     expect(textNode.marks?.[0]?.type).toBe('bold')
   })
 
-  it('should keep content when the document holds only a real non-breaking space (#7495)', () => {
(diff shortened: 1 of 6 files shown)


packages/markdown/__tests__/empty-content.spec.ts
```diff
@@ -85,4 +85,16 @@ describe('empty markdown content', () => {
     expect(textNode.text).toBe('bold')
     expect(textNode.marks?.[0]?.type).toBe('bold')
   })
+
+  it('should keep content when the document holds only a real non-breaking space (#7495)', () => {
```
bdbch, line 89:
```suggestion
  it('should keep content when the document holds only a real non-breaking space', () => {
```

packages/markdown/__tests__/paragraph.spec.ts
```diff
@@ -135,8 +139,10 @@ describe('Paragraph Markdown Rendering', () => {
         ],
       }
 
+      // Updated for #7495 (see comment above): whole-document structural emptiness
```
bdbch:
```suggestion
      // whole-document structural emptiness
```

packages/markdown/__tests__/paragraph.spec.ts
```diff
@@ -150,8 +156,10 @@ describe('Paragraph Markdown Rendering', () => {
...
```
bdbch, line 159:
```suggestion
      // whole-document structural emptiness
```

bdbch (changes requested):

bdbch (approved):

bdbch:
@alexvcasillas there are some conflicts you'd need to resolve here.

PR #8113, opened 2026-07-22
fix(react): update useEditorState when editable changes (#7498)

### LINKED ISSUE #7498: useEditorState stale when isEditable is updated
### Affected Packages

react

### Version(s)

^3.19.0

### Bug Description

Changing `editable`—even when using a dependency array to trigger reactive updates—does not seem to update `editor.isEditable` in the `useEditorState` `selector` function (though the function _is_ invoked).

### Browser Used

Chrome

### Code Example URL

### Expected Behavior

`ctx.editor.isEditable` to be current inside `useEditorState`

_No response_

### Dependency Updates

alexvcasillas:
## Fixes

Fixes #7498

## Changes and Review

`useEditorState` did not re-run its selector when the editor's editable state changed via `editor.setEditable(...)`, so `ctx.editor.isEditable` (and anything else affected by editability) stayed stale in React.

Root cause: `setEditable()` emits only the `update` event (it doesn't dispatch a transaction), but `useEditorState`'s internal subscription listened only to `transaction`.

- `useEditorState` now also subscribes to `update`, so editable changes propagate.
- A document-changing transaction emits both `transaction` and `update` with the same transaction object, so notifications are deduped by transaction identity to keep exactly one update per change (verified by a test that fails with a double count). `setEditable`'s `update` carries a distinct transaction and still gets through; selection-only transactions are unaffected.
- Scope is `@tiptap/react` only (patch).

## Checklist

### Responsibility

### DIFF
--- a/packages/react/src/useEditorState.ts
+++ b/packages/react/src/useEditorState.ts
@@ -95,9 +95,7 @@ class EditorStateManager<TEditor extends Editor | null = Editor | null> {
        * This is to support things like `editor.can().toggleBold()` in components that `useEditor`.
        * This could be more efficient, but it's a good trade-off for now.
        */
-      // A document change emits both `transaction` and `update` carrying the same
-      // transaction; dedupe by identity so consumers are notified once. `setEditable()`
-      // emits only `update` (with a distinct transaction), so it still gets through.
+      // Changes can emit both events, so notify only once
       let lastTransaction: unknown
       const fn = (props?: { transaction?: unknown }) => {
         if (props?.transaction !== undefined && props.transaction === lastTransaction) {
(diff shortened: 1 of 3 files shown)


alexvcasillas:
Good call @coderabbitai, added it. New test dispatches a selection-only transaction (which emits no `update`, only `transaction`) and asserts the counter still increments by exactly 1, locking in that the dedupe doesn't drop selection-change notifications.

packages/react/src/useEditorState.ts
```diff
@@ -95,16 +95,27 @@ class EditorStateManager<TEditor extends Editor | null = Editor | null> {
        * This is to support things like `editor.can().toggleBold()` in components that `useEditor`.
        * This could be more efficient, but it's a good trade-off for now.
        */
-      const fn = () => {
+      // A document change emits both `transaction` and `update` carrying the same
+      // transaction; dedupe by identity so consumers are notified once. `setEditable()`
+      // emits only `update` (with a distinct transaction), so it still gets through.
```
bdbch:
```suggestion
      // Changes can emit both events, so notify only once
```

bdbch (dismissed):

bdbch (approved):

PR #8127, opened 2026-07-27
fix(react): fixes caret placement after split blocks in React NodeViews.

### LINKED ISSUE #8117: React node views: caret does not move into new block on Enter since 3.28.0 (portal notification batching)
### Affected Packages

@tiptap/react

### Tiptap Version

3.28.0

### Browser Used

Chrome

## Description

  When the `paragraph` node (or any block node) is rendered through `ReactNodeViewRenderer`, pressing <kbd>Enter</kbd> splits the node correctly in  the ProseMirror document, but the DOM selection does not follow into the newly created block. The caret visually stays at the end of the original  paragraph and continued typing is inserted there instead of the new paragraph — e.g. `Hello` + <kbd>Enter</kbd> + typing `Second` produces a single  paragraph `HelloSecond`. This makes editors with React-node-view block nodes unusable for normal multi-paragraph text entry.

  Native (non-node-view) nodes are unaffected. Marks are unaffected.

  ## Steps to reproduce

  1. Replace StarterKit's paragraph with a React node view:

      ```tsx
      import Paragraph from "@tiptap/extension-paragraph";
      import { EditorContent, NodeViewContent, NodeViewWrapper, ReactNodeViewRenderer, useEditor } from "@tiptap/react";

      const ParagraphView = () => (
          <NodeViewWrapper as="p">
              <NodeViewContent />
          </NodeViewWrapper>
      );

      const NodeViewParagraph = Paragraph.extend({
          addNodeView() {
              return ReactNodeViewRenderer(ParagraphView);
          },
      });

      function App() {
          const editor = useEditor({
              extensions: [StarterKit.configure({ paragraph: false }), NodeViewParagraph],
              content: "<p>Hello</p>",
          });
          return <EditorContent editor={editor} />;
      }
      ```

  2. Click at the end of "Hello".
  3. Press <kbd>Enter</kbd>.
  4. Type anything.

### Expected Behavior

**Expected:** a new empty paragraph is created, the caret moves into it, and the typed text lands in the new paragraph.

  **Actual:** the document splits, but the caret stays in the old paragraph; the typed text is appended to "Hello" (the document collapses back to a  single paragraph containing `HelloSecond`).

  Immediately after step 3, `window.getSelection()` anchors on the node view's content element (`[data-node-view-content-react]`) at offset 0 with an  empty client rect, instead of a renderable caret position inside the new block.

  ## Root cause analysis

  The regression is the *"Batch portal store notifications during simultaneous mounts"* change in 3.28.0. Diff of the portal store between 3.27.4 and  3.28.0:

  ```js
  // 3.27.4 — subscribers notified synchronously in setRenderer/removeRenderer
  subscribers.forEach((subscriber) => subscriber());

  // 3.28.0 — deferred to a microtask
  const notifySubscribers = () => {
      if (isNotificationQueued || !subscribers.size) return;
      isNotificationQueued = true;
      queueMicrotask(() => {
          isNotificationQueued = false;
          subscribers.forEach((subscriber) => subscriber());
      });
  };
  ```

  Because the `Portals` component (`useSyncExternalStore`) is now notified in a microtask, a newly created `ReactNodeView`'s component — and  therefore its `contentDOM` — does not exist until **after** ProseMirror's current view update cycle has finished. When <kbd>Enter</kbd> splits a  block, ProseMirror tries to map the selection into the new node view's `contentDOM` during that same update cycle, finds none, and the DOM  selection is left behind in the old block. The browser then keeps inserting text at the stale caret position.

  3.27.4 flushed the portal synchronously inside `setRenderer`, so `contentDOM` was available in time.

  ## Environment

  - `@tiptap/core` / `@tiptap/react` / `@tiptap/pm` / `@tiptap/starter-kit` / `@tiptap/extension-paragraph` 3.28.0
  - react / react-dom 19.2.6
  - Vite dev server, Chromium (also reported on Chrome stable)
  - Reproduced both with a plain paragraph node view (above) and in a production setup: https://github.com/vivid-planet/comet/issues/6037

  ## Workaround

  Pin `@tiptap/*` to 3.27.4.

  One note: I verified the caret desync end-to-end in the Comet setup (TextBlockStyleParagraph); the minimal snippet in the issue is the standard  pattern distilled from it but wasn't run standalone — if you want, I can spin it up quickly (e.g. in the cms-admin Storybook stories, which already  have a plain node-view story) before you submit, so the repro is guaranteed against the exact snippet.

_No response_

_No response_

bdbch:
## Fixes

- Fixes #8117 

## Changes and Review

Changes introduced by 3.28.0 changed portal-render batching to avoid errors being thrown by React when too many changes were rendered. This caused problems with the caret not being placed into newly attached React contentDom elements. This PR fixes it by temporarily attaching the contentDom to the editor dom before placing it into the React component.

## Checklist

### Responsibility

### DIFF
--- a/demos/src/GuideNodeViews/ReactComponentContent/React/index.jsx
+++ b/demos/src/GuideNodeViews/ReactComponentContent/React/index.jsx
@@ -13,9 +13,7 @@ export default () => {
     <p>
       This is still the text editor you’re used to, but enriched with node views.
     </p>
-    <react-component>
-      <p>This is editable. You can create a new component by pressing Mod+Enter.</p>
-    </react-component>
+    <react-component>This is editable. You can create a new component by pressing Mod+Enter.</react-component>
     <p>
       Did you see that? That’s a React component. We are really living in the future.
     </p>
--- a/demos/src/GuideNodeViews/VueComponentContent/Vue/index.vue
+++ b/demos/src/GuideNodeViews/VueComponentContent/Vue/index.vue
@@ -27,7 +27,7 @@ export default {
           This is still the text editor you’re used to, but enriched with node views.
         </p>
         <vue-component>
-          <p>This is editable.</p>
+          This is editable.
         </vue-component>
         <p>
           Did you see that? That’s a Vue component. We are really living in the future.
--- a/.changeset/2026-07-25-fix-react-nodeview-caret.md
+++ b/.changeset/2026-07-25-fix-react-nodeview-caret.md
@@ -1,0 +1,5 @@
+---
+'@tiptap/react': patch
+---
+
+Fix caret placement after splitting a block rendered with a React NodeView.
--- a/demos/src/GuideNodeViews/ReactComponentContent/index.spec.ts
+++ b/demos/src/GuideNodeViews/ReactComponentContent/index.spec.ts
@@ -1,0 +1,15 @@
+import { expect, test } from '@playwright/test'
+
+const demoPath = '/src/GuideNodeViews/ReactComponentContent/React/'
+
+test.describe('GuideNodeViews/ReactComponentContent/React', () => {
+  test.beforeEach(async ({ page }) => {
+    await page.goto(demoPath)
+  })
+
+  test('renders the initial NodeView content', async ({ page }) => {
+    await expect(page.locator('.tiptap .react-component .content')).toHaveText(
+      'This is editable. You can create a new component by pressing Mod+Enter.',
+    )
+  })
+})
--- a/demos/src/GuideNodeViews/VueComponentContent/index.spec.ts
+++ b/demos/src/GuideNodeViews/VueComponentContent/index.spec.ts
@@ -1,0 +1,13 @@
+import { expect, test } from '@playwright/test'
+
+const demoPath = '/src/GuideNodeViews/VueComponentContent/Vue/'
+
+test.describe('GuideNodeViews/VueComponentContent/Vue', () => {
+  test.beforeEach(async ({ page }) => {
+    await page.goto(demoPath)
+  })
+
+  test('renders the initial NodeView content', async ({ page }) => {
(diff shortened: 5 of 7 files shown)


alexvcasillas (dismissed):

demos/src/GuideNodeViews/ReactComponentContent/React/index.jsx
```diff
@@ -13,9 +13,7 @@ export default () => {
     <p>
       This is still the text editor you’re used to, but enriched with node views.
     </p>
-    <react-component>
-      <p>This is editable. You can create a new component by pressing Mod+Enter.</p>
-    </react-component>
+    <react-component>This is editable. You can create a new component by pressing Mod+Enter.</react-component>
```
arnaugomez, line 16: How is this change related to the fix?

demos/src/GuideNodeViews/VueComponentContent/Vue/index.vue
```diff
@@ -27,7 +27,7 @@ export default {
           This is still the text editor you’re used to, but enriched with node views.
         </p>
         <vue-component>
-          <p>This is editable.</p>
+          This is editable.
         </vue-component>
```
arnaugomez, line 31: How is this change related to the fix?

arnaugomez (approved):

demos/src/GuideNodeViews/ReactComponentContent/React/index.jsx
bdbch: The paragraph around the content was invalid. The schema for the demo only allowed inline-text. I'm not sure when it was introduced exactly but that caused empty content inside the node view.

demos/src/GuideNodeViews/VueComponentContent/Vue/index.vue
bdbch: see above

PR #8129, opened 2026-07-28
feat(extension-find-and-replace): support capture groups in replacements

Aslam97:
## Changes and review

This PR improves the standalone, headless Find and Replace extension:

- Regex replacements can reuse complete matches and capture groups.
- Whole word now works in Regex mode while respecting Match case.
- Literal-mode replacement behavior remains unchanged.
- Update documentation https://github.com/ueberdosis/tiptap-docs/pull/931

## Checklist

- [x] I have added regression tests.
- [x] I have documented the behavioral change and migration path.
- [x] I have manually tested my changes.

### Responsibility

## AI assistance

This contribution was developed with assistance from OpenAI Codex. I reviewed all generated changes and manually verified the resulting behavior.

## UI Component requirement
Screenshot 2026-07-30 at 20 02 09


Aslam97:
@coderabbitai review

packages/extension-find-and-replace/src/search/find-safe-matcher-matches.ts
```diff
@@ -0,0 +1,48 @@
+import type { Matcher, RE2JS } from 're2js'
+
+function advanceStringIndex(text: string, index: number): number {
+  return index + 1 + Number((text.codePointAt(index) ?? 0) > 0xffff)
```
bdbch: what exactly is happening here? this could use a comment, otherwise a human reviewer doesn't understand immediately what this line is doing exactly. The `0xffff` is the biggest problem.

bdbch (dismissed):

packages/extension-find-and-replace/src/search/find-safe-matcher-matches.ts
Aslam97: I didn’t understand it at first either, but it is used to skip characters that use more than one UTF-16 unit, like emojis, and move to the correct next index. I am updating the code to add comment so that it more understandable and not forget

bdbch:
@coderabbitai please do another review of the changes since your last review

bdbch (approved):

PR #8137, opened 2026-07-28
fix(extension-table): snapshot column widths on first resize

### LINKED ISSUE #456: Disable Paste
Is there an easy way to disable being able to paste things into the editor?

### LINKED ISSUE #7560: [Bug] Table columns resize leads to broken behavior of rightmost columns.
### Affected Packages

extension-table

### Version(s)

3.19.0

### Bug Description

When a new table is inserted, the resize handler works unexpectedly. This causes a state where the rightmost columns do not behave correctly (cannot be adjusted or calculated properly) until the first column of the table receives a fixed width.

Steps to reproduce:
- Insert a new table (e.g., 3 columns).
- Attempt to interact with or adjust the borders of the second or third columns.
- Observe that the behavior is broken or unresponsive.
- Adjust the width of the first column.
- Notice that after the first column is set, the subsequent columns start behaving correctly.

https://github.com/user-attachments/assets/ea142178-a624-47d4-ab1e-2e48d923f75a

### Browser Used

Chrome

### Code Example URL

_No response_

### Expected Behavior

All columns should be interactive and maintain correct internal state immediately after table creation, without depending on the state of the first column.

The issue stems from prosemirror-tables logic where the absence of a defined colwidth for the leading column prevents correct delta calculations for the rest of the row. The question is can it be somehow hacked? Maybe it is possible to handle the count of columns/rows?

### Dependency Updates

sverg84:
## Fixes

Closes #7560 

## Changes and Review

- Newly inserted resizable tables had no stored `colwidth`, so the first resize of a non-leading column let the browser redistribute untouched columns.
- Snapshot rendered column widths on resize `mousedown` (before ProseMirror updates the dragged column), including colspan cells, and keep that snapshot out of undo history.
- How to verify: insert a 3-column table → resize column 2 or 3 first → other columns keep their widths; undo undoes only the drag. Unit coverage is in `columnResizeSnapshot.spec.ts`.

## Checklist

### Responsibility

### DIFF
--- a/packages/extension-table/src/table/columnResizingSnapshot.ts
+++ b/packages/extension-table/src/table/columnResizingSnapshot.ts
@@ -1,0 +1,98 @@
+import { Plugin } from '@tiptap/pm/state'
+import { columnResizingPluginKey, TableMap } from '@tiptap/pm/tables'
+import type { EditorView } from '@tiptap/pm/view'
+
+function findTableElement(view: EditorView, tableStart: number) {
+  let dom: Node | null = view.domAtPos(tableStart).node
+
+  while (dom && dom.nodeName !== 'TABLE') {
+    dom = dom.parentNode
+  }
+
+  return dom instanceof HTMLTableElement ? dom : null
+}
+
+function readRenderedColumnWidths(view: EditorView, tableStart: number, columnCount: number) {
+  const tableElement = findTableElement(view, tableStart)
+
+  if (!tableElement) {
+    return null
+  }
+
+  const columns = tableElement.querySelectorAll<HTMLTableColElement>(':scope > colgroup > col')
+
+  if (columns.length !== columnCount) {
+    return null
+  }
+
+  const widths = Array.from(columns, col => Math.round(col.getBoundingClientRect().width))
+
+  if (widths.some(width => width <= 0)) {
+    return null
+  }
+
+  return widths
+}
+
+function snapshotColumnWidths(view: EditorView, cellPos: number) {
+  const $cell = view.state.doc.resolve(cellPos)
+  const table = $cell.node(-1)
+  const tableStart = $cell.start(-1)
+  const map = TableMap.get(table)
+  const cellPositions = [...new Set(map.map)]
+
+  const needsSnapshot = cellPositions.some(pos => {
+    const cell = table.nodeAt(pos)
+    const colwidth = cell?.attrs.colwidth as number[] | null
+
+    return !colwidth || colwidth.length !== cell?.attrs.colspan || colwidth.some(width => !width)
+  })
+
+  if (!needsSnapshot) {
+    return
+  }
+
+  const widths = readRenderedColumnWidths(view, tableStart, map.width)
+
+  if (!widths) {
+    return
+  }
+
+  const tr = view.state.tr
+
+  cellPositions.forEach(pos => {
+    const cell = table.nodeAt(pos)
+
+    if (!cell) {
+      return
+    }
+
+    const column = map.colCount(pos)
+    const colwidth = widths.slice(column, column + cell.attrs.colspan)
+
+    tr.setNodeMarkup(tableStart + pos, null, { ...cell.attrs, colwidth })
+  })
+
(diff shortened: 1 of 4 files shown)


bdbch (approved):

packages/extension-table/src/table/columnResizingSnapshot.ts
```diff
@@ -0,0 +1,104 @@
...
+
+  cellPositions.forEach(pos => {
+    const cell = table.nodeAt(pos)
+
+    if (!cell || !cellNeedsColwidthSnapshot(cell)) {
+      return
+    }
+
+    const column = map.colCount(pos)
+    const colwidth = widths.slice(column, column + cell.attrs.colspan)
```
bdbch, line 77: Small bug here: if a cell spans several columns and only some of its widths are set, the good ones get thrown away. A cell with `colwidth: [100, null]` gets the whole array replaced by measured values, so the `100` is lost.

Can you keep what's already there and only fill the gaps?

```suggestion
    const existing = cell.attrs.colwidth as number[] | null
    
    const colwidth = Array.from(
      { length: cell.attrs.colspan },
      (_, i) => existing?.[i] || widths[column + i],
    )
```

A test for that case would be good too, e.g. a `colspan="2"` cell where only the first width is set.

packages/extension-table/src/table/columnResizingSnapshot.ts
```diff
@@ -0,0 +1,104 @@
...
+    return null
+  }
+
+  const columns = tableElement.querySelectorAll<HTMLTableColElement>(':scope > colgroup > col')
+
+  if (columns.length !== columnCount) {
+    return null
+  }
+
+  const widths = Array.from(columns, col => Math.round(col.getBoundingClientRect().width))
```
bdbch, line 28: One thing I'd like to check before this goes in: we measure the `<col>` elements here, but prosemirror-tables measures the cells. Those two can differ by a few pixels because of cell borders and border-spacing. Since we now set a width on every column at once, those small differences add up and the table can visibly jump the moment you press the mouse down.

Did you try this in Firefox and Safari? If you see a jump, measuring the cells of the first row instead would match what the resize plugin itself uses.

packages/extension-table/src/table/columnResizingSnapshot.ts
```diff
@@ -0,0 +1,104 @@
...
+    tr.setMeta('addToHistory', false)
+    view.dispatch(tr)
+  }
+}
+
+export function columnResizingSnapshot() {
+  return new Plugin({
+    props: {
+      handleDOMEvents: {
+        mousedown(view) {
```
bdbch, line 92: Small thing: the event isn't used, so this also runs on right click. Right-clicking near a column border would write widths into the document without any resize happening. Can we bail out for anything that isn't a left click?

```suggestion
        mousedown(view, event) {
          if (event.button !== 0) {
            return false
          }
```

packages/extension-table/src/table/table.ts
```diff
@@ -567,6 +568,7 @@ export const Table = Node.create<TableOptions>({
     return [
       ...(isResizable
         ? [
+            columnResizingSnapshot(),
```
bdbch, line 571: Could you add a short comment that this has to stay before `columnResizing()`? The fix only works because our handler runs first, and that's easy to break by accident later.

bdbch (changes requested):
Nice fix, the approach makes sense and I like that the snapshot stays out of the undo history.

Two things about behaviour rather than code:

After the first resize every column has a width, so the table gets a fixed pixel width and stops shrinking with its container. That's the trade-off we want, but the changeset doesn't say it. Can you extend it a little?

Undo restores the old column width, but the snapshot itself stays in the document. So dragging and undoing still leaves the document changed. In a collaborative document that means the first person to resize writes their own pixel widths for everyone else too. I think that's fine, I just want it to be a conscious decision.

Nit: the source file is `columnResizingSnapshot.ts` but the test is `columnResizeSnapshot.spec.ts`, would rename the test.

Rest is inline.

PR #8139, opened 2026-07-29
fix(markdown): collapse soft line breaks into spaces

### LINKED ISSUE #8136: Markdown: CommonMark soft line breaks become literal \n in text nodes and render as visible line breaks
### Affected Packages

markdown

### Version(s)

`@tiptap/markdown` 3.29.0, and verified unchanged in 3.29.2 (latest) — `marked` 15.0.12

### Description of the Bug

A hard-wrapped paragraph parses into text nodes containing literal `\n`, which the editor renders as visible line breaks. Every CommonMark renderer displays the same source as one flowing paragraph.

```md
You can try CommonMark here. This dingus is powered by
[commonmark.js](https://github.com/commonmark/commonmark.js), the
JavaScript reference implementation.
```

`MarkdownManager.parse` returns one paragraph — the block structure is right:

```json
{"type":"paragraph","content":[
  {"type":"text","text":"You can try CommonMark here. This dingus is powered by\n"},
  {"type":"text","text":"commonmark.js","marks":[{"type":"link","attrs":{"href":"https://github.com/commonmark/commonmark.js"}}]},
  {"type":"text","text":", the\nJavaScript reference implementation."}
]}
```

The problem is the `\n` inside the text nodes. `@tiptap/core` injects `.ProseMirror { white-space: break-spaces }`, so the paragraph renders on three lines instead of one. Pasting a normally-wrapped README into an editor shreds it.

### Why this isn't a marked bug

CommonMark explicitly permits what marked does:

> A softbreak may be rendered in HTML either as a line ending or as a space. The result will be the same in browser display.

marked picks the line ending — correct for an HTML target, where whitespace collapses. It leans on that guarantee quite hard: `marked.parse('foo \n baz')` returns `"<p>foo \n baz</p>"`, flanking spaces intact.

The equivalence only holds for HTML. `MarkdownManager` doesn't render to HTML — it lowers marked's token text straight into a ProseMirror text node, a model with no whitespace collapsing, and Tiptap then displays it under `break-spaces`, the opposite of the whitespace model the spec's latitude assumes.

marked also offers no signal to consume: it has no `softbreak` token, and its `br` token only fires on the hard-break spellings (two or more trailing spaces, or a backslash) unless `breaks: true`. So the newline stays inside the text token and the collapse has to be re-derived downstream.

### Suggested fix

Collapse it in the two places where a markdown `text` token becomes a text node — the `token.type === 'text'` branch of `parseInlineTokens`, and the block-level `case 'text'` (both currently `text: decodeHtmlEntities(token.text || '')`):

```js
text.replace(/[ \t]*\n[ \t]*/g, ' ')
```

Taking the flanking whitespace with the break matches the spec's "spaces at the end of the line and beginning of the next line are removed".

Scope is contained: code blocks and code spans don't route through either branch (marked already converts newlines inside code spans to spaces per spec), so pre-formatted content keeps its newlines. Explicit hard breaks are unaffected because marked emits those as `br` → `hardBreak`.

Trade-off worth naming: a markdown → doc → markdown round trip loses the author's original wrap columns. The output stays semantically identical, just re-wrapped.

### Code Example

```ts

const manager = new MarkdownManager({ extensions: [StarterKit] })

manager.parse('foo\nbar')
// → paragraph > text "foo\nbar"      renders on two lines
// expected: paragraph > text "foo bar"

manager.parse('```js\nconst a = 1;\nconst b = 2;\n```')
// → codeBlock keeps its newlines — correct, and must stay that way
```

### Expected Behavior

A soft-wrapped paragraph parses to a single-line paragraph, matching how CommonMark renders it.

`markedOptions: { breaks: true }` is not a workaround — it makes the break explicit (`hardBreak`), which is the opposite of what CommonMark specifies for a soft break. A custom `marked` instance with `walkTokens` doesn't work either: `MarkdownManager` calls `parseLexer.lex(markdown)` directly, and `walkTokens` only runs inside `marked.parse()`.

### Additional Context

Related but distinct: #7107 (hard breaks from double-space), #7170 (line feeds on plain-text paste), #7269 (multiple newlines not preserved).

### Dependency Updates

lazerg:
## Fixes

Fixes #8136

## Changes and Review

marked keeps a soft-wrapped paragraph's newlines inside its `text` tokens, leaning on the CommonMark rule that a soft break may render as either a space or a line ending in HTML, where whitespace collapses. `MarkdownManager` lowers that text straight into a ProseMirror node, which has no whitespace collapsing, so under `white-space: break-spaces` a wrapped paragraph renders across several lines instead of one. This collapses each soft break and its flanking spaces to a single space in the two branches that turn a `text` token into a text node; code blocks and code spans don't route through those branches so they keep their newlines, and explicit hard breaks stay `hardBreak`. One existing round-trip fixture is updated to the collapsed form, reflecting the trade-off that a markdown to doc to markdown round trip loses the author's original wrap columns.

To verify, parse `foo\nbar` and check it yields a single paragraph `foo bar`, or run `pnpm vitest run packages/markdown`.

## Checklist

### Responsibility

### DIFF
--- a/.changeset/2026-07-29-collapse-soft-line-breaks.md
+++ b/.changeset/2026-07-29-collapse-soft-line-breaks.md
@@ -1,0 +1,5 @@
+---
+'@tiptap/markdown': patch
+---
+
+Collapse CommonMark soft line breaks into spaces when parsing, so a soft-wrapped paragraph no longer renders with visible line breaks.
--- a/packages/markdown/__tests__/conversion-files/soft-break-marks.ts
+++ b/packages/markdown/__tests__/conversion-files/soft-break-marks.ts
@@ -1,10 +1,8 @@
 export const name = 'Soft Break with Marks'
 
-export const expectedInput = `**Speaker:**
-John Doe.
+export const expectedInput = `**Speaker:** John Doe.
 
-**Speaker:**
-**John Doe**`.trim()
+**Speaker:** **John Doe**`.trim()
 
 export const expectedOutput = {
   type: 'doc',
@@ -13,14 +11,14 @@ export const expectedOutput = {
       type: 'paragraph',
       content: [
         { type: 'text', marks: [{ type: 'bold' }], text: 'Speaker:' },
-        { type: 'text', text: '\nJohn Doe.' },
+        { type: 'text', text: ' John Doe.' },
       ],
     },
     {
       type: 'paragraph',
       content: [
         { type: 'text', marks: [{ type: 'bold' }], text: 'Speaker:' },
-        { type: 'text', text: '\n' },
+        { type: 'text', text: ' ' },
         { type: 'text', marks: [{ type: 'bold' }], text: 'John Doe' },
       ],
     },
--- a/packages/markdown/__tests__/soft-line-break.spec.ts
+++ b/packages/markdown/__tests__/soft-line-break.spec.ts
@@ -1,0 +1,63 @@
+import { CodeBlock } from '@tiptap/extension-code-block'
+import { Document } from '@tiptap/extension-document'
+import { Link } from '@tiptap/extension-link'
+import { Paragraph } from '@tiptap/extension-paragraph'
+import { Text } from '@tiptap/extension-text'
+import { MarkdownManager } from '@tiptap/markdown'
+import { beforeEach, describe, expect, it } from 'vitest'
+
+describe('Soft line breaks', () => {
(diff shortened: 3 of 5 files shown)


lazerg:
@bdbch could you take a look at this markdown fix whenever you get a chance?

lazerg:
Closing as no one reviewed and this has been open more than a month with no activity. Happy to reopen if there's interest — thanks for your time.

PR #8140, opened 2026-07-29
fix(markdown): escape leading block syntax when serializing paragraphs

### LINKED ISSUE #8134: Markdown serialization does not preserve escaped blocks after round-trip
### Affected Packages

markdown

### Tiptap Version

3.29.1

### Browser Used

Not browser-specific

### Description

Escaped block syntax is parsed correctly as paragraph text, but loses the required escape on serialization.

### Reproduction

In the full Markdown demo, enter:
```
\# not a heading

1\. not a list
```

1. Click **Parse Markdown**
    -> Both lines correctly appear as paragraph text.
2. Click **Extract Markdown**
    -> Extracted Markdown is
```
# not a heading

1. not a list
```
3. Click **Parse Markdown** again
    -> The first line becomes a heading and the second line becomes a list item.

The same round-trip issue also affects other escaped blocks, including `\-`, `\+`, `\---`, Setext Headings and GFM table pipes.

### Expected Behavior

The parse, serialize, parse round-trip should preserve the document structure.

_No response_

_No response_

lazerg:
## Fixes

Fixes #8134

## Changes and Review

The serializer already escaped inline markdown, but not block markers at the start of a line. So a paragraph holding `# not a heading` was written back without its backslash and re-parsed as a real heading (same for `-`, `+`, ordered-list numbers and `---`). Serializing now escapes a leading block marker on the first line of a top-level paragraph, so escaped block syntax stays a paragraph across a parse → serialize → parse round-trip. To check, parse `\# not a heading`, extract the markdown, and parse it again, or run the new round-trip tests in `conversion.spec.ts`.

## Checklist

### Responsibility

### DIFF
--- a/packages/markdown/__tests__/conversion.spec.ts
+++ b/packages/markdown/__tests__/conversion.spec.ts
@@ -779,6 +779,57 @@ describe('Markdown Conversion Tests', () => {
       })
     })
 
+    describe('serializing: leading block syntax → backslash-escaped markdown', () => {
+      const paragraph = (text: string) => ({
+        type: 'doc',
+        content: [{ type: 'paragraph', content: [{ type: 'text', text }] }],
+      })
+
+      it('should escape a leading heading marker', () => {
+        expect(markdownManager.serialize(paragraph('# not a heading'))).toBe('\\# not a heading')
+      })
+
+      it('should escape a leading multi-level heading marker', () => {
+        expect(markdownManager.serialize(paragraph('## not a heading'))).toBe('\\## not a heading')
+      })
+
+      it('should escape a leading bullet-list marker', () => {
+        expect(markdownManager.serialize(paragraph('- not a list'))).toBe('\\- not a list')
+        expect(markdownManager.serialize(paragraph('+ not a list'))).toBe('\\+ not a list')
+      })
+
+      it('should escape a leading ordered-list marker', () => {
+        expect(markdownManager.serialize(paragraph('1. not a list'))).toBe('1\\. not a list')
+        expect(markdownManager.serialize(paragraph('1) not a list'))).toBe('1\\) not a list')
+      })
+
+      it('should escape a thematic-break line', () => {
+        expect(markdownManager.serialize(paragraph('---'))).toBe('\\---')
+      })
+
+      it('should not escape a block marker in the middle of a line', () => {
+        expect(markdownManager.serialize(paragraph('a # b'))).toBe('a # b')
+      })
+
+      it('should not escape a heading marker without a following space', () => {
+        expect(markdownManager.serialize(paragraph('#tag'))).toBe('#tag')
+      })
+    })
+
+    describe('round-trip: leading block syntax stays a paragraph', () => {
+      const roundTripType = (input: string) => {
+        const md = markdownManager.serialize(markdownManager.parse(input))
+        return markdownManager.parse(md).content[0].type
+      }
+
+      it('should keep escaped block syntax as a paragraph after round-trip', () => {
+        expect(roundTripType('\\# not a heading')).toBe('paragraph')
+        expect(roundTripType('1\\. not a list')).toBe('paragraph')
+        expect(roundTripType('\\- not a list')).toBe('paragraph')
+        expect(roundTripType('\\+ not a list')).toBe('paragraph')
+      })
+    })
+
     describe('edge cases: code blocks and nested syntax', () => {
       it('should NOT backslash-escape inside inline code marks (parsing)', () => {
         const json = markdownManager.parse('`\\*not italic\\*`')
--- a/packages/markdown/__tests__/conversion.spec.ts
+++ b/packages/markdown/__tests__/conversion.spec.ts
@@ -908,6 +908,41 @@ describe('Markdown Conversion Tests', () => {
         expect(markdownManager.serialize(input)).toBe('\\# not a heading')
       })
 
+      it('should escape a block marker after a newline inside a later text node', () => {
+        const input = {
+          type: 'doc',
+          content: [
+            {
+              type: 'paragraph',
+              content: [
+                { type: 'text', text: 'text' },
+                { type: 'text', text: 'bold', marks: [{ type: 'bold' }] },
+                { type: 'text', text: '\n# not a heading' },
+              ],
+            },
+          ],
+        }
+
+        expect(markdownManager.serialize(input)).toBe('text**bold**\n\\# not a heading')
+      })
+
+      it('should not escape the first line of a text node that starts mid-line', () => {
+        const input = {
+          type: 'doc',
+          content: [
+            {
+              type: 'paragraph',
+              content: [
+                { type: 'text', text: 'a', marks: [{ type: 'bold' }] },
+                { type: 'text', text: '# b' },
+              ],
+            },
+          ],
+        }
+
+        expect(markdownManager.serialize(input)).toBe('**a**# b')
+      })
+
       it('should escape a leading block marker inside a list item', () => {
         expect(markdownManager.serialize(listItem('# not a heading'))).toBe('- \\# not a heading')
         expect(markdownManager.serialize(listItem('- not a list'))).toBe('- \\- not a list')
@@ -948,6 +983,25 @@ describe('Markdown Conversion Tests', () => {
           paragraph('| a | b |\n| - | - |'),
         )
         expect(roundTrip(paragraph('a | b\n--- | ---'))).toEqual(paragraph('a | b\n--- | ---'))
+      })
+
+      it('should keep a block marker after a mark and a newline as text after round-trip', () => {
+        const input = {
+          type: 'doc',
+          content: [
+            {
+              type: 'paragraph',
+              content: [
+                { type: 'text', text: 'text' },
+                { type: 'text', text: 'bold', marks: [{ type: 'bold' }] },
+                { type: 'text', text: '\n# not a heading' },
+              ],
+            },
+          ],
+        }
+
+        expect(roundTrip(input).content).toHaveLength(1)
+        expect(roundTrip(input).content[0].type).toBe('paragraph')
       })
 
       it('should keep block syntax in a list item as text after round-trip', () => {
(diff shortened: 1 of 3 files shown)


packages/markdown/src/MarkdownManager.ts
```diff
@@ -1119,6 +1125,22 @@ export class MarkdownManager {
...
+   * passes `isLineStart` for the first inline child of a top-level paragraph.
+   */
+  private escapeBlockSyntax(text: string): string {
+    return text
+      .replace(/^(#{1,6})(\s|$)/, '\\$1$2')
+      .replace(/^([-+])(\s|$)/, '\\$1$2')
+      .replace(/^(\d{1,9})([.)])(\s|$)/, '$1\\$2$3')
+      .replace(/^(-{3,})(\s*)$/, '\\$1$2')
+  }
+
```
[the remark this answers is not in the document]
lazerg: Good catch, fixed. escapeBlockSyntax now runs per line instead of just at the start of the string, and I added handling for setext underlines, table rows, and up to three spaces of indent. I also extended isLineStart to cover the text right after a hard break, since it's the same bug.

packages/markdown/src/MarkdownManager.ts
```diff
@@ -1119,6 +1125,30 @@ export class MarkdownManager {
...
+   * text node keeps its meaning when the serialized markdown is parsed again.
+   *
+   * Relevant at the start of every line, not just the start of the whole
+   * string: a text node can contain literal `\n` characters (e.g. pasted or
+   * API-constructed content), and each of those lines is its own block-start
+   * context once serialized. Up to three leading spaces are still considered
+   * "line start" per CommonMark's block-indentation rule. The caller passes
+   * `isLineStart` for the first inline child of a top-level paragraph, and
+   * for the text run right after a hard break.
+   */
```
bdbch, line 1141: please shorten this. way to verbose and could easily be explained in 1-2 lines

packages/markdown/src/MarkdownManager.ts
```diff
@@ -1233,7 +1263,12 @@ export class MarkdownManager {
       }
 
       if (node.type === 'text') {
-        let textContent = this.encodeTextForMarkdown(node.text || '', node, parentNode)
+        const isLineStart =
+          level === 0 &&
```
bdbch: Is it correct that we only check level 0 - what happens with nested paragraphs that start with something like `# hello`?

packages/markdown/src/MarkdownManager.ts
```diff
@@ -1233,7 +1263,12 @@ export class MarkdownManager {
...
+          parentNode?.type === 'paragraph' &&
+          !node.marks?.length &&
+          (i === 0 || nodes[i - 1]?.type === 'hardBreak')
```
bdbch: This assumes that the start of a line is at index 0 right? This sounds unstable, what if the first node inside at this position doesn't have text content? Then the position would start at `i === 1`.

packages/markdown/src/MarkdownManager.ts
```diff
@@ -1119,6 +1125,30 @@ export class MarkdownManager {
...
+   * for the text run right after a hard break.
+   */
+  private escapeBlockSyntax(text: string): string {
+    const escapeLine = (line: string): string =>
+      line
+        .replace(/^( {0,3})(#{1,6})(\s|$)/, '$1\\$2$3')
+        .replace(/^( {0,3})([-+])(\s|$)/, '$1\\$2$3')
+        .replace(/^( {0,3})(\d{1,9})([.)])(\s|$)/, '$1$2\\$3$4')
+        .replace(/^( {0,3})(-{2,}|={1,})(\s*)$/, '$1\\$2$3')
+        .replace(/^( {0,3})(\|)/, '$1\\$2')
```
bdbch: small nitpick: on the very first line there's nothing above it to underline, so `---` / `===` can't be a heading there anyway. Same with a leading `|` if no delimiter row follows. So we're adding backslashes that don't do anything.

packages/markdown/__tests__/conversion.spec.ts
```diff
@@ -779,6 +779,94 @@ describe('Markdown Conversion Tests', () => {
...
+      })
+
+      it('should escape a leading ordered-list marker', () => {
+        expect(markdownManager.serialize(paragraph('1. not a list'))).toBe('1\\. not a list')
+        expect(markdownManager.serialize(paragraph('1) not a list'))).toBe('1\\) not a list')
+      })
+
+      it('should escape a thematic-break line', () => {
+        expect(markdownManager.serialize(paragraph('---'))).toBe('\\---')
+      })
```
bdbch, line 835: the tests look good for me, but two of the cases from #8134 only get half-tested: the thematic break (`---`) and the table pipe just check the serialized string, they never parse it back. Would be good to have those in the round-trip block too, since that's where the actual bug showed up.

also nothing seems covers lists or blockquotes anywhere.

packages/markdown/__tests__/conversion.spec.ts
```diff
@@ -779,6 +779,94 @@ describe('Markdown Conversion Tests', () => {
...
+        expect(markdownManager.serialize(paragraph('Title\n==='))).toBe('Title\n\\===')
+      })
+
+      it('should escape a two-dash underline (invalid as a thematic break, valid as setext)', () => {
+        expect(markdownManager.serialize(paragraph('Title\n--'))).toBe('Title\n\\--')
+      })
+
+      it('should escape a leading table-row pipe on a later line', () => {
+        expect(markdownManager.serialize(paragraph('a\n| b | c |'))).toBe('a\n\\| b | c |')
+      })
```
bdbch: see above

bdbch (changes requested):
Thanks, I have a few things that can be improved or fixed.

packages/markdown/src/MarkdownManager.ts
lazerg: Shortened it to two lines.

packages/markdown/src/MarkdownManager.ts
lazerg: It was broken there. A paragraph inside a list item serialized `# hello` unescaped, and parsing that back gave a heading inside the item. Blockquotes were fine by accident, their handler does not indent so level stayed 0.

Dropped the level check, every paragraph escapes its line starts now. One exception I kept: `1.` inside a list item, marked keeps that as text (`- 123.` in the bullet-list-numeric-punctuation fixture), so escaping it there would only add a backslash for nothing. Tests for list item and blockquote, serialize and round-trip.

packages/markdown/src/MarkdownManager.ts
lazerg: Yes, and it broke exactly like that. With an empty text node first, `# hello` at i === 1 serialized unescaped and parsed back as a heading.

The flag now comes from what the paragraph has rendered so far: line start means nothing written yet, or the output ends with a newline. That also drops the special case for hard breaks, since a hard break already ends with one. Test added.

packages/markdown/src/MarkdownManager.ts
lazerg: Agreed on the underline. `---` and `===` on the first line are left alone now, only a thematic break gets escaped there.

The pipe was worse than noise though. marked ignores a backslash on the header row and still builds the table, so `\| a | b |` + `| - | - |` parsed as a table and the paragraph was gone. The escape has to break the delimiter row instead, so I escape its first dash: `| a | b |` + `| \- | - |`. Same for the form without leading pipes (`a | b` + `--- | ---`). A row with no delimiter row under it is left untouched.

packages/markdown/__tests__/conversion.spec.ts
lazerg: Both are in the round-trip block now, comparing the whole doc instead of just the type. The thematic break already survived. The table did not: only the header row was escaped, marked still made a table of it and the paragraph was dropped. Fixed by escaping the delimiter row, see the other thread.

Lists and blockquotes are covered now too, serialize and round-trip, including the list item case that was broken.

packages/markdown/__tests__/conversion.spec.ts
lazerg: Same, covered by the round-trip tests.

packages/markdown/src/MarkdownManager.ts
```diff
@@ -1233,7 +1272,20 @@ export class MarkdownManager {
...
+          parentNode?.type === 'paragraph' &&
+          !node.marks?.length &&
+          (renderedSoFar === '' || renderedSoFar.endsWith('\n'))
+        let textContent = this.encodeTextForMarkdown(
+          node.text || '',
+          node,
+          parentNode,
+          isLineStart,
+          level > 0,
+        )
```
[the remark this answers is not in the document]
lazerg: Confirmed, a paragraph of `text` + bold + `\n# heading` serialized to a real heading. The line state is now a `startsMidLine` flag passed into `escapeBlockSyntax`, so every line inside a node gets checked and only the first one is skipped when earlier nodes already wrote to that line. Setext and delimiter-row handling stays as it was. Added a serialize test and a round-trip test for the mark + newline + marker case, both failed before the change. Markdown suite 285 pass.

packages/markdown/__tests__/conversion.spec.ts
```diff
@@ -779,6 +811,211 @@ describe('Markdown Conversion Tests', () => {
...
+                { type: 'text', text: 'bold', marks: [{ type: 'bold' }] },
+                { type: 'text', text: '\n# not a heading' },
+              ],
+            },
+          ],
+        }
+
+        expect(roundTrip(input).content).toHaveLength(1)
+        expect(roundTrip(input).content[0].type).toBe('paragraph')
+      })
```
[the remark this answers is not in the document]
lazerg: Good catch, this only checked node type. Now asserts full round-trip equality against input. Commit b2d8c97d5.

lazerg:
Hey @bdbch, I've addressed the review feedback and CodeRabbit approved after the fixes. Could you take another look when you have a moment?

PR #8150, opened 2026-07-31
fix(extension-table): keep cursor inside table when deleting last row or column

### LINKED ISSUE #8149: deleteRow() moves cursor outside table when deleting last row
## Description
When deleting the last row in a table using `editor.chain().focus().deleteRow().run()`, the cursor jumps outside the table (to the paragraph below) instead of staying within the table in the row above.

## Expected Behavior
The cursor should remain within the table, moving to the row above the deleted row (similar to how spreadsheet applications handle this).

## Current Behavior
When deleting the last row of a table that has content below it (e.g., a paragraph), the cursor moves outside the table to that content below.

## Reproduction Steps
1. Create a table with multiple rows (e.g., 3 rows)
2. Add a paragraph below the table
3. Place cursor in the last row of the table
4. Call `editor.chain().focus().deleteRow().run()`
5. Observe: cursor jumps to the paragraph below instead of staying in the table

## Environment
- TipTap version: 3.29.2
- Extensions used: Table, TableRow, TableCell, TableHeader
- Table configuration: `Table.configure({ resizable: false })`

## Minimal Reproduction
```javascript
// Create table with content below
editor.commands.insertTable({ rows: 3, cols: 3, withHeaderRow: true })
editor.commands.insertContentAt(editor.state.doc.content.size, '<p>Text below table</p>')

// Move to last row and delete
editor.commands.setTextSelection(/* last row position */)
editor.chain().focus().deleteRow().run()

// Result: cursor is now in the paragraph below, not in the table
console.log(editor.isActive('table')) // false (expected: true)
```

## Notes
- This only happens when there is content below the table
- If the table is at the end of the document (no content below), the behavior is correct
- The same issue may affect `deleteColumn()` but hasn't been tested

https://github.com/user-attachments/assets/a68cd0a3-a532-45e0-807b-477f5d2ae2a4

lazerg:
## Fixes

Fixes #8149

## Changes and Review

When you delete the last row or column of a table that has content below it, the cursor jumps out of the table and into that content. prosemirror-tables doesn't set a new selection after deleting, so the old cursor position maps to just outside the table. `deleteRow` and `deleteColumn` now move the selection back into the table when the cursor would otherwise land outside it.

To verify: put the cursor in the last row of a table that has a paragraph below it and run `deleteRow()`. The cursor stays inside the table.

## Checklist

### Responsibility

### DIFF
--- a/packages/extension-table/src/table/table.ts
+++ b/packages/extension-table/src/table/table.ts
@@ -4,12 +4,13 @@ import {
   type JSONContent,
   type MarkdownToken,
   callOrReturn,
+  findParentNodeClosestToPos,
   getExtensionField,
   mergeAttributes,
   Node,
 } from '@tiptap/core'
 import type { DOMOutputSpec, Node as ProseMirrorNode } from '@tiptap/pm/model'
-import { TextSelection } from '@tiptap/pm/state'
+import { type EditorState, TextSelection, type Transaction } from '@tiptap/pm/state'
 import {
   addColumnAfter,
   addColumnBefore,
@@ -256,6 +257,40 @@ declare module '@tiptap/core' {
   }
 }
 
+// prosemirror-tables does not move the selection after a deletion, so removing the
+// last row or column can push the cursor out of the table into the content below.
+const keepCursorInTable =
+  (command: (state: EditorState, dispatch?: (tr: Transaction) => void) => boolean) =>
+  (state: EditorState, dispatch?: (tr: Transaction) => void): boolean => {
+    if (!dispatch) {
+      return command(state)
+    }
+
+    const table = findParentNodeClosestToPos(
+      state.selection.$from,
+      node => node.type.name === 'table',
+    )
+
+    return command(state, tr => {
+      const stillInTable = findParentNodeClosestToPos(
+        tr.selection.$from,
+        node => node.type.name === 'table',
+      )
+
+      if (table && !stillInTable) {
+        const tableNode = tr.doc.nodeAt(table.pos)
+
+        if (tableNode) {
+          const endOfTable = table.pos + tableNode.nodeSize - 1
+
+          tr.setSelection(TextSelection.near(tr.doc.resolve(endOfTable), -1))
+        }
+      }
+
+      dispatch(tr)
+    })
+  }
+
 /**
  * This extension allows you to create tables.
  * @see https://www.tiptap.dev/api/nodes/table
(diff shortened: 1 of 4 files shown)


packages/extension-table/src/table/table.ts
```diff
@@ -256,6 +257,40 @@ declare module '@tiptap/core' {
...
+        node => node.type.name === 'table',
+      )
+
+      if (table && !stillInTable) {
+        const tableNode = tr.doc.nodeAt(table.pos)
+
+        if (tableNode) {
+          const endOfTable = table.pos + tableNode.nodeSize - 1
+
+          tr.setSelection(TextSelection.near(tr.doc.resolve(endOfTable), -1))
```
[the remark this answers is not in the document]
lazerg: This finding is correct. When a second table follows the first table, the delete command can move the cursor into the second table. `stillInTable` then matches a table, but not the correct one. The code skips the reposition step in this case.

The fix maps the original table position through the transaction. It then compares this position to the new cursor position. The fallback runs only after the cursor leaves the edited table.

I added row and column regression tests. Both use a table below the edited one.

packages/extension-table/src/table/table.ts
```diff
@@ -445,7 +492,7 @@ export const Table = Node.create<TableOptions>({
       deleteColumn:
         () =>
         ({ state, dispatch }) => {
-          return deleteColumn(state, dispatch)
+          return keepCursorInTable(deleteColumn)(state, dispatch)
```
bdbch: I think it would be clearer if `keepCursorInTable` was run inside `deleteColumn` / `deleteRow`. That way `keepCursorInTable` would also not have to be a wrapping command but a simple util that places the selection on the mapped transaction.

packages/extension-table/src/table/table.ts
```diff
@@ -460,7 +507,7 @@ export const Table = Node.create<TableOptions>({
       deleteRow:
         () =>
         ({ state, dispatch }) => {
-          return deleteRow(state, dispatch)
+          return keepCursorInTable(deleteRow)(state, dispatch)
```
bdbch: See above

bdbch (changes requested):
I don't like the wrapping command API you went for, could you make the cursor position handling into a utility that can be called from inside the `deleteRow` / `deleteColumn`?

packages/extension-table/src/table/table.ts
lazerg: Moved it into its own util (`keepCursorInTable.ts`) that `deleteRow` and `deleteColumn` call directly now.

bdbch (approved):

PR #8151, opened 2026-08-01
chore: migrate over to vite+ for tooling

bdbch:
## Fixes

- N/A

## Changes and Review

This PR migrates the whole project from our individual setup (vite + vitest + tsup + oxc) to one solution using [Vite+](https://viteplus.dev/). A few pros and contras on why I would do this:

**Pros**

- Less things to maintain individually
- Simpler onboarding and less commands to remember
- Less configuration fragmentation (just configure the global vite config once)
- Workspace orchestration without something like turborepo or nx
- Still uses the existing tools we used prior
- No need to upgrade individual packages or make sure things work together
- VP comes with a great caching layer and most tooling is build in Rust making it very fast
- Caching works great (after a few runs my build times went from 60+ seconds down to 1 second and a 97% cache hit rate)

**Cons**

- We rely on one solution from the Vite team so if something is not to our likings we need to move out of it again (even though it's not so bad because we can just not use individual vp features)
- One upgrade in the future has a larger blast radius (we now have to check everything => testing, building, formatting, linting, etc.)
- One more layer on top of the existing tools we used which may complicate debugging when something goes wrong

The review for this PR will be tough so the best to make sure this is valid:

1. Check the PR out
2. run the tests => should still run successsful using vitest
3. run the build => should still build all packages to dist folders correctly
4. check that dist paths are consistent to the data in package.json
5. run the formatter + linter => should use the previous formatting, no linting issues

One thing that this PR will also show is if the CI actions run correctly

**extra note**: the idea for this PR came from me trying out a new AI model and I thought a larger migration would be fun to test it on - keep that in mind in review.

## Checklist

- [ ] I have added tests if possible.

### Responsibility


Aslam97 (dismissed):

ozdemircibaris (dismissed):
LGTM

pack.config.mts
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * Base pack config for a single-entry package, matching the old tsup config.
+ */
+export const basePackConfig = (): PackUserConfig => ({
+  tsconfig: '../../tsconfig.build.json',
+  outDir: 'dist',
+  dts: true,
+  clean: true,
+  sourcemap: true,
```
alexvcasillas, line 25: `tsup` picked up `target: "es2019"` from `tsconfig.json` and `vp pack` doesn't, so the published output changed.

On `main` the whole `dist` has 8 `?.` and 0 `??`. On this branch it's 1128 `?.`, 510 `??` and two files with ES2022 `static {}` blocks (`packages/core/dist/index.js:4999`). Anyone bundling us against an ES2019 target gets a syntax error now.

I tried `target: 'es2019'` in `packages/html/vite.config.ts` and it puts the output back exactly where it was. Should we add it here and to the 11 packages that write their own `pack` block (`core`, `pm`, `react`, `vue-2`, `vue-3`, `html`, `extensions`, `extension-list`, `extension-table`, `extension-text-style`, `static-renderer`)?

alexvcasillas, line 25: Separate thing on the same line: `sourcemap: true` makes `vp pack` append `//# sourceMappingURL=index.d.ts.map` to the declaration files, but it never emits the maps. 190 `.d.ts` and `.d.cts` files point at something that isn't there, `main` has zero.

Should we emit them or drop the dts sourcemap? 🤔

package.json
```diff
@@ -79,58 +76,17 @@
     "tinyglobby": "0.2.10",
     "ts-loader": "9.3.1",
     "tsconfig-paths": "^4.2.0",
-    "tsup": "^8.5.1",
-    "turbo": "2.9.14",
     "typescript": "^5.7.3",
-    "vitest": "^4.1.2",
+    "vite": "catalog:",
+    "vite-plus": "catalog:",
```
alexvcasillas, line 81: 31 spec files still import from `vitest` while the other 159 moved to `vite-plus/test`, and `vitest` isn't a dependency anymore. `packages/core/__tests__/decorations.spec.ts` is one of them.

Tests pass because `vp` aliases it at runtime, but `node_modules/vitest` doesn't exist, `tsc` gives `TS2307` on all 31 and the IDE marks them unresolved. Should we finish the sweep?

patches/@changesets__assemble-release-plan@6.0.10.patch
```diff
@@ -0,0 +1,15 @@
+diff --git a/dist/changesets-assemble-release-plan.cjs.js b/dist/changesets-assemble-release-plan.cjs.js
```
alexvcasillas, line 1: Now that only `6.0.10` is referenced, `patches/@changesets__assemble-release-plan.patch` and `patches/@changesets__assemble-release-plan@6.0.9.patch` are dead. Can we drop them while we're here? :)

pnpm-workspace.yaml
```diff
@@ -26,6 +29,13 @@ overrides:
   'terser-webpack-plugin>serialize-javascript': '7.0.5'
```
alexvcasillas: Now that we're on rolldown, `rollup: '4.60.1'` and this `terser-webpack-plugin>serialize-javascript` pin don't match anything in the lockfile anymore. Should we clean those up while we're in here? 🤔

.github/instructions/tiptap.instructions.md
```diff
@@ -121,22 +125,22 @@ Two layers:
...
+vp run test:e2e:firefox       # Playwright headless (Firefox)
+vp run test:e2e:all           # both browsers — every test twice
+vp run test:e2e:open          # UI mode (Chromium tests)
+vp run test:e2e:open:firefox  # UI mode (Firefox tests)
+vp run test:e2e:open:all      # UI mode, switch between browsers in the project picker
+vp run test:e2e:report        # open the HTML report from the last run
 ```
 
-Playwright auto-starts the demo dev server (`pnpm -C demos run start:e2e` on port 4080) via `playwright.config.ts` — no separate terminal needed. Shared helpers live in `demos/test/helpers.ts`: `getEditor`, `setEditorContent`, `clickButton`. Use `demos/src/Commands/Cut/index.spec.ts` as a canonical template when adding new specs.
+Playwright auto-starts the demo dev server (`vp run -C demos start:e2e` on port 4080) via `playwright.config.ts` — no separate terminal needed. Shared helpers live in `demos/test/helpers.ts`: `getEditor`, `setEditorContent`, `clickButton`. Use `demos/src/Commands/Cut/index.spec.ts` as a canonical template when adding new specs.
```
alexvcasillas: Small one: this now says `vp run -C demos start:e2e`, but `playwright.config.ts` still runs `pnpm -C demos run start:e2e`. Which one do we want? 🤔

.github/workflows/publish.yml
```diff
@@ -50,13 +50,15 @@ jobs:
         with:
           node-version: ${{ env.NODE_VERSION }}
           pnpm-version: ${{ env.PNPM_VERSION }}
```
alexvcasillas, line 52: `cache: 'true'` got removed from both `setup-env` calls in this file, while `build.yml` gained it. Was that deliberate? Publishes will do a cold install every time.

demos/package.json
```diff
@@ -60,12 +60,11 @@
     "react": "^19.0.0",
     "react-dom": "^19.0.0",
     "sass-embedded": "^1.99.0",
-    "svelte": "^4.2.19",
+    "svelte": "^5.56.8",
```
alexvcasillas, line 63: The Svelte 5 bump is covered by the Default demo e2e (it iterates React, Vue and Svelte) so I'm not worried about it.

The demo itself is still Svelte 4 syntax though (`on:click`, `editor = editor` for reactivity), so it runs in legacy mode. Should we port it to runes separately? Definitely not for this PR 👍

alexvcasillas (commented):
Nice work, the pack configs map over cleanly and the caching holds up.

One blocker: the published JS went from ES2019 to ES2022. `vp pack` doesn't pick up `target` from `tsconfig.json` like `tsup` did. Details on `pack.config.mts`.

Rest is small and inline. Two that don't fit on a line:

- `vp run --filter` warns instead of failing when nothing matches, so `build:demos` in publish would exit 0 silently if that filter ever breaks. Should we add `--fail-if-no-match`? 🤔
- Heads-up for when this lands, not a change request: `vp config` writes `core.hooksPath` into `.git/config`, so switching back to `main` or `maintenance/v3` kills commit hooks until you re-install.

.github/instructions/tiptap.instructions.md
bdbch: fixed, uses pnpm now

pnpm-workspace.yaml
bdbch: dropped! :)

patches/@changesets__assemble-release-plan@6.0.10.patch
bdbch: I rebased with main - as it was changed there already

.github/actions/setup-env/action.yml
```diff
@@ -24,12 +24,13 @@ runs:
...
-      with:
-        node-version: ${{ inputs.node-version }}
-        cache: ${{ inputs.cache == 'true' && 'pnpm' || '' }}
-
     - name: Configure pnpm store
       if: inputs.store-dir != ''
       shell: bash
       run: pnpm config set store-dir ${{ github.workspace }}/${{ inputs.store-dir }}
+
+    - uses: voidzero-dev/setup-vp@v1
```
fengmk2: A pinned version is required, which is the way we recommend.  https://viteplus.dev/guide/ci#setup-vp-versioning


```suggestion
    - uses: voidzero-dev/setup-vp@v1.17.0
```

bdbch: Thank you for taking a look into it! 🤝

PR #8163, opened 2026-08-05
fix(core): insert content when prosemirror-model is loaded more than once

### LINKED ISSUE #6762: Can not convert <image> to a Fragment (looks like multiple versions of prosemirror-model were loaded)
### Affected Packages

"@tiptap/core": "^3.0.9",

### Version(s)

3.0.9

### Bug Description

Can not convert <image> to a Fragment (looks like multiple versions of prosemirror-model were loaded) 
at src/components/tiptap-node/image-upload-node/image-upload-node.tsx (473:27) @ <unknown>
 .insertContentAt(pos, imageNodes)
using nextjs 15

### Browser Used

Chrome

### Code Example URL

_No response_

### Expected Behavior

can upload image 

_No response_

### Dependency Updates

ozdemircibaris:
Closes #6762

`Fragment.from` recognises a fragment with `instanceof`, so loading prosemirror-model twice makes it reject valid fragments. One version is enough — an ESM/CJS split does it, which is what the Next.js reports hit.

Fixed:

- `insertContentAt` passed a fragment to `tr.replaceWith`. That is the crash in the issue.
- Its text branch did the same, so a fragment holding only text inserted nothing.
- `createNodeFromContent` matched nodes with `instanceof`, so a node from the other copy silently became empty content.
- The editor warns once when it finds a duplicate.

Not fixed: wrapping and splitting nodes (`toggleBlockquote`, `toggleBulletList`, `sinkListItem`, `splitListItem`). Those build the fragment inside prosemirror-transform and prosemirror-schema-list, so core cannot reach them.

This does not remove the duplicate. Users still need to dedupe their lock file or alias prosemirror-model in their bundler.

Tests load the CJS build of prosemirror-model as the second copy.


packages/core/src/helpers/warnOnDuplicatedProseMirrorModel.ts
```diff
@@ -0,0 +1,29 @@
...
+let hasWarned = false
+
+/**
+ * Warns once when prosemirror-model is loaded more than once. ProseMirror then
+ * rejects valid fragments with a `Can not convert <…> to a Fragment` error, and
+ * wrapping or splitting nodes stops working.
+ * @param doc A document created from the editor schema
+ * @example warnOnDuplicatedProseMirrorModel(editor.state.doc)
+ */
+export function warnOnDuplicatedProseMirrorModel(doc: ProseMirrorNode): void {
```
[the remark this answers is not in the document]
ozdemircibaris: This part is outdated, but interestingly it doesn’t show as outdated

packages/core/__tests__/duplicatedProseMirrorModel.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+    it('keeps a node built by the other copy instead of dropping it', () => {
+      editor = createEditor()
+      const paragraph = editor.schema.nodeFromJSON({
+        type: 'paragraph',
+        content: [{ type: 'text', text: 'kept' }],
+      })
+      const fragment = otherCopy.Fragment.fromArray([paragraph])
+
+      expect(createNodeFromContent(fragment, editor.schema).toJSON()).toEqual(fragment.toJSON())
+    })
```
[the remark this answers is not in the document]
ozdemircibaris: This part is outdated, but interestingly it doesn’t show as outdated

.changeset/stupid-singers-wonder.md
```diff
@@ -0,0 +1,5 @@
+---
+"@tiptap/core": patch
+---
+
+Fixed `insertContent`, `insertContentAt` and `setContent` failing when prosemirror-model is loaded more than once. They threw `Can not convert <…> to a Fragment` or silently inserted nothing.
```
bdbch:
```suggestion
Fixed `insertContent`, `insertContentAt` and `setContent` failing when prosemirror-model is loaded more than once.
```

packages/core/__tests__/duplicatedProseMirrorModel.spec.ts
```diff
@@ -0,0 +1,139 @@
...
+
+import { Editor, createNodeFromContent } from '@tiptap/core'
+import Document from '@tiptap/extension-document'
+import Image from '@tiptap/extension-image'
+import Paragraph from '@tiptap/extension-paragraph'
+import Text from '@tiptap/extension-text'
+import { afterEach, describe, expect, it, vi } from 'vitest'
+
+// Put @tiptap/core on the CJS build and the rest of ProseMirror on the ESM one,
+// the split a bundler produces when part of the graph is required, part imported.
```
bdbch:
```suggestion
// Put @tiptap/core on the CJS build and the rest of ProseMirror on the ESM one
```

The first line already explains what is done perfectly. I found the second line just more confusing.

packages/core/__tests__/warnOnDuplicatedProseMirrorModel.spec.ts
```diff
@@ -0,0 +1,31 @@
+import { Editor } from '@tiptap/core'
+import Document from '@tiptap/extension-document'
+import Paragraph from '@tiptap/extension-paragraph'
+import Text from '@tiptap/extension-text'
+import { afterEach, describe, expect, it, vi } from 'vitest'
+
+// No module mock here, so this is the ordinary single copy install.
```
bdbch:
```suggestion
```

imo no need to explain if nothing is used here

packages/core/src/commands/insertContentAt.ts
```diff
@@ -197,7 +187,8 @@ export const insertContentAt: RawCommands['insertContentAt'] =
           from = Math.max(0, from - 1)
         }
 
-        tr.replaceWith(from, to, newContent)
+        // Nodes cross a duplicated prosemirror-model, a fragment does not.
```
bdbch:
```suggestion
```

not really sure what this comment tells me here without any context and why, i'd drop it. In general I'd make sure comments are really just added when needed.

packages/core/src/commands/setContent.ts
```diff
@@ -60,7 +61,10 @@ export const setContent: RawCommands['setContent'] =
       })
 
       if (dispatch) {
-        tr.replaceWith(0, doc.content.size, document).setMeta('preventUpdate', !emitUpdate)
+        // Nodes cross a duplicated prosemirror-model, a fragment does not.
```
bdbch:
```suggestion
```

same here - a reader of `setContent.ts` really doesn't care about prosemirror-model, fragments or nodes. This information is just necessary in the context of the bug we're fixing and just pollutes the command file.

packages/core/src/helpers/warnOnDuplicatedProseMirrorModel.ts
```diff
@@ -0,0 +1,33 @@
+import type { Node as ProseMirrorNode } from '@tiptap/pm/model'
+import { Fragment } from '@tiptap/pm/model'
+import { Transform } from '@tiptap/pm/transform'
+
+let hasChecked = false
+
+/**
+ * Warns once when prosemirror-model is loaded twice, which breaks wrapping and splitting nodes.
```
bdbch:
```suggestion
 * Warns once when prosemirror-model is loaded twice
```

not necessary to mention what breaks - the comment should just describe what this is.

packages/core/src/helpers/warnOnDuplicatedProseMirrorModel.ts
```diff
@@ -0,0 +1,33 @@
...
+ * @param doc A document created from the editor schema
+ * @returns Nothing, the warning goes to the console
```
bdbch:
```suggestion
 * @returns void
```

packages/core/src/helpers/warnOnDuplicatedProseMirrorModel.ts
bdbch, line 1: I'm not so sure about this whole warning function. Right now it rather checks if fragments can be inserted, not necessarily that the prosemirror-model is loaded twice.

Sure - it will throw this warning if that is caused by two instances of the module being loaded - but if this fails because of another reason it would throw an invalid warning to the user as far as I see.

bdbch (changes requested):
Some thoughts and changes I'd do.

packages/core/src/helpers/warnOnDuplicatedProseMirrorModel.ts
ozdemircibaris: hmm yeah it was testing behaviour, not the actual condition. It now compares the fragment prosemirror-transform builds with its own copy of prosemirror-model against ours, so the warning fires only when those two classes differ, which is exactly the duplicate case. If the check itself fails for any other reason it stays quiet rather than guessing.

Happy to drop the helper from this PR if you’d rather keep it to the fix.

bdbch (approved):
Looks good to me, thanks

PR #8177, opened 2026-08-11
fix(extension-table): stop rendering default colspan and rowspan on cells

### LINKED ISSUE #8176: TableCell and TableHeader serialize default colspan="1" and rowspan="1" attributes
### Affected Packages

@tiptap/extension-table

### Tiptap Version

3.29.2

### Browser Used

Chrome

### What happened?

TableCell and TableHeader serialize colspan="1" and rowspan="1" for every table cell, even when no spanning is used.

For example, a normal table cell is serialized as:

<td colspan="1" rowspan="1"><p>Cell</p></td>

instead of:

<td><p>Cell</p></td>

The same happens for <th> elements.

Both attributes have a default value of 1 in TableCell and TableHeader. These defaults are then included in HTMLAttributes and rendered to the element.

This adds redundant markup to every cell and also causes existing table HTML to be changed when it is parsed and serialized again.

### Expected Behavior

Default colspan and rowspan values should not be serialized.

A normal cell should be:

<td><p>Cell</p></td>

and colspan or rowspan should only be output when their value differs from 1.

This is also how prosemirror-tables serializes these attributes.

_No response_

I originally encountered this through Umbraco CMS 17.

Umbraco 17.6.0 uses @tiptap/extension-table 3.22.3 and extends Tiptap's TableCell and TableHeader without overriding their HTML serialization, so the behavior originates in Tiptap.

I also checked the current Tiptap 3.29.2 source and the behavior is still present.

prosemirror-tables keeps colspan=1 and rowspan=1 as defaults in its document model, but its serializer explicitly omits them from the generated HTML when the value is 1.

lazerg:
## Fixes

Fixes #8176

## Changes and Review

Every `<td>` and `<th>` came out with `colspan="1" rowspan="1"`, because both attributes default to 1 in the schema and that default was rendered like any other value. Parsing existing table HTML and serializing it again therefore changed the markup. Both attributes now render only when they are greater than 1, which is what prosemirror-tables does.

To verify: load a table with plain cells into the editor and call `getHTML()`. Cells come back as `<td><p>Cell</p></td>`, and cells that really span still carry `colspan`/`rowspan`.

## Checklist

### Responsibility

### DIFF
--- a/.changeset/2026-08-11-table-default-spans.md
+++ b/.changeset/2026-08-11-table-default-spans.md
@@ -2,4 +2,4 @@
 '@tiptap/extension-table': patch
 ---
 
-Table cells and headers no longer render `colspan="1"` and `rowspan="1"`. Both attributes are only written to the HTML when a cell actually spans more than one column or row, which matches how prosemirror-tables serializes them.
+Table cells and headers only serialize `colspan` and `rowspan` when a cell spans more than one column or row.
--- a/.changeset/2026-08-11-table-default-spans.md
+++ b/.changeset/2026-08-11-table-default-spans.md
@@ -1,0 +1,5 @@
+---
+'@tiptap/extension-table': patch
+---
+
+Table cells and headers no longer render `colspan="1"` and `rowspan="1"`. Both attributes are only written to the HTML when a cell actually spans more than one column or row, which matches how prosemirror-tables serializes them.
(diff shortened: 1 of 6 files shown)


lazerg:
Closing as no one reviewed and this has been open more than a month with no activity. Happy to reopen if there's interest — thanks for your time.

Issue #5889, opened 2024-11-29
[Bug]: Invalid regular expression: invalid group specifier name (introduced in 2.10.3)


erwamartin:
### Affected Packages

extension-code

### Version(s)

2.10.3

### Bug Description

The TipTap code extension fails to load on Safari versions below 16 due to the following error:

Invalid regular expression: invalid group specifier name.

This issue prevents the file from being loaded and stems from an incompatible regex introduced in this commit:
ueberdosis/tiptap commit f79c05e.

The regex uses named capturing groups and lookbehind assertions, which are not supported in Safari versions earlier than 16.

### Browser Used

Safari

### Expected Behavior

The file should load correctly.

**Rewritten Regex**
_Original:_
```
export const inputRegex = /(?<!`)`([^`]+)`(?!`)/;
export const pasteRegex = /(?<!`)`([^`]+)`(?!`)/g;
```

_Rewritten Without Lookbehind:_
```
export const inputRegex = /(^|[^`])`([^`]+)`(?!`)/;
export const pasteRegex = /(^|[^`])`([^`]+)`(?!`)/g;
```

### Dependency Updates

- [X] Yes, I've updated all my dependencies.

wallisonfelipe:
Same here:
Invalid regular expression: invalid group specifier name

nperez0111:
Would happily accept a PR to address this

lino-levan:
This issue was fixed in #5916. I guess we're waiting on that to be merged.

nperez0111:
Yes, we queue releases. If you need it now, you cna use the PR's version: https://github.com/ueberdosis/tiptap/pull/5916#issuecomment-2519498744

lino-levan:
Any timeline on when the next version of tiptap will be released?

nperez0111:
No, we release when we feel ready. Please refer to my previous comment above
> If you need it now, you can use the PR's version

ccreusat:
Hi guys !

For your information, we have the same problem with Tiptap and the code extension in version `2.3.0` in Safari 15.6 and below (but that's ok I think). Works in Safari 16.5, 17.6, ... 👍🏼

Nantris:
Whoops I see I've duplicated this issue in #5948.

I've closed it as a duplicated, but I've suggested a more robust approach there that brings the best of both worlds.

Nantris:
[...]

PR #8180, opened 2026-08-11
fix(extension-table): make markdown pipe-escape regex parseable on WebKit < Safari 16.4

### LINKED ISSUE #8179: [Bug]: @tiptap/extension-table >= 3.27.2 fails to parse on iOS/Safari < 16.4, breaking the whole bundle chunk
### Affected Packages

extension-table

### Tiptap Version

3.27.2 – 3.30.0 (current latest)

### Browser Used

Safari (WebKit before Safari 16.4, including WKWebView on older iOS/macOS)

### What happened?

`escapeTableCellPipes` in `packages/extension-table/src/table/utilities/markdown.ts` escapes pipes inside code spans with:

```js
spanContent.replace(/(?<!\\)\|/g, '\\|')
```

WebKit only supports regex lookbehind from Safari 16.4. Because this is a regex **literal**, older WebKit fails at **parse time** with:

```
SyntaxError: Invalid regular expression: invalid group specifier name
```

and the failure takes down the **entire chunk** containing the extension, not just the code path that uses the regex. Any application bundling `@tiptap/extension-table` >= 3.27.2 (where the markdown serializer landed) serves a bundle that is completely broken on those browsers.

We hit this in production: after a routine bump from 3.27.1, every iOS/Safari < 16.4 visitor lost all client-rendered content on pages containing the table extension's chunk. Sentry showed the SyntaxError with culprit `parseModule` and a browser breakdown that was 100% Safari/iOS.

This is the same failure class as #5889 (extension-code, fixed by #5916 "support safari versions 16.4 and older") and #6727 (extension-mathematics, fixed by #6731).

### Expected Behavior

The published dist parses on WebKit < 16.4, as it did up to 3.27.1. The escaping behavior is expressible without lookbehind — I've opened a PR replacing the regex with an exact-parity lookbehind-free pattern (behavioral equivalence verified over a 200k-string randomized corpus).

Not reproducible in a sandbox on a current browser — the failure happens at script parse time on old WebKit only. To observe the mechanism, run `new Function('/(?<!a)b/')` in Safari <= 16.3 (or any WKWebView on iOS < 16.4): it throws the same SyntaxError. Loading any bundle that includes `@tiptap/extension-table` >= 3.27.2 fails the same way.

### Additional context

Two more lookbehind occurrences currently on `main`, listed for completeness:

- `packages/extension-mathematics/src/extensions/InlineMath.ts` (~line 223): `find: /(?<!\$)(\$\$([^$\n]+?)\$\$)(?!\$)/` — also a regex literal, so the same parse-time chunk failure applies to extension-mathematics (the #6727 class again). My PR does not touch it: rewriting input-rule regexes shifts match offsets/captures and deserves maintainer judgment.
- `packages/extension-find-and-replace/src/search/regex.ts` (~line 62) builds a lookbehind inside a `new RegExp(...)` string — this fails only at runtime when whole-word search runs, so it degrades that feature on old WebKit rather than killing the chunk.

jansalwowski:
## Fixes

Fixes #8179

## Changes and Review

- `escapeTableCellPipes` used a negative-lookbehind regex literal; WebKit < Safari 16.4 cannot parse lookbehind, and a regex literal fails the **whole chunk at parse time** — any bundle including `@tiptap/extension-table` >= 3.27.2 was broken on those browsers (same class as #5889/#5916 and #6727/#6731).
- Replaced with an exact-equivalent lookbehind-free pattern: an alternation that consumes already-escaped `\|` pairs untouched and escapes bare `|`. Behavioral parity with the old regex verified over a 200k-string randomized corpus (`a` / `\` / `|` / backtick / space).
- Added unit tests pinning the escape behavior — including the already-escaped-pipe branches the suite didn't cover — and verified the built dist is free of lookbehind. All 92 extension-table tests pass.

## Checklist

### Responsibility

https://claude.ai/code/session_01UHw12prLZYBUmTwQKv1VjB

### DIFF
--- a/.changeset/fix-table-markdown-lookbehind-safari.md
+++ b/.changeset/fix-table-markdown-lookbehind-safari.md
@@ -1,0 +1,5 @@
+---
+'@tiptap/extension-table': patch
+---
+
+Replace the negative lookbehind in the table markdown pipe-escape regex with an equivalent lookbehind-free pattern. WebKit before Safari 16.4 cannot parse lookbehind assertions, and since the regex was a literal, the entire chunk containing the table extension failed at parse time on those browsers — any bundle importing `@tiptap/extension-table` ≥ 3.27.2 was unusable on iOS/Safari < 16.4.
--- a/.changeset/fix-table-markdown-lookbehind-safari.md
+++ b/.changeset/fix-table-markdown-lookbehind-safari.md
@@ -2,4 +2,4 @@
 '@tiptap/extension-table': patch
 ---
 
-Replace the negative lookbehind in the table markdown pipe-escape regex with an equivalent lookbehind-free pattern. WebKit before Safari 16.4 cannot parse lookbehind assertions, and since the regex was a literal, the entire chunk containing the table extension failed at parse time on those browsers — any bundle importing `@tiptap/extension-table` ≥ 3.27.2 was unusable on iOS/Safari < 16.4.
+Fixed table markdown pipe escaping so the extension loads on older Safari and iOS versions (WebKit before Safari 16.4).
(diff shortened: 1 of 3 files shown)


bdbch (approved):
LGTM, thank you!

PR #8186, opened 2026-08-12
feat(core): add ProseMirror type guard functions

bdbch:
## Fixes

- N/A

## Changes and Review

This PR includes a few new helper functions that are exported by the core package which can be used to safely check if a value is a specific ProseMirror type. In the past we used `instanceof` for this, but this can cause issues when third-party editors are involved where different class instances could be used internally.

## Checklist

### Responsibility


packages/core/src/helpers/isProseMirrorSlice.ts
```diff
@@ -0,0 +1,43 @@
...
+    return false
+  }
+
+  const fragment = content as Record<string, unknown>
+
+  if (!isProseMirrorFragment(fragment)) {
+    return false
+  }
+
+  return true
```
Aslam97, line 42: Just a small nit pick, I would just return `slice.size === content.size - slice.openStart - slice.openEnd` cause of how size calculated here https://code.haverbeke.berlin/prosemirror/prosemirror-model/src/commit/09098e3b00a2e36843040bcde1b7af9adf76816e/src/replace.ts#L34

bdbch: What's the reasoning behind this besides just being a bit more stricter? I don't really see a lot of value adding this (and also makes the function function harder to review as reviewers will need to know that this calculation is actually 1:1 from the prosemirror getter.

I'd say in most cases just checking the object structure should be enough.

Aslam97: Ok I agree. My intention was only to make the check a bit stricter, since the goal is just to identify a ProseMirror Slice across bundles not Slice validity validator

Aslam97:
@bdbch Also shouldnt it include some kind of manual create **Fragment and Slice**? Because of the error `can not convert <"sugges..."> to a Fragment etc.`?

 Screenshot 2026-08-13 at 14 30 42

- I would expect APIs like `createManualFragmen` creates a new `Fragment` containing nodes, using the constructor that owns source.
- I would also expect `createManualSlice` which creates a new `Slice` using the constructor that owns source, and pass the Fragment value (created from `createManualFragment`) and as well as the `openStart` and `openEnd` that want to be convert into Slice 

Just like https://github.com/ueberdosis/tiptap-pro/pull/980/changes#diff-1c656e8e49bb7a1fddda54c61cc8bc7dbb8fa2cc6c676d4016f476ab4add6095 and https://github.com/ueberdosis/tiptap-pro/pull/980/changes#diff-e8b14ce1596a4afe6d3b23e1f41fe7d608adf0e43b8e714030eac4bc44fade75

bdbch:
I wouldn't really bother with creating Fragments or Slices right now.

The most common problems we encounter right now are `instanceof` checks which break cross-module, let's say I have a third-party editor and I create my own Extension outside of it with my own ProseMirror package versions the host editor should not care what module created Slices or Fragments.

If we avoid strict instanceof checks in the future, the host should not care if a Slice-like object comes from one or another Prosemirror instance.

Otherwise we have to literally rebuild all ProseMirror classes to make sure everything is completely separated from the ProseMirror instance used by the host.

Aslam97 (approved):

PR #8189, opened 2026-08-14
fix(extension-floating-menu): re-evaluate menu after IME composition ends

### LINKED ISSUE #6838: The 'floating menu' does not respond to Korean.
### Affected Packages

core

### Version(s)

3

### Bug Description

I have been testing on `https://tiptap.dev/docs/editor/getting-started/style-editor/custom-menus`.

The ‘menu’ appears and disappears correctly for English, numbers, and special characters.

However, it does not respond to other languages (Korean).

It should respond to the regular expressions `/[가-힣]+/g` or `[가-힣ㄱ-ㅎㅏ-ㅣ]+/g`.

Image

### Browser Used

Chrome

### Code Example URL

_No response_

### Expected Behavior

It should respond to the regular expressions `/[가-힣]+/g` or `[가-힣ㄱ-ㅎㅏ-ㅣ]+/g`.

It should respond correctly to other languages (Korean) and the menu should disappear.

_No response_

### Dependency Updates

g1mn:
Fixes #6838.

`FloatingMenuView.updateHandler` bails out on `view.composing`, and there is no `compositionend` listener anywhere in the plugin, so every doc/selection change that happens during an IME composition is skipped. The flush at the end of the composition usually carries no further doc or selection change either, so it hits the `isSame` early return as well — the result is that `shouldShow` is never re-evaluated. Hangul composes every syllable, which is why this only shows up for Korean (and other IME input) while English, numbers and punctuation behave.

The fix adds a `compositionend` listener on `view.dom` that re-runs the update one tick later, following the existing `focusHandler` pattern (passing no `oldState` makes `isSame` false so the check actually runs). It is guarded with `editor.isDestroyed` — same guard `bubble-menu-plugin.ts` already uses — and removed in `destroy()`.

Two regression tests:

- `shouldShow` is not called while `view.composing` is true, and is called after a `compositionend`
- the listener is gone after `destroy()`

Reverting only the `src/` change makes the first test fail, so it does cover the bug.

Checks: `vitest run packages/extension-floating-menu` 11 passed (9 pre-existing + 2 new), `oxlint` 0 warnings/errors, `oxfmt --check` clean, changeset added.

`bubble-menu-plugin.ts` has the same early return, but I left it out to keep this scoped to the reported issue — happy to send a follow-up if you want the same handling there.

<sub>Drafted with AI assistance; reviewed and tested by me.</sub>

### DIFF
--- a/packages/extension-floating-menu/src/floating-menu-plugin.ts
+++ b/packages/extension-floating-menu/src/floating-menu-plugin.ts
@@ -157,6 +157,8 @@ export class FloatingMenuView {
 
   private resizeDebounceTimer: number | undefined
 
+  private compositionEndTimer: number | undefined
+
   private isVisible = false
 
   private scrollTarget: HTMLElement | Window = window
@@ -385,7 +387,9 @@ export class FloatingMenuView {
     // flush at the end of the composition usually carries no further doc or selection change, so it
     // hits the `isSame` early return as well. Without this the menu is never re-evaluated for
     // composed input (e.g. Hangul, where every syllable goes through a composition).
-    setTimeout(() => {
+    clearTimeout(this.compositionEndTimer)
+
+    this.compositionEndTimer = window.setTimeout(() => {
       if (this.editor.isDestroyed) {
         return
       }
@@ -580,6 +584,7 @@ export class FloatingMenuView {
     this.editor.off('blur', this.blurHandler)
     this.editor.off('transaction', this.transactionHandler)
     this.view.dom.removeEventListener('compositionend', this.compositionEndHandler)
+    clearTimeout(this.compositionEndTimer)
 
     if (this.floatingUIOptions.onDestroy) {
       this.floatingUIOptions.onDestroy()
--- a/packages/extension-floating-menu/src/floating-menu-plugin.ts
+++ b/packages/extension-floating-menu/src/floating-menu-plugin.ts
@@ -320,6 +320,7 @@ export class FloatingMenuView {
     this.editor.on('transaction', this.transactionHandler)
     window.addEventListener('resize', this.resizeHandler)
     this.scrollTarget.addEventListener('scroll', this.resizeHandler)
+    this.view.dom.addEventListener('compositionend', this.compositionEndHandler)
 
     this.update(view, view.state)
 
@@ -377,6 +378,20 @@ export class FloatingMenuView {
 
   mousedownHandler = () => {
     this.preventHide = true
+  }
+
+  compositionEndHandler = () => {
+    // While an IME composition is running `updateHandler` bails out on `view.composing`, and the
+    // flush at the end of the composition usually carries no further doc or selection change, so it
+    // hits the `isSame` early return as well. Without this the menu is never re-evaluated for
+    // composed input (e.g. Hangul, where every syllable goes through a composition).
+    setTimeout(() => {
+      if (this.editor.isDestroyed) {
+        return
+      }
+
+      this.update(this.view)
+    })
   }
 
   focusHandler = () => {
@@ -564,6 +579,7 @@ export class FloatingMenuView {
     this.editor.off('focus', this.focusHandler)
     this.editor.off('blur', this.blurHandler)
     this.editor.off('transaction', this.transactionHandler)
+    this.view.dom.removeEventListener('compositionend', this.compositionEndHandler)
 
     if (this.floatingUIOptions.onDestroy) {
       this.floatingUIOptions.onDestroy()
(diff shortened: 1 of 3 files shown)


g1mn:
Thanks — all three were valid, fixed in the follow-up commit.

The destroy race was the real one: I had only guarded with `editor.isDestroyed`, but `FloatingMenuView.destroy()` can run while the editor is still alive, so a timeout scheduled by `compositionend` would still fire. The timeout id is now tracked (same pattern as `resizeDebounceTimer`), cleared on re-entry, and cleared in `destroy()`.

Added a test for exactly that lifecycle — view destroyed, editor alive — and it fails if the `clearTimeout` in `destroy()` is removed.

The IME test now asserts the visible result instead of just the callback: `shouldShow` returns `true` before the composition and `false` after, and the test asserts `onHide` fires. Changeset reworded to describe the user-visible behaviour.

Checks: 12 passed (9 pre-existing + 3 new), `oxlint` 0, `oxfmt --check` clean.

packages/extension-floating-menu/src/floating-menu-plugin.ts
```diff
@@ -379,6 +382,22 @@ export class FloatingMenuView {
     this.preventHide = true
   }
 
+  compositionEndHandler = () => {
+    // While an IME composition is running `updateHandler` bails out on `view.composing`, and the
+    // flush at the end of the composition usually carries no further doc or selection change, so it
+    // hits the `isSame` early return as well. Without this the menu is never re-evaluated for
+    // composed input (e.g. Hangul, where every syllable goes through a composition).
```
bdbch: please shorten this comment. could be a simple plain one-liner. also try to use simple, plain english. claude tends to overdo it with comments which makes them harder to read/parse through.

bdbch (changes requested):

packages/extension-floating-menu/src/floating-menu-plugin.ts
g1mn: Done — one line now. I also trimmed the two long comments in the test file for the same reason.

packages/extension-floating-menu/src/floating-menu-plugin.ts
```diff
@@ -379,6 +382,19 @@ export class FloatingMenuView {
     this.preventHide = true
   }
 
+  compositionEndHandler = () => {
```
bdbch, line 385: Do we need to do the same for the bubble menu?

g1mn: Yes — I checked and it reproduces there too.

`bubble-menu-plugin.ts` has the same `composing || isSame` early return and no `compositionend` listener. The flow that breaks: select some text so the menu shows, then type over it with an IME. The selection gets replaced during the composition, so the menu should hide, but the update is skipped while composing and the flush at the end carries no further change. Doing the same thing without an IME hides it correctly, so it is the composing branch rather than something else in that path.

Want me to add it to this PR, or send it as a follow-up once this one lands?

g1mn: Added it here rather than a follow-up, since it is the same change.

`bubble-menu-plugin.ts` now gets the same `compositionend` listener, timer guard and cleanup as the floating menu, plus three tests mirroring the ones there. The tests are set to `updateDelay: 0` so they take the direct `update()` path instead of the debounce.

I checked that the behavioural test fails on the unpatched file and passes with it. Both packages: 24 tests passing.

packages/extension-bubble-menu/src/bubble-menu-plugin.ts
```diff
@@ -425,6 +428,19 @@ export class BubbleMenuView implements PluginView {
...
+  compositionEndHandler = () => {
+    // Re-check after IME input — updates are skipped while composing.
+    clearTimeout(this.compositionEndTimer)
+
+    this.compositionEndTimer = window.setTimeout(() => {
+      if (this.editor.isDestroyed) {
+        return
+      }
+
+      this.update(this.view)
```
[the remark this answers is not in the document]
g1mn: Good catch, fixed in bf2cd66.

`destroy()` now clears `updateDebounceTimer` as well. The leak was not limited to the composition path: any `update()` with a non-empty selection and a positive `updateDelay` could leave a timer pending at destroy, so clearing it in `destroy()` covers every caller rather than just the one I added.

I also added a test for the debounced path. My earlier tests used `updateDelay: 0` to keep them deterministic, which meant they never exercised the branch the bug lives in. The new test uses a real selection with `updateDelay: 20` and asserts nothing runs after `destroy()`; it fails without the added line.

On the two nitpicks: the changeset already reads "Floating menus now update correctly after IME input, including Korean text", and the floating-menu test already asserts `onHide` alongside `shouldShow`. Both were addressed in an earlier revision.

PR #8202, opened 2026-08-15
fix(react): update useEditorState immediately when the editor instance arrives or is replaced

merong:
## The problem

`useEditorState`'s `EditorStateManager` is created once with the first render's editor value. `watch(nextEditor)` reassigns `this.editor` and moves the event subscriptions, but it never bumps `transactionNumber` or notifies subscribers when the editor **identity** changes. Since `getSnapshot()` returns the cached snapshot while `transactionNumber` is unchanged, selectors stay pinned to the previous editor's snapshot until the *new* editor emits its first transaction.

Observable symptom: with `useEditor({ immediatelyRender: false })`, every `useEditorState` consumer (e.g. toolbar buttons built on it) sees `{ editor: null }` after the editor arrives — buttons stay disabled on cold load until the user types or clicks into the document. The same stall happens when an editor instance is replaced (A→B) or torn down (A→null).

## The fix

Detect the identity change at the top of `watch()`, invalidate the cached snapshot and notify subscribers:

- first mount is unaffected — the manager was constructed with the same editor, so `editorChanged` is `false`
- `watch` is called from a layout effect, so the notify → re-render scheduling is safe
- replacement (A→B) and teardown (A→null) share the same path

## Tests

`packages/react/__tests__/useEditorStateEditorIdentity.spec.ts` (4 cases): arrival without a transaction, instance replacement, pre-arrival `setEditable(..., false)` state visible on arrival, and a regression guard that transactions still re-render. `pnpm vitest run packages/react` → 25 passed.

A changeset (`@tiptap/react` patch) is included.

### DIFF
--- a/.changeset/2026-08-10-useeditorstate-editor-arrival.md
+++ b/.changeset/2026-08-10-useeditorstate-editor-arrival.md
@@ -1,0 +1,5 @@
+---
+'@tiptap/react': patch
+---
+
+useEditorState now updates immediately when the editor instance becomes available or is replaced, instead of waiting for the editor's first transaction.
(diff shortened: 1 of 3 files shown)


merong:
Closing — we've moved this change to our own maintained fork. Thanks for the automated review feedback.

PR #8211, opened 2026-08-18
fix(core): make clearNodes unwrap every level of a nested list

### LINKED ISSUE #8175: clearNodes() only unwraps one list level, leaves nested lists behind (and can throw `RangeError`)
### Affected Packages

tiptap/core

### Tiptap Version

3.29.2 

### Browser Used

Chrome

### What happened?

`clearNodes()` walks the document **as it was before its own transaction** (`state.doc.nodesBetween(...)`) while it mutates `tr`. A lift that only becomes legal *after* an earlier lift in the same transaction is therefore never performed, and positions taken from the stale document can end up pointing at the wrong node.

Two symptoms follow from this:

**1. Nested lists are only partly unwrapped**

A `listItem` that still holds a nested list cannot be lifted: `listItem`'s content spec is `paragraph block*`, so removing the leading paragraph would leave an invalid node and `liftTarget()` returns `null`. Once the nested list has been lifted out, the lift *would* be legal — but there is no second pass, so the item stays wrapped.

This is what a user sees when selecting an indented list and pressing "clear formatting": every item is flattened **except** the one that has children.

**2. `clearNodes()` followed by another node command can throw**

`editor.chain().focus().clearNodes().toggleOrderedList().run()` on the same document throws:

```
RangeError: Invalid content for node type paragraph
    at setNodeMarkup (prosemirror-transform)
    at Transaction.setNodeMarkup (prosemirror-transform)
    at clearNodes.ts:47
```

`setNodeMarkup(nodeRange.start, defaultType)` is applied at a position derived from the stale document and hits a node that cannot take that type.

### Steps to Reproduce

```ts
import { Editor } from '@tiptap/core'

const content =
  '<ul><li><p>one</p><ul><li><p>one.one</p></li></ul></li><li><p>two</p></li></ul>'

const editor = new Editor({ element, extensions: [StarterKit], content })
editor.commands.selectAll()
editor.chain().focus().clearNodes().run()

console.log(editor.getHTML())
```

### Expected Behavior
`clearNodes()` should turn the whole selection into plain paragraphs, no matter how deep the lists are nested. For the document in the reproduction:

```html
<p>one</p><p>one.one</p><p>two</p>
```

And a node command chained after `clearNodes()`, like `.clearNodes().toggleOrderedList()`, should convert the content instead of throwing.

_No response_

Workaround: run the command repeatedly in separate transactions until the document stops changing. Chaining several `clearNodes()` into one transaction does not work — it throws the `RangeError` above.

```ts
for (let pass = 0; pass < 10; pass++) {
  const before = editor.state.doc
  editor.chain().focus().clearNodes().run()
  if (editor.state.doc.eq(before)) break
}
```

This costs one undo step per pass, so a fix inside the command would be better.

solomonaustin:
Fixes #8175. The diagnosis in that issue is @McGery's and it was exact.

### Problem

`clearNodes()` walked `state.doc` — the document as it was *before* its own transaction — while mutating `tr`. Two symptoms follow.

**Nested lists are only partly unwrapped.** A lift becomes legal only after an earlier lift in the same transaction: `listItem` is `paragraph block*`, so an item that still holds a nested list can't be lifted until that list is out of it. With one sweep there's no second chance, so "clear formatting" on an indented list flattens every item *except* the ones with children:

```html
<!-- in -->
<ul><li><p>one</p><ul><li><p>one.one</p></li></ul></li><li><p>two</p></li></ul>
<!-- out, before -->
<ul><li><p>one</p></li></ul><p>one.one</p><p>two</p>
```

**Chaining could throw.** Positions taken from the stale document could land on a node that can't take the default type, so `setNodeMarkup` threw `RangeError: Invalid content for node type paragraph` — reachable from `.clearNodes().toggleOrderedList()` in a single chain.

### Fix

Each range is swept against the transaction's *current* document until it stops changing, so a lift unblocked by an earlier one still happens. Positions within a sweep are mapped through only the steps that sweep adds (`tr.mapping.slice(passMappingStart)`) rather than the whole transaction.

The retype is additionally skipped when the target can't hold its own content as the default type — the guard that turns the `RangeError` above into a no-op.

The loop normally exits on the `doc.eq` check after one settling pass; `MAX_PASSES` is only a runaway guard.

I deliberately left the retype unconditional otherwise (rather than skipping when the node is already the default type): `setNodeMarkup` also resets attributes, and normalising those looks intended for a "clear" command, so skipping would be a silent behaviour change.

### Testing

New `packages/core/src/commands/__tests__/clearNodes.test.ts` covering one-level, two-level and three-level lists, the chained-command case, and plain paragraphs as a no-op.

Verified against unmodified source first — 3 of the 5 fail for exactly the reported reasons (partial unwrap, and the `RangeError`), while the flat-list and plain-paragraph cases already passed, so the existing behaviour is pinned.

- `vitest run` across the repo — **192 files / 1856 tests pass**
- `oxlint` and `oxfmt --check` clean on both files
- Patch changeset included


solomonaustin:
Both review points applied in 8fb14a1.

**Fixed sweep limit — removed.** The loop now ends only on `tr.doc.eq(docBefore)`. That is safe to leave unbounded: a sweep either lifts something, which strictly reduces nesting, or changes nothing and ends the loop. `setNodeMarkup` is idempotent, so it cannot oscillate. The cap was an arbitrary ceiling on how deep a document may be, which is exactly the class of bug this PR is fixing.

One correction to the review's premise, since I checked rather than assumed: a 30-level list still passes **with** a 20-sweep cap in place. A sweep lifts every legal position in the range, not one level, so the sweeps needed are far fewer than the nesting depth. I could not construct a list deep enough to hit 20 sweeps.

So I renamed the new case to `flattens a deeply nested list` — the original name claimed it was a regression test for the cap, and it demonstrably is not. It is still worth keeping as a deep-nesting guard, but I did not want a test asserting something it does not prove. Removing the cap remains correct on principle.

**Changeset.** Rewritten to user-visible behaviour with the internals dropped, and renamed to the dated form `AGENTS.md` specifies (`2026-08-18-clear-nodes-nested-lists.md`) — the original filename was missing the date prefix.

Verification after the changes: 192 files / 1857 tests pass, `oxlint` and `oxfmt --check` clean.

solomonaustin:
I reviewed my own diff again before waiting on another round, and found a defect in it. Pushed as a16ce4d.

**The ranges were mapped from the wrong baseline.** `ranges` comes from `tr.selection`, so the positions already point into the *current* document — but I was mapping them through `tr.mapping`, the whole transaction. When an earlier command in the same chain had already added steps, that mapped them a second time and the sweep ran past the end of the document.

```ts
editor.chain().focus().toggleBulletList().clearNodes().run()
```

That is the reverse of the order in #8175, and it throws.

Worth being clear about the scope: **this is not a regression from this PR — it fails on `main` too**, with `Position 22 out of range`. My version only changed which error you get. I fixed it here because this PR is already rewriting exactly that position handling, and shipping a rewrite that leaves the same crash in place seemed worse than a slightly wider diff. Happy to split it into its own PR if you would rather keep this one to the reported bug.

The fix records `tr.mapping.maps.length` when the command starts and maps the ranges from there, so it is correct whether or not the chain has already changed the document.

While in there I also stopped rebuilding the per-sweep `Mapping` for every visited node — it is now rebuilt only when a step actually lands, which avoids O(nodes × steps) copying on a large selection.

New test: `runs correctly after another node command in the same chain`.

192 files / 1858 tests pass, `oxlint` and `oxfmt --check` clean.

solomonaustin:
Both CodeRabbit threads are fixed and resolved. I also gave the diff another read myself and found something worth changing, pushed as a6eca4c.

`setNodeMarkup` does not diff — it writes a step whether or not the markup actually differs. So every textblock in range was retyped on each sweep even when it already carried the default markup, and those steps end up in the undo stack and go out to collaborating clients. Skipping the retype when `target.hasMarkup(defaultType)` fixes it, and it also removes most of the cost of the extra sweep this PR introduces.

Measured on the same two documents:

| | already-flat doc | nested list |
| --- | --- | --- |
| `main` | 3 steps | 6 steps, output still wrapped |
| this PR before the change | 3 steps | 13 steps |
| this PR now | 0 steps | 4 steps |

So clearing a selection that is already plain is now a no-op instead of three steps that change nothing, and the nested case ends up cheaper than `main` while actually finishing the job. I kept the retype unconditional in every other case — `setNodeMarkup` also resets attributes, and normalising those looks intended for a "clear" command, so only the exact-match case is skipped.

Added one more test for a selection that spans part of the document rather than everything, since every case so far went through `selectAll`.

192 files / 1859 tests pass, `oxlint` and `oxfmt --check` clean.

packages/core/src/commands/clearNodes.ts
```diff
@@ -16,41 +16,84 @@ declare module '@tiptap/core' {
...
     const { selection } = tr
     const { ranges } = selection
 
     if (!dispatch) {
       return true
     }
 
+    // The ranges point into the document as it is now, which is not always the
+    // one the transaction started from: an earlier command in the same chain
+    // may already have changed it. Map them from here on, not from the start.
```
bdbch:
```suggestion
    // map the position through the transaction
    // earlier commands may have already changed the document
```

packages/core/src/commands/clearNodes.ts
```diff
@@ -16,41 +16,84 @@ declare module '@tiptap/core' {
...
+      const to = $to.pos
 
-        const { doc, mapping } = tr
-        const $mappedFrom = doc.resolve(mapping.map(pos))
-        const $mappedTo = doc.resolve(mapping.map(pos + node.nodeSize))
-        const nodeRange = $mappedFrom.blockRange($mappedTo)
+      // A lift can only become legal once an earlier one has run: a list item
+      // holding a nested list can't be lifted until that list is out of it. So
+      // sweep until nothing changes. Each sweep either lifts something, which
+      // makes the range shallower, or ends the loop.
```
bdbch:
```suggestion
      // keep sweeping until no more lifts become valid
      // earlier lifts can make nested items liftable
```

packages/core/src/commands/clearNodes.ts
```diff
@@ -16,41 +16,84 @@ declare module '@tiptap/core' {
...
+      for (;;) {
+        const docBefore = tr.doc
+        const sweepStart = tr.mapping.maps.length
+        const rangeMapping = tr.mapping.slice(commandStart)
 
-        if (!nodeRange) {
-          return
-        }
+        // Positions come from `docBefore`, so they need the steps this sweep
+        // adds. Rebuild that mapping only when a step lands, not per node.
```
bdbch:
```suggestion
        // rebuild the mapping after each sweep changes the document
        // positions still refer to `docBefore`
```

packages/core/src/commands/clearNodes.ts
```diff
@@ -16,41 +16,84 @@ declare module '@tiptap/core' {
...
+          const targetLiftDepth = liftTarget(nodeRange)
+
+          if (node.type.isTextblock) {
+            const { defaultType } = $mappedFrom.parent.contentMatchAt($mappedFrom.index())
+            const target = tr.doc.nodeAt(nodeRange.start)
+
+            // `setNodeMarkup` does not diff, so retyping a node that already
+            // has the default markup only adds a step that changes nothing.
+            // An earlier lift can also leave this position on a node that
+            // cannot hold its own content as that type, which would throw.
```
bdbch:
```suggestion
            // skip nodes that already have the default markup
            // invalid nodes may throw after an earlier lift
```

packages/core/src/commands/clearNodes.ts
```diff
@@ -16,41 +16,84 @@ declare module '@tiptap/core' {
...
     const { ranges } = selection
 
     if (!dispatch) {
       return true
     }
 
+    // The ranges point into the document as it is now, which is not always the
+    // one the transaction started from: an earlier command in the same chain
+    // may already have changed it. Map them from here on, not from the start.
+    const commandStart = tr.mapping.maps.length
```
bdbch: find a better naming for this variable please, `commandStart` sounds like it's the start of a command, not the length of the mapping' maps.

packages/core/src/commands/clearNodes.ts
```diff
@@ -16,41 +16,84 @@ declare module '@tiptap/core' {
...
-        const $mappedFrom = doc.resolve(mapping.map(pos))
-        const $mappedTo = doc.resolve(mapping.map(pos + node.nodeSize))
-        const nodeRange = $mappedFrom.blockRange($mappedTo)
+      // A lift can only become legal once an earlier one has run: a list item
+      // holding a nested list can't be lifted until that list is out of it. So
+      // sweep until nothing changes. Each sweep either lifts something, which
+      // makes the range shallower, or ends the loop.
+      for (;;) {
+        const docBefore = tr.doc
+        const sweepStart = tr.mapping.maps.length
```
bdbch: Rename this as well - it's the maps length for this sweep

packages/core/src/commands/clearNodes.ts
```diff
@@ -16,41 +16,84 @@ declare module '@tiptap/core' {
...
-        if (!nodeRange) {
-          return
-        }
+        // Positions come from `docBefore`, so they need the steps this sweep
+        // adds. Rebuild that mapping only when a step lands, not per node.
+        let mappedThrough = sweepStart
+        let sweepMapping = tr.mapping.slice(sweepStart)
 
-        const targetLiftDepth = liftTarget(nodeRange)
+        docBefore.nodesBetween(rangeMapping.map(from), rangeMapping.map(to), (node, pos) => {
```
bdbch: I'd like to move the from / to out of the nodesBetween arguments - easier to read

```suggestion
        const mappedFrom = rangeMapping.map(from)
        const mappedTo = rangeMapping.map(to)

        docBefore.nodesBetween(mappedFrom, mappedTo, (node, pos) => {
```

packages/core/src/commands/clearNodes.ts
```diff
@@ -16,41 +16,84 @@ declare module '@tiptap/core' {
...
 
-        const { doc, mapping } = tr
-        const $mappedFrom = doc.resolve(mapping.map(pos))
-        const $mappedTo = doc.resolve(mapping.map(pos + node.nodeSize))
-        const nodeRange = $mappedFrom.blockRange($mappedTo)
+      // A lift can only become legal once an earlier one has run: a list item
+      // holding a nested list can't be lifted until that list is out of it. So
+      // sweep until nothing changes. Each sweep either lifts something, which
+      // makes the range shallower, or ends the loop.
+      for (;;) {
```
bdbch: This looks dangerous, I don't see a direct way to break this, but since this script will run on the main thread it could potentially start to completely crash the browser tab if this loop somehow is never interrupted.

bdbch (changes requested):

packages/core/src/commands/clearNodes.ts
solomonaustin: Applied.

packages/core/src/commands/clearNodes.ts
solomonaustin: Applied.

packages/core/src/commands/clearNodes.ts
solomonaustin: Applied.

packages/core/src/commands/clearNodes.ts
solomonaustin: Applied.

packages/core/src/commands/clearNodes.ts
solomonaustin: Renamed to `startMapsLength`.

packages/core/src/commands/clearNodes.ts
solomonaustin: Renamed to `sweepMapsLength`, and the cache counter is `cachedMapsLength`.

packages/core/src/commands/clearNodes.ts
solomonaustin: Applied — `mappedFrom` / `mappedTo` are lifted out now.

packages/core/src/commands/clearNodes.ts
solomonaustin: Agreed. Bounded by the nesting depth of the range now: a sweep only helps if it lifts something, and a lift removes a level, so `n` deep can't need more than `n + 1` sweeps. It still exits on the `doc.eq` check — the bound is only there so a tab can't hang.

solomonaustin:
@bdbch all eight are addressed in ff8f861 — ready for another look when you have time. 193 files / 1873 tests pass, lint and format clean.

solomonaustin:
@bdbch gentle nudge on this one whenever you have a moment.

All eight of your review points were addressed in `ff8f861`, and I've since merged `main` in (`d1164ca`). All 14 checks are green — unit, lint/format, and all four e2e shards — and there are no unresolved review threads left. The `changes requested` status is just the original review still standing, so it needs a re-review to clear.

Happy to rebase, split it up, or trim the scope if any of that would make it easier to land.

solomonaustin:
CI went red here, so to be clear about which half was mine:

**Mine — fixed in `d1c084f`.** `@tiptap/extension-list-item` was removed from the repo, and this suite still imported it, so it failed at collection:

```
Failed to resolve import "@tiptap/extension-list-item" from "packages/core/src/commands/__tests__/clearNodes.test.ts"
```

Now imports `{ ListItem }` from `@tiptap/extension-list`, matching `nodePos.spec.ts` and `getPreviousBlockSibling.spec.ts`. I also switched the test to `vite-plus/test` — it was the only file in `packages/core` still on `vitest` (22 to 1). Full suite: 198 files, 1974 tests pass.

**Not mine.** The `Check linting \& formatting` failure is on `main`, not this branch:

```console
$ git checkout main && vp check
error: Formatting issues found
.changeset/fix-ios-ignore-mutation-contentdom.md
```

That changeset came in with #8221 and uses double quotes where oxfmt wants single. It fails on a pristine `main` checkout, so it is currently red for every open PR. I opened #8303 with the one-character fix rather than carrying an unrelated file in this PR. Once that lands this branch should be fully green.

solomonaustin:
All checks pass now.

Reversing what I said above: rather than wait on #8303, I carried the same one-character changeset formatting fix into this branch so CI is green here and now. It is byte-identical to #8303, so whichever lands first the other merges without a conflict, and the duplicate collapses to nothing.

If you would rather this PR contain only its own change, say so and I will drop that commit once #8303 is in.

solomonaustin:
@bdbch this one is ready whenever you have a moment — everything you raised is done and it is fully green now.

- All 8 review points addressed in `ff8f861`
- The `@tiptap/extension-list-item` import that broke collection is fixed in `d1c084f` (now `{ ListItem }` from `@tiptap/extension-list`, matching `nodePos.spec.ts`)
- **12/12 checks passing** — unit, build, lint/format and all four e2e shards
- **0 unresolved review threads**

Thanks for merging #8303, that cleared the formatting failure on `main`. This branch still carries its own copy of that one-character fix so CI could go green before #8303 landed; it is byte-identical, so it merges cleanly either way. Happy to drop that commit if you would rather the PR contain only its own change.

The only thing left is the `changes requested` review from 19 Aug, which stays blocking until you re-review.

PR #8214, opened 2026-08-18
fix(extension-youtube): embed YouTube live URLs

### LINKED ISSUE #8209: YouTube live URLs (/live/<id>) render an iframe without src in v3 — v2 fix from #6234 was never ported
### Affected Packages

extension-youtube

### Tiptap Version

3.x (reproduced on 3.22.5 and 3.30.1)

### Browser Used

Not browser-specific

### What happened?

Embedding a YouTube livestream URL like `https://www.youtube.com/live/<videoId>` breaks silently.

`isValidYoutubeUrl` accepts the URL, so `setYoutubeVideo` and the paste rule insert the node without any error. But `getEmbedUrlFromYoutubeUrl` returns `null` for it, because `videoIdRegex` only knows `v=`, `list=` and `shorts/`. The result is an iframe rendered without a `src`, and the original URL is lost on the next HTML round trip (the node comes back with `src: null`).

This was reported in #5447 and fixed by #6234, but that PR was merged into the v2 `develop` branch and shipped in v2.13.0 only. The v3 line never got the fix — the extraction regex is unchanged on `main` and in every published 3.x release up to 3.30.1. #5447 was closed based on the v2 release, so nothing currently tracks the gap for v3.

### Expected Behavior

`https://www.youtube.com/live/<videoId>` embeds the same way `watch?v=<videoId>` does, like it has in v2 since 2.13.0. Happy to open a PR porting the #6234 approach to `main` if that helps.

solomonaustin:
Fixes #8209. @RyanFischback's report is accurate, including the pointer to the v2 fix.

### Problem

`isValidYoutubeUrl` accepts `https://www.youtube.com/live/<id>`, so `setYoutubeVideo` and the paste rule insert the node without complaint. But `getEmbedUrlFromYoutubeUrl` returns `null` for it, because the id regex only knew `v=`, `list=` and `shorts/`:

```ts
const videoIdRegex = /(?:(v|list)=|shorts\/)([-\w]+)/gm
```

The node therefore renders an iframe with **no `src`**, and because the attribute is empty the original URL is lost on the next HTML round trip. It fails silently — no error, no warning, just a blank embed.

### Fix

Match `live/` the way `shorts/` is already matched:

```ts
const videoIdRegex = /(?:(v|list)=|(?:shorts|live)\/)([-\w]+)/gm
```

The added group is non-capturing, so `matches[1]` / `matches[2]` keep their meaning and the `matches[1] === 'list'` playlist check is untouched.

### On porting the v2 fix

#6234 fixed this in v2 by replacing the regex with `new URL()` parsing plus a `getYoutubeVideoOrPlaylistId` helper. I deliberately did **not** port that here.

That refactor also covered `youtu.be`, but v3 already handles `youtu.be` separately a few lines above, so the only thing actually broken on v3 is `/live/`. Porting the whole thing would swap the parsing strategy for every URL shape — including moving to `new URL()`, which throws on the protocol-relative URLs `YOUTUBE_REGEX` accepts — to fix one missing case. The targeted change keeps the blast radius at the reported bug. Happy to do the full port instead if you would rather have v2 and v3 converge.

### Testing

Five tests in a `YouTube Live URL handling` block, mirroring the existing Shorts block: plain live URL, without `www`, with multiple params, with `nocookie`, and one with a query string after the id.

All five fail without the source change and pass with it.

Full suite: 191 files / 1856 tests pass. `oxlint` and `oxfmt --check` clean. Patch changeset included.


solomonaustin:
Both points applied in a634c09.

**Path boundary.** Confirmed before changing anything — all three near misses did produce an embed:

```
https://www.youtube.com/notlive/EkRHhOCdZjw            -> embed/EkRHhOCdZjw
https://www.youtube.com/watch?redirect=live/EkRHhOCdZjw -> embed/EkRHhOCdZjw
https://www.youtube.com/notshorts/EkRHhOCdZjw          -> embed/EkRHhOCdZjw
```

Worth noting the third one: `notshorts/` behaves this way on `main` today, so the loose match is pre-existing rather than something this PR introduced — adding `live` only widened it. The alternatives now have to start at a path separator, which fixes both:

```ts
const videoIdRegex = /(?:(v|list)=|\/(?:shorts|live)\/)([-\w]+)/gm
```

Three regression tests added for those URLs. All the existing shorts cases still pass, since they carry a real `/shorts/` path.

**Changeset.** Reworded to `YouTube live URLs (/live/<id>) now embed the video.` — the iframe and `src` detail is gone.

191 files / 1859 tests pass, `oxlint` and `oxfmt --check` clean.

bdbch (approved):
great, thank you!

Issue #4133, opened 2023-06-15
[Bug]: Cannot toggle nested list types


Slapbox:
### Which packages did you experience the bug in?

core

### What Tiptap version are you using?

2.1.0-rc.8

### What’s the bug you are facing?

The `toggleList` command seems to imply it can take a prescribed `itemType` and so one would think that the `toggleList` command would be capable of changing the items to their requested `itemType`s for the requested `listType`, but instead it fails on this line and forcibly un-nests the lists rather than converting the lists as expected:

https://github.com/ueberdosis/tiptap/blob/b0198eb14b98db5ca691bd9bfe698ffaddbc4ded/packages/core/src/commands/toggleList.ts#L96-L100

### What browser are you using?

Chrome

### What did you expect to happen?

I should be able to toggle a nested list to a different type without it mangling the list nesting completely (which is the current behavior)

### Did you update your dependencies?

- [X] Yes, I’ve updated my dependencies to use the latest version of all packages.

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

Issue #2492, opened 2022-02-06
Task item toggling causes various problems on list item structure 


bkempe:
### What’s the bug you are facing?

Toggling task items inside a list item structure leads to several issues with inconsistent confusing behavior.

### How can we reproduce the bug on our side?

Put the HTML into a file `test.html`, open it in a browser and copy&paste it into the example editor on https://tiptap.dev

When you toggle the task list on:
- item1 => nothing happens
- item1.1 => no task item created, nodes get shifted around but 
- item1.1p1 => a task item is created, nodes get shifted around 
- item1.1p2 => a task item created inside a new nested list item 

The schema below has several paragraphs in one listItem, but that is valid according to the schema and is possible to achieve via some copy&paste inside the editor.
HTML:
```html
<html>
<body>
<ul>
  <li>
    <p>item1</p>
    <ul>
      <li>
        <p>item1.1</p>
        <ul>
          <li>
            <p>item1.1p1</p>
            <p>item1.1p2</p>
            <p>item1.1p3</p>
            <ul>
              <li>
                <p>item1.1.1</p>
                <ul>
                  <li>
                    <p>item1.1.1.1</p>
                  </li>
                </ul>
              </li>
            </ul>
          </li>
        </ul>
      </li>
    </ul>
  </li>
</ul>
</body>
</html>
```

### What did you expect to happen?

One consistent behavior, e.g. task items created under the root node.

### Anything to add? (optional)

Somewhat related to https://github.com/ueberdosis/tiptap/issues/2428, which is a milder version of the problem.

### Did you update your dependencies?

- [X] Yes, I’ve updated my dependencies to use the latest version of all packages.

### Are you sponsoring us?

- [X] Yes, I’m a sponsor. 💖

bdbch:
Thanks for the issue. I'll see when I can start looking into this issue!

bdbch:
I think the issue here is that list items can have `paragraph block*` as content which allows you to have multiple paragraphs inside one list item.

From what I think list items and task items should be separated and block* in the future should be replaced by more strict content that resembled the text + optional sublist structure.

The correct fix for this would probably to have tasks / tasklists that are essentially lists with a taskitem as the first child instead of a paragraph?

bdbch:
This is a duplicate of #2428, which is being actively worked on. Closing in favor of the parent issue.

Issue #2428, opened 2022-01-24
toggleTaskList flattens nested bullet lists


bkempe:
### What’s the bug you are facing?

The `toggleTaskList` command flattens an existing nested bullet list it's invoked in.

### How can we reproduce the bug on our side?

Go to https://tiptap.dev/
In the example editor, create a bullet list like this:
- 1
  - 2
    - 3

With the cursor on the "3", click the "Task List" button.

### What did you expect to happen?

Only the bullet item itself or only the nested bullet item list should be converted to a task list.

### Did you update your dependencies?

- [X] Yes, I’ve updated my dependencies to use the latest version of all packages.

### Are you sponsoring us?

- [X] Yes, I’m a sponsor. 💖

itsmingjie:
Hi! I'm one of the program coordinators of a nonprofit program called [CodeDay Init](https://labs.codeday.org/), and we partner with colleges to help CS students make their first open-source contribution. 

We'll claim this issue for now and a group of students will be on it soon with mentorship! If this issue is no longer relevant, please let me know.

philippkuehn:
Oh great. Good luck :)

max-nextcloud:
Hi,

I've been looking into this problem some. Would love to see it fixed. I'll add my thoughts here in case they might be helpful.

I'd like to see `TaskLists` behave like `OrderedLists` and `BulletLists` when switching between them. When the cursor is on a bullet list item and i trigger `toggleOrderedList` the entire list becomes an ordered list. Parent lists and sublists remain what they are.

For ordered and bullet lists this switch is (somewhat) easy because they only need to switch the tag for the entire list and not the tags for the items.
Task lists however by default may not contain listitems - only taskitems. And bullet and ordered lists may only contain listitems.
So it's not easy to convert one into the other (to my knowledge). If i first convert the entire List the items are invalid content and if i try to convert an item it's invalid as well. I am not aware of a mechanism to convert both at the same time.

So I looked into building all three list types with a single item type. The item type would need a `checked` attribute for the todo list items to properly work - but it would only apply in todo lists. So i extended ListItem like that and configured the TaskList to contain ListItems.

The List items in the Task list should still render differently - they need to display the checkbox etc. In order to achieve that i tried to change the `renderHTML` function. But so far there is no way (that i know of) for the list item to tell it's inside a task list. The `renderHTML` function only gets the node and the HTML Attributes - no information about the parent node or the state the document is in.

[...]

Issue #166, opened 2019-01-17
Todo items mixed with regular list items


holtwick:
I am currently looking into todo item support and as far as I understand the current implementation, it looks although it behaves like a list, it cannot be mixed with other list items of numbered or bulleted lists. Is that correct?

For my project I would like to mix these and therefore am considering two options, which might be interesting for TipTaps feature as well:

1. Make the `TodoList` accept `ListItem` nodes. Difficulty: How can the ListItems become aware, that they should work like a todo entry?
2. Don't wrap `ListItem` etc. at all and make them live side by side with `Paragraph` or `Header` on the top level and make their indentation just another attribute. Numbering could be achieved with some CSS tricks. Difficulty: Parsing and generating HTML.

What do you think? Is there some "golden" way to do it?

holtwick:
Another little issue with list is, that you cannot change the list type intuitively: https://tiptap.scrumpy.io/  This is an issue as well in the regular ProseMirror demo.

![listchange](https://user-images.githubusercontent.com/49557/51380292-b2393b80-1b11-11e9-8fa9-260edec54297.gif)

holtwick:
I implemented my second idea and think it is quite alright for my specific needs. It needs some more fine tuning, but this is how it looks in action:

![screenflow](https://user-images.githubusercontent.com/49557/51431816-a2fbe000-1c2e-11e9-96eb-1f03b9315097.gif)

In case this is interesting for TipTap as well I can share the code.

The drawbacks as mentioned are the parsing and DOM creation. I'm planning to look into these issues next.

exnihilo-creatio:
@holtwick Nice, I would expect thought that increasing the indentation to also change the list type. For example, the first tab inscrease in the example to check the textbox to an `a` list and merge the `a` -> `b` etc.

Alecyrus:
@holtwick  Nice work !!! That is just what I wanna to implement. Could you share your code as a documentation example?  That will be helpful.

holtwick:
I'm not super happy with the results as well yet, but the other approaches have different issues as well. 

You can experiment with the current implementation here:  https://collect-app.com/web/?ref=tiptap

The extensions and SCSS files are attached as ZIP file here [TodoList.zip](https://github.com/scrumpy/tiptap/files/2778295/TodoList.zip). It needs some handwork to adopt but the idea should become clear.

[...]

PR #8225, opened 2026-08-20
feat(core): sink and lift list items across list types

### LINKED ISSUE #4424: [Bug]: Able to have different types of list items in same level

core

2.1.6

### Current behavior

We're able to have different types of list elements in a list according to the below gif.
Not sure if this is a bug or expected behavior though.
different list types

### Steps to reproduce
1. Create a bullet list enter something and press enter
2. Press tab for nested list and enter something and select numberered list then press enter
3. Press shift+tab to go to previous level and press tab to come back to same level
4. Newly created element will be of type bullet list item

### Expected behavior
As in the 4step above we should be having numbered list 

Note: I haven't added any code dependencies and this can be tried out in https://tiptap.dev/editor

Chrome

### Code example

_No response_

When we replicate the steps as in the gif attached we should have the list item of the nested list and not the parent level list type

_No response_

### Are you sponsoring us?

- [ ] Yes, I’m a sponsor. 💖

jordanh:
sinkListItem joins a previous item's sublist of another type instead of starting a second one, and liftListItem lifts into an outer list of another type as one of its items, carrying following siblings as a sublist. Both paths are schema-driven and fall back to prosemirror-schema-list, so single-type lists behave as before.

## Fixes

Fixes #4424

See also these related items: #166, #2492, #2428, #4133, [Mix bullet and ordered list](https://discuss.prosemirror.net/t/mix-bullet-and-ordered-list/5515), [Question regarding sinkListItem, liftListItem commands](https://discuss.prosemirror.net/t/question-regarding-sinklistitem-liftlistitem-commands/5322)

## Changes and Review

- `sinkListItem`: when the previous item ends with a list of another type (e.g. ordered list inside a bullet list, task list inside a bullet list, …), the selected items join that list and take its item type instead of a second sublist being created next to it. See repro in #4424

- `liftListItem`: when the item's list is nested inside an item of another type, the selected items become items of that outer list, taking its item type; following siblings move along as a sublist, the same shape `liftToOuterList` produces for single-type lists. Previously the item was lifted out into a bare paragraph inside the parent item

- Both paths use the schema only (`isList`, `contentMatch.defaultType`, `validContent`/`canReplace`) and return to the prosemirror-schema-list commands whenever they cannot produce a valid document, so nothing changes for lists of one type. `can()` reports the new cases without mutating

**Verification:**

```
pnpm vp test \
   packages/core/__tests__/sinkListItemMixedLists.spec.ts \
   packages/core/__tests__/liftListItemMixedLists.spec.ts`
```

```
pnpm playwright test demos/src/Extensions/ListKeymap --project=chromium
```

...or open the ListKeymap demo and press Tab on the last item / Shift+Tab on the nested task.

AI disclosure: implemented with the aide of Claude Code (Anthropic); reviewed and tested by me.

## Checklist

### Responsibility

### DIFF
--- a/.changeset/2026-08-17-list-item-sink-lift-across-list-types.md
+++ b/.changeset/2026-08-17-list-item-sink-lift-across-list-types.md
@@ -1,0 +1,5 @@
+---
+'@tiptap/core': minor
+---
+
+`sinkListItem` and `liftListItem` now work across list types. Pressing Tab on an item right after a sublist of another type (for example a task list inside a bullet list) moves the item into that sublist and gives it the sublist's item type instead of starting a second sublist. Pressing Shift+Tab on an item nested in a list of another type lifts it into that list as one of its items, and any items after it move along as its sublist. Nothing changes for lists of a single type.
--- a/.changeset/2026-08-17-list-item-sink-lift-across-list-types.md
+++ b/.changeset/2026-08-17-list-item-sink-lift-across-list-types.md
@@ -2,4 +2,4 @@
 '@tiptap/core': minor
 ---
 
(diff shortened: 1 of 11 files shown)


jordanh:
@bdbch I believe I've addressed the coderabbit review comments now

jordanh:
@bdbch just checking in here, is this something you'd be able to merge?

PR #8230, opened 2026-08-23
fix(extension-mathematics): drop the lookbehind from the inline math input rule

### LINKED ISSUE #6727: [v3] Mathematics extension would crash TipTap in older WebKit environments

giaBaoJS:
## Fixes

Fixes #6727

## Changes and Review

The inline math input rule's `find` was a regex **literal** containing a lookbehind. WebKit older than Safari 16.4 rejects lookbehind while *parsing* the literal, so the failure takes down the entire chunk `@tiptap/extension-mathematics` is bundled into, not just this one rule — the same class as #8179/#8180 and #5889/#5916. It is the only lookbehind left in the package (`BlockMath` and the migration util use lookahead only, which is fine everywhere), and it does reach the published bundle: building `main` yields one `(?<!\$)` in both `dist/index.js` and `dist/index.cjs`; building this branch yields zero in both.

**Why not rewrite the pattern.** #6731 replaced the lookbehind with a consuming `(^|[^$])` group and was reverted by #7481. The reason is structural, not cosmetic: `InputRule.ts` derives the replacement range purely from the match length (`from - (match[0].length - text.length)`), so any group that consumes the preceding character makes that character part of the range and it gets replaced along with the math. On `x$$a$$` the `(^|[^$])` form matches from index 0 and swallows the `x`. Any lookbehind-free *regex* has the same problem, because asserting on preceding context without consuming it is exactly what a lookbehind is for.

**The fix.** `InputRuleFinder` is `RegExp | ((text: string) => InputRuleMatch | null)`, so a function finder can inspect the preceding character without consuming it and return a `text` that is exactly the `$$latex$$` sequence the range is computed from. One consequence worth flagging in review: for a function finder `inputRuleMatcherHandler` builds the match array from the returned `text` alone, so `match[2]` no longer exists and the latex moves to `match.data`.

One subtlety I got wrong on the first attempt and want to point out, since it is easy to reproduce accidentally: when a candidate is rejected for being preceded by a `$`, the scan must resume from the character *after the rejected match's start*, not from its end. A valid match can begin inside a rejected one, at its trailing `$$` — on `$$$a$$b$$` the old regex matches `$$b$$`, and a finder that skipped the whole rejected match would return nothing. I checked equivalence between the old regex and the new finder exhaustively over all 88,572 strings of length ≤ 10 in the `{$, a, \n}` alphabet plus 400,000 random strings over `{$, a, b, x, space, \n, \}`: the resume-by-one finder has zero mismatches in index, matched text and captured latex; the resume-past-the-match variant has 8 and 179 respectively.

**On the evidence, plainly:** I cannot give you a behavioural counterfactual for this one. The bug is that a JS engine refuses to *parse* the file, and the engine running the test suite parses lookbehind fine, so the behavioural tests pass identically before and after — I ran the four new ones against unmodified `main` and they were green there too. The only assertion that actually goes red on `main` is the structural one in the new `lookbehind.spec.ts`, which fails with `expected [ '../src/extensions/InlineMath.ts' ] to deeply equal []`. Please read the behavioural tests as a regression net for the next person rather than as proof this change works — they pin the cases #6731 got wrong (a non-space character before the math, math at the start of a paragraph, the rejected-then-valid `$$$a$$b$$` case, and a trailing third `$`), so a future rewrite that swallows the preceding character fails loudly instead of shipping.

The structural test scans raw source text and so cannot tell a regex literal from a comment; that is called out in the file, and it means the sequence must not appear in prose in this package either.

Baseline on a clean checkout of `main` is 193 files / 1920 tests, and the mathematics package is 2 files / 7 tests. With this branch it is 194 / 1925 and 3 / 12. `pnpm lint` is clean and `oxfmt` reports no issues in the touched files.

## Checklist

### Responsibility

### Issue assignment

Per `CONTRIBUTING.md`, I have asked to be assigned #6727 (the issue is currently unassigned) and am opening this now rather than sitting on the work; happy to close it if a maintainer would rather it went to someone else.

### AI usage disclosure

Per `CONTRIBUTING.md`: I used an AI coding assistant while preparing this change.

Everything stated above is measured rather than asserted — the equivalence counts come from actually running both finders against the old regex, the lookbehind counts come from building the package with and without the change, and the note that the behavioural tests pass on `main` comes from running them there rather than from assuming it.

### DIFF
--- a/.changeset/2026-08-23-inline-math-lookbehind.md
+++ b/.changeset/2026-08-23-inline-math-lookbehind.md
@@ -1,0 +1,5 @@
+---
+'@tiptap/extension-mathematics': patch
+---
+
+The inline math input rule no longer uses a regex lookbehind, which WebKit older than Safari 16.4 rejects at parse time, taking down every chunk the extension is bundled into.
--- a/.changeset/2026-08-23-inline-math-lookbehind.md
+++ b/.changeset/2026-08-23-inline-math-lookbehind.md
@@ -2,4 +2,4 @@
 '@tiptap/extension-mathematics': patch
 ---
 
-The inline math input rule no longer uses a regex lookbehind, which WebKit older than Safari 16.4 rejects at parse time, taking down every chunk the extension is bundled into.
+The mathematics extension no longer crashes the editor in older WebKit browsers and WKWebView.
--- a/packages/extension-mathematics/src/extensions/InlineMath.ts
+++ b/packages/extension-mathematics/src/extensions/InlineMath.ts
@@ -67,19 +67,8 @@ declare module '@tiptap/core' {
 }
 
 /**
- * Finds the `$$latex$$` sequence that the inline math input rule should replace.
- *
- * This is a function finder rather than a regex literal because the rule has to
- * reject a sequence preceded by a third `$`, and there is no way to express that
- * in a regex without a lookbehind. A lookbehind inside a regex *literal* fails at
- * parse time on WebKit older than Safari 16.4, which takes down the whole chunk
- * rather than just this rule.
- *
- * Rewriting the lookbehind as a consuming group such as `(^|[^$])` is not
- * equivalent: the preceding character becomes part of the match, and the input
- * rule range is derived from the match length, so that character gets replaced
- * along with the math. A function finder can look at the preceding character
- * without consuming it.
+ * A negative lookbehind would crash old WebKit, and a consuming group like `(^|[^$])`
+ * would get replaced along with the math, so reject the preceding `$` in code instead.
  */
 const findInlineMath = (text: string): InputRuleMatch | null => {
   const pattern = /\$\$([^$\n]+?)\$\$(?!\$)/g
@@ -87,13 +76,11 @@ const findInlineMath = (text: string): InputRuleMatch | null => {
   let match = pattern.exec(text)
 
   while (match !== null) {
-    // Stands in for a negative lookbehind rejecting a preceding `$`.
     if (match.index === 0 || text[match.index - 1] !== '$') {
       return { index: match.index, text: match[0], data: { latex: match[1] } }
     }
 
-    // Resume from the character after the rejected match rather than from its end:
-    // a valid match can begin inside a rejected one, at its trailing `$$`.
+    // A valid match can start inside a rejected one, at its trailing `$$`.
     pattern.lastIndex = match.index + 1
     match = pattern.exec(text)
   }
--- a/packages/extension-mathematics/src/extensions/InlineMath.ts
+++ b/packages/extension-mathematics/src/extensions/InlineMath.ts
@@ -1,3 +1,4 @@
+import type { InputRuleMatch } from '@tiptap/core'
 import { InputRule, mergeAttributes, Node } from '@tiptap/core'
 import type { Node as PMNode } from '@tiptap/pm/model'
 import katex, { type KatexOptions } from 'katex'
@@ -63,6 +64,41 @@ declare module '@tiptap/core' {
       updateInlineMath: (options?: { latex?: string; pos?: number }) => ReturnType
     }
   }
+}
+
+/**
+ * Finds the `$$latex$$` sequence that the inline math input rule should replace.
+ *
+ * This is a function finder rather than a regex literal because the rule has to
+ * reject a sequence preceded by a third `$`, and there is no way to express that
+ * in a regex without a lookbehind. A lookbehind inside a regex *literal* fails at
+ * parse time on WebKit older than Safari 16.4, which takes down the whole chunk
+ * rather than just this rule.
+ *
+ * Rewriting the lookbehind as a consuming group such as `(^|[^$])` is not
+ * equivalent: the preceding character becomes part of the match, and the input
+ * rule range is derived from the match length, so that character gets replaced
+ * along with the math. A function finder can look at the preceding character
+ * without consuming it.
+ */
+const findInlineMath = (text: string): InputRuleMatch | null => {
+  const pattern = /\$\$([^$\n]+?)\$\$(?!\$)/g
+
+  let match = pattern.exec(text)
+
+  while (match !== null) {
+    // Stands in for a negative lookbehind rejecting a preceding `$`.
+    if (match.index === 0 || text[match.index - 1] !== '$') {
+      return { index: match.index, text: match[0], data: { latex: match[1] } }
+    }
+
+    // Resume from the character after the rejected match rather than from its end:
+    // a valid match can begin inside a rejected one, at its trailing `$$`.
+    pattern.lastIndex = match.index + 1
+    match = pattern.exec(text)
+  }
+
+  return null
 }
 
 /**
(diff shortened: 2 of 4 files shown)


.changeset/2026-08-23-inline-math-lookbehind.md
```diff
@@ -0,0 +1,5 @@
+---
+'@tiptap/extension-mathematics': patch
+---
+
+The inline math input rule no longer uses a regex lookbehind, which WebKit older than Safari 16.4 rejects at parse time, taking down every chunk the extension is bundled into.
```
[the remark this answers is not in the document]
giaBaoJS: Reworded in d12c6ec, but not to the sentence you proposed.

"Inline math input now works in older WebKit browsers" understates it. The lookbehind is in a regex *literal*, so the engine rejects it while parsing the file, not while running the rule. Nothing in the chunk `@tiptap/extension-mathematics` is bundled into ever executes. #6727 is titled "Mathematics extension would crash TipTap in older WebKit environments" and reports `Invalid regular expression: invalid group specifier name` from WKWebView. The editor is gone, not just the input rule.

There is precedent for saying so: `packages/extension-mathematics/CHANGELOG.md:745` for #6731 reads "Fixed an issue with the mathematics regex using modern negative lookups causing crashes in older Safari versions."

New text:

> The mathematics extension no longer crashes the editor in older WebKit browsers and WKWebView.

I left out a version number on purpose. I can source the failure mode and the affected browser family from #6727, but not a precise cutoff, so I would rather not put one in a changelog.

packages/extension-mathematics/__tests__/lookbehind.spec.ts
```diff
@@ -0,0 +1,26 @@
...
+
+/**
+ * A lookbehind inside a regex *literal* is rejected at parse time by WebKit older
+ * than Safari 16.4, which takes down the entire chunk the module is bundled into
+ * rather than only the code path that uses the regex. This package has regressed
+ * on that twice already, so pin it rather than relying on review to catch it.
+ *
+ * The scan is deliberately coarse — it looks at raw source text and cannot tell a
+ * regex literal from a comment, so the sequence must not appear in prose either.
+ */
```
bdbch: please unslop this comment / shorten it to 1-2 lines and make it easier for a human to read.

packages/extension-mathematics/src/extensions/InlineMath.ts
```diff
@@ -65,6 +66,41 @@ declare module '@tiptap/core' {
...
+ * in a regex without a lookbehind. A lookbehind inside a regex *literal* fails at
+ * parse time on WebKit older than Safari 16.4, which takes down the whole chunk
+ * rather than just this rule.
+ *
+ * Rewriting the lookbehind as a consuming group such as `(^|[^$])` is not
+ * equivalent: the preceding character becomes part of the match, and the input
+ * rule range is derived from the match length, so that character gets replaced
+ * along with the math. A function finder can look at the preceding character
+ * without consuming it.
+ */
```
bdbch: same here

bdbch (changes requested):
small nitpicks, otherwise we can go forward with this.

packages/extension-mathematics/__tests__/lookbehind.spec.ts
giaBaoJS: Cut to two lines in ba6dfb4.

packages/extension-mathematics/src/extensions/InlineMath.ts
giaBaoJS: Down to two lines. I also trimmed the other comments this PR adds, in the same commit.

giaBaoJS:
Both comments are down to two lines in ba6dfb44, and the changeset now just says the editor no longer crashes in older WebKit. Ready for another look whenever you have a moment.

giaBaoJS:
Rebased onto current `main`. One conflict, in the structural test only: the test migration moved this package's tests into `src/`, so `__tests__/lookbehind.spec.ts` is now `src/extensions/lookbehind.test.ts`. Its glob had to move with it (`../src/**/*.ts` to `../**/*.ts`), and since tests now live inside `src/` it skips `*.test.ts` / `*.spec.ts` so a future test may still write a lookbehind on purpose.

Verified the moved scan is not vacuous: with `InlineMath.ts` restored from `main` it fails with `expected [ './InlineMath.ts' ] to deeply equal []`, and adding a lookbehind to `src/utils.ts` flags `../utils.ts`, so it still covers the whole package source. Source change and behavioural tests are unchanged.

packages/extension-mathematics/src/extensions/InlineMath.ts
```diff
@@ -65,6 +66,28 @@ declare module '@tiptap/core' {
   }
 }
 
+/**
+ * A negative lookbehind would crash old WebKit, and a consuming group like `(^|[^$])`
+ * would get replaced along with the math, so reject the preceding `$` in code instead.
+ */
+const findInlineMath = (text: string): InputRuleMatch | null => {
+  const pattern = /\$\$([^$\n]+?)\$\$(?!\$)/g
```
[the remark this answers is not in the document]
giaBaoJS: Good catch on the range math, but I do not think this PR changes it.

`inputRuleMatcherHandler` sends a regex finder straight through `find.exec(text)` with no end anchor, so the literal that was there before behaves the same way on the same input:

```
"$$x$$y"
  before  /(?<!\$)(\$\$([^$\n]+?)\$\$)(?!\$)/   ->  "$$x$$" at 0, ends at cursor: false
  after   findInlineMath                        ->  "$$x$$" at 0, ends at cursor: false
```

On the path that actually fires the rule, where the closing `$` is the character just typed, both end at the cursor and the range is correct.

Whether the finder should reject a match that does not end at the cursor is a fair question on its own, since `range` in `InputRule.ts` is `from - (match[0].length - text.length)` and that only holds when the match ends there. I left it out because it is a behaviour change on top of dropping the lookbehind, and this PR is meant to be about the WebKit parse failure. Happy to add the guard here if you would rather have it in one go.

giaBaoJS:
Both nitpicks are addressed on the current head (f040271). The commits I named in my earlier replies (ba6dfb44, d12c6ec) were rebased away, so the two comments now live at:

- `packages/extension-mathematics/src/extensions/InlineMath.ts:70-71`
- `packages/extension-mathematics/src/extensions/lookbehind.test.ts:3-4`

The test path in your review no longer resolves because the test migration moved this package's tests from `__tests__/` into `src/`. Each comment is two lines now and gives the reason the code is shaped that way instead of restating it.

Nothing else changed since your review apart from that rebase. CI is green, and the mathematics package is 3 files / 12 tests locally.

PR #8251, opened 2026-08-25
chore(vue): remove vue-2 package

bdbch:
## Fixes

- N/A

## Changes and Review

This PR drops the `@tiptap/vue-2` and it's relevant packages (drag-handle) from the repository & renames `@tiptap/vue-3` to `@tiptap/vue`. With Tiptap v4 we want to deprecate our Vue 2 support. [Read more here](https://v2.vuejs.org/eol/).

It also adds a new codemod package for the Tiptap 3 to 4 migration which will help our users to make the migration between majors a lot easier.

**Info**:

Most changed files in this PR are deletions (of vue-2 and vue-3) and additions (the addition of the new vue package).

To make the review easier you should make sure that package names on the new vue package are correct, that the configuration files etc. are fine, etc.

No need to go through all vue-3 and vue-2 deletions or all vue additions.

## Checklist

### Responsibility


arnaugomez (commented):
What if we leave this change to Tiptap 4? This way, developers don't need to migrate twice: once to migrate from @tiptap/vue-3 to @tiptap/vue, and another to migrate to Tiptap 4

bdbch:
This is for Tiptap 4. Main is already our "next" branch now.

PR #8254, opened 2026-08-25
Do not use globalThis to reference the DOM Node interface

arnaugomez:
## Changes and Review

After renaming ProseMirror Node class to PMNode, it is not necessary to reference the DOM Node interface as `globalThis.Node`, instead it can be referenced as just `Node`

This is a refactor, and it does not change the behaviour of the code, so it does not require a changeset 

## Checklist

### Responsibility


bdbch (commented):
Does this need a changeset?

bdbch (approved):

arnaugomez:
> Does this need a changeset?

No, it is a refactor and it has no change in functionality.

PR #8262, opened 2026-08-26
fix(ci): filter released packages before passing through to mapping

bdbch:
## Fixes

- the currently broken ci for publishing new packages 🙃

## Changes and Review

The CI fails because the new codemod package is not providing the metadata the changelog aggregator expects - so we just filter out non-valid packages now instead of throwing

## Checklist

### Responsibility

### DIFF
--- a/scripts/aggregate-changeset.js
+++ b/scripts/aggregate-changeset.js
@@ -374,12 +374,12 @@ async function main() {
 
     const packageMap = await loadWorkspacePackages(cwd)
     const releasesWithPackages = plan.releases
+      .filter(release => {
+        const pkg = packageMap.get(release.name)
+        return !!pkg
+      })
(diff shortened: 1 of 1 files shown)


alexvcasillas (approved):

arnaugomez (approved):

scripts/aggregate-changeset.js
```diff
@@ -374,13 +374,13 @@ async function main() {
 
     const packageMap = await loadWorkspacePackages(cwd)
     const releasesWithPackages = plan.releases
+      .filter(release => {
+        const pkg = packageMap.get(release.name)
+        return !!pkg
```
arnaugomez, line 379: No need to convert pkg to a boolean. The filter function already coerces the return type to a boolean.

PR #8277, opened 2026-08-28
add new visual selection handler & refactored react node view selection handling

bdbch:
## Fixes

- refactors how React NodeViews create their `selected` state

## Changes and Review

See above. The NodeView solution was pretty straight forward. Instead of doing this **per nodeview** we moved this up into the React editor class which now has a nodeview selection manager that checks potential selection states once & then gives that prop down to the right nodeview.

## Checklist

- [x] This is a critical bug or security fix that also needs to land on the current stable release (see [Branching](../CONTRIBUTING.md#branching) in CONTRIBUTING.md).

### Responsibility


packages/core/src/VisualSelectionStore.ts
```diff
@@ -0,0 +1,138 @@
...
+   * @returns A function that unsubscribes the callback.
+   */
+  subscribe = (callback: () => void): (() => void) => {
+    this.subscribers.add(callback)
+    return () => {
+      this.subscribers.delete(callback)
+    }
+  }
+
+  destroy(): void {
```
alexvcasillas: Other simple methods like `clear` or `subscribe` have `JSDoc` but this one lacks of it 🤔 is intentional?

packages/core/src/VisualSelectionStore.ts
```diff
@@ -0,0 +1,138 @@
...
+    this.editor.off('transaction', this.handleTransaction)
+    this.editor.off('selectionUpdate', this.handleSelectionUpdate)
+    this.subscribers.clear()
+  }
+
+  private notify(snapshot: VisualSelectionSnapshot | null): void {
```
alexvcasillas: Same here about the `JSDoc` 🤔

packages/core/src/VisualSelectionStore.ts
```diff
@@ -0,0 +1,138 @@
...
+  getSnapshot = (): VisualSelectionSnapshot | null => {
+    return this.snapshot
+  }
+
+  /**
+   * Marks the node at `pos` as visually selected.
+   *
+   * @param pos The absolute position of the node to select.
+   */
+  set(pos: number): void {
```
alexvcasillas: How do you expect this to be called? The store clears on any real selection change, so a click that sets a `NodeSelection` would wipe it if `set()` lands first 🤔 Wondering if the ordering needs documenting 🤔

packages/core/src/VisualSelectionStore.ts
```diff
@@ -0,0 +1,138 @@
...
+    this.snapshot = snapshot && Object.freeze(snapshot)
+    this.subscribers.forEach(callback => callback())
+  }
+
+  // Avoid a stale visual selection next to a new real one.
+  private handleSelectionUpdate = () => {
+    this.clear()
+  }
+
+  private handleTransaction = ({ transaction }: EditorEvents['transaction']) => {
```
alexvcasillas: This maps through the root transaction only, so appended transactions get skipped. I tested it with a plugin whose `appendTransaction` inserts 4 characters: the snapshot stayed at `{ pos: 9, nodeSize: 8 }` while the paragraph had moved to 13, and `pos` 9 landed inside a text node. Any `appendTransaction` that changes the doc without moving the selection hits this, so collaboration and trailing node too.

Could we combine first, the way `extensions/delete.ts` does?

```ts
combineTransactionSteps(transaction.before, [transaction, ...appendedTransactions])
```

Two things come with it: `transaction.doc` on line 128 is the intermediate doc, so we would read the node from `this.editor.state.doc`, and the `!transaction.docChanged` guard misses the case where only the appended transaction touched the doc.

packages/core/src/VisualSelectionStore.ts
```diff
@@ -0,0 +1,138 @@
...
+    if (unchanged) {
+      return
+    }
+
+    this.snapshot = snapshot && Object.freeze(snapshot)
+    this.subscribers.forEach(callback => callback())
+  }
+
+  // Avoid a stale visual selection next to a new real one.
+  private handleSelectionUpdate = () => {
```
alexvcasillas: `transaction` fires before `selectionUpdate` (`Editor.ts:710` and `:717`), so a doc change that also moves the cursor notifies twice: first the remapped position, then `null` from here. I measured 2 calls. On `useSyncExternalStore` that is two renders and the first one is thrown away.

Could `handleTransaction` skip the notify when the transaction changes the selection as well? 🤔

packages/core/src/VisualSelectionStore.ts
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  /**
+   * The current visual selection, or `null` if nothing is visually selected.
+   * Returns the same object reference until the selection changes, so it's
+   * safe to use as a `useSyncExternalStore` snapshot.
+   *
+   * @returns The current visual selection.
+   */
+  getSnapshot = (): VisualSelectionSnapshot | null => {
```
alexvcasillas: `notify` freezes the object but this returns the mutable `VisualSelectionSnapshot`. Should `pos` and `nodeSize` be `readonly` so the type matches? Right now a consumer can assign to them and it silently does nothing.

packages/react/src/ReactNodeViewSelectionManager.ts
```diff
@@ -0,0 +1,67 @@
...
+
+import type { ReactNodeView } from './ReactNodeViewRenderer.js'
+
+/**
+ * One selection listener per editor, instead of one per node view.
+ */
+class ReactNodeViewSelectionManager {
+  private editor: Editor
+
+  private views = new Set<ReactNodeView<any>>()
```
alexvcasillas: Could we drop the `any`? The manager only calls one method, so `Set<{ checkSelection: () => void }>` would type it without the generic.

packages/core/src/Editor.ts
```diff
@@ -825,6 +826,21 @@ export class Editor extends EventEmitter<EditorEvents> {
...
   }
 
+  private _visualSelection: VisualSelectionStore | null = null
+
+  /**
+   * A store for marking a node as selected without dispatching a transaction.
+   *
+   * @returns The shared `VisualSelectionStore` for this editor.
+   */
+  public get visualSelection(): VisualSelectionStore {
```
alexvcasillas: `destroy()` resets `_visualSelection` to `null`, so reading this afterwards builds a fresh store and calls `editor.on(...)` on a destroyed editor. Worth a guard here?

alexvcasillas (dismissed):
Looks good to me ✅

I've added a couple questions that are non-blocking and would love to hear your thoughts on them 🙇🏻

packages/core/src/VisualSelectionStore.ts
bdbch: I just added JSDocs to public API that users will see. Tiptap users should usually not call destroy on the visual selection manager itself manually. I can add a JSDoc for internal developers but I wanted to avoid unnecessary comment bloat.

packages/core/src/VisualSelectionStore.ts
bdbch: See above, notify is called internally (in contrast to subscribe which is the user-facing function to get notifications).

packages/core/src/VisualSelectionStore.ts
bdbch: This is wanted. The usecase is that you want to set a visual selection on an element without a selection change on the editor itself. Example: You click on an element, prevent default & call the editors `editor.visualSelection.set()` to set the visual representation of an selection.

I think this will not be relevant to a lot of users. This is an escape hatch for users with deeply-nested nodes that want to handle active/selection states on nodes without causing transactions that potentially update multiple hundred of nodes (which is quite slow).

bdbch:
After thinking about it, I think I'll remove the visual selection manager. Feels out of place & leaks to much into user space. If users want to add custom visual selections, they should manage them themselves.

The only time this is really relevant is for the rare case that a user is putting a LOT of prosemirror nodes nested into each other.

PR #8283, opened 2026-08-30
fix(pm): bump prosemirror-view to ^1.42.3 to patch paste XSS

lazerg:
## Fixes

Fixes #8282

## Changes and Review

`prosemirror-view` before 1.42.3 has a high severity XSS (GHSA-c8x8-7fp4-3x9w): pasting attacker-supplied HTML into the editor can run arbitrary JavaScript, and there is no workaround other than upgrading.

- Raise the `prosemirror-view` floor in `@tiptap/pm` and the demos app from `^1.41.9` to `^1.42.3`, so consumers get the patched version.
- Add a `prosemirror-view` override in `pnpm-workspace.yaml`, same as the existing `prosemirror-model` one, so the published `@tiptap/pm` copies pulled in as dev dependencies stop resolving 1.42.2 as well. The lockfile now has no pre-1.42.3 copy left.
- To verify: `pnpm build` and `pnpm test:unit` pass (1956/1956), and `grep prosemirror-view@1.42.2 pnpm-lock.yaml` returns nothing.

Backport onto `maintenance/v3`: #8284

## Checklist

- [x] I have added a [changeset](https://github.com/changesets/changesets) if necessary.
- [ ] I have added tests if possible.
- [x] I have made sure to test my changes myself.
- [x] This is a critical bug or security fix that also needs to land on the current stable release (see [Branching](../CONTRIBUTING.md#branching) in CONTRIBUTING.md).

### Responsibility

- [x] I have reviewed and understand these changes, and I take responsibility for this PR, even if an AI agent created it.

Issue #8282, opened 2026-08-30
prosemirror-view XSS pre-1.42.3


Nantris:
### Affected Packages

pm

### Tiptap Version

Latest

### Browser Used

Chrome

### What happened?

https://discuss.prosemirror.net/t/xss-vulnerability-found-in-prosemirror-view/9067

### Expected Behavior

Upgrade `prosemirror-view`

PR #8284, opened 2026-08-30
fix(pm): bump prosemirror-view to ^1.42.3 to patch paste XSS

lazerg:
## Fixes

Backport of #8283 onto `maintenance/v3`. Related to #8282.

## Changes and Review

`prosemirror-view` before 1.42.3 has a high severity XSS (GHSA-c8x8-7fp4-3x9w): pasting attacker-supplied HTML into the editor can run arbitrary JavaScript, and there is no workaround other than upgrading. `maintenance/v3` publishes `latest`, so the fix has to land here too.

- Raise the `prosemirror-view` floor in `@tiptap/pm` and the demos app from `^1.41.9` to `^1.42.3`.
- Add a `prosemirror-view` override in `pnpm-workspace.yaml`, same as the existing `prosemirror-model` one, so nothing in the lockfile resolves a pre-1.42.3 copy.
- To verify: `pnpm build` and `pnpm test:unit` pass (1966/1966), and `grep 1.42.2 pnpm-lock.yaml` returns nothing.

## Checklist

- [ ] I have added tests if possible.

### Responsibility

### DIFF
--- a/.changeset/2026-08-31-bump-prosemirror-view.md
+++ b/.changeset/2026-08-31-bump-prosemirror-view.md
@@ -1,0 +1,5 @@
+---
+'@tiptap/pm': patch
+---
+
+Bump `prosemirror-view` to `^1.42.3`, which fixes an XSS vulnerability where pasting crafted HTML could run arbitrary JavaScript (GHSA-c8x8-7fp4-3x9w).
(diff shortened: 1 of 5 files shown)


lazerg:
The lint failure is in `renderNestedMarkdownContent.spec.ts` and `list-indentation.spec.ts`, which this PR does not touch. Those two already fail on `maintenance/v3`, and it is the same `vitest` import rule that #8276 fixed on `main`.

bdbch (approved):

PR #8289, opened 2026-08-31
Refactor React node view selection tracking

bdbch:
## Fixes

- N/A

## Changes and Review

This PR refactors selection handling on React NodeViews by using Prosemirrors selected state instead of inferring it ourselves. Also introduces a new `selectionInside` prop which users can use to check if the selection is fully inside this nodeview.

The goal of the PR to heavily improve the performance in larger React based editors when selection changes.

## Checklist

- [x] This is a critical bug or security fix that also needs to land on the current stable release (see [Branching](../CONTRIBUTING.md#branching) in CONTRIBUTING.md).

### Responsibility


.changeset/2026-08-31-react-nodeview-selection-state.md
```diff
@@ -0,0 +1,5 @@
+---
+'@tiptap/react': minor
```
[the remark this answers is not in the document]
bdbch: @coderabbitai I will mark the new selectionInside option optional to make this a non-breaking change.

packages/react/src/ReactNodeViewRenderer.tsx
```diff
@@ -5,21 +5,26 @@ import type {
...
 import { ReactRenderer } from './ReactRenderer.js'
 import type { ReactNodeViewProps } from './types.js'
 import type { ReactNodeViewContextProps } from './useReactNodeView.js'
 import { ReactNodeViewContext } from './useReactNodeView.js'
 
 export interface ReactNodeViewRendererOptions extends NodeViewRendererOptions {
+  /**
+   * @deprecated Read `selectionInside` from the node view props instead.
+   */
+  selectedOnTextSelection?: boolean
```
arnaugomez, line 27: If this property is deprecated, why is it a new property?

packages/react/src/types.ts
```diff
@@ -3,4 +3,6 @@ import type React from 'react'
 
 export type ReactNodeViewProps<T = HTMLElement> = CoreNodeViewProps & {
   ref: React.RefObject<T | null>
+  /** Whether a text selection is fully inside the node view. Always provided by ReactNodeViewRenderer. */
+  selectionInside?: boolean
```
arnaugomez, line 7: If this property is always provided, why is it optional?

arnaugomez (approved):

packages/react/src/ReactNodeViewRenderer.tsx
bdbch: It's not new, the option was there previously.

packages/react/src/types.ts
bdbch: Because it was not there previously, users may not have this provided on their versions of Tiptap - making this required would cause this to be a breaking change.

PR #8295, opened 2026-09-01
fix(extension-table): escape pipe characters in serialized table cells

### LINKED ISSUE #8294: [Bug] @tiptap/markdown: escaped pipe in a table cell is lost on serialize, then destroys cell content on the next round trip
## Description

A GFM table cell containing an escaped pipe (`\|`) round-trips destructively through `@tiptap/markdown`: parsing reads the cell correctly, but serialization emits the pipe unescaped, so the next parse splits the cell at that pipe and the content after it is silently lost.

## Affected versions

- `@tiptap/markdown` 3.31.0
- `@tiptap/extension-table` 3.31.0
- `@tiptap/starter-kit` 3.31.0
- Node 20.20.2, no DOM involved

## Minimal reproduction

```ts
import { MarkdownManager } from "@tiptap/markdown";
import { TableKit } from "@tiptap/extension-table";

const manager = new MarkdownManager({ extensions: [StarterKit, TableKit] });

const source = "| Col | Value |\n| --- | --- |\n| a | uses \| pipe |";

const pass1 = manager.serialize(manager.parse(source));
const pass2 = manager.serialize(manager.parse(pass1));

console.log(pass1);
console.log(pass2);
```

## Actual output

Pass 1 (the escape is gone, so the cell markup is now broken):

```
| Col | Value       |
| --- | ----------- |
| a   | uses | pipe |
```

Pass 2 (the broken markup now parses as an extra column and the text after the pipe is gone):

```
| Col | Value |
| --- | ----- |
| a   | uses  |
```

Walking the parsed documents' text nodes shows the data loss happens at the pass-1 parse already: the source parses to cell text `uses | pipe`, but re-parsing pass 1 yields only `uses`.

## Expected

`serialize` should re-escape a literal `|` inside table cell text as `\|` (the GFM spelling the parser itself understands), so `serialize(parse(x))` reaches a fixpoint instead of losing content.

## Notes on the likely cause

`MarkdownManager.escapeMarkdownSyntax` escapes ``[\`*_[\]~]`` but not `|`, and the table serialization path does not appear to apply `escapeTableCellPipes` (exported by `@tiptap/extension-table`) to text children, so cell text containing a literal pipe is emitted raw.

Thanks for the markdown work in v3; outside this edge it has held up very well under round-trip testing.

lazerg:
## Fixes

Fixes #8294

## Changes and Review

A literal `|` in a table cell was written out raw, so re-parsing the output split the cell at that pipe and dropped everything after it. Cell text is now pipe-escaped when the table is serialized, with existing backslash escapes left alone so nothing gets escaped twice.

To verify, parse `| a | uses \| pipe |` and serialize it: the pipe stays escaped and a second round trip gives the same output. Covered by unit tests in `packages/extension-table/__tests__/tableMarkdown.spec.ts`.

## Checklist

### Responsibility

### DIFF
--- a/.changeset/2026-09-02-table-cell-pipe-escape.md
+++ b/.changeset/2026-09-02-table-cell-pipe-escape.md
@@ -1,0 +1,5 @@
+---
+'@tiptap/extension-table': patch
+---
+
+Table cells exported to Markdown now escape literal pipe characters, so the cell content survives when the output is read back.
--- a/packages/extension-table/__tests__/tableMarkdown.spec.ts
+++ b/packages/extension-table/__tests__/tableMarkdown.spec.ts
@@ -298,3 +298,34 @@ describe('escapeTableCellPipes — pipe escaping inside code spans', () => {
     expect(escapeTableCellPipes('| `|\\||` |')).toBe('| `\\|\\|\\|` |')
   })
 })
+
+describe('table markdown — pipe escaping when serializing cells', () => {
+  const manager = new MarkdownManager({
+    extensions: [Document, Paragraph, Text, Code, TableKit],
+  })
+
+  it('escapes a bare pipe in cell text', () => {
+    const markdown = '| Col | Value |\n| --- | --- |\n| a | uses \\| pipe |'
+    const serialized = manager.serialize(manager.parse(markdown))
+
+    expect(serialized).toContain('uses \\| pipe')
+  })
+
+  it('round trips a cell with an escaped pipe', () => {
+    const markdown = '| Col | Value |\n| --- | --- |\n| a | uses \\| pipe |'
+    const pass1 = manager.serialize(manager.parse(markdown))
+    const pass2 = manager.serialize(manager.parse(pass1))
+
+    expect(pass2).toBe(pass1)
+
+    const cell = manager.parse(pass1).content?.[0]?.content?.[1]?.content?.[1]
+    expect(cell?.content?.[0]?.content?.[0]?.text).toBe('uses | pipe')
+  })
+
+  it('escapes pipes inside a code span in a cell', () => {
+    const markdown = '| H |\n| - |\n| `a || b` |'
+    const serialized = manager.serialize(manager.parse(markdown))
+
+    expect(serialized).toContain('`a \\|\\| b`')
+  })
+})
--- a/packages/extension-table/__tests__/tableMarkdown.spec.ts
+++ b/packages/extension-table/__tests__/tableMarkdown.spec.ts
@@ -304,27 +304,27 @@ describe('table markdown — pipe escaping when serializing cells', () => {
     extensions: [Document, Paragraph, Text, Code, TableKit],
   })
 
+  const markdown = '| Col | Value |\n| --- | --- |\n| a | uses \\| pipe |'
+
   it('escapes a bare pipe in cell text', () => {
-    const markdown = '| Col | Value |\n| --- | --- |\n| a | uses \\| pipe |'
     const serialized = manager.serialize(manager.parse(markdown))
 
     expect(serialized).toContain('uses \\| pipe')
   })
 
   it('round trips a cell with an escaped pipe', () => {
-    const markdown = '| Col | Value |\n| --- | --- |\n| a | uses \\| pipe |'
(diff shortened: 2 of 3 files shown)


solomonaustin:
I hit this independently with a round-trip audit and was about to file it before finding #8294 — yours was first, so this is just verification. I checked the branch out, merged current `main`, and ran it. Three things.

**1. The fix is still needed, and it works.** `main` already has `escapeTableCellPipes`, but its doc comment scopes it to code spans: *"escape any `|` characters that appear inside backtick code spans"*. A bare or escaped pipe in ordinary cell text is still lost on `main` today:

```
input:        | x \| y | 2 |
first parse:  [["a","b"],["x | y","2"]]
serialized:   | x | y | 2   |        <- escape dropped, cell splits
second parse: [["a","b"],["x","y"]]  <- the "2" is gone
```

With your branch merged onto `main`, that round-trips correctly:

```
serialized:   | x \| y | 2   |
second parse: [["a","b"],["x | y","2"]]   STABLE
```

**2. It is currently `CONFLICTING`, and the cause is a rename rather than your code.** The branch is 24 commits behind. `main` moved `packages/extension-table/__tests__/tableMarkdown.spec.ts` to `packages/extension-table/tests/integration/tableMarkdown.test.ts`, so that one file is the only conflict. Everything in `src/` merges cleanly.

**3. Resolving it naively leaves 5 failing tests, which is the part worth knowing before you rebase.** After taking both sides you get:

```
FAIL  tests/integration/tableMarkdown.test.ts
  × escapes a bare pipe inside a code span
  × escapes every pipe of a consecutive run
  × keeps an already-escaped pipe untouched
  × keeps a pipe preceded by a backslash pair untouched
  × handles mixed escaped and bare pipes in one span

ReferenceError: escapeTableCellPipes is not defined
```

The reason is that `main` extracted those helper tests into a co-located unit test at `packages/extension-table/src/table/utilities/markdown.test.ts`, with byte-identical assertions to the `escapeTableCellPipes` block in your integration file. Your branch predates that, so after the merge the block is duplicated and has lost its import.

**Suggested resolution:** drop the `escapeTableCellPipes` describe block from the integration file — `main` covers it in the co-located unit test now — and keep your serialize/round-trip tests, which are the ones that are genuinely new. With exactly that resolution the package is green:

```
Test Files  11 passed (11)
     Tests  95 passed (95)
```

Nothing here is a problem with the fix itself. It is drift from a test-layout change that landed after you opened this. I hit the same thing on two of my own PRs last week, so it seems to be catching several open branches.

solomonaustin:
Since the blocker here is drift rather than anything in your change, I did the rebase and pushed it to my fork so you can take it directly instead of redoing it:

**`solomonaustin/tiptap` branch `help/8295-rebased`**

```bash
git remote add solomonaustin https://github.com/solomonaustin/tiptap.git
git fetch solomonaustin help/8295-rebased
git checkout fix/table-cell-pipe-escape        # your branch
git reset --hard solomonaustin/help/8295-rebased
git push --force-with-lease
```

It is your commits with `main` merged in, plus one commit removing the `escapeTableCellPipes` describe block that `main` now covers in the co-located unit test at `src/table/utilities/markdown.test.ts`. Your `src/` change and your round-trip tests are untouched.

Result against current `main`:

```
vp check       All 2042 files formatted, 0 lint errors
extension-table   11 files, 95 tests passed
full unit suite   200 files, 1975 tests passed
```

Diff is 3 files, +43/-1 — the changeset, `src/table/utilities/markdown.ts`, and the integration test.

Take it, cherry-pick just the dedup commit, or ignore it entirely and rebase yourself — whatever is least annoying. I only pushed it because I had already done the merge locally while verifying the fix, and it seemed wasteful to describe the resolution without handing it over. The PR is yours; I am not going to open a competing one.

lazerg:
Thanks for double-checking and for the offer. I'd already rebased onto main and pushed (c4480e8), same dedup approach you found. Should be unblocked now, just waiting on a maintainer review.

arnaugomez (approved):

arnaugomez:
@lazerg thanks for your contribution. We'd love to know your feedback about Tiptap. What feature or functionality could Tiptap provide that would be game-changing for you?

arnaugomez (approved):

PR #8315, opened 2026-09-04
fix(markdown): keep marks on inline atom nodes

### LINKED ISSUE #8314: Markdown: marks are dropped on inline atom nodes, so emphasis around them is lost on round trip
### Affected Packages

@tiptap/markdown

### Tiptap Version

3.30.2 (also reproduced on 3.31.3)

### Browser Used

Not browser-specific

### What happened?

When an inline atom node sits inside `**...**`, the mark is dropped while parsing. The emphasis is then missing when the document is serialized back to Markdown.

`MarkdownManager.applyMarkToContent` applies a mark only when `node.type === "text"`. For any other node it recurses into `node.content`, which an atom node does not have, so no mark is applied and nothing errors.

Minimal reproduction with a small inline atom node:

```ts
const Chip = Node.create({
  name: 'chip',
  inline: true,
  group: 'inline',
  atom: true,
  addAttributes: () => ({ label: { default: '' } }),
  markdownTokenName: 'chip',
  markdownTokenizer: {
    name: 'chip',
    level: 'inline',
    start: (src) => src.indexOf('[['),
    tokenize(src) {
      const m = /^\[\[([^\]]+)\]\]/.exec(src);
      return m ? { type: 'chip', raw: m[0], label: m[1] } : undefined;
    },
  },
  parseMarkdown: (t) => ({ type: 'chip', attrs: { label: t.label ?? '' } }),
  renderMarkdown: (n) => `[[${n.attrs?.label ?? ''}]]`,
});

const m = new MarkdownManager({ extensions: [StarterKit, Chip] });
m.serialize(m.parse(source));
```

| Source | Parsed | Serialized |
| --- | --- | --- |
| `[[a]]` | chip | `[[a]]` |
| `**[[a]]**` | chip, no marks | `[[a]]` |
| `**x [[a]] y**` | text(bold), chip (no marks), text(bold) | `**x** [[a]] **y**` |

There is a related case on the serializer side. If a document is built where an inline atom node does carry a mark, `serialize()` closes the mark before the node and reopens it after:

```
paragraph[ text("bold ", bold), chip(marks: [bold]), text(" more", bold) ]
  ->  **bold **[[a]]** more**
```

The space before the closing `**` means that output no longer renders as bold when it is parsed again.

### Expected Behavior

A mark spanning an inline atom node survives a parse/serialize round trip, so `**[[a]]**` comes back as `**[[a]]**`.

For reference, `prosemirror-markdown` handles both directions: `addNode()` passes the currently open marks to any node type, and `renderInline()` reads `node.marks` for every inline child regardless of its type.

I looked for an existing report and did not find one, though I may well have missed it. The closest I found were #7827 and #7409, which look like the opposite case — extra marks appearing after an atom — and appear to be fixed.

Happy to help if it is useful: I can put together a failing test case, or take a run at a PR. Just let me know what you would prefer, and whether you would want the parse and serialize sides handled together.

lazerg:
## Fixes

Fixes #8314

## Changes and Review

`applyMarkToContent` only added a mark to `text` nodes and otherwise recursed into `node.content`, which an inline atom does not have, so `**[[chip]]**` parsed to a chip with no marks. It now marks any leaf node and keeps recursing into nodes that have content.

On the serializer side, a marked non-text node closed every open mark and reopened them after itself, producing `**bold **[[chip]]** more**`, which no longer reads as bold when parsed again. Marks the node carries now stay open across it, and a mark carried only by the atom is opened and closed around it.

Verified with new tests in `packages/markdown/tests/integration/inline-atom-marks.test.ts`, covering both directions and the round trip. They fail on `main` and pass here.

## Notes for review

Two things I left alone on purpose, happy to change either:

- `closeMarksBeforeNode` and `reopenMarksAfterNode` in `packages/markdown/src/utils.ts` now have no callers. `src/index.ts` re-exports `./utils.js`, so they are public API and removing them would be a breaking change. Say the word and I will drop them in a separate major-bump PR.
- A mark that spans an atom which carries only part of the surrounding mark set still serializes to broken markdown, for example bold+italic text around a bold-only atom gives `***a*[chip]*b***`. The same shape built from plain text nodes produces the identical output on `main`, so this is an existing delimiter-run limitation of the text path rather than something this change introduces. Fixing it means using the HTML fallback when an inner mark closes while an outer one stays open, which felt out of scope here.

## Checklist

### Responsibility


jheddings:
Thanks for writing this — I filed #8314 for the same defect a few days ago, before I saw this PR. Happy to add a data point from hitting it in an app.

Your diagnosis matches what I found: the mark is dropped at parse time in `applyMarkToContent`, and the serializer is broken symmetrically. The second half is easy to miss, so it seems worth confirming you've covered it — fixing only the parse side turns a silent drop into malformed output, because a marked atom serializes as `**bold **[[chip]]** more**` and the space before the closing `**` voids the emphasis.

Two things from the app side, in case they're useful for test cases:

- A mark reaches the node by two independent routes. Parsing `**[[chip]]**` from markdown goes through `applyMarkToContent`; applying bold from the toolbar marks the node directly in ProseMirror and never touches the parser. They failed separately for me, so both are worth covering.
- The initial content path is easy to miss. `Markdown.onBeforeCreate` parses `editor.options.content` in the same hook that creates the manager, so a document supplied at construction takes a different code path from one supplied later via `setContent`. My first fix worked for `setContent` and not for construction.

In our case the affected node is a Handlebars `{{...}}` expression in a message template, so losing the emphasis silently changed stored content when a template was merely opened and saved.

We're working around it for now by degrading the node back to a text node whenever a mark reaches it — the formatting survives and the node representation is what's given up. That's a workaround for our specific case rather than something general, and I'd much rather drop it in favour of this.

No obligation, but I'm glad to test this branch against our reproduction if that would help — just say the word.

lazerg:
Both sides are fixed. The parse side is `applyMarkToContent`. The serialize side is the non-text branch of `renderNodesWithMarkBoundaries`. It now keeps a mark open across an inline atom, and it opens a mark the atom carries on its own, instead of closing every active mark before the node and reopening after it. The malformed emphasis goes away with that.

Your two routes had no test, so I added four to `packages/markdown/tests/integration/inline-atom-marks.test.ts`:

- markdown given to the `Editor` constructor, which runs through `onBeforeCreate`
- the same markdown through `setContent`
- `setBold` on a selection that holds only the atom
- `setBold` on a selection that holds text and the atom

All four fail on `main` and pass on this branch. The command route never reaches the parser, and it works because the serializer reads the marks off the node.

Thanks for the note about construction time. That one was worth checking. You do not need to run your reproduction against the branch. The tests cover both routes and CI is green.

arnaugomez:
A linked, bold atom followed by bold-only text loses the following text’s bold formatting when saved and loaded as Markdown.

To reproduce in `inline-atom-marks.test.ts`, import `Link` from `@tiptap/extension-link`, add it to the manager’s extensions, and use the existing helpers:

```ts
const markdown = markdownManager.serialize(paragraph([
  atomNode([
    { type: 'link', attrs: { href: 'https://example.com' } },
    { type: 'bold' },
  ]),
  { type: 'text', text: ' tail', marks: [{ type: 'bold' }] },
]))
console.log(markdownManager.parse(markdown))
```

The generated Markdown is:

```md
[**[custom-inline type="bug" id="1234"]](https://example.com) tail**
```

Expected: `tail` stays bold. Actual: it becomes plain text with literal `**`; the base preserves its bold formatting.

The link closes while bold is still open, crossing their boundaries. Please order the atom’s marks by how long they continue, as the text path does, and add this regression test.

lazerg:
Fixed in 0a09acc. The atom branch now reuses `getMarksToOpenForSerialization`, the same ordering the text path uses. A mark that continues past the atom opens outside one that ends on it.

Your example now serializes to:

```md
**[[custom-inline type="bug" id="1234"]](https://example.com) tail**
```

Parsing that back keeps `tail` bold. I added the regression test to `inline-atom-marks.test.ts`.

lazerg:
I rebased on main, so that fix is now decfc36. Code marks keep closing and reopening around an atom, because a code span renders its content literally.

arnaugomez:
An inline atom can also have content, but this change only puts the mark on leaf nodes. For example, an extension with `inline: true`, `atom: true`, and `content: "inline*"` loses the node mark when parsing:

```ts
markdownManager.parse(
  "**[custom-inline-tag]Home[/custom-inline-tag]**",
)
```

This marks `Home` inside the atom instead of the atom itself, so serializing produces `[custom-inline-tag]**Home**[/custom-inline-tag]` rather than preserving the original wrapper. Please apply the mark to content-bearing inline atoms too.

arnaugomez:
A bold atom immediately after unmarked text can cause the following text to lose its bold formatting when saved and loaded as Markdown.

To reproduce in `inline-atom-marks.test.ts`, use the existing helpers:

```ts
const markdown = markdownManager.serialize(paragraph([
  { type: 'text', text: 'aaa' },
  atomNode([{ type: 'bold' }]),
  { type: 'text', text: 'bbb', marks: [{ type: 'bold' }] },
]))
console.log(markdownManager.parse(markdown))
```

The generated Markdown is:

```md
aaa**[custom-inline type="bug" id="1234"]bbb**
```

Expected: the atom and `bbb` stay bold. Actual: both are unmarked, and the asterisks become literal text. The base already loses the atom’s mark, but preserves bold on `bbb`; this change loses that formatting too.

The opening `**` sits between a letter and the atom’s opening `[`, so it cannot open Markdown emphasis. Please handle these boundaries with a valid fallback and add a regression test with no spaces around the atom.

lazerg:
Fixed in 848bc77. If an emphasis delimiter next to an atom would sit between a word and the atom's punctuation, the atom now drops that mark. Your example now serializes to:

```md
aaa[custom-inline type="bug" id="1234"]**bbb**
```

`bbb` stays bold, which is the same output as on `main`. The atom loses its bold only at such a boundary. I did not use `<strong>` as the fallback, because the parser then reads the atom back as plain text. The new test covers both sides, with no spaces around the atom.

PR #8319, opened 2026-09-06
fix(extension-table): stop leaking U+001F into table markdown

### LINKED ISSUE #8152: renderTableToMarkdown leaks the U+001F cell separator into serialized markdown
### Affected Packages

`@tiptap/extension-table`, `@tiptap/markdown`

### Tiptap Version

3.27.3

### Browser Used

N/A - reproduces in plain Node with `MarkdownManager`, no DOM involved.

### What happened?

Serializing a table whose cell contains more than one block child emits a raw `U+001F` (UNIT SEPARATOR) control character into the markdown output.

`renderTableToMarkdown` joins multi-block cells with `DEFAULT_CELL_LINE_SEPARATOR = '\u001F'` so the parts can be split apart again later, but nothing ever splits them: the value flows straight into `collapseWhitespace`, whose `/\s+/` does **not** match `U+001F`, and then into the emitted pipe row.

`packages/extension-table/src/table/utilities/markdown.ts`:

```ts
export const DEFAULT_CELL_LINE_SEPARATOR = '\u001F'

function collapseWhitespace(s: string) {
  return (s || '').replace(/\s+/g, ' ').trim()   // does not match U+001F
}
...
const parts = cellNode.content.map(child => h.renderChildren(child))
raw = parts.join(cellSep)
...
const text = collapseWhitespace(raw)
```

I grepped the package for another consumer of `cellSep` / `DEFAULT_CELL_LINE_SEPARATOR` and found none on the serialize path, so the separator appears to be vestigial.

Practical impact: any app that persists `editor.getMarkdown()` stores the control character. Ours ended up in a database and in diffs, invisible in most editors.

### Reproduction

```js
import { Table } from '@tiptap/extension-table'
import { TableRow } from '@tiptap/extension-table-row'
import { TableCell } from '@tiptap/extension-table-cell'
import { TableHeader } from '@tiptap/extension-table-header'

const m = new MarkdownManager({ extensions: [StarterKit, Table, TableRow, TableCell, TableHeader] })
const p = t => ({ type: 'paragraph', content: [{ type: 'text', text: t }] })

const md = m.serialize({ type: 'doc', content: [{ type: 'table', content: [{ type: 'tableRow', content: [
  { type: 'tableCell', content: [p('line one'), p('line two')] },   // two blocks in one cell
  { type: 'tableCell', content: [p('B')] },
]}]}]})

console.log(JSON.stringify(md))
console.log('contains U+001F:', md.includes(String.fromCharCode(31)))
```

Output:

```
"\n|                   |     |\n| ----------------- | --- |\n| line one\u001fline two | B   |\n"
contains U+001F: true
```

### Expected Behavior

The serialized markdown should contain no control characters. A multi-block cell should collapse to something a markdown table can hold - a space, or `<br>`, or whatever the project prefers - but the internal separator should not survive into the output.

Two directions, depending on what the separator was meant for:

1. If nothing consumes it any more, drop the join or replace `cellSep` with the intended visible separator.
2. If it is meant as an internal marker, strip it in `collapseWhitespace` (e.g. `/[\s\u001F]+/`) before the text reaches the row.

### Additional Context

Found while working around a separate limitation in the same function: `columnCount` is derived from `max(row.length)` and `colspan`/`rowspan` are never read, so a merged-cell table parsed from HTML is silently flattened on serialize. That one is inherent to GFM pipe syntax and we handled it with our own `renderMarkdown` override - happy to open it separately if a span-aware HTML fallback would be in scope.

Yahiro025:
## Fixes

Fixes #8152

## Changes and Review

`renderTableToMarkdown` joined multi-block table cells with `U+001F` (UNIT SEPARATOR). Nothing consumed that marker on the serialize path, and `collapseWhitespace` (`/\s+/`) does not match it, so `editor.getMarkdown()` could persist a control character.

Multi-block cells are now joined with newlines and still serialized as `<br>`. Any leftover `U+001F` is stripped before emit. The public `DEFAULT_CELL_LINE_SEPARATOR` export and optional `cellLineSeparator` argument stay for compatibility.

Verify: serialize a table cell with two paragraphs (`line one`, `line two`) and confirm the markdown contains `line one<br>line two` and no `U+001F`.

## Checklist

### Responsibility

### AI assistance

Drafted with AI assistance. The reproduction, the vestigial-separator change, and the unit/integration tests were reviewed before opening this PR.


packages/extension-table/src/table/utilities/markdown.ts
```diff
@@ -107,25 +105,23 @@ export function renderTableToMarkdown(
...
-          raw = parts.join(cellSep)
+          raw = parts.join('\n')
         } else {
           raw = cellNode.content
             ? h.renderChildren(cellNode.content as unknown as JSONContent[])
             : ''
         }
 
+        if (options.cellLineSeparator) {
+          raw = raw.split(options.cellLineSeparator).join('\n')
```
arnaugomez: This drops an empty paragraph when callers use `cellLineSeparator: '\n\n'`.

Simple reproduction (using `markdownHelpers` from the existing unit test):

1. Create a table cell with three paragraphs: `one`, an empty paragraph (`content: []`), and `two`.
2. Serialize the table with:

```ts
renderTableToMarkdown(table, markdownHelpers, { cellLineSeparator: '\n\n' })
```

Expected cell text: `one<br><br>two` (the output before this PR).
Actual cell text: `one<br>two` — the blank line is lost.

Joining the children first produces `one\n\ntwo`, and this replacement turns those two newlines into one. Could we replace custom separators inside each rendered child **before** joining the children with `\n`? That keeps the paragraph boundaries intact. Please also add this case to the tests.

Reproduced against the base and PR versions of the helper; the default separator is unaffected.

Yahiro025: Thanks — addressed in 916d135.

`replaceCellLineSeparator` now runs on each rendered child **before** joining with `\n`, so an empty paragraph stays a blank line when `cellLineSeparator` is `'\n\n'` (e.g. `one` / empty / `two` → `one<br><br>two`). Default separator behavior is unchanged. Added a regression coverage for that case; table markdown tests are green.

arnaugomez (approved):

PR #8324, opened 2026-09-07
fix(vue): give each node view its own useId() prefix

### LINKED ISSUE #8323: Bug: useId() collides across Vue node views because editor.appContext is a shallow copy, not a shared reference
### Environment

- `@tiptap/vue-3`: 3.31.3
- `@tiptap/core`: 3.31.3
- `vue`: 3.5.12
- Browser: Chrome (reproduces in any browser, this is a JS-level issue, not rendering-related)

### Summary

Every Vue node view rendered by `@tiptap/vue-3` gets its own copy of `editor.appContext`
(`{...instance.appContext, provides: instance.provides}`, see
[`src/EditorContent.ts`](https://github.com/ueberdosis/tiptap/blob/main/packages/vue-3/src/EditorContent.ts)).
Because it's a **shallow copy** and not the same object reference as the parent app's real
`appContext`, Vue's own `useId()` — which is documented to be *"unique per Vue application
instance"* — silently loses that guarantee inside node views: every node view's `setup()` sees a
"fresh" counter and starts generating the same IDs (`v-0`, `v-1`, ...) as every other node view,
instead of continuing the parent app's sequence.

This is surprising and easy to hit for any node view component that itself uses `useId()` (directly,
or indirectly through a child component/composable/third-party library that generates a DOM id this
way) to produce a DOM id it relies on being unique — e.g. to create/register a per-instance resource
keyed by that id (a chart library, a canvas library, anything doing `document.getElementById`-style
registration).

### Minimal reproduction

Full runnable project (Vite + Vue 3.5.12 + @tiptap/vue-3 3.31.3, `npm install && npm run dev`)
available on request — the relevant files:

```vue
<!-- IdBlockView.vue — the component used as the node view -->
<template>
  <node-view-wrapper class="id-block">useId() → <strong>{{ id }}</strong></node-view-wrapper>
</template>

<script setup>
import { useId } from 'vue'
import { NodeViewWrapper, nodeViewProps } from '@tiptap/vue-3'

defineProps(nodeViewProps)
// This is the whole bug: two instances of this exact same component, mounted as sibling Tiptap
// node views in the same editor/app, produce the SAME id here instead of two different ones.
const id = useId()
console.log('useId() ->', id)
</script>
```

```vue
<!-- PlainIdBlock.vue — control group: identical useId() call, but NOT a node view -->
<template>
  <div class="id-block">useId() → <strong>{{ id }}</strong></div>
</template>

<script setup>
import { useId } from 'vue'
const id = useId()
console.log('useId() (plain sibling) ->', id)
</script>
```

```vue
<!-- App.vue -->
<template>
  <!-- Control group: two ordinary sibling components, no Tiptap involved -->
  <PlainIdBlock />
  <PlainIdBlock />

  <!-- Bug: two Tiptap Vue node views, same component, same editor -->
  <EditorContent :editor="editor" />
</template>

<script setup>
import { useEditor, EditorContent, VueNodeViewRenderer } from '@tiptap/vue-3'
import { Node, mergeAttributes } from '@tiptap/core'
import PlainIdBlock from './PlainIdBlock.vue'
import IdBlockView from './IdBlockView.vue'

const IdBlock = Node.create({
  name: 'idBlock',
  group: 'block',
  atom: true,
  // parseHTML/renderHTML needed so the initial `content` string below round-trips at all —
  // without them the custom tag is silently dropped by the HTML parser.
  parseHTML() {
    return [{ tag: 'div[data-id-block]' }]
  },
    return ['div', mergeAttributes(HTMLAttributes, { 'data-id-block': '' })]
  },
  addNodeView() {
    return VueNodeViewRenderer(IdBlockView)
  },
})

const editor = useEditor({
  extensions: [StarterKit, IdBlock],
  content:
    '<p>Two idBlock node views follow:</p><div data-id-block></div><div data-id-block></div>',
})
</script>
```

**Expected console output** — the control group (`PlainIdBlock`, not a node view) gets this, and
it's what the `idBlock` node views below should *also* get, since they're rendered in the same app:
```
useId() (plain sibling) -> v-0
useId() (plain sibling) -> v-1
```

**Actual console output**, captured verbatim from a running instance of this exact reproduction:
```
useId() (plain sibling) -> v-0
useId() (plain sibling) -> v-1
useId() -> v-0
useId() -> v-0
```

Both `idBlock` instances get the exact same id (`v-0`), even though they're two separate component
instances in the same editor, mounted at the same time, in the same document — right next to two
plain sibling components that correctly got `v-0`/`v-1` from the identical `useId()` call.

### Root cause

1. `EditorContent`'s `setup()` sets `editor.appContext` once, from the *current* component instance,
   as a **new plain object**:

   ```js
   // packages/vue-3/src/EditorContent.ts
   if (instance) editor.appContext = {
     ...instance.appContext,
     provides: instance.provides
   }
   ```

2. Every time a node view is rendered, `VueRenderer.renderComponent()` assigns that same cloned object
   as the vnode's `appContext` and renders it with Vue's low-level `render()` API:

   ```js
   // packages/vue-3/src/VueRenderer.ts
   let vNode = h(this.component, this.props)
   if (this.editor.appContext) vNode.appContext = this.editor.appContext
   render(vNode, this.el)
   ```

3. `provides` is explicitly copied over, so `provide`/`inject` correctly cross the node-view boundary
   — that part works as intended and is presumably the reason this clone exists in the first place
   (to keep `provides` up to date with wherever `EditorContent` currently sits in the tree, rather than
   freezing it at that one snapshot forever).

4. But `{...instance.appContext}` only copies `instance.appContext`'s own **enumerable** properties at
   the moment `EditorContent` mounts. Vue's internal `useId()` implementation keeps its counter as
   mutable state that must stay attached to the *same* app-context object identity to behave as
   "unique per app instance" — a shallow copy breaks that invariant. Since every node view's `render()`
   call reuses this one cloned object, and the object itself was never actually threaded through Vue's
   real app-context machinery in the way a literal `instance.appContext` reference would be, `useId()`
   inside a node view behaves as if it's in a brand new, isolated context every time.

### Why this is a bug and not just an edge case

Vue's own docs are explicit about the contract `useId()` provides:

> IDs generated by `useId()` are unique-per-application... For cases where you have multiple Vue
> applications mounted on the same page, you can avoid ID conflicts by giving each app an ID prefix
> via `app.config.idPrefix`.
> — https://vuejs.org/api/composition-api-helpers.html#useid

`@tiptap/vue-3` node views are documented as being part of the *same* application as the rest of the
editor (that's the whole point of forwarding `provides` — so `provide`/`inject` see them as such). A
consumer has no way to know, and no warning anywhere in the docs, that this "same app" guarantee quietly
stops applying to `useId()` (or, most likely, to any other Vue-internal mechanism that similarly
depends on `appContext` being the *same object*, not a copy of its current contents). This is a leaky
abstraction: `provide`/`inject` works as if node views are first-class members of the app, while
`useId()` silently behaves as if they're a separate one — with no way for a component author to detect
this difference without already knowing about this exact implementation detail.

### Suggested fix

Reuse the *same* `appContext` object reference instead of spreading it into a new one, e.g.:

```js
if (instance) {
  editor.appContext = instance.appContext
  editor.appContext.provides = instance.provides
}
```

(mutating `provides` on the real object, rather than creating a new object with a copied `provides`)
would keep `useId()` (and anything else keyed off `appContext` identity) correctly shared, while still
keeping `provides` live-updated the way the current code intends. If mutating the app's real
`appContext` object like this has other side effects that make it undesirable, an alternative is
documenting the limitation explicitly (node views don't share `useId()`'s counter with the rest of the
app) so it isn't a silent trap.

### Workaround (for anyone hitting this in the meantime)

Don't rely on `useId()` (or anything built on it) for uniqueness inside a node view. Use
`crypto.randomUUID()` (or any generator not scoped to Vue's app-context identity) for DOM ids you need
uniqueness guarantees for.

teo-basili:
## Fixes

Fixes #8323

## Changes and Review

Vue node views are mounted standalone via the low-level `render()` API (`VueRenderer.renderComponent()`), so they have no real parent component instance. `useId()`'s per-instance counter (`instance.ids`) is inherited *by reference* from `parent.ids` when a parent exists, and otherwise reset to `["", 0, 0]` — so every node view independently restarts its counter at 0, and any two node views calling `useId()` produce identical ids (`v-0`, `v-0`, ...) instead of unique ones.

This gives each `VueRenderer` its own `appContext.config.idPrefix` — the one piece of that machinery available through a custom `vNode.appContext` — so ids stay unique *across* node views, even though each renderer's own counter still starts over at 0. It's a shallow copy of `config` (same pattern `EditorContent.ts` already uses for `provides`), so the real editor's shared `appContext.config` is never mutated.

Verified with a live reproduction inside `demos/` (two Vue node views + two plain sibling components as a control group): before the fix, the node views logged `useId() -> v-0` / `useId() -> v-0`; after, `useId() -> nv0-0` / `useId() -> nv1-0`, matching the control group's `v-0` / `v-1` behavior.

Added a unit test to `VueRenderer.test.ts` asserting two renderers get distinct `idPrefix`es and that the editor's shared `appContext.config` is left untouched. Full `packages/vue` suite passes (40/40).

## Checklist

### Responsibility


teo-basili:
Addressed in 1967f62:

- **Nitpick (test only checked `idPrefix`, not generated ids)**: fixed — the test now renders a real `useId()`-backed element on each renderer and asserts the two generated ids are distinct, in addition to the existing `appContext.config.idPrefix` assertions.
- **`VueRenderer.ts` null-`editor.appContext` case**: verified against the current code rather than applied blindly. It's real, but only reachable via an `Editor` mounted through a direct `element` option, bypassing `<EditorContent>` — for the standard `useEditor()` + `<EditorContent>` flow, any node views rendered before `appContext` is set are transient (created during the synchronous `Editor` constructor into a detached view) and get torn down and recreated by `editor.createNodeViews()` once `appContext` is set and the real element attaches, before anything is ever painted. I didn't attempt a fix for the bypass case: doing it safely would mean deriving a fallback `AppContext` without one to copy from `editor.appContext`, and the only way I found to construct a real one is `createApp().mount()`'s private, undocumented `_context` (which is what the new test itself uses, acceptable for a test but not something I want to ship in library code depended on by third parties). Documented as a known gap in a code comment instead. Happy to take a different approach here if you have one in mind.

(Note: the earlier changeset-wording nitpick was already fixed in a prior commit before this review ran — that one's based on an older revision.)

packages/vue/src/VueRenderer.ts
```diff
@@ -65,7 +80,26 @@ export class VueRenderer {
...
+      // `editor.appContext` to copy would mean depending on Vue internals with no public API
+      // equivalent (e.g. `createApp().mount()`'s private `_context`), which isn't done here.
+      //
+      // A shallow copy, same as EditorContent.ts already does for `provides` — this must not mutate
+      // `this.editor.appContext.config` in place, since that's the real app's shared config object
+      // (every node view of this editor shares the same `this.editor.appContext`).
+      vNode.appContext = {
+        ...this.editor.appContext,
+        config: {
+          ...this.editor.appContext.config,
```
arnaugomez, line 97: Spreading `appContext.config` triggers an unrelated Vue warning every time a renderer is created or `updateProps()` runs in a runtime-only development build.

A minimal reproduction with that Vue build is:

```ts
import { createApp } from 'vue'

const app = createApp({ render: () => null })
const config = { ...app.config, idPrefix: 'nv0' }
// Unexpected warning: `compilerOptions` is only respected
// when using a build that includes the runtime compiler.
```

Expected: copying the config produces no warnings. Actual: the spread reads Vue's `compilerOptions` getter, which logs the warning. This means normal node-view updates can repeatedly fill the console with a misleading configuration warning.

Please copy the config without evaluating its accessors, then override `idPrefix`. A focused comparison using the PR's renderer produced zero warnings at base and three at this head for two renderer creations and one prop update. This affects development builds; the full test suite was not run.

PR #8329, opened 2026-09-08
perf(core): reuse the inverted mapping in getChangedRanges

### LINKED ISSUE #8328: getChangedRanges rebuilds the inverted mapping for every range
### Affected Packages

core, extension-link

### Tiptap Version

3.31.3

### Browser Used

Not browser-specific

### What happened?

`getChangedRanges` in `@tiptap/core` calls `mapping.invert()` twice and `mapping.slice(index)` twice inside its inner loop, once per range. `invert()` allocates a new Mapping with one map per step, and the Link extension's autolink plugin calls `getChangedRanges` on every transaction.

### Expected Behavior

`mapping.invert()` is computed once per call and `mapping.slice(index)` once per step map.

_No response_

Example repro:
```
const transform = new Transform(doc)
for (let i = 0; i < 4000; i += 1) transform.insert(1 + i, schema.text('a'))
getChangedRanges(transform) // ~300 ms
```

issacgerges:
## Fixes

Fixes #8328

## Changes and Review

`getChangedRanges` called `mapping.invert()` and `mapping.slice(index)` inside the per-range loop. 

This PR computes `invert()` once per call and `slice(index)` once per step map. With 4,000 / 8,000 single-character insert steps, `getChangedRanges` goes from 301 ms / 1,394 ms to 157 ms / 695 ms.

## Checklist

### Responsibility


arnaugomez (approved):
Thanks for your contribution! 

We'd love to know your feedback about Tiptap. **What feature or functionality could Tiptap provide that would be game-changing for you?**

PR #8335, opened 2026-09-10
fix(link): space typed after a link no longer extends it

### LINKED ISSUE #8334: [Bug]: URL that gets pasted never ends and bleeds to text after it
### Affected Packages

extension-link

### Tiptap Version

3.31.3

### Browser Used

Chrome

### What happened?

If i type the link manually and after i hit space it stops with underline and link indicator, but if I paste the link and hit space it continues the link for all the text after it

https://github.com/user-attachments/assets/11e543cb-11a1-4a78-9c81-af9166cb43db

### Expected Behavior

I expected to break the link after i paste it and hit space and continue writing text

_No response_

_No response_

LukaSikic:
## Summary

Fixes https://github.com/ueberdosis/tiptap/issues/8334

- With `autolink` enabled the link mark is inclusive, so a space typed after a pasted URL inherited the mark and everything typed after it grew the link.
- The autolink plugin now removes the link mark from trailing whitespace of a change and clears the stored mark, so typing continues unlinked. A space typed inside multi-word link text keeps the link.
- Pressing Enter was already safe via `keepOnSplit: false`, so only the whitespace path needed the fix.

Also lists `vite-plus` in extension-link devDependencies so the new test file passes the fallow audit gate.

## Test plan

- [x] New unit tests in `packages/extension-link/tests/integration/autolink.test.ts` — regression test fails without the fix, plus guards for the typed-URL flow and spaces inside link text
- [x] New e2e test in `demos/src/Marks/Link/index.spec.ts` covering the real clipboard paste path (passes on React and Vue)
- [x] `vp lint`, `vp run -F @tiptap/extension-link build`, package integration tests (62/62), `fallow audit` verdict pass


arnaugomez:
I reproduced a regression in `unlinkTrailingWhitespace`: it runs for every changed range, including mark-only transactions.

For an existing link containing `"click here "` (including the final space):
- Applying bold to the link removes the link mark from the final space.
- Calling `unsetLink()` and then undoing restores the link only on `"click here"`, leaving the final space unlinked.

Both scenarios preserve the original link before this PR. `getChangedRanges()` includes mark-only steps, so its ranges cannot be treated as newly inserted text.

I suggest limiting whitespace unlinking to actual text edits and preserving formatting during undo/redo. I am working on a fix with regression coverage for formatting changes, history restoration, and the existing typed/pasted URL behavior.

arnaugomez:
Thanks for your contribution! We're reviewing it now.

Thank you image

What would you improve about Tiptap? Create an issue or [share your thoughts here](https://github.com/ueberdosis/tiptap/discussions/7351).

arnaugomez:
The follow-up review found one regression in my fix:

**[P2] Unlink adjacent whitespace insertions together** (`packages/extension-link/src/helpers/autolink.ts`, per-step unlinking loop).

At the end of linked `example`, dispatching `editor.state.tr.insertText(' ').insertText(' ')` produces `<a>example </a> ` instead of `<a>example</a>  `. Processing insertion steps in order means the first space sees the second linked space as continuing link text, so only the second space is unlinked.

I added a regression test covering both insertion orders. The fix processes mapped insertion ranges from right to left, so later spaces are unlinked before earlier spaces are checked. All 73 link integration tests now pass locally; I am completing the remaining checks before pushing the follow-up commit to this PR.

The review found no other actionable issues.

arnaugomez (approved):

PR #8355, opened 2026-09-18
Fix/node view ignore mutation

### LINKED ISSUE #456: Disable Paste

### LINKED ISSUE #8220: iOS `ignoreMutation` branch keys on `this.dom`, catching React node-view chrome → infinite redraw loop (tab freeze)
## Summary

The iOS/Android branch in `NodeView.ignoreMutation` (and its `MarkView` twin) returns `false` for `childList` mutations anywhere inside `this.dom`. For node views whose non-editable chrome is rendered by a framework (`ReactNodeViewRenderer` and equivalents), the framework mounting its content **is** such a mutation — so ProseMirror redraws the node view, the framework remounts, and the two loop synchronously forever. The tab freezes with no exception; iOS kills it.

Present in `@tiptap/core@3.26.1` and unchanged in `3.30.2` (both `dist/index.js` occurrences).

## The branch (src/NodeView.ts / MarkView.ts)

```js
if (this.dom.contains(mutation.target) && mutation.type === "childList"
    && (isiOS() || isAndroid()) && this.editor.isFocused) {
  const changedNodes = [...mutation.addedNodes, ...mutation.removedNodes];
  if (changedNodes.every((node) => node.isContentEditable)) return false;
}
```

It was added (via #1971) so browser-initiated edits from mobile keyboards aren't ignored — but `this.dom` spans the entire node view, including chrome the integration renders around `contentDOM`. React-mounted children report `isContentEditable === true` when the surrounding node view is inside the editable root, so the `every(...)` guard doesn't exclude them, and `editor.isFocused` is true in the common case (inserting from a suggestion menu).

## Loop

1. React mounts the node view's component → `childList` mutation inside `this.dom`
2. branch returns `false` → ProseMirror redraws the node view
3. React remounts → goto 1

Synchronous, rescheduled in a microtask — the main thread never yields, so there is no error, no log, and no recovery.

## Proposed fix

Key the branch on the editable content instead of the whole view:

```diff
- if (this.dom.contains(mutation.target) && mutation.type === "childList" && (isiOS() || isAndroid()) && this.editor.isFocused) {
+ if (this.contentDOM && this.contentDOM.contains(mutation.target) && mutation.type === "childList" && (isiOS() || isAndroid()) && this.editor.isFocused) {
```

This preserves the #1971 behaviour (mobile-keyboard mutations land in `contentDOM`) while excluding node-view chrome. We've been running exactly this as a `patch-package` diff in production (four one-line changes: NodeView + MarkView, ESM + CJS) — it fixes the freeze on real iOS devices with no observed regression to content editing. Happy to open a PR.

## Reproduction (no device needed)

`isiOS()`'s second clause is `navigator.userAgent.includes("Mac") && "ontouchend" in document`, so Chromium with touch emulation on and any Mac-containing UA takes the iOS path. With focus in the editor, insert any node whose node view mounts framework-rendered children (e.g. a TipTap React node view with content) — the renderer wedges immediately; deleting `document.ontouchend` before inserting (flipping `isiOS()` to false) makes the same insertion succeed, isolating the branch as the cause.

Related: #1971 (origin of the branch), #1538 (request to expose `ignoreMutation` in node-view options — integrations like BlockNote hardcode renderer options, so consumers currently have no seam short of patching core). Downstream report: TypeCellOS/BlockNote#2994.

### LINKED ISSUE #8348: [Bug][v3.31.1] iOS and Android that will break node views on enter
### Affected Packages

core

### Tiptap Version

3.31.1

### Browser Used

Chrome

### What happened?

When implementing a code block with `NodeView`, pressing `enter` to start a new line on mobile produces an extra blank line, and the content entered after the line break is lost.

Image

Insert a line break after the end of the first line of code.

Image

### Expected Behavior

It behaves normally when using a PC, or after downgrading to version 3.31.0.

Image

https://github.com/xiaweiss/tiptap-bug-report

The relevant change is in NodeView.ts, line 297. issue #8220

https://github.com/ueberdosis/tiptap/blob/d3bbcf4cdad92470388efd0b6aef53c65b3515ce/packages/core/src/NodeView.ts#L297

xiaweiss:
## Fixes

Revert #8221 
Fixes #8220 
Fixes #8348

#### Revert

When Enter is pressed on mobile, the `contentDOM` gets split first, producing two `contentDOM`s. That's why we need to observe mutations on `this.dom` and `return false` so ProseMirror recreates the DOM — the two will eventually be merged back into a single `contentDOM`.

The change in #8221 meant that only the first `contentDOM` (the one produced after the split) was being observed. As a result, the `NodeView` gets split, and data in the second `contentDOM` can even be lost.

#### Fix

The reproduction provided in #8220 uses a nested NodeView

image

```html
<parent-node-view-wrapper>
    <parent-node-view-content + child-node-view-wrapper>
        <child-node-view-content>
```

During the mutation, the intermediate layer is a contentDOM, a node-view-wrapper, and a node-view-content all at once. As a result, inserting child-node-view-content incorrectly triggers the logic that is supposed to prevent the NodeView from being split, creating a loop.

Here we ignore the case where an insertion occurs on the intermediate layer.

## Checklist

### Responsibility


arnaugomez:
Thanks for your contribution! 

We'd love to know your feedback about Tiptap. **What feature or functionality could Tiptap provide that would be game-changing for you?**

xiaweiss:
> Thanks for your contribution!
> 
> We'd love to know your feedback about Tiptap. **What feature or functionality could Tiptap provide that would be game-changing for you?**

Thanks for the tiptap team's hard work.

Tiptap is highly customizable and fits well with the Vue framework I'm using. Right now I only use the core library, tiptap/core, together with extensions I've developed myself, to build my note-taking functionality.

Thanks to the power of the open-source community, there are not only Tiptap's own demos but also many plugins built by third-party authors, which lowers the barrier to development.

